PR #19, opened 2026-03-30
Overwrite op results with subgraph results.

chinazhangchao:


src/winml/modelkit/pattern/transpose_patterns.py
```diff
@@ -374,7 +374,7 @@ class ReshapeTransposeReshapeOverlyHighDimPatternInputGenerator(
     """PatternInputGenerator for ReshapeTransposeReshapeOverlyHighDim pattern."""
 
     pattern = ReshapeTransposeReshapeOverlyHighDimPattern()
-    registration_name = "ReshapeTransposeReshapeOverlyHighDim"
+    registration_name = "ReshapeTransposeReshapeOverlyHighDimPattern"
```
vortex-captain, line 377: We would need to add `Pattern` suffix to all `registration_name`s in other "*_patterns.py" files

src/winml/modelkit/analyze/core/runtime_checker_query.py
```diff
@@ -1832,77 +1961,231 @@ def run_for_subgraph(
...
+                        compile_result = ret.iloc[0]["compile_run_success"][0]
+                        run_result = ret.iloc[0]["compile_run_success"][1]
+                    else:
+                        logger.info(
+                            "Negative rules passed but properties not found for %s: %s",
+                            pattern_name,
+                            filter_v,
+                        )
+
+                        if run_unknown_op:
```
vortex-captain: should we remove condition `if run_unknown_op:`? If I understand correctly, since `_run_for_subgraph_per_node` calls `run_for_node`, which also looks up data in zip, and this should be applied no matter "run_unknown_op" is true or not.

src/winml/modelkit/analyze/core/runtime_checker_query.py
```diff
@@ -1832,77 +1961,231 @@ def run_for_subgraph(
...
+                                    "table_file": table_file,
+                                },
                             ),
                             alternatives=self.alternatives,
                             pattern_match=pattern_match,
                         )
-                    # Pattern exists but no negative rules
+                else:
+                    # No table data — fallback to per-node check if allowed
+                    if run_unknown_op:
```
vortex-captain: remove `if run_unknown_op:`? Similar considerations as in the other comment

vortex-captain (dismissed):

vortex-captain (approved):

PR #23, opened 2026-03-30
Normalize input_constraints before computing case signature

xieofxie:
## Summary

- Calls `normalize_constraint_dict` on each `input_constraints` value in `_compute_case_signature` so that the compact `same_value`/`same_value_shape` representation (emitted by `InputValueConstraint.to_dict()` for uniform arrays) produces the same signature as the equivalent fully-expanded `value` list stored by older code.
- Adds unit tests in `tests/unit/analyze/test_check_ops.py` constructed via real `InputValueConstraint` instances, covering backward-compatible signature stability and explicit verification that dict key insertion order has no effect (`sort_keys=True`).

## Test plan

- [ ] `python -m pytest tests/unit/analyze/test_check_ops.py -v` — 9 tests pass
- [ ] `test_all_same_value_array_compact_matches_expanded` — compact vs expanded form produce same signature
- [ ] `test_key_insertion_order_does_not_affect_signature` — dict key order has no effect
- [ ] `test_top_level_input_constraints_key_order_does_not_matter` — input key ordering has no effect


xieofxie:
So test cases for Where will save size and we could use large sizes

src/winml/modelkit/pattern/op_input_gen/op_input_gen.py
```diff
@@ -177,6 +185,21 @@ def to_dict(self) -> dict[str, Any]:
...
+    """Expand same_value/same_value_shape back to the canonical value list form.
+
+    Converts the compact representation produced by InputValueConstraint.to_dict()
+    when all values are equal, back to the full nested value list. Use this when
+    consuming serialized constraint dicts to ensure consistent handling regardless
+    of which representation was saved.
+    """
+    if "same_value" in c and "same_value_shape" in c:
+        normalized = {k: v for k, v in c.items() if k not in ("same_value", "same_value_shape")}
+        normalized["value"] = np.full(c["same_value_shape"], c["same_value"]).tolist()
```
fangyangci:
```suggestion
        dtype = np.dtype(c["dtype"]) if "dtype" in c else None
        normalized["value"] = np.full(c["same_value_shape"], c["same_value"], dtype=dtype).tolist()

```

fangyangci: Use the stored dtype to reconstruct the array so that .tolist() produces the same Python scalar types (e.g. float vs int) as the original array, which keeps JSON serialization and case_index signatures stable.

fangyangci (changes requested):

src/winml/modelkit/analyze/runtime_checker/check_ops.py
```diff
@@ -846,6 +846,9 @@ def build_parser():
 
 def run_from_args(args: Any) -> None:
     """Run check_ops from parsed CLI args."""
+    from ... import winml
+
+    winml.register_execution_providers(ort=True)
```
xieofxie, line 851: move here so test could run on github workflow

fangyangci (approved):

PR #200, opened 2026-04-01
fix: resolve charmap codec error in wmk analyze and wmk inspect on Windows

### LINKED ISSUE #191: wmk analyze / wmk inspect: Windows console crashes with charmap codec error due to emoji in output
## Summary

On Windows with the default cp1252 console encoding, \`wmk analyze\` and \`wmk inspect\` crash with a \`UnicodeEncodeError\` when rendering Rich output that contains emoji characters (e.g., 📊, ✓).

## Repro

\`\`\`bash
wmk analyze --model resnet_onnx/model.onnx
# or
wmk inspect -m microsoft/resnet-50
\`\`\`

**Error:**
\`\`\`
UnicodeEncodeError: 'charmap' codec can't encode character '\U0001f4ca'
  in position N: character maps to <undefined>
\`\`\`

## Workaround

\`\`\`bash
PYTHONIOENCODING=utf-8 wmk analyze --model resnet_onnx/model.onnx
\`\`\`

## Root Cause

The CLI entry point does not set \`PYTHONIOENCODING\` or configure the stdout/stderr stream to use UTF-8, so Python falls back to the Windows default cp1252 encoding which cannot encode Unicode emoji used in Rich console output.

## Acceptance Criteria

- [ ] \`wmk analyze\` and \`wmk inspect\` run without encoding errors on Windows with default console settings
- [ ] Fix applied at CLI entry point (e.g., \`reconfigure(encoding='utf-8')\` on stdout/stderr, or replace emoji with ASCII-safe fallbacks)
- [ ] No \`PYTHONIOENCODING\` env var required as workaround

DingmaomaoBJTU:
## Summary

Fixes #191

- **Root cause (analyze)**: `StaticAnalyzerConsoleWriter` passed `legacy_windows=False` to `Console()`. This flag bypasses Rich's Win32 `WriteConsoleW` path and forces output through Python's IO layer, which on a default Windows console uses cp1252 — a codec that cannot encode any emoji or Unicode check marks. `wmk perf` and other commands use a plain `Console()` with no such flag, which is why they work fine.

- **Root cause (inspect)**: `inspect/formatter.py` uses `✓`/`✗`/`○`/`◇` as status icons; these are all outside cp1252 and crash even with the default Console path.

**Fix:**
1. Remove `legacy_windows=False` (and the unrelated `force_terminal=True`) from `StaticAnalyzerConsoleWriter` — matching the pattern used by `wmk perf`.
2. Replace all non-cp1252 symbols in `analyze/console_writer.py`, `commands/analyze.py`, and `inspect/formatter.py` with ASCII equivalents (`+`/`x`/`!`/`?`/`-`/`*`).

image

### DIFF
--- a/src/winml/modelkit/analyze/console_writer.py
+++ b/src/winml/modelkit/analyze/console_writer.py
@@ -201,7 +201,7 @@ class StaticAnalyzerConsoleWriter:
             unique_operator_types: Total number of unique operator types
         """
         # Header with support status
-        status_icon = "[+]" if ihv_result.runtime_support else "[x]"
+        status_icon = "+" if ihv_result.runtime_support else "x"
         status_text = (
             self._bright_green("SUPPORTED")
             if ihv_result.runtime_support
--- a/src/winml/modelkit/analyze/console_writer.py
+++ b/src/winml/modelkit/analyze/console_writer.py
@@ -43,8 +43,6 @@ class StaticAnalyzerConsoleWriter:
         self.verbose = verbose
         self.console = console or Console(
             width=self.CONSOLE_WIDTH,
-            force_terminal=True,
-            legacy_windows=False,
             highlight=False,
         )
         # Ensure we have a reliable width value (Rich may return None in some cases)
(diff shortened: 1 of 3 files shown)


src/winml/modelkit/analyze/console_writer.py
```diff
@@ -464,7 +462,7 @@ def _write_information_items(self, information_list: list[Information], ihv_name
     def _write_footer(self, analysis: AnalysisOutput) -> None:
         """Write analysis footer with summary."""
         self.console.print("=" * self.SEPARATOR_LENGTH)
-        self.console.print(f"📈 {self._bold('ANALYSIS SUMMARY')}")
+        self.console.print(f"{self._bold('ANALYSIS SUMMARY')}")
```
timenick, line 465: Bug: Rich markup collision with `[+]` and `[x]`

`console.print()` parses square brackets as Rich markup tags. `highlight=False` only disables syntax highlighting — it does **not** disable markup parsing. So `"[+]"` and `"[x]"` will be interpreted as (invalid) style tags, potentially swallowed silently or causing a `MarkupError` at runtime.

The `classification_info` block just below (lines 231–234) correctly uses bare characters (`"+"`, `"!"`, `"?"`, `"x"`) without brackets. Suggest matching that pattern here:

```python
status_icon = "+" if ihv_result.runtime_support else "x"
```

Or escape them explicitly:

```python
from rich.markup import escape
status_icon = escape("[+]") if ihv_result.runtime_support else escape("[x]")
```

🤖 Generated with Claude Code

DingmaomaoBJTU: Fixed in a321715 — changed to bare `+`/`x` without brackets, matching the pattern in `classification_info`.

timenick (approved):

PR #208, opened 2026-04-01
fix: resolve charmap codec error in remaining command files on Windows

### LINKED ISSUE #207: fix: charmap codec error in remaining command files on Windows
## Bug

Several command files under `src/winml/modelkit/commands/` crash on Windows with `UnicodeEncodeError` when printing non-ASCII characters (`→`, `—`, `•`). The default `cp1252` encoding on Windows cannot encode these characters.

```
UnicodeEncodeError: 'charmap' codec can't encode character '\u2192' in position 1: character maps to <undefined>
```

## Root Cause

PR #200 (commit 92325bf) fixed the same class of bug in `analyze` and `inspect`, but **missed the following files**:

- `optimize.py` — `→` and `•` in comments and `console.print()`
- `build.py` — `—` in comment
- `config.py` — `—` in string literal (runtime crash)
- `eval.py` — `→` in comment
- `live_chart.py` — `→` and `—` in comments
- `perf.py` — `—` and `→` in comments

## Fix

Replace `→` with `->`, `—` with `--`, and `•` with `*` in all affected files, matching the approach in PR #200.

## Environment

- Windows 11, Python 3.10.9, cp1252 console encoding

timenick:
## Summary
- Replace non-ASCII characters (`→`, `—`, `•`) with ASCII equivalents (`->`, `--`, `*`) in all command files under `src/winml/modelkit/commands/`
- Extends the fix from #200 (which covered `analyze` and `inspect`) and the original scope of `optimize.py` to also cover `build.py`, `config.py`, `eval.py`, `live_chart.py`, and `perf.py`

Closes #207


timenick:
PR looks good — clean ASCII replacement matching the #200 approach, no remaining non-ASCII in `optimize.py`.

However, found the same `→` (U+2192) issue in two more files:
- `src/winml/modelkit/commands/build.py:231` — `→` in comment
- `src/winml/modelkit/commands/config.py:102` — `→` in string literal (will crash at runtime on cp1252)

`config.py:102` is a runtime string, so it'll hit the same `UnicodeEncodeError`. Recommend fixing those here or in a follow-up.

🤖 Generated with Claude Code

DingmaomaoBJTU (approved):

PR #209, opened 2026-04-01
fix: op-tracing preflight, eval samples merge, compile device label, HTP tagged_nodes stats

xieofxie:
## Summary

- **Bug A (P2)** `wmk perf --op-tracing`: moved `is_qnn_profiling_available()` check to before `PerfBenchmark.run()` so the command fails fast instead of wasting a full benchmark run
- **Bug B (P2)** `wmk eval --samples N`: replaced wholesale `DatasetConfig` overwrite with a merge that preserves user-specified `samples`, `split`, `shuffle`, and `seed` when falling back to a default dataset
- **Bug C (P3)** `wmk compile --ep dml`: derive the displayed `Device:` label from `_EP_TO_DEVICE[provider]` instead of the raw CLI `--device` default (`npu`)
- **Bug D (P3)** `wmk export --no-hierarchy`: gate `tagged_nodes` and `coverage_percentage` stats on `embed_hierarchy_attributes` so they report `0` when hierarchy tagging is disabled

Each bug is covered by a new regression test that failed before the fix and passes after.


src/winml/modelkit/export/htp/exporter.py
```diff
@@ -306,8 +307,7 @@ def export(
 
             # Calculate coverage percentage
             total_nodes = len(onnx_model.graph.node)
-            tagged_nodes = len(self._tagged_nodes)
-            coverage = (tagged_nodes / total_nodes * 100.0) if total_nodes > 0 else 0.0
+            coverage = (embedded_count / total_nodes * 100.0) if total_nodes > 0 else 0.0
             self._export_stats["coverage_percentage"] = coverage
```
xieofxie: should coverage_percentage be 0 in this case ? i.e. embed_hierarchy_attributes = false

tests/unit/export/test_htp_exporter_stats.py
```diff
@@ -0,0 +1,66 @@
...
+# Licensed under the MIT License.
+# --------------------------------------------------------------------------
+"""Tests for HTPExporter export statistics correctness."""
+
+from __future__ import annotations
+
+from unittest.mock import MagicMock
+
+from winml.modelkit.export.htp.exporter import HTPExporter
+
```
timenick, line 12: `HTPExporter` is already exported in `winml.modelkit.export.htp.__init__.py`, so per project import rules this should be:

```python
from winml.modelkit.export.htp import HTPExporter
```

Test code should import from the package level, not reach into internal submodules for publicly exported symbols.

🤖 Generated with Claude Code

src/winml/modelkit/eval/evaluate.py
```diff
@@ -152,7 +152,18 @@ def evaluate(config: WinMLEvaluationConfig) -> EvalResult:
...
         config.dataset = deepcopy(default)
+        # Preserve user-specified values that differ from DatasetConfig class defaults
+        _cls_default = DatasetConfig()
+        if user_dataset.samples != _cls_default.samples:
+            config.dataset.samples = user_dataset.samples
+        if user_dataset.split != _cls_default.split:
+            config.dataset.split = user_dataset.split
+        if user_dataset.shuffle != _cls_default.shuffle:
+            config.dataset.shuffle = user_dataset.shuffle
+        if user_dataset.seed != _cls_default.seed:
```
timenick: This approach compares user values against `DatasetConfig()` class defaults to detect "user-specified" fields. It breaks when a user intentionally passes a value that matches the class default — e.g. `--samples 100` or `--seed 42` would be silently dropped if the default dataset defines different values.

A more robust approach: have the CLI layer (or config builder) track which fields were explicitly set, e.g. via a `set()` of field names or sentinel defaults like `None`, so the merge can distinguish "user passed 100" from "nobody touched this field".

🤖 Generated with Claude Code

src/winml/modelkit/export/htp/exporter.py
```diff
@@ -296,7 +296,8 @@ def export(
             self._export_stats["export_time"] = export_time
             self._export_stats["hierarchy_modules"] = len(self._hierarchy_data)
             self._export_stats["onnx_nodes"] = len(onnx_model.graph.node)
-            self._export_stats["tagged_nodes"] = len(self._tagged_nodes)
+            embedded_count = len(self._tagged_nodes) if self.embed_hierarchy_attributes else 0
```
timenick: This `embedded_count` + coverage calculation is duplicated between `export()` (here) and `_apply_hierarchy_tags()` (line 514). If one is updated without the other, stats will silently diverge. Consider extracting a small helper like `_compute_tag_stats(total_nodes)` to keep them in sync.

🤖 Generated with Claude Code

src/winml/modelkit/export/htp/exporter.py
xieofxie: if it means real tagged / total not could tagged / total, then it is correct

src/winml/modelkit/commands/eval.py
```diff
@@ -61,22 +61,19 @@
 @click.option(
     "--samples",
     type=int,
-    default=100,
-    show_default=True,
-    help="Number of dataset samples.",
+    default=None,
```
xieofxie, line 64: @zhenchaoni

src/winml/modelkit/export/htp/exporter.py
```diff
@@ -493,6 +489,18 @@ def _get_optimum_patcher(model: nn.Module, task: str | None) -> Any:
...
+    def _update_tag_stats(self, total_nodes: int) -> None:
+        """Update tagged_nodes and coverage_percentage in export stats.
+
+        Centralises the embed-aware calculation so _apply_hierarchy_tags and
+        the final stats block in export() always stay in sync.
+        """
+        embedded_count = len(self._tagged_nodes) if self.embed_hierarchy_attributes else 0
+        self._export_stats["tagged_nodes"] = embedded_count
+        self._export_stats["coverage_percentage"] = (
+            embedded_count / total_nodes * 100.0 if total_nodes > 0 else 0.0
```
xieofxie, line 501: @tezheng

src/winml/modelkit/commands/eval.py
zhenchaoni: Looks good to me

zhenchaoni (approved):
The change related to eval and evaluate looks good to me

src/winml/modelkit/commands/eval.py
tezheng, line 1: I dont quite get the default value changes here, why remove default values from cli args, and delay to the function call?

xieofxie: I have updated the previous bugbash report in description

Bug B — P2: wmk eval --samples N silently ignored when no --dataset is given
Command: wmk eval -m microsoft/resnet-50 --samples 20
Symptom: Output always shows 'samples': 100 regardless of --samples value.
Root cause: evaluate.py:149-155 — when config.dataset.path is None (no --dataset flag), the entire DatasetConfig is replaced wholesale with the hardcoded _DEFAULT_DATASETS entry (which has samples=100), discarding the user's value:

if config.dataset.path is None:
    config.dataset = deepcopy(default)   # overwrites samples, split, shuffle
Fix: Merge user-specified fields (samples, split, shuffle, seed) onto the default rather than replacing it entirely.
Location: src/winml/modelkit/eval/evaluate.py:155

The None could let user decide either user value or default config value could take effect

PR #212, opened 2026-04-01
Fix facebook/sam2.1-hiera-small.

chinazhangchao:


timenick (dismissed):

src/winml/modelkit/optim/api.py
```diff
@@ -256,6 +256,7 @@ def optimize_onnx(
     logger.info("Starting optimization pipeline...")
     optimizer = Optimizer()
     optimized_model = optimizer.optimize(loaded_model, **optimizer_kwargs)
+    optimized_model = optimizer.optimize(optimized_model, **optimizer_kwargs)
```
vortex-captain, line 260: would be great to add a comment like 
```python
# more than one "optimize" call is necessary for certain models for thorough optimization
```

vortex-captain (dismissed):

vortex-captain (approved):

PR #224, opened 2026-04-02
 feat: Conv2DInplaceLinear patterns (4D/3D/2D)

AChenreddy24:
Adds three Conv2D pattern variants that convert MatMul+Add linear layers to 1x1 Conv2D for Qualcomm HTP acceleration. The patterns work through PatternRewriter.

4D (NHWC): Transpose → Conv(1x1) → Transpose (for models like ConvNeXT)
3D (sequence): Transpose → Unsqueeze → Conv(1x1) → Squeeze → Transpose (for transformer models)
2D (Gemm): Reshape → Conv(1x1) → Reshape (for SAM2-style Gemm layers)
Validated on ConvNeXT (36 MatMul+Add → Conv2D) and SAM2 (5 Gemm → Conv2D, max diff 2.98e-06).

As the comments were made on the PR of old repo, below I summarized them along with the responses.

**@reny_microsoft: "It'd be great to add some UTs, please refer to tests\unit\static_analyzer\pattern"**
Added test_conv2d_inplace_linear.py with 23 tests covering schema consistency, self-matching, model generation, dimension checks, and PatternRewriter round trips. Also updated test_pattern_input_generator.py registered count. Full pattern test suite passes (301 passed).

**@reny_microsoft: "Once the pattern classes of source patterns and target pattern are defined, the rewrite should already work without adding rewrite-specific code, by using PatternRewriter"**
Confirmed! all patterns share _MATMUL_ADD_SCHEMA so PatternRewriter handles the rewriting directly. Reverted all changes to modelkit/optim/pipes/surgery.py and modelkit/optim/capabilities/surgery.py. The weight transformation (B: 2D matrix → 4D conv weight) is embedded in each skeleton through Transpose + Reshape nodes on the B input path. Tests test_rewrite_matmuladd_to_conv2d_4d, test_rewrite_matmuladd_to_conv2d_3d, and test_rewrite_matmuladd_to_conv2d_2d verify the full round-trip.

**@reny_microsoft: "Patterns sharing same inputs and outputs should also share identical schemas so that PatternRewriter can rewrite between these patterns"**
Removed the 4 separate schemas from the original PR. All 3 Conv2D patterns now import and return _MATMUL_ADD_SCHEMA from gemm_patterns.py, the same schema object used by MatMulAddPattern and ReshapeGemmReshapePattern. Test test_all_patterns_share_matmuladd_schema verifies this with identity check.

**@reny_microsoft: "Ideally, one implementation of PatternInputGenerator could be shared across different patterns which share the same pattern schema"**
Conv2DInplaceLinearInputGeneratorBase holds all the generator logic. The three registered subclasses only set pattern, registration_name, and override _get_a_shapes() to return shapes appropriate for their topology. Follows the same pattern as GemmPatternInputGenerator shared between MatMulAddPatternInputGenerator and ReshapeGemmReshapePatternInputGenerator.

**@reny_microsoft: "If the 3D patterns work, could you help migrate 3D patterns... check_model would not be a hard requirement"**
Added back Conv2DInplaceLinear3DPattern. The earlier issue was that Unsqueeze/Squeeze need axis constants as inputs and that is fixed by declaring them as internal constants in get_internal_constants_and_attributes (axes=[-1] for both). The generated model self-matches and rewrites correctly.

**@reny_microsoft: "For each conv2d pattern, please consider checking input dimension in get_internal_constants_and_attributes, and raise PatternMismatchedException in case of mismatch"**
Each pattern now checks inputs["A"].ndim in get_internal_constants_and_attributes and raises PatternMismatchedError if the rank doesn't match (e.g., 4D pattern rejects 3D input). Tests test_4d_pattern_rejects_3d_input, test_3d_pattern_rejects_4d_input, test_2d_pattern_rejects_3d_input verify this.

**@reny_microsoft: "Please help update modelkit\pattern\base.py to handle the PatternMismatchedException raised in get_internal_constants_and_attributes"**
Updated PatternRewriter.rewrite() in base.py to wrap new_pattern.get_onnx_model(...) in a try/except block. if PatternMismatchedError is raised, the rewrite is skipped with continue. Test test_rewrite_skips_dimension_mismatch verifies that rewriting a 2D MatMulAdd model to 4D Conv2D is skipped gracefully (MatMulAdd is still present after).

**@reny_microsoft: "Also generator class for 3D pattern"**
Added Conv2DInplaceLinear3DPatternInputGenerator with 3D input shapes (1, 10, 4) and (2, 5, 4).

**@reny_microsoft: "Please refer to #302 and use CLI command wmk optimize --list-rewrites to confirm that new patterns shown in the list"**
Added all three Conv2D variants as alternatives to the MatMulAdd entry in default.json. They appear as matmuladd-conv2d4d, matmuladd-conv2d3d, and matmuladd-conv2d2d in the rewrite groups.


AChenreddy24:
@microsoft-github-policy-service agree company="Microsoft"

src/winml/modelkit/pattern/base.py
```diff
@@ -1994,16 +1994,19 @@ def rewrite(
...
+                        attributes=match_result.attributes,
+                        is_constant_map=is_constant_map,
+                        output_dtypes=output_dtypes,
+                        domain_versions=self.domain_versions,
+                        prefix=prefix,
+                        input_names=input_names,
+                        output_names=output_names,
+                    )
+                except PatternMismatchedError:
+                    continue
```
timenick, line 2010: The `except PatternMismatchedError: continue` silently swallows the exception with no logging. When `get_internal_constants_and_attributes` raises due to dimension mismatch, there's no trace in logs. Other patterns calling `get_onnx_model` could also raise `PatternMismatchedError` for different reasons.

Consider adding a debug log so rewrite skip events are visible when troubleshooting:
```python
except PatternMismatchedError as e:
    logger.debug("Skipping rewrite %s: %s", new_pattern_class.__name__, e)
    continue
```

🤖 Generated with Claude Code

src/winml/modelkit/pattern/conv2d_inplace_linear_patterns.py
```diff
@@ -0,0 +1,407 @@
...
+
+        Raises PatternMismatchedError if A is not 4D.
+        """
+        if "A" in inputs and inputs["A"] is not None and inputs["A"].ndim != 4:
+            raise PatternMismatchedError(
+                f"Conv2DInplaceLinear4D requires 4D input A, got {inputs['A'].ndim}D"
+            )
+
+        internal_constants: list[tuple[int, int, np.ndarray]] = []
+
```
timenick, line 136: `_weight_reshape_constant` returns `None` when B is missing or `b.ndim != 2`, which silently produces a model without the weight reshape constant. Since `_check_bc_shapes` already validates B is 2D at match time, the `b.ndim != 2` guard here is redundant.

If B somehow arrives as non-2D at model generation time, that indicates a bug upstream — better to let it surface as an error rather than silently generating a broken model. Consider removing the `ndim` guard or raising an explicit error.

🤖 Generated with Claude Code

src/winml/modelkit/pattern/conv2d_inplace_linear_patterns.py
```diff
@@ -0,0 +1,407 @@
...
+class Conv2DInplaceLinearInputGeneratorBase(PatternInputGenerator):
+    """Shared input generator for Conv2DInplaceLinear patterns.
+
+    Subclasses override ``_get_a_shapes`` for topology-appropriate A shapes.
+    """
+
+    def _get_a_shapes(self) -> list[tuple[int, ...]]:
+        """Return A shapes compatible with this pattern's topology."""
+        raise NotImplementedError
```
timenick: `_get_a_shapes` raises `NotImplementedError` at runtime. If a subclass forgets to override it, the failure only surfaces when `get_input_and_infinite_attribute_combinations` is called, not at class instantiation.

Consider using `@abc.abstractmethod` so missing overrides are caught at import/instantiation time:
```python
from abc import abstractmethod

@abstractmethod
def _get_a_shapes(self) -> list[tuple[int, ...]]:
    """Return A shapes compatible with this pattern's topology."""
```

🤖 Generated with Claude Code

timenick (commented):
Review of Conv2D inplace linear patterns PR.

src/winml/modelkit/pattern/base.py
AChenreddy24: Makes sense, added [logger.debug("Skipping rewrite %s: %s", new_pattern_class.__name__, e)](vscode-file://vscode-app/c:/Users/achenreddy/AppData/Local/Programs/Microsoft%20VS%20Code/cfbea10c5f/resources/app/out/vs/code/electron-browser/workbench/workbench.html) so skipped rewrites are visible during troubleshooting.

src/winml/modelkit/pattern/conv2d_inplace_linear_patterns.py
AChenreddy24: Agreed, the silent return None was hiding potential upstream issues. Changed to [raise ValueError](vscode-file://vscode-app/c:/Users/achenreddy/AppData/Local/Programs/Microsoft%20VS%20Code/cfbea10c5f/resources/app/out/vs/code/electron-browser/workbench/workbench.html) with a descriptive message.

src/winml/modelkit/pattern/conv2d_inplace_linear_patterns.py
AChenreddy24: Thanks, switched from raise NotImplementedError to [@abc.abstractmethod](vscode-file://vscode-app/c:/Users/achenreddy/AppData/Local/Programs/Microsoft%20VS%20Code/cfbea10c5f/resources/app/out/vs/code/electron-browser/workbench/workbench.html). Now a missing [_get_a_shapes](vscode-file://vscode-app/c:/Users/achenreddy/AppData/Local/Programs/Microsoft%20VS%20Code/cfbea10c5f/resources/app/out/vs/code/electron-browser/workbench/workbench.html) override will be caught at class instantiation instead of at runtime.

src/winml/modelkit/pattern/conv2d_inplace_linear_patterns.py
```diff
@@ -0,0 +1,412 @@
...
+and raises PatternMismatchedError if the input rank is incompatible,
+so that PatternRewriter can skip infeasible rewrites gracefully.
+"""
+
+from abc import abstractmethod
+from typing import Any
+
+import numpy as np
+
+from ..onnx import ONNXDomain
```
timenick, line 30: nit: `_MATMUL_ADD_SCHEMA` is a `_`-prefixed (private) symbol, but it's now shared across two modules (`gemm_patterns.py` and `conv2d_inplace_linear_patterns.py`). Consider promoting it to a public name (e.g., `MATMUL_ADD_SCHEMA`) and exporting it from `__init__.py`, since it's becoming a shared contract rather than an implementation detail.

🤖 Generated with Claude Code

src/winml/modelkit/pattern/conv2d_inplace_linear_patterns.py
```diff
@@ -0,0 +1,412 @@
...
+        inputs: dict[str, np.ndarray],
+        attributes: dict[str, Any],
+        is_constant_map: dict[str, bool],
+        domain_versions: dict[ONNXDomain, int],
+    ) -> tuple[list[tuple[int, int, np.ndarray]], dict[tuple[int, str], Any]]:
+        """Return weight reshape shape and node attributes.
+
+        Raises PatternMismatchedError if A is not 4D.
+        """
+        if "A" in inputs and inputs["A"] is not None and inputs["A"].ndim != 4:
```
timenick: nit: The guard `if "A" in inputs and inputs["A"] is not None` means that if `A` is missing or None, the dimension check is silently skipped and no `PatternMismatchedError` is raised. Is that intentional? If `get_onnx_model` / `get_internal_constants_and_attributes` is always called with concrete inputs during rewriting, this may never matter in practice — but if inputs can be incomplete, this would bypass the validation and produce a less clear error downstream.

🤖 Generated with Claude Code

timenick (commented):
Review comments posted inline.

🤖 Generated with Claude Code

timenick (dismissed):

src/winml/modelkit/pattern/conv2d_inplace_linear_patterns.py
AChenreddy24: Thanks, renamed to [MATMUL_ADD_SCHEMA](vscode-file://vscode-app/c:/Users/achenreddy/AppData/Local/Programs/Microsoft%20VS%20Code/cfbea10c5f/resources/app/out/vs/code/electron-browser/workbench/workbench.html) and exported from [__init__.py](vscode-file://vscode-app/c:/Users/achenreddy/AppData/Local/Programs/Microsoft%20VS%20Code/cfbea10c5f/resources/app/out/vs/code/electron-browser/workbench/workbench.html) since it's now a shared contract across [gemm_patterns.py](vscode-file://vscode-app/c:/Users/achenreddy/AppData/Local/Programs/Microsoft%20VS%20Code/cfbea10c5f/resources/app/out/vs/code/electron-browser/workbench/workbench.html) and [conv2d_inplace_linear_patterns.py](vscode-file://vscode-app/c:/Users/achenreddy/AppData/Local/Programs/Microsoft%20VS%20Code/cfbea10c5f/resources/app/out/vs/code/electron-browser/workbench/workbench.html)

src/winml/modelkit/pattern/conv2d_inplace_linear_patterns.py
AChenreddy24: Fixed it. Changed from if "A" in inputs and inputs["A"] is not None to [a = inputs.get("A")](vscode-file://vscode-app/c:/Users/achenreddy/AppData/Local/Programs/Microsoft%20VS%20Code/cfbea10c5f/resources/app/out/vs/code/electron-browser/workbench/workbench.html) with an explicit [raise PatternMismatchedError](vscode-file://vscode-app/c:/Users/achenreddy/AppData/Local/Programs/Microsoft%20VS%20Code/cfbea10c5f/resources/app/out/vs/code/electron-browser/workbench/workbench.html) when A is missing or None. Applied to all 3 patterns.

timenick (approved):

PR #241, opened 2026-04-03
Fix test transient failure.

chinazhangchao:


tezheng (commented):
Pushed a follow-up commit that moves the retry logic from the test into `PdhQuery.collect()` itself.

**What changed:**

- `PdhQuery.collect()` now retries until all counters return non-None (default: 1s timeout, 100ms interval). This handles the known PDH quirk where rate counters transiently return `None` right after `prime()`.
- The old single-shot behavior is preserved as `_collect_once()`, used internally by `PdhPoller._poll_loop()` and `stop()` where `None` is a legitimate "no data yet" signal and retrying would stall the polling thread.
- `PdhPoller.start()` simplified: replaces `time.sleep(0.05)` + raw `collect()` with `collect(interval=0.05)`.
- Test simplified from 12-line retry loop to 3 lines: `prime()` → `collect()` → assert.

**New `collect()` logic:**

```python
def collect(self, *, timeout=1.0, interval=0.1):
    deadline = time.monotonic() + timeout
    while True:
        values = self._collect_once()          # 1. query PDH
        if all(v is not None for v in values.values()):  # 2. all counters have data?
            return values                       #    → yes: done
        if time.monotonic() >= deadline:        # 3. out of time?
            return values                       #    → yes: return best-effort
        time.sleep(interval)                    # 4. wait, then retry
```

```
prime() called
    │
    ▼
┌─ _collect_once() ◄─────────────┐
│                                │
├─ all values non-None? ─YES──► return ✓
│                                │
├─ past deadline? ─────YES──► return (with Nones)
│                                │
└─ sleep(interval) ──────────────┘
```

- **Happy path**: first `_collect_once()` returns all valid data → returns immediately, zero sleep.
- **Post-prime path** (the bug this fixes): first 1-2 calls return `None` → retries every 100ms → typically succeeds within 200ms.
- **Worst case**: retries for 1s, then returns with `None` values. Caller decides what's fatal.

This way every caller of `collect()` gets the fix for free, not just this one test.

tezheng (approved):

vortex-captain (approved):

timenick (approved):

PR #244, opened 2026-04-03
Fix Slice and Split derive_properties for dynamic axes

vortex-captain:
## Summary
- Fix `Slice` `derive_properties` crash when `data_shape` contains symbolic dimension names (e.g. `"time"` from dynamic axes). Filters to fixed-shape axes only when computing `starts_equal_shape` and `slice_all`.
- Fix `Split` `derive_properties` to use `n_outputs` from caller when `split_value` and `attr_num_outputs` are both absent.

### DIFF
--- a/src/winml/modelkit/analyze/core/runtime_checker_query.py
+++ b/src/winml/modelkit/analyze/core/runtime_checker_query.py
@@ -724,9 +724,12 @@ def get_query_conditions_for_node(
             update_conditions_(conditions, input_name, is_variadic, is_constant, shape, None)
             conditions[f"{input_name}_is_none"] = False
 
+    conditions["n_outputs"] = len(node.output)
+
     # Try to derive properties, but catch errors for incomplete/invalid model information
     try:
         conditions = runtime_checker_op.derive_properties(conditions)
+        conditions.pop("n_outputs", None)
     except (KeyError, TypeError, IndexError) as e:
         # KeyError: missing required property (e.g., 'input_value', 'input_shape')
         # TypeError: invalid property value (e.g., None when expecting iterable)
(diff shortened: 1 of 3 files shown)


DingmaomaoBJTU:
### Code review

Found 1 issue:

1. **Pattern path injects incorrect `n_outputs` for variadic-output operators like Split**

The node path correctly uses `len(node.output)` (the actual number of output edges on the matched node). The pattern path uses `len(pattern_match.skeleton_match_result.pattern.get_schema().outputs)`, which returns the number of *schema-defined output parameters* — not the number of actual output edges. For the Split operator, the ONNX schema defines a single variadic output parameter, so `get_schema().outputs` always has length 1 regardless of how many ways the tensor is split. Any Split node matched via a pattern will receive `n_outputs = 1`, causing `SplitInputGenerator.derive_properties` to derive `num_outputs = 1` instead of the real value.

https://github.com/microsoft/ModelKit/blob/1114c0c3da63e6970a1880ddff273a6f1c6bfeee/src/winml/modelkit/analyze/core/runtime_checker_query.py#L820-L824

The node path (used in `get_query_conditions_for_node`) may be the right model here — something like the number of matched output edges rather than the schema parameter count.

vortex-captain:
> ### Code review
> Found 1 issue:
> 
> 1. **Pattern path injects incorrect `n_outputs` for variadic-output operators like Split**
> 
> The node path correctly uses `len(node.output)` (the actual number of output edges on the matched node). The pattern path uses `len(pattern_match.skeleton_match_result.pattern.get_schema().outputs)`, which returns the number of _schema-defined output parameters_ — not the number of actual output edges. For the Split operator, the ONNX schema defines a single variadic output parameter, so `get_schema().outputs` always has length 1 regardless of how many ways the tensor is split. Any Split node matched via a pattern will receive `n_outputs = 1`, causing `SplitInputGenerator.derive_properties` to derive `num_outputs = 1` instead of the real value.
> 
> https://github.com/microsoft/ModelKit/blob/1114c0c3da63e6970a1880ddff273a6f1c6bfeee/src/winml/modelkit/analyze/core/runtime_checker_query.py#L820-L824
> 
> The node path (used in `get_query_conditions_for_node`) may be the right model here — something like the number of matched output edges rather than the schema parameter count.
> 
> 

Matching subgraph patterns with variadic outputs is out of scope of pattern matcher, as it complicates things tremendously and has no practical use cases for now. The current `conditions["n_outputs"]` works properly in the current scope.

DingmaomaoBJTU (approved):

PR #249, opened 2026-04-07
Optimize analyze performance.

chinazhangchao:
image

1. pattern/utils.py — make_hashable fast-path for sequences of simple types (avoids 57M unnecessary recursive calls)
2. analyze/core/runtime_checker_query.py — Replaced pandas .apply(make_hashable) with direct list comprehension over numpy array, bypassing pandas per-cell dispatch overhead (27s → ~2s)
3. analyze/core/doc_constraint_checker.py — Added skip_shape_inference parameter to avoid redundant infer_shapes() call (~3.3s saved)
4. analyze/core/information_engine.py + analyze/analyzer.py + analyze/core/runtime_checker.py — Pass shape-inferred model proto from RuntimeCheckerQuery to DocConstraintChecker
5. export/init.py — Lazy-load io and pytorch submodules (avoids importing optimum.exporters.tasks + transformers at startup, ~29s)
6. compiler/init.py — Lazy-load Compiler, stages (avoids importing session → torch, ~4s)
7. core/init.py — Lazy-load generate_dummy_inputs_from_specs (avoids importing torch, ~3.6s)
8. config/build.py — Lazy-import ONNXConfigNotFoundError (avoids importing export/io.py at module load)
9. analyze/utils/json_utils.py — Lazy-import jsonschema (~3.5s saved)


src/winml/modelkit/export/__init__.py
```diff
@@ -17,15 +17,31 @@
...
+        "resolve_io_specs",
+    }
+    if name in _io_names:
+        from . import io
+
+        return getattr(io, name)
+
+    _pytorch_names = {"export_pytorch", "export_onnx"}
+    if name in _pytorch_names:
+        from .pytorch import export_pytorch
```
timenick, line 40: Missing `globals()` caching in `__getattr__` — unlike `compiler/__init__.py` and `core/__init__.py` which store resolved symbols via `globals().update(...)` / `globals()[name] = ...`, this `__getattr__` just returns without caching. That means every access to e.g. `export.export_pytorch` will re-trigger `__getattr__` and the import lookup, defeating much of the lazy-load benefit.

Suggested fix: cache into `globals()` before returning, e.g.:
```python
if name in _io_names:
    from . import io
    val = getattr(io, name)
    globals()[name] = val
    return val
```

🤖 Generated with Claude Code

DingmaomaoBJTU:
### Code review

Found 1 issue:

1. `export/__init__.py`'s `__getattr__` never caches loaded symbols back into the module's `__dict__` via `globals().update(...)`, so every subsequent attribute access re-triggers `__getattr__` instead of hitting the module dict directly. The same PR's `compiler/__init__.py` and `core/__init__.py` both call `globals().update(...)` / `globals()[name] = val` after loading to avoid this, but `export/__init__.py` does not.

https://github.com/microsoft/ModelKit/blob/2a263478058c40d3ef4d0cc2494ea9787a5d9dda/src/winml/modelkit/export/__init__.py#L29-L46

For comparison, `compiler/__init__.py` (same PR) correctly caches:

https://github.com/microsoft/ModelKit/blob/2a263478058c40d3ef4d0cc2494ea9787a5d9dda/src/winml/modelkit/compiler/__init__.py#L35-L47

DingmaomaoBJTU (approved):

PR #262, opened 2026-04-08
fix: suppress false-alarm WinMLModel pipeline warning in winml eval (#258)

### LINKED ISSUE #258: bug: winml eval emits misleading warning that WinMLModelForImageClassification is not supported
## Summary

`winml eval` emits a long, alarming warning that `WinMLModelForImageClassification` is not supported for image-classification, listing 50+ supported model classes. Despite this warning, evaluation completes successfully and returns correct results (e.g. 74% accuracy for ResNet-50).

## Repro

```bash
winml eval -m microsoft/resnet-50 --device npu --samples 100
```

## Observed output (excerpt)

```
The model 'WinMLModelForImageClassification' is not supported for image-classification.
Supported models are ['BeitForImageClassification', 'BitForImageClassification', ...,
'ViTMSNForImageClassification'].
```

Followed by a successful evaluation:

```
{'model_id': 'microsoft/resnet-50', 'task': 'image-classification', 'device': 'npu',
 'metrics': {'accuracy': 0.74, 'samples_per_second': 58.04, ...}}
```

## Root Cause

The warning is emitted by HuggingFace's `pipeline()` internals, which do not recognize the `WinMLModelForImageClassification` wrapper class (it is not in HF's built-in registry). The WinML eval stack works correctly regardless, but the warning text implies a fatal unsupported-model condition.

## Expected behavior

The warning should either be:
- Suppressed (since the evaluation proceeds correctly), or
- Replaced with a debug-level log: \"WinML wrapper class not in HF registry — this is expected and does not affect evaluation.\"

## Notes

- Reported by multiple users during bug bash
- Severity: P3 (UX) — evaluation produces correct results; the issue is the misleading log output

DingmaomaoBJTU:
## Summary

- Adds a logging filter on `transformers.pipelines.base` to suppress the misleading `WinMLModel* is not supported for image-classification` message emitted during `winml eval`
- The warning is a false alarm: HuggingFace's `pipeline()` logs an error when the model class is not in its built-in registry, but WinML wrapper classes are intentionally not registered with HF and evaluation works correctly regardless

Closes #258

## Change

[`src/winml/modelkit/_warnings.py`](src/winml/modelkit/_warnings.py) — added `_WinMLPipelineFilter` logging filter following the existing `_DiffusersDistributionFilter` pattern.

### DIFF
--- a/src/winml/modelkit/_warnings.py
+++ b/src/winml/modelkit/_warnings.py
@@ -44,9 +44,20 @@ def _configure() -> None:
         def filter(self, record: logging.LogRecord) -> bool:
             return "Multiple distributions found" not in record.getMessage()
 
-    logging.getLogger("diffusers.utils.import_utils").addFilter(
-        _DiffusersDistributionFilter()
-    )
+    logging.getLogger("diffusers.utils.import_utils").addFilter(_DiffusersDistributionFilter())
+
+    class _WinMLPipelineFilter(logging.Filter):
+        """Filter false-alarm 'WinMLModel* is not supported' from transformers pipeline.
+
+        HuggingFace's pipeline emits an error-level log when the model class is not
+        in its built-in registry. WinML wrapper classes are intentionally not
+        registered with HF; evaluation works correctly regardless.
+        """
+
+        def filter(self, record: logging.LogRecord) -> bool:
+            return "WinMLModel" not in record.getMessage()
+
+    logging.getLogger("transformers.pipelines.base").addFilter(_WinMLPipelineFilter())
 
     # =========================================================================
     # Warning filters (for warnings.warn() calls)
--- a/src/winml/modelkit/commands/hub.py
+++ b/src/winml/modelkit/commands/hub.py
@@ -117,7 +117,7 @@ def _fmt_model_id(model_id: str) -> Text:
     Returns:
         Rich Text with org in dim and model name in cyan bold.
     """
-    t = Text(overflow="ellipsis", no_wrap=True)
+    t = Text(overflow="crop", no_wrap=True)
     if "/" in model_id:
         org, name = model_id.split("/", 1)
         t.append(org + "/", style="dim")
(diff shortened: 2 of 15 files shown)


src/winml/modelkit/_warnings.py
```diff
@@ -44,9 +44,20 @@ class _DiffusersDistributionFilter(logging.Filter):
...
-    )
+    logging.getLogger("diffusers.utils.import_utils").addFilter(_DiffusersDistributionFilter())
+
+    class _WinMLPipelineFilter(logging.Filter):
+        """Filter false-alarm 'WinMLModel* is not supported' from transformers pipeline.
+
+        HuggingFace's pipeline emits an error-level log when the model class is not
+        in its built-in registry. WinML wrapper classes are intentionally not
+        registered with HF; evaluation works correctly regardless.
+        """
```
timenick, line 55: The filter matches any message containing `"WinMLModel"`, which is broader than needed. The existing `_DiffusersDistributionFilter` checks for a very specific substring (`"Multiple distributions found"`), but this one would suppress *any* log from `transformers.pipelines.base` that mentions a WinML model class — not just the "is not supported" false alarm.

A tighter check would be safer:

```python
def filter(self, record: logging.LogRecord) -> bool:
    msg = record.getMessage()
    return not ("WinMLModel" in msg and "is not supported" in msg)
```

🤖 Generated with Claude Code

timenick:
**FYI — remaining `→` (U+2192) in log/print statements**

This PR correctly replaced `→` with `->` in all capability `description=` strings that surface in `--help` / `--list-capabilities`. However there are a couple of remaining spots where `→` appears in runtime output that could hit the same cp1252 encoding crash on Windows terminals:

- `src/winml/modelkit/optim/pipes/surgery.py:203` — `logger.warning()` format string: `"Clamped tensor '%s': [%.2e, %.2e] → [%.2e, %.2e]"`
- `src/winml/modelkit/core/onnx_node_bucketizer.py:114` — `print()` call with `→`

Not blocking for this PR, but might be worth a follow-up sweep.

🤖 Generated with Claude Code

DingmaomaoBJTU:
Fixed in c7e0b27 — replaced `→` with `->` in both spots:
- `surgery.py:203` logger.info format string
- `onnx_node_bucketizer.py:114` print() call

DingmaomaoBJTU:
Closing in favor of a cleaner branch — see replacement PR.

PR #278, opened 2026-04-09
Support MODELKIT_RULES_DIR with env-first runtime rule lookup

fangyangci:
Add MODELKIT_RULES_DIR support for runtime rule zip lookup (multi-path supported).
Change lookup order: env paths first, built-in rules directory second.
Update analyzer warnings, unit tests, and README to match new behavior.
Validation: pytest tests/unit/analyze/models/test_rule_loader.py -q, 22 passed.


src/winml/modelkit/analyze/core/runtime_checker_query.py
```diff
@@ -141,7 +142,8 @@ def _ensure_loaded(self) -> None:
         if not self._zip_path.exists():
             logger.warning(
                 "Rule zip not found: %s. "
-                "Run 'uv run python scripts/download_rules.py' to download rule files.",
+                "Run 'uv run python scripts/download_rules.py' to download rule files, "
```
xieofxie, line 145: @timenick I think we could remove ` scripts/download_rules.py` and move into artifacts repo. because artifacts repo is internal so this file is useless for public users

timenick: This is a temporary solution, as we have not yet decided how to handle these zip files once they go public. We should also make the artifact repository accessible to the Microsoft organization.

src/winml/modelkit/analyze/rules/runtime_check_rules/README.md
```diff
@@ -22,6 +22,24 @@ Use `--force` to re-download all files even if they already exist locally.
 
 Copy all `*.zip` files from [`gim-home/ModelKitArtifacts/op_check_results/rules/`](https://github.com/gim-home/ModelKitArtifacts/tree/main/op_check_results/rules) into this directory.
 
+### Option 3: Use external rules directory via environment variable
+
+Set `MODELKIT_RULES_DIR` to one or more directories containing runtime rule zip files.
+
+- Windows (PowerShell): `$env:MODELKIT_RULES_DIR="D:\\rules;E:\\more_rules"`
+- Linux/macOS (bash/zsh): `export MODELKIT_RULES_DIR="/opt/rules:/data/more_rules"`
```
timenick: Please remove any docstrings to Linux and Mac, as Modelkit only supports the Windows platform.

src/winml/modelkit/analyze/core/runtime_checker_query.py
timenick: When the repo becomes public, external users will download the zip file from the release. Only internal users can access the latest zip from the artifact repository. This script should remain here.

tests/unit/analyze/models/test_rule_loader.py
```diff
@@ -444,110 +445,72 @@ def test_load_real_qc_information_rules(self, mock_data_rules_dir):
...
-        # First alternative: {"OP/ai.onnx/QLinearConv": "QDQ"}
-        assert "OP/ai.onnx/QLinearConv" in alternatives[0]
-        assert alternatives[0]["OP/ai.onnx/QLinearConv"] == "QDQ"
+class TestResolveRuleZipPath:
+    """Test resolve_rule_zip_path and get_runtime_rules_search_dirs."""
 
-        # Second alternative: {"OP/ai.onnx/ConvInteger": "equivalent"}
-        assert "OP/ai.onnx/ConvInteger" in alternatives[1]
-        assert alternatives[1]["OP/ai.onnx/ConvInteger"] == "equivalent"
+    def test_default_search_dir_included(self, monkeypatch):
```
timenick, line 455: Import violation: `get_runtime_rules_search_dirs` and `resolve_rule_zip_path` are exported in `analyze/utils/__init__.py`, so tests should import from the package level:

```python
from winml.modelkit.analyze.utils import get_runtime_rules_search_dirs, resolve_rule_zip_path
```

not from the internal submodule `winml.modelkit.analyze.utils.rule_loader`. Per project convention: "Never reach into internal submodules for symbols exported by `__init__.py`." The `_DEFAULT_RUNTIME_RULES_DIR` import is fine since it is `_`-prefixed.

Also, these imports should be at the module top level rather than inside each test method body.

🤖 Generated with Claude Code

timenick (approved):

PR #282, opened 2026-04-09
fix: move CheckResultWriter to op_utils and fix delta_only input_constraints serialization

xieofxie:
## Summary

- Moves `CheckResultWriter` from `analyze/runtime_checker/check_ops.py` to `analyze/utils/op_utils.py` so it can be imported without triggering `winml.register_execution_providers` (which runs at module level in `check_ops.py` and blocks the test pipeline)
- Fixes `reuse_existing_result`: when delta_only reuses an existing case, `input_constraints` is now updated from the current generator output (which went through `InputValueConstraint.to_dict()`), upgrading old `value: [...]` array format to the compact `same_value` form
- Removes unused imports from `check_ops.py` after the move (`json`, `numpy`, `onnx`, `json_format`, `compute_case_signature`, `hash_case_signature`)
- Adds `TestReuseExistingResultInputConstraints` unit tests covering the serialization upgrade


src/winml/modelkit/analyze/utils/op_utils.py
```diff
@@ -95,3 +96,321 @@ def _is_empty_top_level(value: Any) -> bool:
 def hash_case_signature(signature: str) -> str:
     """Return a stable hash value for a case signature."""
     return hashlib.sha256(signature.encode("utf-8")).hexdigest()
+
+
+class CheckResultWriter:
```
timenick, line 102: `CheckResultWriter` is a new public class imported by `check_ops.py` and tests, but it isn't exported in `analyze/utils/__init__.py`.

Per CLAUDE.md: *"Every symbol that external code needs must be exported in the package's `__init__.py`"* and *"When adding new public classes/functions, add them to both the import and `__all__`"*.

Please add `CheckResultWriter` to `src/winml/modelkit/analyze/utils/__init__.py`.

Once exported, the import in `check_ops.py` should also go through `__init__.py`:
```python
from ..utils import CheckResultWriter
```
instead of the current:
```python
from ..utils.op_utils import CheckResultWriter
```

🤖 Generated with Claude Code

tests/unit/analyze/test_check_ops.py
```diff
@@ -144,3 +144,103 @@ def test_empty_attrs_excluded_from_signature(self) -> None:
...
         )
+
+
+class TestReuseExistingResultInputConstraints:
+    """Tests that reuse_existing_result upgrades input_constraints to current format."""
+
+    @staticmethod
+    def _make_writer(existing_cases: list[dict], tmp_path):
+        """Create a CheckResultWriter with pre-loaded existing_signatures from cases."""
+        import json
```
timenick: `import json` and the `CheckResultWriter` import (line 157) are inside the method body, but the same module (`op_utils`) is already imported at the file's top level (line 9). Since the whole point of this PR is that `op_utils` is now free of module-level side effects, there's no reason to lazy-import. Both should be moved to the top of the file:

```python
import json
import numpy as np

from winml.modelkit.analyze.utils.op_utils import CheckResultWriter, compute_case_signature
from winml.modelkit.pattern.op_input_gen import InputValueConstraint
```

🤖 Generated with Claude Code

src/winml/modelkit/analyze/runtime_checker/check_ops.py
```diff
@@ -36,7 +32,7 @@
 from ...sysinfo import SysInfo
 from ...utils import constants
 from ..utils.model_utils import get_op_since_version
-from ..utils.op_utils import compute_case_signature, hash_case_signature
+from ..utils.op_utils import CheckResultWriter
```
timenick: Per CLAUDE.md, source code should use relative imports through `__init__.py`, not directly into submodules:

```python
from ..utils import CheckResultWriter
```

instead of:
```python
from ..utils.op_utils import CheckResultWriter
```

This requires `CheckResultWriter` to be exported in `analyze/utils/__init__.py` first (see other comment on `op_utils.py`).

🤖 Generated with Claude Code

fangyangci (approved):

src/winml/modelkit/analyze/utils/op_utils.py
```diff
@@ -95,3 +97,321 @@ def _is_empty_top_level(value: Any) -> bool:
...
+            if hash_case_signature(sig) not in self._filter_case_index_set:
+                return False
+
+        existing_case = self.existing_signatures.get(sig)
+        if existing_case:
+            self.used_signatures.add(sig)
+            if sig in self.output_signatures:
+                self.duplicate_skipped_count += 1
+                return False  # duplicate reuse should not count as reused
+            # Re-apply current serialization to input_constraints so reused cases
```
xieofxie, line 252: Update here

PR #312, opened 2026-04-10
Save op columns to json files.

chinazhangchao:


src/winml/modelkit/analyze/core/runtime_checker_query.py
```diff
@@ -178,6 +236,26 @@ def get(self, key: str, default: pd.DataFrame | None = None) -> pd.DataFrame | N
...
             return default
 
+    def get_columns(self, key: str) -> list[str] | None:
+        """Get ordered table column names for an operator.
+
+        Prefers explicit metadata from columns_file_name. Falls back to deriving
+        column names from the raw table payload when metadata is not available.
+        """
+        if key in self._loaded_tables:
+            return self._loaded_tables[key].columns.to_list()
```
DingmaomaoBJTU: **Bug:** When column metadata is absent (e.g. older rule zips without `*_table_columns.json`), this fallback returns all columns **including `compile_run_success`**. The same issue applies to line 255 (`list(raw_table.keys())`).

Downstream, `_build_table_filter_conditions` iterates these column names looking each up in `conditions`. Since `compile_run_success` is never in `conditions`, it raises `OpOptionalInputSupportError` — turning every table match into a false "unsupported" result for any zip that predates the new column files.

The old code explicitly handled this:
```python
match_keys.remove("compile_run_success")
```

Suggested fix — filter in `get_columns` so every caller is safe:
```python
def get_columns(self, key: str) -> list[str] | None:
    ...
    if key in self._loaded_tables:
        return [c for c in self._loaded_tables[key].columns if c != "compile_run_success"]
    ...
    if isinstance(raw_table, dict):
        return [c for c in raw_table if c != "compile_run_success"]
```

src/winml/modelkit/analyze/core/runtime_checker_query.py
```diff
@@ -2008,7 +2163,21 @@ def run_for_subgraph(
...
+                table_filter_conditions, _ = get_query_conditions_for_pattern(
+                    pattern_match,
+                    pattern_name,
+                    self.opset_versions,
+                    dynamic_axis_strict_mode=self.dynamic_axis_strict_mode,
+                )
+            except Exception:
+                return self._run_for_subgraph_per_node(pattern_match, pattern_name, run_unknown_op)
+
+            return self._run_for_subgraph_per_node(
```
DingmaomaoBJTU, line 2167: **Dead code:** `table_filter_conditions` is computed but never used. Both the success path (line 2176) and the except path (line 2174) call `_run_for_subgraph_per_node` with the same arguments, and neither passes `table_filter_conditions`.

This looks like incomplete scaffolding — was the intent to pass it through to the per-node call? If not needed, the try/except can be removed to avoid confusion.

src/winml/modelkit/analyze/core/runtime_checker_query.py
```diff
@@ -151,6 +191,24 @@ def _ensure_loaded(self) -> None:
                     self._raw_data = json.loads(zf.read(self._file_name).decode("utf-8"))
                 else:
                     logger.debug(f"Table file not found in zip: {self._file_name}")
+
+                if self._columns_file_name:
```
DingmaomaoBJTU: **Nit — duplicate logic:** This column-parsing block in `_ensure_loaded` is nearly identical to the standalone `_load_table_columns_file` (line 118). Both read the same `columns_file_name` from the same zip with the same validation/coercion. As a result, every column file is read and parsed **twice** — once eagerly via `_load_table_columns_file` into `self.df_table_columns`, and once lazily here into `self._raw_columns`.

Consider having one canonical source. For example, remove this block and let `get_columns` fall back to the eagerly-loaded `df_table_columns` dict when `_raw_columns` is empty, or vice versa.

src/winml/modelkit/analyze/runtime_checker/result_processor.py
```diff
@@ -797,17 +822,40 @@ def get_opset_version_range(op_name: str, start_opset_version: int, op_domain: s
...
+            is_qdq,
+        ), op_tables in tables_by_ep_domain_opset.items():
+            domain_str = op_domain if op_domain else "ai.onnx"
+            qdq_suffix = "_qdq" if is_qdq else ""
+            output_file = output_dir / (
+                f"{ep_name}_{device}_{domain_str}"
+                f"_opset{opset_version}_table_columns{qdq_suffix}.json"
+            )
+
+            op_columns = _extract_op_table_columns(op_tables)
```
DingmaomaoBJTU: **Minor:** `op_columns` is already being collected into `table_columns_by_ep_domain_opset` during the main loop (line 718: `table_columns_by_ep_domain_opset[key][op_name] = df.columns.to_list()`). Here you re-derive it by calling `_extract_op_table_columns(op_tables)` over the table data.

You could iterate `table_columns_by_ep_domain_opset` directly instead, avoiding the redundant re-extraction. Just note that the main loop stores the raw column list (including `compile_run_success`) while `_extract_op_table_columns` filters it out — so you'd need to apply the filter at collection time.

src/winml/modelkit/analyze/runtime_checker/result_processor.py
```diff
@@ -797,17 +822,40 @@ def get_opset_version_range(op_name: str, start_opset_version: int, op_domain: s
...
-            f"and {len(tables_by_ep_domain_opset)} table file(s)."
+            f"and {len(tables_by_ep_domain_opset)} table file(s), "
+            f"plus {len(table_columns_by_ep_domain_opset)} table-column file(s)."
         )
 
         if args.update_zip:
             for group_name, file_list in zip_group.items():
-                rule_zip_path = Path(__file__).parent.joinpath(
-                    f"../rules/runtime_check_rules/{group_name}_{domain_str_for_filename}_opset{current_opset_version}.zip"
+                rule_zip_path = input_dir.parent.joinpath(
```
DingmaomaoBJTU, line 839: This changes the zip path resolution from `Path(__file__).parent / "../rules/runtime_check_rules/..."` to `input_dir.parent / "rules/..."`. That's a behavioral change unrelated to column saving — worth calling out in the PR description (or splitting into its own commit) so reviewers know it's intentional.

src/winml/modelkit/analyze/core/runtime_checker_query.py
```diff
@@ -315,6 +393,27 @@ def _format_list_preview(items: Any, max_items: int = 10) -> list[Any]:
     return lst
 
 
+def _build_table_filter_conditions(
```
DingmaomaoBJTU, line 401: **Missing test coverage:** This new helper has non-trivial behavior (raises `OpOptionalInputSupportError` when a column is missing from conditions). It would benefit from direct unit tests, especially for:
- Happy path (all columns present in conditions)
- Column missing → raises with descriptive message
- Columns in `infinite_properties` are skipped
- Empty `column_names` → returns empty dict

DingmaomaoBJTU (commented):
Thanks for the PR! The column metadata pre-loading is a nice optimization. I have one potential bug and a few suggestions — see inline comments.

src/winml/modelkit/analyze/runtime_checker/result_processor.py
chinazhangchao: zip file has been moved to artifacts repo. So just generate it to artifacts directory.

src/winml/modelkit/analyze/core/runtime_checker_query.py
chinazhangchao: Add https://github.com/microsoft/ModelKit/pull/312/changes#diff-a7ebe26ee87a1f326113d89f6a0737a20062f7428242540f2605506f286e08ad

src/winml/modelkit/analyze/runtime_checker/result_processor.py
timenick: Should we move this logic to the artifact repo, since the repo is accessible only internally?

chinazhangchao: Good idea. Maybe in another pr.

vortex-captain (approved):

PR #316, opened 2026-04-13
Add -C/--config option to analyze, compile, eval, export, optimize, perf, quantize commands

ssss141414:
Title: Add -C/--config option to analyze, compile, eval, export, optimize, perf, quantize commands

Body:

## Summary

Add optional `-C`/`--config` flag to seven CLI commands so they can read default parameter values from a `WinMLBuildConfig` JSON file (the same file produced by `winml config` and consumed by `winml build`).

## Requirements

1. `-C` is **optional** — every command works exactly as before without it.
2. Each command reads only the config sections relevant to its parameters.
3. If the user also passes an explicit CLI flag, the **explicit flag wins** (checked via `click.Context.get_parameter_source`).

### Shared infrastructure — `utils/cli.py`

| Addition | Purpose |
|---|---|
| `config_option` | Reusable `-C`/`--config` Click decorator |
| `load_config()` | Load + validate a `WinMLBuildConfig` JSON file |
| `is_cli_provided()` | Detect whether a parameter was explicitly passed on the command line |

### Per-command config mapping

| Command | Parameters read from config |
|---|---|
| **analyze** | `ep` |
| **compile** | `ep`, `compiler`, `embed`, `validate`, `verbose` |
| **eval** | `task`, `samples`, `dataset_name` |
| **export** | full export dict as base layer, `task`, `dynamo`, `no_hierarchy` |
| **optimize** | all optimization capability flags (inserted between defaults and presets in the precedence chain) |
| **perf** | `task`, `ep` |
| **quantize** | `samples`, `method`, `weight_type`, `activation_type`, `per_channel`, `symmetric` |

## Testing

All 198 command unit tests pass. All 3819 unit tests pass (75 skipped for hardware requirements).


src/winml/modelkit/utils/cli.py
```diff
@@ -110,3 +118,60 @@ def verbosity_options(f):
         help="Enable verbose logging to stderr",
     )(f)
     return f  # noqa: RET504
+
+
+def config_option(func):
+    """Add -C/--config option for WinMLBuildConfig JSON file."""
+    return click.option(
+        "-C",
+        "--config",
```
DingmaomaoBJTU, line 127: **Bug: `--config` name collision with `optimize`'s existing `--config`.**

The `optimize` command already defines `--config` / `-c` (line 214-219 in optimize.py) for its YAML/JSON optimization config file. Adding another `--config` here (via `-C/--config`) creates two Click options claiming the same `--config` long-form flag on the same command. This will cause a Click error or undefined behavior at runtime.

Consider using a different long name to avoid the collision, e.g. `--build-config` or `--wmk-config`:
```python
click.option(
    "-C",
    "--build-config",
    "config_file",
    ...
)
```

src/winml/modelkit/commands/compile.py
```diff
@@ -148,6 +151,22 @@ def compile(
...
+            cc = build_cfg.compile
+            if not cli_utils.is_cli_provided(ctx, "ep"):
+                ep = ep or cc.ep_config.provider
+            if not cli_utils.is_cli_provided(ctx, "compiler"):
+                compiler = cc.ep_config.compiler
+            if not cli_utils.is_cli_provided(ctx, "embed"):
+                embed = cc.ep_config.embed_context
+            if not cli_utils.is_cli_provided(ctx, "validate"):
+                validate = cc.validate
+            if not cli_utils.is_cli_provided(ctx, "verbose"):
```
DingmaomaoBJTU, line 167: **Bug: `verbose` override comes after `configure_logging(verbose=verbose)` on line 152.**

If the config file sets `verbose: true` but the user doesn't pass `--verbose`, logging is already configured with `verbose=False` before this override runs. The reassignment on line 168 updates the local variable but doesn't reconfigure logging.

Fix: move the config-application block **before** `configure_logging`, or call `configure_logging` again after the override.

src/winml/modelkit/commands/export.py
```diff
@@ -265,12 +282,17 @@ def export(
...
-        **export_config_dict,
-    }
+    config_kwargs = {}
+    # Layer 1: build config defaults (lowest precedence)
+    if _build_export_cfg is not None:
+        config_kwargs.update(_build_export_cfg.to_dict())
+    # Layer 2: --export-config file overrides
+    config_kwargs.update(export_config_dict)
+    # Layer 3: explicit CLI options (highest precedence)
+    config_kwargs["enable_hierarchy_tags"] = not no_hierarchy
```
DingmaomaoBJTU, line 292: **Bug: Layer 3 unconditionally overwrites `enable_hierarchy_tags` and `verbose`, defeating config defaults.**

These two lines always run, ignoring whether the user actually passed `--no-hierarchy` or `--verbose` on the CLI. If the config file sets `enable_hierarchy_tags: false` and the user doesn't pass `--no-hierarchy`, the Click default (`no_hierarchy=False`) produces `enable_hierarchy_tags=True` here, overriding the config value.

Contrast with `dynamo` on line 294, which is correctly guarded by `is_cli_provided`. Apply the same pattern:
```python
if cli_utils.is_cli_provided(ctx, "no_hierarchy"):
    config_kwargs["enable_hierarchy_tags"] = not no_hierarchy
if cli_utils.is_cli_provided(ctx, "verbose"):
    config_kwargs["verbose"] = verbose
if cli_utils.is_cli_provided(ctx, "dynamo"):
    config_kwargs["dynamo"] = dynamo
```

src/winml/modelkit/commands/eval.py
```diff
@@ -157,6 +161,17 @@ def eval(
...
+    # Apply build config defaults (CLI explicit options take precedence)
+    if config_file is not None:
+        build_cfg = cli_utils.load_config(config_file)
+        if build_cfg.loader and not cli_utils.is_cli_provided(ctx, "task"):
+            task = task or build_cfg.loader.task
+        if build_cfg.quant:
+            if not cli_utils.is_cli_provided(ctx, "samples"):
+                samples = build_cfg.quant.samples
+            if not cli_utils.is_cli_provided(ctx, "dataset_path"):
+                dataset_path = dataset_path or build_cfg.quant.dataset_name
```
DingmaomaoBJTU, line 173: **Semantic mismatch?** The `--dataset` CLI option (`dataset_path`) is the HF dataset path (e.g. `'imagenet-1k'`, `'glue'`), while `build_cfg.quant.dataset_name` is the dataset *config name* for multi-config datasets (e.g. `'mrpc'`). The eval command has a separate `--dataset-name` parameter for that.

Should this be mapping to `dataset_name` instead? Or is `quant.dataset_name` actually the dataset path despite the field name?

src/winml/modelkit/commands/compile.py
```diff
@@ -148,6 +151,22 @@ def compile(
 
     configure_logging(verbose=verbose)
 
+    # Apply build config defaults (CLI explicit options take precedence)
+    if config_file is not None:
+        build_cfg = cli_utils.load_config(config_file)
+        if build_cfg.compile:
+            cc = build_cfg.compile
+            if not cli_utils.is_cli_provided(ctx, "ep"):
+                ep = ep or cc.ep_config.provider
```
DingmaomaoBJTU, line 160: **Nit — inconsistent override style.** This line uses `ep = ep or cc.ep_config.provider` (with `or` guard), while lines 162-168 use direct assignment (`compiler = cc.ep_config.compiler`). Both are inside an `is_cli_provided` check, so the `or` is redundant when the Click default is `None`.

Not a bug, but the inconsistency makes it harder to reason about precedence. Pick one pattern and use it consistently. I'd suggest always using direct assignment since `is_cli_provided` already gates the block:
```python
if not cli_utils.is_cli_provided(ctx, "ep"):
    ep = cc.ep_config.provider
```

src/winml/modelkit/commands/quantize.py
```diff
@@ -142,6 +145,24 @@ def quantize(
...
     configure_logging(verbose=verbose)
 
+    # Apply build config defaults (CLI explicit options take precedence)
+    if config_file is not None:
+        build_cfg = cli_utils.load_config(config_file)
+        if build_cfg.quant:
+            qc = build_cfg.quant
+            if not cli_utils.is_cli_provided(ctx, "samples"):
+                samples = qc.samples
+            if not cli_utils.is_cli_provided(ctx, "method"):
```
DingmaomaoBJTU, line 155: Same `or`-vs-direct inconsistency as compile.py: `samples` and `method` use direct assignment, while `weight_type` and `activation_type` use `or`. Since all are inside `is_cli_provided` guards, the behavior is the same (Click defaults are `None` for the `or` cases), but the mixed style is confusing.

src/winml/modelkit/commands/optimize.py
```diff
@@ -417,6 +420,13 @@ def optimize(
     for cap_name, cap_def in all_caps.items():
         final_config[cap_name] = cap_def.default
 
+    # 1.5. Apply build config optim section if specified (overrides defaults)
+    if config_file is not None:
+        build_cfg = cli_utils.load_config(config_file)
+        if build_cfg.optim:
+            final_config.update(build_cfg.optim.to_dict())
+            console.print(f"[dim]Applied config optim section: {config_file}[/dim]")
```
DingmaomaoBJTU, line 428: **Nit — step numbering.** Inserting this as step "1.5" between 1 and 2 works but is fragile. Consider renumbering all steps to keep them clean (1 → defaults, 2 → build config, 3 → preset, 4 → config file, 5 → CLI, 6 → auto-deps, 7 → validate).

DingmaomaoBJTU (commented):
Nice work adding config file support across all commands — the shared infra in cli.py is clean. Found a few issues to address before merging, see inline comments.

PR #320, opened 2026-04-13
fix: delete stale external data sidecar before save_onnx overwrites it

xieofxie:
## Summary

- `quantize_onnx` calls ORT's `quantize()` (Step 3) which writes `<output>.onnx.data`, then calls `save_onnx` (Step 8) which tries to write the same sidecar again
- `onnx.save_model` has no overwrite option and raises `FileExistsError: External data file exists in <location>`
- Fix: `save_onnx` now deletes the existing `.data` sidecar before writing, so re-saving over an ORT-generated file works cleanly

### DIFF
--- a/src/winml/modelkit/onnx/persistence.py
+++ b/src/winml/modelkit/onnx/persistence.py
@@ -106,6 +106,12 @@ def save_onnx(
 
     if save_external:
         ext_location = location or f"{path.name}.data"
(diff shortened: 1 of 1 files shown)


xieofxie:
Found when trying quantize

fangyangci (approved):

DingmaomaoBJTU (approved):

PR #330, opened 2026-04-14
fix(perf): op-tracing preflight check before benchmark run

xieofxie:
## Summary

- Move `is_qnn_profiling_available()` check to **before** `PerfBenchmark.run()` so `wmk perf --op-tracing` fails fast instead of wasting a full benchmark run when `onnxruntime-qnn` is not installed.

## Test plan

- [ ] `TestOpTracingPreflight::test_benchmark_does_not_run_when_qnn_unavailable` — verifies `PerfBenchmark.run()` is never called when QNN profiling is unavailable and exit code is non-zero

Extracted from microsoft/ModelKit#209 (Bug A only).


tezheng (approved):
LGTM. The preflight check is correctly placed at the try-block level (covers both ONNX and HF paths) before `benchmark.run()`. Old duplicate check properly removed. Ruff clean.

Minor note (non-blocking):
- Could add an ONNX-path test to complement the HF-path test, but not a blocker.

xieofxie:
dup of https://github.com/microsoft/WinML-ModelKit/pull/397

PR #333, opened 2026-04-14
Fix sam models perf issue.

chinazhangchao:


src/winml/modelkit/models/hf/sam.py
```diff
@@ -45,7 +45,7 @@
     DummyInputGenerator,
     DummyVisionInputGenerator,
 )
-from transformers import Sam2Model
+from transformers import Sam2Model, SamModel
```
timenick, line 48: nit: Now that this file covers both SAM v1 and SAM2, the module docstring (lines 5-32) should be updated. It still says "SAM2 HuggingFace Model Patches and Export Configs" with no mention of SAM v1.

🤖 Generated with Claude Code

src/winml/modelkit/models/hf/sam.py
```diff
@@ -812,13 +926,105 @@ def outputs(self) -> dict[str, dict[int, str]]:
...
+        - mask_input:       {0: "batch_size"} [B, 1, 256, 256]
+        - use_mask_input:   {0: "batch_size"} [B]
+
+    Outputs:
+        - masks:          {0: "batch_size"} [B, 3, 1024, 1024]
+        - iou_scores:     {0: "batch_size"} [B, 3]
+        - low_res_masks:  {0: "batch_size"} [B, 3, 256, 256]
+    """
+
+    NORMALIZED_CONFIG_CLASS = Sam2NormalizedVisionConfig
```
timenick, line 995: nit: `NORMALIZED_CONFIG_CLASS = Sam2NormalizedVisionConfig` reuses a SAM2-named config for SAM v1. Both architectures use 1024 image_size so this works functionally, but the naming is misleading. Consider renaming to a shared name (e.g., `SamNormalizedVisionConfig`) or adding a comment noting the intentional cross-version reuse.

🤖 Generated with Claude Code

src/winml/modelkit/models/hf/sam.py
```diff
@@ -812,13 +926,105 @@ def outputs(self) -> dict[str, dict[int, str]]:
...
+    DUMMY_INPUT_GENERATOR_CLASSES = (
```
timenick, line 998: nit: `DUMMY_INPUT_GENERATOR_CLASSES` reuses `Sam2PointsInputGenerator` and `Sam2MaskInputGenerator`, whose docstrings explicitly reference "SAM2 decoder." Functionally correct since the tensor shapes match, but could confuse future readers. Consider either renaming to shared generator names or adding a comment noting the cross-version compatibility.

🤖 Generated with Claude Code

src/winml/modelkit/models/hf/sam.py
```diff
@@ -812,13 +926,105 @@ def outputs(self) -> dict[str, dict[int, str]]:
...
+        self.batch_size = batch_size
+
+    def generate(
+        self,
+        input_name: str,
+        framework: str = "pt",
+        int_dtype: str = "int64",
+        float_dtype: str = "fp32",
+    ):
+        shape = [self.batch_size, 256, 64, 64]
```
timenick, line 960: nit: Hardcodes `[B, 256, 64, 64]` dimensions rather than reading from `normalized_config`. This is consistent with the existing `Sam2EmbeddingsInputGenerator` pattern, but worth noting as a deviation from the project's "No Hardcoded Logic" cardinal rule. Could read `hidden_size` and `image_embedding_size` from config if those fields are reliably available.

🤖 Generated with Claude Code

timenick (commented):
Review comments below.

vortex-captain (approved):

PR #334, opened 2026-04-14
feat: multi-model support with KV cache (T5, Qwen, Mu2)

vortex-captain:
## Summary

Multi-model composite pipeline: build, optimize, and run inference on models composed of multiple ONNX sub-components (encoder+decoder, prefill+gen) through a unified `WinMLAutoModel` interface.

### Composite model framework
- `WinMLCompositeModel` base class with `_SUB_MODEL_CONFIG` mapping component names to HF tasks
- `@register_composite_model(model_type, task)` registry for `wmk config` multi-config generation
- `from_pretrained()` builds all sub-components via `WinMLAutoModel` with per-component `sub_model_kwargs`
- `from_onnx(dict)` loads pre-built ONNX files, resolves concrete class from registry

### Polymorphic KV cache (`WinMLCache` hierarchy)
- `WinMLStaticCache` (index_copy_, left-aligned mask) — T5, learned position bias models
- `WinMLSlidingWindowCache` (FIFO slice+concat, right-aligned mask) — Qwen3, Mu2, RoPE models
- Shared interface: `build_decoder_mask(num_new_tokens)`, `prepare_prefill_chunk()`, `update_all_layers()`
- Decoder-only model is fully cache-agnostic — padding, mask, and cache update all delegated to cache class
- Both cache types verified with e2e generation on Qwen3

### Models
- **T5 encoder-decoder** (`WinMLT5Model`): registered for `translation` and `summarization` tasks
- **Qwen3 decoder-only** (`WinMLQwen3Model`): prefill/gen split with chunked prefill, `WinMLSlidingWindowCache`
- **Mu2 encoder-decoder** (`WinMLMu2Model`): custom `trust_remote_code` model with `WinMLSlidingWindowCache`

### WinMLAutoModel routing
- `from_pretrained()`: checks `PIPELINE_MODEL_REGISTRY` before config phase, delegates to `WinMLCompositeModel.from_pretrained` for composite models
- `from_onnx(dict)`: delegates to `WinMLCompositeModel.from_onnx` for dict `onnx_path`
- Circular import fix: `WinMLCompositeModel` import moved under `TYPE_CHECKING` + lazy import

### Optimization
- `remove_isnan_in_attention_mask` surgery: removes dead `Softmax→IsNaN→Where` NaN guard patterns
- Build configs for T5, Qwen3, Mu2: `gelu_fusion`, `fuse_rmsnorm`, `matmul_add_fusion`, `clamp_constant_values`, `remove_isnan_in_attention_mask`

### run_eval.py composite support
- Generalized `_run_build`: `winml config` produces list of config JSONs (1 for single, N for composite), common build loop
- Generalized `run_model` (perf): takes list of ONNX paths, runs perf for each, merges results
- Fixed `_extract_onnx_path` to match `"Artifact:"` and `"Existing artifact found:"` markers

### Other
- `trust_remote_code` forwarded through CLI, `WinMLAutoModel`, `WinMLCompositeModel`
- `session_options` passthrough from `WinMLAutoModel.from_onnx` to `WinMLSession`
- Architecture-agnostic KV dtype from ONNX metadata (supports fp16 models)
- DepthPro ONNX registration enabled

## Verified

- **T5-small translation**: exact match vs PyTorch (`Bonjour, comment êtes-vous ?`) — build + e2e generation
- **T5-small summarization**: correct summary generation with `sequence_length=512` — build + e2e generation + `run_eval` perf PASS
- **Qwen3-0.6B text generation**: correct (`8×7=56`, `9+6=15`) with both `WinMLSlidingWindowCache` and `WinMLStaticCache` — build + e2e generation. `run_eval` perf skipped (OOM segfault in quantization stage)
- **Mu2 translation**: correct on all 6 queries (EN/CN/ES/DE/FR) with both cache types, streaming generation — build + e2e generation
- **resnet-50**: `run_eval` perf PASS (no regression for single-model path)
- **Unit tests**: 15 new tests for KV cache methods and composite `from_onnx` dispatch (79 total in modified files, all passing)


docs/design/pipeline-model.md
```diff
@@ -0,0 +1,133 @@
+# Multi-Model Pipeline Design
```
timenick: This document should not be checked in.

vortex-captain: This PR is currently a draft and will be cleaned up

scripts/e2e_eval/run_eval.py
```diff
@@ -378,59 +382,91 @@ def _run_build(
...
+    last_proc = config_proc
 
-    # Extract ONNX path from build output
-    # winml build prints "Final artifact: <path>" in stderr
-    onnx_path = None
-    for line in build_proc["stderr"].splitlines():
-        if "Final artifact:" in line:
-            onnx_path = line.split("Final artifact:")[-1].strip()
-            break
+    for sub_cfg in sub_configs:
```
zhenchaoni, line 412: Once `winml build` supports building a model ID into multiple onnx files, then we don't need the loop here.

src/winml/modelkit/commands/config.py
```diff
@@ -328,6 +328,31 @@ def config(
         else:
             _is_onnx_mode = False
 
+            # Check pipeline model registry: (model_type, task) → multi-config
```
zhenchaoni: Should we keep the naming consistent? I think this should be called `composite model`

src/winml/modelkit/models/winml/composite_model.py
```diff
@@ -0,0 +1,276 @@
...
+Per-component kwargs
+--------------------
+``sub_model_kwargs`` in ``from_pretrained`` allows different ``shape_config``
+per sub-model (e.g., different ``max_cache_len`` for prefill vs gen)::
+
+    WinMLCompositeModel.from_pretrained(model_id, task="text-generation",
+        sub_model_kwargs={
+            "decoder_prefill": {"shape_config": {"max_cache_len": 256, "seq_len": 64}},
+            "decoder_gen":     {"shape_config": {"max_cache_len": 256, "seq_len": 1}},
+        })
```
zhenchaoni, line 32: Can these configs be loaded from huggingface? If not, I think it is a bit hard for the user to construct this kwargs.

src/winml/modelkit/models/winml/composite_model.py
```diff
@@ -0,0 +1,276 @@
...
+            sub_models[name] = WinMLAutoModel.from_pretrained(
+                model_id,
+                task=component_task,
+                device=device,
+                use_cache=use_cache,
+                force_rebuild=force_rebuild,
+                **merged,
+            )
+
+        return cls(sub_models=sub_models, config=hf_config)
```
zhenchaoni, line 191: I think the instance creation should be delegated to the resolved_cls which is the concrete class. This is also "from_onnx" implemented. Can you help me to understand why we fallback to `cls` for instance creation?

src/winml/modelkit/models/winml/composite_model.py
```diff
@@ -0,0 +1,276 @@
...
+            component_task = resolved_cls._SUB_MODEL_CONFIG.get(name)
+            merged = {**kwargs, "task": component_task, **sub_model_kwargs.get(name, {})}
+            sub_models[name] = WinMLAutoModel.from_onnx(Path(path), **merged)
+
+        return resolved_cls(sub_models=sub_models, config=hf_config)
+
+    @property
+    def device(self) -> torch.device:
+        """Device (CPU — ORT handles actual placement)."""
+        return torch.device("cpu")
```
zhenchaoni, line 258: This is a known issue that we need to hardcode device to be string type and to be "cpu". I suggest to do this in the base model class and add a `ort_device` property. Please see #189

src/winml/modelkit/models/winml/composite_model.py
```diff
@@ -0,0 +1,276 @@
...
+    def __call__(self, **kwargs: Any) -> Any:
+        """Inference entry point."""
+        return self.forward(**kwargs)
+
+    def forward(self, **kwargs: Any) -> Any:
+        """Subclasses implement task-specific logic."""
+        raise NotImplementedError
+
+    @staticmethod
+    def _pad_inputs(
```
zhenchaoni: Is the pad_input logic generic enough for all models? If so, then I think even the single-onnx models can also have the same padding logic.

src/winml/modelkit/models/winml/composite_model.py
vortex-captain: take qwen3-0.6b for example: https://huggingface.co/Qwen/Qwen3-0.6B/blob/main/config.json
Since hf pipeline mainly focuses on CPU and GPU inference, making shape static is not a main concern to them. As a result, the default config does not reflect all the parameters we need (context length, prefill length, generation length). Loading these params via HF would require user to pass customized shape config, similar to current approach

src/winml/modelkit/models/winml/composite_model.py
vortex-captain: This is the case where `not cls._SUB_MODEL_CONFIG` is False, meaning `cls` is a concrete subclass (like `WinMLT5Model`) with `_SUB_MODEL_CONFIG` defined which can be instantiated

zhenchaoni:
@vortex-captain Below are generated comments. They focus more on the generation mixin part and the composite model part of this PR. You can take as a reference.


# Review of PR #334 — Encoder-Decoder & Composite Model
---

## 🔴 Critical Issues

### C1. Beam search will silently break — `past_key_values` lacks `reorder_cache`

`forward()` returns `Seq2SeqLMOutput(past_key_values=cache)`. HF `GenerationMixin` calls `past_key_values.reorder_cache(beam_idx)` (or `_reorder_cache` on the model) during beam search. Looking at `kv_cache.py` there's no `reorder_cache` on `WinMLCache`.

**Impact**: `generate(num_beams=4)` — the most common seq2seq path for T5 translation and BLIP captioning — will either crash or produce wrong results.

**Fix**: Either implement `WinMLCache.reorder_cache(beam_idx)` (index_select on `keys`/`values` along batch dim; for sliding-window, reorder the ring buffer), or explicitly raise when `num_beams > 1` with a clear message. For image-to-text this is a blocker — BLIP captioning benchmarks use `num_beams=3`.

### C2. `prepare_inputs_for_generation` hardcodes `input_ids[:, -1:]` — no prefill path

This is correct for pure single-step KV-cached decode after step 0, but:

- **Decoder-only prefill**: if a user passes a prompt (`input_ids` with seq_len > 1) on step 0, only the last token is fed. The cache is empty, so the model sees one token without prompt context. Silent wrong output.
- **Assisted / speculative decoding**: passes multiple candidate tokens per step — incompatible.
- **Cache re-use across `generate()` calls**: forbidden silently.

Optimum's `ORTModelForSeq2SeqLM.prepare_inputs_for_generation` handles this with:
```python
if past_key_values is not None:
    input_ids = input_ids[:, -1:]
```

**Fix**: Add the `past_key_values is not None` guard so step-0 behaves correctly. Document any intentional single-token-only constraint in the class docstring.

### C3. `WinMLCompositeModel.__init__` doesn't call `super().__init__()`

```python
def __init__(self, sub_models, config) -> None:
    self.sub_models = sub_models
    self.config = config
```

`PreTrainedModel`/`nn.Module.__init__` is never invoked. Consequences:

- `sub_models` is a plain `dict`, not `nn.ModuleDict` → `.modules()`, `.parameters()`, `.state_dict()`, `.eval()`/`.train()` all silently skip sub-components.
- HF's `GenerationMixin` sometimes does `self._modules.get(...)` → `AttributeError` is possible.
- `torch.compile`, `accelerate`, and `pipeline()` introspection will break.

**Fix**:
```python
def __init__(self, sub_models, config):
    torch.nn.Module.__init__(self)  # or super().__init__(config) if PreTrainedModel
    self.sub_models = torch.nn.ModuleDict(sub_models)
    self.config = config
```

---

## 🟡 HF-Contract / Usage-Experience Issues

### U1. `_EncoderWithInputPadding.forward(self, **kwargs)` — no positional args

HF idiom is `encoder(input_ids, attention_mask=...)`. `GenerationMixin._prepare_encoder_decoder_kwargs_for_generation` currently uses `**kwargs` so this works today, but:

- Any downstream code doing `model.get_encoder()(pixel_values)` (which `WinMLModelForImageToText` from our earlier design does) will `TypeError`.
- `pipeline("image-to-text")` calls `get_encoder()(**features)` so it's fine, but the constraint is undocumented.

**Fix**: Accept `*args` too, or document the contract clearly on the wrapper. For image-to-text reuse, we need positional-compatible.

### U2. Wrapping the encoder hides its attributes

`get_encoder()` returns `_EncoderWithInputPadding`, not the raw `WinMLAutoModel`. Common access patterns broken:

- `model.get_encoder().config` → AttributeError
- `model.get_encoder().main_input_name` → AttributeError (this is **critical for image-to-text**; `GenerationMixin._prepare_model_inputs` reads `encoder.main_input_name` to pick between `input_ids` and `pixel_values`)
- `model.get_encoder().device` → AttributeError

**Fix**: Proxy attribute access via `__getattr__`, or set the key attributes (`config`, `main_input_name`, `device`) on the wrapper in `__init__`. Without this, our upcoming `WinMLModelForImageToText` cannot ride on top.

### U3. `forward(input_ids=..., **model_kwargs)` re-runs the encoder as a fallback

```python
if encoder_outputs is None and input_ids is not None:
    encoder_outputs = self._encoder(input_ids=input_ids, **model_kwargs)
```

This is a footgun: HF `generate()` always pre-computes `encoder_outputs`. If a user calls `model(input_ids=...)` directly (e.g., for loss computation / debugging), `**model_kwargs` contains decoder-only keys (`decoder_input_ids`, `decoder_attention_mask`, `past_key_values`, `position_id`...) which the encoder wrapper will quietly drop via `_pad_inputs`. Works, but debugging why outputs are wrong is painful.

**Fix**: Split encoder kwargs vs decoder kwargs explicitly, or simply document that direct `forward(input_ids=...)` is not supported and raise.

### U4. `encoder_outputs["last_hidden_state"]` fails when HF passes a tuple

HF `generate()` sometimes wraps/unwraps to `tuple` (e.g., when `return_dict_in_generate=False`). Safer:
```python
enc_h = encoder_outputs[0] if isinstance(encoder_outputs, tuple) else encoder_outputs.last_hidden_state
```

### U5. Cross-attention mask is ambiguous

`feeds` passes through `model_kwargs` which contains `attention_mask` (encoder-side per HF convention). If the decoder ONNX expects a separate `encoder_attention_mask`, it'll be dropped silently. If the ONNX name is `attention_mask`, this conflicts with the decoder self-attention mask naming. Document the contract and consider renaming to match common HF exports (`encoder_attention_mask` for cross-attn). For padded encoder sequences (BLIP text encoder, T5 prompt), missing this produces subtly wrong outputs.

### U6. `_resolve_cache` silently discards user-passed cache state

If a user passes a `DynamicCache` from a prior `generate()` call, `_resolve_cache` creates a fresh one — their state is gone. For continued generation (chat loops), this is wrong. Either copy state across or raise.

### U7. `_supports_cache_class = False` disables `cache_implementation=` kwarg

Users can't do `generate(cache_implementation="static", max_cache_len=256)` to control cache sizing at runtime — they must pass `shape_config` at build time. OK for WinML's static-shape world, but document it clearly and reject the kwarg with a helpful message instead of silently ignoring.

---

## 🟡 Composite Model / Usability

### U8. `to()` and `device` silently lie

```python
def to(self, *args, **kwargs): return self
```

HF `pipeline(device=0)` and `model.to("cuda")` become silent no-ops. ORT picks its own EP regardless, but users think they're on GPU. Recommend logging a warning:
```python
logger.warning("WinMLCompositeModel uses ORT EPs; .to(%r) is a no-op. Configure providers via session_options.", args)
```

### U9. `_pad_inputs` silently drops oversized inputs by producing `pad=[0,0,...]`

```python
deficit = expected_shape[dim] - val.shape[dim]
pad.extend([0, max(deficit, 0)])
```

If `val.shape[dim] > expected_shape[dim]` (user ran `generate(max_new_tokens=1024)` on a model built for 256), no padding is applied — ORT then errors deep inside the session with a cryptic shape-mismatch. **Fix**: Raise explicitly with the offending name and shapes. Huge UX win for the common "I didn't set max_cache_len correctly" mistake.

### U10. `_pad_inputs` pads `decoder_input_ids` with zeros

Token 0 is usually `<pad>` in T5 but a valid token elsewhere. If a decoder ONNX expects `[B, N]` (non-cached prefill), padding one token with zeros corrupts the sequence unless `position_id`/`cache_position` tells the model where the real token lives. Document this assumption or expose a per-input pad-value. For image-to-text where the decoder might be exported with a multi-token prefill shape, this matters.

### U11. Registry key `(model_type, task)` is too narrow

T5 has `translation`, `summarization`, `text2text-generation`. Registering one tuple forces users to use the exact string. Either:
- Normalize tasks to a family (`_TASK_FAMILY = {"translation": "text2text", "summarization": "text2text", ...}`), or
- Allow `register_composite_model("t5", ("translation", "summarization", "text2text-generation"))`.

Otherwise users hit `No pipeline model registered for ('t5', 'summarization')` and have to read source to discover the magic string.

### U12. `from_onnx(onnx_path: dict[str, str])` has no key validation

If a user typos `"encder"` instead of `"encoder"`, the sub_model is silently missing and `__init__` crashes at `sub_models["encoder"]` with `KeyError`. Validate keys against `resolved_cls._SUB_MODEL_CONFIG` and raise a helpful error listing the expected names.

### U13. `from_pretrained` signature drops HF's `pretrained_model_name_or_path`

HF convention: `from_pretrained(pretrained_model_name_or_path, *args, **kwargs)`. Using `model_id` breaks tooling that introspects the signature (e.g., `accelerate`, `optimum-cli`). Alias it:
```python
def from_pretrained(cls, pretrained_model_name_or_path=None, *, model_id=None, task, ...):
    model_id = model_id or pretrained_model_name_or_path
```

### U14. `sub_models` dict order depends on Python dict insertion

In `from_pretrained`, the loop iterates `_SUB_MODEL_CONFIG.items()`. For prefill+gen models, the order matters for disk IO parallelism (prefill is larger). Minor — but worth declaring that the registration order is the build order, and optionally building sub-models in parallel with a `ThreadPoolExecutor` for big models.

---

## 🟢 Nice Design Points

- `_resolve_cache` correctly unwraps `EncoderDecoderCache` — matches HF 4.41+ internals well.
- Separating `WinMLStaticCache` vs `WinMLSlidingWindowCache` with `position_input_name` polymorphism avoids isinstance checks in `forward()`. Clean.
- `_SUB_MODEL_CONFIG` + `sub_model_kwargs` per-component design naturally extends to prefill/gen and encoder/decoder/adapter splits — perfect foundation for image-to-text (`{"vision_encoder": ..., "text_decoder": ...}`).
- `EncoderDecoderInputGenerator` as shared `DummyInputGenerator` subclass is idiomatic Optimum and reusable.

scripts/e2e_eval/run_eval.py
vortex-captain: indeed, added a TODO here

src/winml/modelkit/models/winml/composite_model.py
vortex-captain: added property `ort_device`

src/winml/modelkit/commands/config.py
vortex-captain: fixed naming

src/winml/modelkit/models/winml/composite_model.py
vortex-captain: It could be generic for all models, moved to new file `src\winml\modelkit\utils\data_utils.py`

zhenchaoni (approved):

src/winml/modelkit/models/winml/composite_model.py
```diff
@@ -0,0 +1,259 @@
...
+# Maps (model_type, task) → pipeline class with _SUB_MODEL_CONFIG.
+# Used by `wmk config` to generate one config file per sub-component.
+PIPELINE_MODEL_REGISTRY: dict[tuple[str, str], type] = {}
+
+
+def register_composite_model(model_type: str, task: str):
+    """Class decorator that registers a composite model for `wmk config`."""
+
+    def decorator(cls: type) -> type:
+        PIPELINE_MODEL_REGISTRY[(model_type, task)] = cls
```
tezheng: **[C3 — FIXED] Registry decorator silently overwrites duplicate `(model_type, task)` keys**

A second `@register_composite_model("t5", "translation")` anywhere in the codebase silently replaces the first entry with no warning. The resulting class is determined by import order, making the registry fragile to refactors.

```suggestion
    def decorator(cls: type) -> type:
        key = (model_type, task)
        if key in PIPELINE_MODEL_REGISTRY:
            raise ValueError(
                f"Composite model already registered for {key!r}: "
                f"{PIPELINE_MODEL_REGISTRY[key].__name__}. "
                f"Cannot register {cls.__name__}."
            )
        PIPELINE_MODEL_REGISTRY[key] = cls
        return cls
```

tezheng (commented):
Follow-up review on PR #334 after a local fix pass rebased onto 3c766019.
This review documents 12 fully-addressed findings and 2 partially addressed
(1 docstring follow-on from the T5 refactor introduced in this commit).

Fix-pass state: 10 files, uncommitted. ruff check + ruff format clean.
All 3869 unit tests pass. Full context in `pr_334_verdicts.md` §10.

Items addressed in the companion branch:
- [C3] Registry silent overwrite → ValueError on duplicate key
- [C6] Silent task=None for unknown component → ValueError
- [C8] Broken from_onnx(dict) dispatch (config vs hf_config) → dedicated
  hf_config: PretrainedConfig parameter + un-mocked regression test
- [I1] WinMLCache.reset() no longer leaks self.captured
- [I2] Registry renamed PIPELINE_MODEL_REGISTRY → COMPOSITE_MODEL_REGISTRY
- [I9] pad_inputs mode: Literal["left", "right"]
- [I11-I14] Docstring contradictions across mu2/qwen/decoder_only/encoder_decoder
- [NI-6] Phantom position_id in Qwen forward docstring
- [NI-8] Un-mocked regression test for composite dict dispatch
- [NM-2 partial] pad_inputs explicit ValueError on invalid mode
- [NEW] encoder_decoder.py docstring updated to reflect T5→sliding-window
  refactor (introduced in this commit)

Still open (27 items, full list in pr_334_verdicts.md §10):
- Critical blockers: C1 exports, C2 skip-list policy, C4 silent fp16,
  C5 surgery precondition, C9 position-encoding footgun
- Six items need PR-author design decisions (Phase 2 in §10.2)

Recommendation: do not merge as-is until the 5 Criticals are addressed.

src/winml/modelkit/models/winml/composite_model.py
```diff
@@ -0,0 +1,259 @@
...
+                    f"Registered: {list(PIPELINE_MODEL_REGISTRY.keys())}"
+                )
+        else:
+            resolved_cls = cls
+
+        from ..auto import WinMLAutoModel
+
+        sub_models: dict[str, Any] = {}
+        for name, path in onnx_path.items():
+            component_task = resolved_cls._SUB_MODEL_CONFIG.get(name)
```
tezheng, line 244: **[C6 — FIXED] `from_onnx` silently passes `task=None` for unrecognized component names**

`_SUB_MODEL_CONFIG.get(name)` returns `None` for any key not declared on the subclass (e.g., a typo like `"encoder_"` instead of `"encoder"`). That `None` is forwarded as `task=None` to `WinMLAutoModel.from_onnx` with no error, producing a silently misconfigured sub-model.

```suggestion
            component_task = resolved_cls._SUB_MODEL_CONFIG.get(name)
            if component_task is None:
                valid = list(resolved_cls._SUB_MODEL_CONFIG.keys())
                raise ValueError(
                    f"Unknown component {name!r}. "
                    f"Valid names for {resolved_cls.__name__}: {valid}"
                )
```

src/winml/modelkit/models/auto.py
```diff
@@ -129,6 +130,24 @@ def from_onnx(
         Returns:
             WinMLPreTrainedModel inference wrapper.
         """
+        if isinstance(onnx_path, dict):
+            from .winml.composite_model import WinMLCompositeModel
+
+            return WinMLCompositeModel.from_onnx(
+                onnx_path,
+                task=task,
+                config=config,
```
tezheng: **[C8 — FIXED] `from_onnx(dict)` passes `config` where callee expects `hf_config` — entire dict API is broken**

The caller passes `config=config` (a `WinMLBuildConfig`), but `WinMLCompositeModel.from_onnx` declares the parameter as `hf_config`. The keyword falls through to `**kwargs`, so `hf_config` is always `None` on entry, `model_type` resolves to `None`, and the registry lookup always raises `ValueError("No composite model for (None, ...)")`. Every call to `WinMLAutoModel.from_onnx(dict, ...)` fails at runtime. The existing `TestFromOnnxDictDispatch` test mocks the callee entirely and never exercises the real body.

```suggestion
            return WinMLCompositeModel.from_onnx(
                onnx_path,
                task=task,
                hf_config=config,
```

src/winml/modelkit/models/winml/kv_cache.py
```diff
@@ -0,0 +1,372 @@
...
+            k = outputs[f"present_{i}_key"]
+            v = outputs[f"present_{i}_value"]
+            k = k if isinstance(k, torch.Tensor) else torch.tensor(k)
+            v = v if isinstance(v, torch.Tensor) else torch.tensor(v)
+            n = k.size(2)
+            ck = {"cache_position": torch.arange(self.step, self.step + n, dtype=torch.int64)}
+            self.update(k, v, i, cache_kwargs=ck)
+        self.step += n
+
+    def reset(self) -> None:
```
tezheng, line 150: **[I1 — FIXED] `WinMLCache` missing `abc.ABC` — abstract contract is unenforced; `reset()` leaks `captured`**

`WinMLCache` uses `@abstractmethod` but inherits only `StaticCache`, not `ABC`. Python won't enforce the abstract contract — a concrete subclass that omits `build_decoder_mask` or `prepare_prefill_chunk` instantiates silently. Separately, `reset()` zeroes the KV buffers but never calls `self.captured.clear()`, so stale tensor references from the previous generation persist into the next.

```suggestion
class WinMLCache(StaticCache, ABC):
```

For `reset()`:
```suggestion
    def reset(self) -> None:
        """Zero out all layers and reset step (start of new generation)."""
        self.step = 0
        self.captured.clear()
        for i in range(self.num_layers):
            self.layers[i].keys.zero_()
            self.layers[i].values.zero_()
```

src/winml/modelkit/models/winml/composite_model.py
```diff
@@ -0,0 +1,259 @@
...
+logger = logging.getLogger(__name__)
+
+
+# =========================================================================
+# composite model Registry
+# =========================================================================
+
+# Maps (model_type, task) → pipeline class with _SUB_MODEL_CONFIG.
+# Used by `wmk config` to generate one config file per sub-component.
+PIPELINE_MODEL_REGISTRY: dict[tuple[str, str], type] = {}
```
tezheng: **[I2 — FIXED] Registry name `PIPELINE_MODEL_REGISTRY` does not reflect its actual scope**

The registry holds composite models (encoder+decoder, prefill+gen), not generic pipelines. The name `PIPELINE_MODEL_REGISTRY` is misleading — it implies HF pipeline compatibility rather than the multi-component architecture this module implements.

Fixed in companion branch: renamed to `COMPOSITE_MODEL_REGISTRY` throughout (composite_model.py, auto.py, commands/config.py).

src/winml/modelkit/utils/data_utils.py
```diff
@@ -0,0 +1,58 @@
...
+
+from typing import Any
+
+import torch
+
+
+def pad_inputs(
+    source: dict[str, Any],
+    expected: dict[str, list[int]],
+    mode: str = "right",
```
tezheng: **[I9 — FIXED] `mode` in `pad_inputs` should be `Literal["left", "right"]`**

```suggestion
    mode: Literal["left", "right"] = "right",
```

Also add `from typing import Literal` to the imports. The runtime `ValueError` guard at line 35 is good but the type annotation should match.

src/winml/modelkit/models/hf/mu2.py
```diff
@@ -0,0 +1,335 @@
...
+        """Run decoder with FIFO KV cache (Slice+Concat).
+
+        Positional args (order matches OnnxConfig.inputs):
+            decoder_input_ids, encoder_hidden_states, attention_mask (encoder),
+            decoder_attention_mask, position_id,
+            past_0_key, past_0_value, past_1_key, past_1_value, ...
+
+        Returns:
+            (logits, present_0_key, present_0_value, ...) where each
+            present KV is the full updated buffer [batch, n_kv_head, max_cache_len, head_dim]
```
tezheng: **[I11 — FIXED] Mu2 decoder docstring claims present-KV is "full updated buffer" — it is new-token only**

```suggestion
        Returns:
            (logits, present_0_key, present_0_value, ...) where each
            present KV is the new-token slice only [batch, n_kv_head, seq_len, head_dim]
            (raw key_states/value_states captured before Slice+Concat in WinMLSlidingWindowCache).
```

src/winml/modelkit/models/hf/qwen.py
```diff
@@ -0,0 +1,370 @@
...
+
+class QwenDecoderWrapper(nn.Module):
+    """Wraps Qwen3ForCausalLM with static KV cache I/O.
+
+    Used for both prefill and generation ONNX export — same forward logic,
+    different OnnxConfig determines the static input shapes.
+
+    Input KV: full static buffer ``[batch, kv_heads, max_cache_len, head_dim]``.
+    Output KV: new positions only ``[batch, kv_heads, seq_len, head_dim]``.
+    Logits: last position only ``[batch, 1, vocab_size]`` (both prefill and gen).
```
tezheng: **[I12 — FIXED] Qwen3 three-way shape contradiction: module docstring, class docstring, and actual code disagree on logits shape**

The module docstring (line 13) says prefill logits are `[1, 64, vocab]`. The class docstring (line 129) says "last position only `[batch, 1, vocab_size]` (both prefill and gen)". The actual code (line 205) returns `out.logits` — all positions. The class docstring is wrong for both claims.

```suggestion
    Logits: all input positions ``[batch, seq_len, vocab_size]`` (both prefill and gen).
    The caller selects the relevant position (last for gen, all for perplexity evaluation).
```

src/winml/modelkit/models/winml/decoder_only.py
```diff
@@ -0,0 +1,366 @@
...
+     Right-pads the last chunk; only writes real tokens' KV into the cache
+     (padding positions are discarded).  Returns logits for ALL real
+     positions ``[1, seq_len, vocab]`` — matches HF convention, enabling
+     both generation (last-token selection) and perplexity evaluation
+     (shifted cross-entropy over all positions).
+
+   - **Generation** (``input_ids`` has 1 token): runs the gen ONNX model
+     with the single token + full KV cache buffer as input.
+
+4. KV cache uses HF ``StaticCache`` — same class as T5.  ``get_seq_length()``
```
tezheng: **[I13 — FIXED] `decoder_only.py` module docstring describes `WinMLStaticCache` only; Qwen3 uses sliding-window**

```suggestion
4. KV cache is cache-agnostic — ``WinMLDecoderOnlyModel`` delegates mask
   construction, position encoding, and cache writes to the ``WinMLCache``
   subclass (``WinMLStaticCache`` for T5/index_copy_, ``WinMLSlidingWindowCache``
   for Qwen3/Slice+Concat).  The cache persists across generate() steps via
   ``CausalLMOutputWithPast.past_key_values``.
```

src/winml/modelkit/models/winml/encoder_decoder.py
```diff
@@ -0,0 +1,340 @@
...
+                auto-padded to match the ONNX model's expected input shape.
+        """
+        # Encoder hidden states
+        if encoder_outputs is None and input_ids is not None:
+            encoder_outputs = self._encoder(input_ids=input_ids, **model_kwargs)
+        if encoder_outputs is None:
+            raise ValueError("Either encoder_outputs or input_ids required")
+        enc_h = encoder_outputs["last_hidden_state"]
+
+        # Resolve or create cache (subclasses override _create_cache).
```
tezheng: **[I14 — FIXED] Stale `_create_cache` reference in comment — method is `get_cache_class`**

```suggestion
        # Resolve or create cache (subclasses override get_cache_class).
```

src/winml/modelkit/models/hf/qwen.py
```diff
@@ -0,0 +1,370 @@
...
+
+    def get_export_args(self, inputs: dict[str, torch.Tensor]) -> tuple[torch.Tensor, ...]:
+        """Convert dict inputs to positional args for torch.onnx.export."""
+        return tuple(inputs.values())
+
+    def forward(self, *args: torch.Tensor) -> tuple[torch.Tensor, ...]:
+        """Run decoder with static KV cache.
+
+        Positional args (order matches OnnxConfig.inputs):
+            input_ids, attention_mask, position_ids, position_id,
```
tezheng: **[NI-6 — FIXED] `QwenDecoderWrapper.forward` docstring lists 4 pre-KV args; code uses 3**

The docstring lists `input_ids, attention_mask, position_ids, position_id` (4 args before KV), but the implementation uses only 3 (`input_ids`, `attention_mask`, `position_ids` at indices 0-2, KV at index 3). The phantom `position_id` arg is misleading.

```suggestion
    def forward(self, *args: torch.Tensor) -> tuple[torch.Tensor, ...]:
        """Run decoder with static KV cache.

        Positional args (order matches OnnxConfig.inputs):
            input_ids, attention_mask, position_ids,
            past_0_key, past_0_value, past_1_key, past_1_value, ...

        Returns:
            (logits, present_0_key, present_0_value, ...) where:
            - logits is ``[batch, seq_len, vocab_size]`` (all positions)
            - present KV is ``[batch, kv_heads, seq_len, head_dim]``
        """
```

tests/unit/models/auto/test_auto_onnx.py
```diff
@@ -195,3 +195,30 @@ def test_passes_ep_from_kwargs(self, fake_onnx: Path, tmp_path: Path):
...
+# from_onnx dict dispatch → WinMLCompositeModel.from_onnx
+# =============================================================================
+
+
+class TestFromOnnxDictDispatch:
+    """from_onnx with dict onnx_path delegates to WinMLCompositeModel.from_onnx."""
+
+    def test_dict_dispatches_to_composite(self, tmp_path: Path):
+        """Dict onnx_path calls WinMLCompositeModel.from_onnx."""
+        with patch(
```
tezheng, line 210: **[NI-8 — FIXED] `TestFromOnnxDictDispatch` only checks two kwargs — would not catch the `config`/`hf_config` mismatch**

The test mocks `WinMLCompositeModel.from_onnx` entirely and only asserts `task="translation"` and `skip_build=True`. It would pass even if the caller passed `config=<wrong_type>` instead of `hf_config=<correct_type>`, which is exactly the C8 bug. Replace the mock with an un-mocked integration test (using a lightweight fake ONNX) so the real callee body is exercised.

Fixed in companion branch: added an un-mocked regression test that calls the real `WinMLCompositeModel.from_onnx` with a minimal dict and a real `PretrainedConfig`, asserting that `hf_config.model_type` is used for the registry lookup. Also added a negative-path test verifying `ValueError` on unknown component name.

src/winml/modelkit/utils/data_utils.py
```diff
@@ -0,0 +1,58 @@
...
+    Args:
+        source: Input tensors keyed by name.
+        expected: ONNX expected shapes keyed by input name.
+        mode: Padding side — ``"right"`` (default, pad at end) or
+            ``"left"`` (pad at start).
+
+    Returns:
+        Filtered and padded tensors matching *expected* keys.
+    """
+    if mode not in ("right", "left"):
```
tezheng, line 35: **[NM-2 — PARTIALLY FIXED] `pad_inputs` crashes on symbolic ONNX dims (`str` in `expected_shape`)**

`expected_shape[dim] - val.shape[dim]` raises `TypeError` when `expected_shape[dim]` is a string (e.g., `"sequence_length"` from a dynamic-shape ONNX model). The existing `TODO` comment acknowledges this but the fallback is missing — it currently raises rather than skipping.

```suggestion
                if not isinstance(expected_shape[dim], int):
                    pad.extend([0, 0])
                    continue
                deficit = max(expected_shape[dim] - val.shape[dim], 0)
```

**Partially fixed in companion branch:** the explicit `ValueError` on invalid `mode` is in place (this line). The symbolic-dim `isinstance` guard is a follow-on — added as a TODO with the exact fix above so it is not lost.

src/winml/modelkit/models/winml/encoder_decoder.py
```diff
@@ -0,0 +1,340 @@
...
+
+    def forward(
+        self,
+        *,
+        encoder_outputs: BaseModelOutput | tuple | None = None,
+        past_key_values: Cache | None = None,
+        input_ids: torch.Tensor | None = None,
+        **model_kwargs: Any,
+    ) -> Seq2SeqLMOutput:
+        """Run decoder with static KV cache.
```
tezheng: **[NEW — FIXED] Docstring became stale after the T5→`WinMLSlidingWindowCache` refactor**

This commit (`3c766019`) switched T5's `get_cache_class()` to return `WinMLSlidingWindowCache` (`t5.py:351`). This forward docstring still says "Run decoder with static KV cache" and "StaticCache (or wrapper)" — incorrect for T5.

**Fixed in companion branch:**
```python
"""Run decoder with a ``WinMLCache`` (``WinMLStaticCache`` or
``WinMLSlidingWindowCache``, selected by the subclass via
``get_cache_class()``).
...
past_key_values: ``WinMLCache`` (or ``EncoderDecoderCache`` wrapper) from previous step.
"""
```

Sibling fix also applied in `decoder_only.py` module docstring (see [I13] comment).

src/winml/modelkit/utils/data_utils.py
vortex-captain: The point on dynamic shape is interesting. IMHO we should return the tensors unpadded for which the expected shapes are dynamic, since most likely these inputs would not need padding.

tezheng (commented):
Second follow-up review: Phase 1 quick wins from pr_334_verdicts.md section 10.2.

This commit addresses 6 findings plus 2 bugs caught by an independent
critic pass on the initial fix attempt (I7 variable shadowing and NM-2
dim-pair misalignment — both would have crashed in production).

3876 unit tests pass (unchanged from baseline). ruff check + ruff format
clean on all 6 touched files. See inline comments for finding-by-finding
diffs.

Remaining work (Phase 2 — needs author input): C2 skip-list policy,
C5 surgery precondition, C9 position-encoding footgun, I3 trust_remote_code
typing, I10 cache in-place semantics, NI-5 Mu2 num_hidden_layers alias.
Full list in pr_334_verdicts.md section 10.

src/winml/modelkit/models/winml/__init__.py
```diff
@@ -175,21 +175,38 @@ def register_specialization(model_type: str, task: str, class_name: str) -> None
 # =============================================================================
 
 from .base import WinMLModelForGenericTask, WinMLPreTrainedModel
+from .composite_model import (
```
tezheng, line 179: **[C1 — FIXED] Composite classes now importable via the package public API**

Previously external callers had to reach into `winml.modelkit.models.winml.composite_model` directly (violating the project import rule in CLAUDE.md). Classes now exported from `models/winml/__init__.py`:

- `WinMLCompositeModel`, `COMPOSITE_MODEL_REGISTRY`, `register_composite_model`
- `WinMLCache`, `WinMLStaticCache`, `WinMLSlidingWindowCache`
- `WinMLEncoderDecoderModel`, `WinMLDecoderOnlyModel`

Concrete T5/Qwen3/Mu2 model classes remain in `models/hf/` — that package already imports the modules for registration side effects.

Smoke test: `uv run python -c "from winml.modelkit.models.winml import WinMLCompositeModel, COMPOSITE_MODEL_REGISTRY, ..."` passes.

src/winml/modelkit/models/winml/encoder_decoder.py
```diff
@@ -0,0 +1,356 @@
...
+            1 for n in self._dec_expected if n.startswith("past_") and n.endswith("_key")
+        )
+        # Resolve KV cache dtype from ONNX input types (fp32 or fp16)
+        dec_type_map = dict(
+            zip(dec_io.get("input_names", []), dec_io.get("input_types", []), strict=False)
+        )
+        import numpy as np
+
+        if "past_0_key" not in dec_type_map:
+            raise KeyError(
```
tezheng, line 203: **[C4 — FIXED] Silent fp16→fp32 coercion replaced with explicit KeyError** (`encoder_decoder.py`)

Previous code was `_np_dtype = <type_map>.get("past_0_key", np.float32)`. If the decoder ONNX's `input_types` lacked `past_0_key` for any reason (malformed io_config, renamed key, past-less decoder) the fallback silently picked float32, resulting in shape/dtype mismatches deep in ORT with no hint the dtype derivation had failed.

Now raises `KeyError` with an actionable message naming `PastKeyValueInputGenerator` as the producer to check.

src/winml/modelkit/models/winml/decoder_only.py
```diff
@@ -0,0 +1,375 @@
...
+            1 for n in self._gen_expected if n.startswith("past_") and n.endswith("_key")
+        )
+        # Resolve KV cache dtype from ONNX input types (fp32 or fp16)
+        gen_type_map = dict(
+            zip(gen_io.get("input_names", []), gen_io.get("input_types", []), strict=False)
+        )
+        import numpy as np
+
+        if "past_0_key" not in gen_type_map:
+            raise KeyError(
```
tezheng, line 214: **[C4 — FIXED] Silent fp16→fp32 coercion replaced with explicit KeyError** (`decoder_only.py`)

Mirror of the encoder_decoder.py fix. Previous code was `_np_dtype = <type_map>.get("past_0_key", np.float32)`. If the decoder ONNX's `input_types` lacked `past_0_key` for any reason (malformed io_config, renamed key, past-less decoder) the fallback silently picked float32, resulting in shape/dtype mismatches deep in ORT with no hint the dtype derivation had failed.

Now raises `KeyError` with an actionable message naming `PastKeyValueInputGenerator` as the producer to check.

src/winml/modelkit/models/winml/kv_cache.py
```diff
@@ -0,0 +1,374 @@
...
+    from optimum.utils import NormalizedConfig
+    from transformers import PretrainedConfig
+
+
+# =============================================================================
+# WinMLCache — common interface
+# =============================================================================
+
+
+class WinMLCache(StaticCache, ABC):
```
tezheng, line 62: **[I1 partial — FIXED] WinMLCache abstract contract now enforced**

Before: `class WinMLCache(StaticCache):` — `@abstractmethod` decorators on `build_decoder_mask` and `prepare_prefill_chunk` were unenforced because the class lacked `abc.ABC` in its MRO. An incomplete subclass (or `WinMLCache` itself) would instantiate and fail later with runtime `AttributeError` or `NotImplementedError` far from the construction site.

Also: `position_input_name: str` was a bare annotation with no default. Accessing it on the base raised `AttributeError`.

**Fixed:**
- `class WinMLCache(StaticCache, ABC):` — abstractness enforced at instantiation
- `position_input_name: ClassVar[str] = ""` — default sentinel; subclasses override with concrete values

Sub-defect I1-C (`reset()` leaking `self.captured`) was fixed in the previous follow-up commit.

scripts/e2e_eval/run_eval.py
```diff
@@ -1048,8 +1133,10 @@ def main() -> None:
                         if _should_skip_existing(existing, retry_types, args.eval_type):
                             skipped_count += 1
                             continue
-                    except Exception:
-                        pass  # Corrupt result file — include model for re-evaluation
+                    except (OSError, json.JSONDecodeError, KeyError) as exc:
```
tezheng, line 1146: **[I7 — FIXED] except Exception: pass blocks narrowed and made observable**

Two sites in run_eval.py (registry enrichment + corrupt result file detection) swallowed every exception type silently. Now narrowed to specific expected exceptions with diagnostic output:

- `except (OSError, json.JSONDecodeError, KeyError) as exc: safe_print(...)` for corrupt result files
- Registry-optional catch similarly narrowed

**Critic-caught bug during review:** original fix used `as e:`, which shadowed the outer `for e in entries:` loop variable. Python 3 deletes `e` after the except block, so `filtered.append(e)` on the next line would have raised `NameError` on the exact corrupt-file path this fix was supposed to improve. Patched to `as exc:` before commit.

(Watchdog-cleanup `except Exception: pass` at run_eval.py:275 left as-is — its context genuinely tolerates any failure mode.)

src/winml/modelkit/models/winml/encoder_decoder.py
```diff
@@ -0,0 +1,356 @@
...
+        input_ids: torch.LongTensor,
+        past_key_values: Cache | None = None,
+        attention_mask: torch.Tensor | None = None,
+        encoder_outputs: BaseModelOutput | None = None,
+        **kwargs: Any,
+    ) -> dict[str, Any]:
+        """Build decoder inputs for each generate() step."""
+        from .kv_cache import WinMLCache
+
+        if isinstance(past_key_values, WinMLCache) and past_key_values.get_seq_length() > 0:
```
tezheng, line 249: **[NI-4 — FIXED] prepare_inputs_for_generation no longer truncates multi-token decoder prompts on first step**

Previous code: `"decoder_input_ids": input_ids[:, -1:]` — always slices to last token regardless of cache state. For standard T5 (single `decoder_start_token_id`) this is fine, but any encoder-decoder with a multi-token decoder prompt (forced BOS + prefix, or a seeded decoder sequence) would silently lose all but the last token on the first decode step.

Now mirrors the decoder-only pattern: slice only when `past_key_values` is a populated `WinMLCache`; otherwise pass the full `input_ids` through.

src/winml/modelkit/utils/data_utils.py
```diff
@@ -0,0 +1,64 @@
...
+            ndim = min(len(val.shape), len(expected_shape))
+            # torch.nn.functional.pad takes pairs (low, high) from the LAST
+            # dim backwards. Skip batch dim (dim 0).
+            pad: list[int] = []
+            for dim in reversed(range(1, ndim)):
+                exp = expected_shape[dim]
+                # Dynamic ONNX dims may be None or a string symbol; emit a
+                # (0, 0) pair so later pairs stay aligned with their dim index.
+                if not isinstance(exp, int):
+                    pad.extend([0, 0])
```
tezheng, line 54: **[NM-2 — FIXED] pad_inputs handles non-int dynamic dims without silent misalignment**

Previously `max(expected_shape[dim] - val.shape[dim], 0)` would raise `TypeError` when `expected_shape[dim]` is `None` or a string symbol (dynamic ONNX dim).

**Critic-caught bug during review:** my initial fix used `continue` to skip non-int dims, but `torch.nn.functional.pad` interprets the pair list positionally (last-dim pair first, then second-to-last, etc.). Skipping omits the pair and misaligns later pairs with their dims — for 3D+ tensors with a dynamic dim between static dims, padding gets applied to the wrong dimension.

**Fixed:** non-int dims now emit `pad.extend([0, 0])` to preserve alignment, with a comment explaining why. Behavioral sanity test run against three cases (dynamic-mid, dynamic-last, None dim) — all produce correct shapes.

tezheng (commented):
Three questions on the open Critical items — each has multiple acceptable
paths and needs your design call. No fixes in this review, just questions.

The other 17 remaining items (Important + Minor) will be addressed in a
follow-up commit alongside this review; those don't require your input.

scripts/e2e_eval/cache/timeout_skip_list.json
```diff
@@ -58,5 +43,25 @@
...
     "task": "summarization",
     "reason": "Build hangs >10000s, bart enc-dec export issue"
+  },
+  {
+    "hf_id": "apple/DepthPro-hf",
+    "task": "depth-estimation",
+    "reason": "OOM in quantization (model too large for in-process quantize)"
+  },
+  {
+    "hf_id": "Qwen/Qwen3-0.6B",
```
tezheng, line 53: **[C2 — Question] Is there a path to making Qwen3 (0.6B/1.7B/8B) and DepthPro-hf actually run on NPU, or should these skips be permanent?**

As written, these skip entries hide quantizer OOMs (segfault on 2.9GB Qwen3-0.6B; OOM on 1.7B/8B/DepthPro). Three options that come to mind:

1. File tracking issues per model and reference them in the `reason` field.
2. Add an unquantized fp16/fp32 path and mark quantized-only as blocked.
3. Scope this PR's Qwen3 support to CPU-only and defer NPU to a follow-up.

Side note: `Qwen/Qwen3-8B` appears twice in the file (once at line ~18 as "Build hangs", once at line ~63 as "OOM in quantization"). First-hit wins per loader semantics — the OOM entry gets shadowed.

Which direction do you want to go?

src/winml/modelkit/optim/pipes/surgery.py
```diff
@@ -139,6 +146,9 @@ def process(self, model: onnx.ModelProto, config: SurgeryPipeConfig) -> onnx.Mod
                 model_copy, config.clamp_min, config.clamp_max, config.verbose
             )
 
+        if config.remove_isnan_in_attention_mask:
```
tezheng, line 149: **[C5 — Question] Should `_remove_isnan_in_attention_mask` hard-require `clamp_constant_values=True`?**

The surgery removes `Softmax → IsNaN → Where` guards that are "dead code when `clamp_constant_values` has replaced -inf with a finite value". But there's no runtime check — enabling `remove_isnan_in_attention_mask=True` with `clamp_constant_values=False` silently removes a live NaN guard and can produce NaN outputs.

Three options:

1. Hard precondition: `if not config.clamp_constant_values: raise ValueError(...)` — simplest, prevents misuse.
2. Auto-enable clamp when isnan-removal is set — convenient but hides coupling.
3. Document as convention; keep all three model configs (T5/Mu2/Qwen) setting both flags and trust future authors to match.

A regression test for the pass would also help — there are currently zero. Any preference?

src/winml/modelkit/models/winml/decoder_only.py
```diff
@@ -0,0 +1,375 @@
...
+            attn_mask = cache.build_decoder_mask(self._max_cache_len, chunk_len)
+
+            feeds: dict[str, Any] = {
+                "input_ids": padded_ids,
+                "attention_mask": attn_mask,
+                "position_ids": position_ids,
+            }
+            if "cache_position" in self._prefill_expected:
+                feeds["cache_position"] = position_ids.squeeze(0)
+            for i in range(self._num_kv_layers):
```
tezheng, line 352: **[C9 — Question] Should the `cache_position` feed branch be deleted, or fixed?**

For `WinMLSlidingWindowCache`, `prepare_prefill_chunk` returns left-padded `position_ids` (zeros at front, real tokens at end). The ONNX graph for Qwen3 is traced with a right-aligned `arange(max_cache_len - seq_len, max_cache_len)` — mismatch would be silent numerical corruption.

Currently dead for Qwen3 (not in `_prefill_expected`). But `qwen.py:48-58` instructs readers to add `cache_position` to the OnnxConfig when switching to StaticCache — following those instructions activates the bug.

Two reasonable paths:

1. Delete the dead branch and remove/update the switching instructions.
2. Compute `cache_position` independently from `position_ids` using the right-aligned `arange` to match the export trace.

Which do you prefer?

scripts/e2e_eval/cache/timeout_skip_list.json
tezheng, line 53: **[C2 — Question] Is there a path to making Qwen3 (0.6B/1.7B/8B) and DepthPro-hf actually run on NPU, or should these skips be permanent?**

As written, these skip entries hide quantizer OOMs (segfault on 2.9GB Qwen3-0.6B; OOM on 1.7B/8B/DepthPro). Three options that come to mind:

1. File tracking issues per model and reference them in the `reason` field.
2. Add an unquantized fp16/fp32 path and mark quantized-only as blocked.
3. Scope this PR's Qwen3 support to CPU-only and defer NPU to a follow-up.

Side note: `Qwen/Qwen3-8B` appears twice in the file (once at line ~18 as "Build hangs", once at line ~63 as "OOM in quantization"). First-hit wins per loader semantics — the OOM entry gets shadowed.

Which direction do you want to go?

src/winml/modelkit/optim/pipes/surgery.py
tezheng, line 149: **[C5 — Question] Should `_remove_isnan_in_attention_mask` hard-require `clamp_constant_values=True`?**

The surgery removes `Softmax → IsNaN → Where` guards that are "dead code when `clamp_constant_values` has replaced -inf with a finite value". But there's no runtime check — enabling `remove_isnan_in_attention_mask=True` with `clamp_constant_values=False` silently removes a live NaN guard and can produce NaN outputs.

Three options:

1. Hard precondition: `if not config.clamp_constant_values: raise ValueError(...)` — simplest, prevents misuse.
2. Auto-enable clamp when isnan-removal is set — convenient but hides coupling.
3. Document as convention; keep all three model configs (T5/Mu2/Qwen) setting both flags and trust future authors to match.

A regression test for the pass would also help — there are currently zero. Any preference?

src/winml/modelkit/models/winml/decoder_only.py
tezheng, line 352: **[C9 — Question] Should the `cache_position` feed branch be deleted, or fixed?**

For `WinMLSlidingWindowCache`, `prepare_prefill_chunk` returns left-padded `position_ids` (zeros at front, real tokens at end). The ONNX graph for Qwen3 is traced with a right-aligned `arange(max_cache_len - seq_len, max_cache_len)` — mismatch would be silent numerical corruption.

Currently dead for Qwen3 (not in `_prefill_expected`). But `qwen.py:48-58` instructs readers to add `cache_position` to the OnnxConfig when switching to StaticCache — following those instructions activates the bug.

Two reasonable paths:

1. Delete the dead branch and remove/update the switching instructions.
2. Compute `cache_position` independently from `position_ids` using the right-aligned `arange` to match the export trace.

Which do you prefer?

tezheng:
## Deferred items — tracked for follow-up / design discussion

The following 7 findings from the review pass are acknowledged but **not** addressed in the companion fix commits. Each needs design input, new test infrastructure, or a project-wide decision outside this PR's scope:

- **I5** — mixed `torch.nn.Module` / `PreTrainedModel` paradigm inside `WinMLEncoderDecoderModel`. Major refactor; not PR-specific.
- **I10** — `WinMLSlidingWindowCache.update()` reassigns `.keys`/`.values` (non-in-place), deviating from HF `StaticCache` in-place contract. Touches cache semantics — needs design call before changing.
- **NI-1** — No ORT `InferenceSession` lifecycle (no `close`/`__del__`/context manager) on `WinMLCompositeModel`. Project-wide gap, not specific to this PR.
- **NI-2 + NI-3** — Surgery-pass graph hygiene: `_remove_isnan_in_attention_mask` leaves orphan `graph.value_info` / dead initializers, and rewires all `Where` consumers without verifying input position (condition vs. value_if_true vs. value_if_false). Needs ONNX domain expertise; not low-risk.
- **NI-5** — `WinMLCache.__init__` reads `config.num_hidden_layers`, while `Mu2DecoderWrapper` uses `config.n_decoder_layer`. If the remote-code `Mu2Config` doesn't expose both attributes, Mu2 inference crashes at cache construction. Needs author confirmation on the `Mu2Config` shape.
- **NM-3** — `Path.with_stem()` usage in `src/winml/modelkit/commands/config.py` requires Python 3.9+. Project-wide Python version floor decision.
- **NM-4** — KV cache unit tests assert shapes only; no value-level tests for `update` / `captured` / `reset`. Needs new test infrastructure.

These are all real findings, just not mechanically fixable in a review-follow-up branch. Tracking as follow-up issues (or resolving inline here) is equally fine with me.

tezheng (commented):
Second-follow-up review (Phase B + C): flags 9 Important+Minor findings on the current HEAD. A companion commit in this review addresses each of them. All changes have been verified (3876 unit tests pass, ruff clean) and an independent critic caught and corrected one regression before this commit.

src/winml/modelkit/models/winml/composite_model.py
```diff
@@ -0,0 +1,271 @@
...
+                names from ``_SUB_MODEL_CONFIG`` (e.g., ``"decoder_prefill"``,
+                ``"decoder_gen"``).  Values are dicts merged on top of the
+                shared ``**kwargs``.  Use this to pass different
+                ``shape_config`` per sub-model.
+            **kwargs: Forwarded to ``WinMLAutoModel.from_pretrained()``
+                for every sub-component (overridden by ``sub_model_kwargs``).
+        """
+        from transformers import AutoConfig
+
+        trust_remote_code = kwargs.get("trust_remote_code", False)
```
tezheng: **[I3] `trust_remote_code` should be an explicit parameter, not a kwargs lookup**

Currently `kwargs.get("trust_remote_code", False)` extracts a security-sensitive flag from untyped `**kwargs` — no IDE support, no signature hint, no type check.

**Fixed in companion commit:** promoted to an explicit keyword-only parameter on `from_pretrained` and threaded through to both the recursive `from_pretrained` call and `WinMLAutoModel.from_pretrained` / `AutoConfig.from_pretrained`. Docstring updated.

src/winml/modelkit/models/auto.py
```diff
@@ -278,6 +306,35 @@ def from_pretrained(
...
 
+        # =====================================================================
+        # COMPOSITE MODEL CHECK — delegate to WinMLCompositeModel.from_pretrained
+        # when (model_type, task) is a registered composite (e.g., T5 translation,
+        # Qwen text-generation).  AutoConfig is lightweight (~config.json only).
+        # =====================================================================
+        if task is not None:
+            from transformers import AutoConfig
+
+            _hf_cfg = AutoConfig.from_pretrained(model_id, trust_remote_code=trust_remote_code)
```
tezheng: **[I4] `AutoConfig.from_pretrained` probe runs on every `from_pretrained` call, even for non-composite models**

For a simple `WinMLAutoModel.from_pretrained("microsoft/resnet-50", task="image-classification")`, the composite-model registry check fires here and loads the HF config, even though "image-classification" is not a composite task. The main pipeline then loads the config again.

**Fixed in companion commit:** `KNOWN_COMPOSITE_TASKS` derived from `COMPOSITE_MODEL_REGISTRY` gates the probe — only fires when `task` matches a composite task.

scripts/e2e_eval/run_eval.py
```diff
@@ -1182,70 +1269,53 @@ def main() -> None:
...
+            build_result = _run_build(
+                entry,
+                args.device,
+                _DEFAULT_PRECISION,
+                args.timeout,
+                model_dir,
+            )
+            onnx_paths = build_result["onnx_paths"] if build_result["success"] else {}
+            # First ONNX path for accuracy phase (TODO: composite model support)
+            first_path = next(iter(onnx_paths.values()), None) if onnx_paths else None
```
tezheng: **[I6] Composite accuracy picks an arbitrary sub-model — result is meaningless**

For a composite model (e.g., T5: encoder + decoder sub-components), this `next(iter(...))` grabs whichever sub-model the dict iterates first. The accuracy phase then runs an unrelated-task eval against that arbitrary sub-model — garbage result.

**Fixed in companion commit:** skip accuracy for composite with explicit `skip_reason="composite_model_not_supported"` and a log message. The existing TODO is preserved; proper composite accuracy evaluation remains future work.

src/winml/modelkit/models/winml/decoder_only.py
```diff
@@ -0,0 +1,375 @@
...
+        self._prefill_seq_len = self._prefill_expected["input_ids"][1]
+
+    # ----- Cache + GenerationMixin interface -----
+
+    @classmethod
+    def get_cache_class(cls) -> type:
+        """Return the WinMLCache subclass. Subclasses must override."""
+        raise NotImplementedError
+
+    def _resolve_cache(self, past_key_values: Any) -> Any:
```
tezheng, line 232: **[I16] `_resolve_cache` asymmetry — decoder-only missing the `EncoderDecoderCache` unwrap**

`WinMLEncoderDecoderModel._resolve_cache` unwraps `hasattr(past_key_values, "self_attention_cache")`. The decoder-only version lacks this branch entirely — fragile if GenerationMixin semantics change.

**Fixed in companion commit:** add the unwrap with a comment clarifying it's defensive symmetry (never fires under the current GenerationMixin flow for decoder-only models, but kept aligned with encoder_decoder.py).

src/winml/modelkit/commands/config.py
```diff
@@ -486,3 +511,91 @@ def config(
...
+        )
+        _apply_stage_overrides(cfg, no_quant=no_quant, no_compile=no_compile)
+
+        config_json = json.dumps(cfg.to_dict(), indent=2)
+
+        if output:
+            out_path = Path(output)
+            suffixed = out_path.with_stem(f"{out_path.stem}_{component_name}")
+            suffixed.parent.mkdir(parents=True, exist_ok=True)
+            suffixed.write_text(config_json)
```
tezheng: **[NI-7] Per-sub-component config writes are non-atomic**

For composite models, one `wmk config` invocation writes N files sequentially. A crash after `foo_encoder.json` but before `foo_decoder.json` leaves inconsistent state with no sentinel. The next run sees partial output and may silently proceed with one stale + one fresh config.

**Fixed in companion commit:** write to `suffixed.with_suffix(".json.tmp")`, then `tmp.replace(suffixed)`. Atomic on POSIX (rename) and Windows (`os.replace` >= Python 3.3).

scripts/e2e_eval/run_eval.py
```diff
@@ -378,59 +382,93 @@ def _run_build(
...
-        "build",
-        "-c",
-        str(config_path),
-        "-m",
-        entry.hf_id,
-        "--use-cache",
-    ]
+    # Collect config files: composite models produce suffixed files
+    # (e.g., build_config_encoder.json); single models produce config_path itself.
+    sub_configs = sorted(config_path.parent.glob(f"{config_path.stem}_*.json"))
```
tezheng, line 402: **[NI-9] Stale-config cleanup — initial fix was wrong, critic caught it**

My first pass added this cleanup loop AFTER `wmk config` ran. But `wmk config` for composite models writes files matching `{stem}_*.json` — so the cleanup was deleting the freshly-written configs, silently degrading every composite build to single-model.

**Fixed in companion commit:** move the cleanup BEFORE the `wmk config` invocation, so it removes prior-run artifacts only, without touching current-run output.

src/winml/modelkit/models/winml/composite_model.py
```diff
@@ -0,0 +1,271 @@
...
+    def ort_device(self) -> str:
+        """ORT execution provider target (e.g. "npu", "gpu", "cpu", "auto")."""
+        return self._device
+
+    @property
+    def dtype(self) -> torch.dtype:
+        """Model dtype for HF compatibility."""
+        return torch.float32
+
+    def to(self, *args: Any, **kwargs: Any) -> WinMLCompositeModel:
```
tezheng, line 270: **[NM-1] `.to()` silently discards the device arg**

`model.to("npu")` returns `self` unchanged, sub-models stay on their original device. HF pipelines expect `.to()` to have an effect; callers get no signal that the call was a no-op.

**Fixed in companion commit:** log the no-op at `DEBUG` level (not `WARNING` — HF pipelines routinely call `.to("cpu")` as setup, and warning-level would spam normal usage). Users debugging device placement can enable debug logging to see the message.

src/winml/modelkit/models/winml/kv_cache.py
```diff
@@ -0,0 +1,374 @@
...
+        """Left-pad: padding at start, real tokens at END."""
+        import torch
+
+        chunk_len = chunk_ids.shape[1]
+        pad_len = prefill_seq_len - chunk_len
+
+        padded_ids = torch.zeros(1, prefill_seq_len, dtype=chunk_ids.dtype)
+        padded_ids[0, pad_len:] = chunk_ids[0]
+
+        position_ids = torch.zeros(1, prefill_seq_len, dtype=torch.int64)
```
tezheng, line 323: **[NM-5] `position_ids` padding at position 0 relies on attention mask — undocumented**

For left-padded prefill chunks, padding positions get `position_id=0` (a valid RoPE position). RoPE computes real embeddings for those padding tokens. The safety argument is that `build_decoder_mask` masks them out before softmax — but that was not documented here.

**Fixed in companion commit:** one-line comment spelling out the invariant. Documentation-only; no behavior change.

src/winml/modelkit/models/winml/decoder_only.py
```diff
@@ -0,0 +1,375 @@
...
+                self._prefill_seq_len,
+            )
+            attn_mask = cache.build_decoder_mask(self._max_cache_len, chunk_len)
+
+            feeds: dict[str, Any] = {
+                "input_ids": padded_ids,
+                "attention_mask": attn_mask,
+                "position_ids": position_ids,
+            }
+            if "cache_position" in self._prefill_expected:
```
tezheng, line 350: **[NM-6] Dead `cache_position` branch for Qwen3 — previously just `# see C9 review comment`**

Qwen3's OnnxConfig omits `cache_position` (computed internally by the export wrapper), so these two branches (one in `_run_prefill`, one in `_run_gen`) never fire. My initial follow-up added `# NOTE: see C9 review comment; delete or fix after author decision.` — that's meaningless outside the review context.

**Fixed in companion commit:** replace both notes with a self-contained explanation referencing the StaticCache switching instructions in `hf/qwen.py` and the position-alignment caveat that blocks activating this branch.

scripts/e2e_eval/cache/timeout_skip_list.json
vortex-captain: I'd prefer option 1, since this skip list is only used in e2e perf/eval test targeting NPUs with default quantization, and neither unquantized model test nor CPU model test is in the current scope of the e2e test.

vortex-captain: I'd prefer option 1, since this skip list is only used in e2e perf/eval test targeting NPUs with default quantization, and neither unquantized model test nor CPU model test is in the current scope of the e2e test.

src/winml/modelkit/optim/pipes/surgery.py
vortex-captain: I'd prefer option 1, with dedicated UTs and doc.
Note: I tried to clamp relevant -inf values when `remove_isnan_in_attention_mask=True`, but it turns out finding exactly all the "relevant" values can be tricky due to different ways to express the mask and the "-inf" values, so we might have to stick with `remove_isnan_in_attention_mask+clamp_constant_values` for a while.

vortex-captain: I'd prefer option 1, with dedicated UTs and doc.
Note: I tried to clamp relevant -inf values when `remove_isnan_in_attention_mask=True`, but it turns out finding exactly all the "relevant" values can be tricky due to different ways to express the mask and the "-inf" values, so we might have to stick with `remove_isnan_in_attention_mask+clamp_constant_values` for a while.

src/winml/modelkit/models/winml/decoder_only.py
vortex-captain: This branch is left for use with StaticCache (e.g. for performance comparison of 2 kv caches in the future), please leave it intact.
There is currently a bit of coupling between export model wrapper and the cache used during `forward()`, which could indeed lead to confusions like this one. We could think of a cleaner separation later.

vortex-captain: This branch is left for use with StaticCache (e.g. for performance comparison of 2 kv caches in the future), please leave it intact.
There is currently a bit of coupling between export model wrapper and the cache used during `forward()`, which could indeed lead to confusions like this one. We could think of a cleaner separation later.

src/winml/modelkit/models/winml/composite_model.py
```diff
@@ -0,0 +1,287 @@
...
+                trust_remote_code=trust_remote_code,
+                **merged,
+            )
+
+        return cls(sub_models=sub_models, config=hf_config)
+
+    @classmethod
+    def from_onnx(
+        cls,
+        onnx_path: dict[str, str | Path],
```
tezheng, line 196: **[I8 — FIXED] Type-annotation mismatch reversed**

Earlier audit passes retracted I8 as cosmetic/non-actionable because runtime `Path(path)` coercion masks the effect. On reflection, a type-annotation mismatch between a public caller (`WinMLAutoModel.from_onnx` declares `dict[str, str | Path]` at `auto.py:99`) and its callee (this signature previously declared `dict[str, str]`) is a real defect — `mypy --strict` / `pyright` would flag it even though the code runs.

**Fixed by widening the callee** to match the caller:

```python
# composite_model.py:196
onnx_path: dict[str, str | Path],
```

Plus `Path` added to the `TYPE_CHECKING` block (no runtime cost) and docstring updated to note both types are accepted. I8 moves from retracted to fixed.

src/winml/modelkit/models/winml/composite_model.py
```diff
@@ -0,0 +1,283 @@
...
+                trust_remote_code=trust_remote_code,
+                **merged,
+            )
+
+        return cls(sub_models=sub_models, config=hf_config)
+
+    @classmethod
+    def from_onnx(
+        cls,
+        onnx_path: dict[str, str],
```
tezheng: **[I8] Type-annotation mismatch — callee narrower than caller**

`WinMLAutoModel.from_onnx` declares `onnx_path: str | Path | dict[str, str | Path]` at `auto.py:99`. The composite callee here declares `dict[str, str]`. When the caller dispatches a dict whose values include `Path`, the callee receives values its type annotation forbids.

Runtime works — `Path(path)` coercion inside the dispatch loop handles both types — but `mypy --strict` / `pyright` flag the mismatch at the call site.

**Ask:** widen the callee to `dict[str, str | Path]` so the annotation matches what the code already accepts. Fix to be addressed in a follow-up commit.

zhenchaoni (approved):
Re approve after comments are addressed.

PR #336, opened 2026-04-14
fix: Enable RTR rewrite autoconf for QNN 6D Reshape models (Swin, Nougat)

ssss141414:
## Problem

Several models fail on QNN EP (NPU) because Swin's window partition creates **6D Reshape operations** (e.g. `[1, 8, 7, 8, 7, 192]`), exceeding QNN's **max tensor rank of 5D**.

The codebase already has a `ReshapeTransposeReshapeOverlyHighDimPattern` rewrite that merges consecutive dimensions to reduce 6D→≤5D, but it was **never auto-triggered** due to two bugs in the autoconf pipeline.

### Affected Models
- `microsoft/swin-large-patch4-window7-224` (image-classification)
- `facebook/nougat-base` (image-to-text, vision-encoder-decoder)
- `naver-clova-ix/donut-base` (image-to-text, vision-encoder-decoder)
- `PekingU/rtdetr_r50vd_coco_o365` (object-detection) — already worked, no 6D Reshapes

### Bug 1: `pattern_id` mismatch

`ReshapeTransposeReshapeOverlyHighDimPattern` and `LowDimPattern` both inherit `pattern_id` from a shared `PatternSchema` (`name="ReshapeTransposeReshapePattern"`), producing `SUBGRAPH/ReshapeTransposeReshapePattern` instead of the class-specific IDs expected by information rules (e.g. `SUBGRAPH/ReshapeTransposeReshapeOverlyHighDimPattern`).

### Bug 2: kebab/snake case key mismatch

The information rule emits `"highdimRTR-lowdimRTR": true` (kebab-case), but `RewritePipe.build_config()` looks up `kwargs.get("highdimRTR_lowdimRTR")` (snake_case via `cap.python_name`).

## Fix

1. **Override `pattern_id`** in both `OverlyHighDimPattern` and `LowDimPattern` to return their class-specific IDs (uses the base class's designed extension point).

2. **Normalize keys** in `AnalysisResult.get_optimization_config()` with `key.replace("-", "_")` — no-op for existing snake_case keys.

### Unit Tests
- **3821 passed**, 73 skipped (hardware-gated), 0 failures

### Perf Verification (`winml perf` on Snapdragon X Elite NPU)

| Model | Task | Before | After | Latency |
|---|---|---|---|---|
| `PekingU/rtdetr_r50vd_coco_o365` | object-detection | ✅ NPU | ✅ NPU | **303ms** |
| `microsoft/swin-large-patch4-window7-224` | image-classification | ❌ `ComposeGraph failed` | ✅ **NPU** | **141ms** |
| `facebook/nougat-base` | image-to-text | ❌ `ComposeGraph failed` | ✅ **NPU** | **459ms** |
| `naver-clova-ix/donut-base` | image-to-text | ❌ `ComposeGraph failed` | ⚠️ `FinalizeGraphs 6020` | Input 2560×1920 exceeds QNN DDR spill limits |

### DIFF
--- a/src/winml/modelkit/analyze/analyzer.py
+++ b/src/winml/modelkit/analyze/analyzer.py
@@ -393,7 +393,10 @@ class AnalysisResult:
                     continue
 
                 if action_item.optimization_options:
-                    optim_options.update(action_item.optimization_options)
+                    # Normalize kebab-case keys to snake_case (python_name)
+                    # so they match the capability system's python_name format.
+                    for key, value in action_item.optimization_options.items():
+                        optim_options[key.replace("-", "_")] = value
 
         # Create and return config from collected options
         return WinMLOptimizationConfig(**optim_options)
--- a/src/winml/modelkit/pattern/transpose_patterns.py
+++ b/src/winml/modelkit/pattern/transpose_patterns.py
@@ -315,7 +315,7 @@ class ReshapeTransposeReshapeOverlyHighDimPattern(Pattern):
     @property
     def pattern_id(self) -> str:
         """Return pattern ID matching the information rule configuration."""
-        return "SUBGRAPH/ReshapeTransposeReshapeOverlyHighDimPattern"
+        return f"SUBGRAPH/{type(self).__name__}"
 
     def get_schema(self) -> PatternSchema:
         """Return the schema definition for ReshapeTransposeReshape pattern.
@@ -617,7 +617,7 @@ class ReshapeTransposeReshapeLowDimPattern(ReshapeTransposeReshapeOverlyHighDimP
     @property
     def pattern_id(self) -> str:
         """Return pattern ID matching the information rule configuration."""
-        return "SUBGRAPH/ReshapeTransposeReshapeLowDimPattern"
+        return f"SUBGRAPH/{type(self).__name__}"
 
     def get_schema(self) -> PatternSchema:
         """Return the schema definition for ReshapeTransposeReshapeLowDim pattern."""
--- a/src/winml/modelkit/pattern/transpose_patterns.py
+++ b/src/winml/modelkit/pattern/transpose_patterns.py
@@ -312,6 +312,11 @@ class ReshapeTransposeReshapeOverlyHighDimPattern(Pattern):
             return None
         return result
 
+    @property
+    def pattern_id(self) -> str:
+        """Return pattern ID matching the information rule configuration."""
+        return "SUBGRAPH/ReshapeTransposeReshapeOverlyHighDimPattern"
+
     def get_schema(self) -> PatternSchema:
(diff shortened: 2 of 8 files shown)


src/winml/modelkit/pattern/transpose_patterns.py
```diff
@@ -609,6 +614,11 @@ def get_internal_constants_and_attributes(
 
         return internal_constants, internal_attributes
 
+    @property
+    def pattern_id(self) -> str:
+        """Return pattern ID matching the information rule configuration."""
+        return "SUBGRAPH/ReshapeTransposeReshapeLowDimPattern"
```
xieofxie: @copilot is there any way to automatically generate it based on class name?

tests/unit/analyze/core/test_unified_pattern_config.py
xieofxie, line 28: what does "multiple Pattern classes can share same pattern_id" mean? I am unfaimilar with this..

src/winml/modelkit/analyze/analyzer.py
```diff
@@ -393,7 +393,10 @@ def get_optimization_config(self, ep: str | None = None) -> WinMLOptimizationCon
                     continue
 
                 if action_item.optimization_options:
-                    optim_options.update(action_item.optimization_options)
+                    # Normalize kebab-case keys to snake_case (python_name)
```
DingmaomaoBJTU, line 396: Consider to update the original information config? Then we could get rid of this transform

src/winml/modelkit/pattern/transpose_patterns.py
xieofxie: @ssss141414 never thought copilot will create pr directly.. Please revert if we should not do it. As I see in the test, multiple Pattern classes can share same pattern_id

DingmaomaoBJTU (approved):

src/winml/modelkit/analyze/analyzer.py
ssss141414: Should we add both config and here? Adding here is to prevent wrong config...

src/winml/modelkit/pattern/transpose_patterns.py
ssss141414: Removed hardcode.

tests/unit/analyze/core/test_unified_pattern_config.py
ssss141414: From copilot:

Multiple Pattern classes can share the same pattern_id when they are semantically equivalent variants of the same subgraph. For example, Gelu1Pattern, Gelu2Pattern, Gelu3Pattern, and Gelu4Pattern all return "SUBGRAPH/GeluPattern"

DingmaomaoBJTU (approved):

PR #343, opened 2026-04-14
add new ops for DeepSeek

xieofxie:


xieofxie:
The indices is dynamic shaped in embeddings.onnx

PR #345, opened 2026-04-14
add --trust-remote-code to inspect & -h

xieofxie:
## Summary

Two small usability improvements to the `winml` CLI:

- **`inspect` command** now accepts `--trust-remote-code`, which is forwarded to `AutoConfig.from_pretrained` and `AutoModel.from_pretrained` calls inside `inspect_model`, `_inspect_model_v2`, and `extract_hierarchy`. This unblocks inspection of models that ship custom code in the repository (e.g. pyannote), which previously failed because `trust_remote_code` was hardcoded to `False`.
- **`-h` is now accepted as an alias for `--help`** across the CLI, via `context_settings={"help_option_names": ["-h", "--help"]}` on the top-level Click group.

## Changes

| File | Change |
|------|--------|
| `cli.py` | Enable `-h`/`--help` on the root group |
| `commands/inspect.py` | Add `--trust-remote-code` flag; thread through `_inspect_model_v2` |
| `inspect/__init__.py` | Add `trust_remote_code` param to `inspect_model` |
| `inspect/hierarchy.py` | Add `trust_remote_code` param to `extract_hierarchy` |

Default remains `False` — no behavior change for existing callers.

## Context

Surfaced while investigating pyannote model inspection, which requires custom code from its HF repo.


src/winml/modelkit/cli.py
```diff
@@ -112,7 +112,7 @@ def format_commands(self, ctx: click.Context, formatter: click.HelpFormatter) ->
                 formatter.write_dl(rows)
 
 
-@click.group(cls=LazyGroup)
+@click.group(cls=LazyGroup, context_settings={"help_option_names": ["-h", "--help"]})
```
xieofxie, line 115: do we need `-h` ? I feel it useful

PR #348, opened 2026-04-14
Fix config not correct on Intel device.

chinazhangchao:
https://github.com/microsoft/WinML-ModelKit/issues/344
image


tezheng:
Thanks for the fix. I'd prefer to hold this PR — we need a proper redesign of the device/EP mapping that also includes versioning support. That work is in progress. Let's fold this fix into that effort rather than merging a workaround we'd need to undo later.

Unless this is a must-have for an upcoming release or is blocking someone, I'd suggest we defer.

PR #354, opened 2026-04-15
feat: add winml serve & winml run with MCP support

DingmaomaoBJTU:
## Summary

Add a local REST inference server (`winml serve`) and one-shot CLI runner (`winml run`) with MCP integration for Claude Desktop.

### Architecture — Phased Server

```
winml serve                                     → Phase 0: CLI-as-API wrapper
winml serve --model microsoft/resnet-50         → Phase 1: warm single-model inference
winml serve --model resnet-50 --idle-timeout 300 → Phase 2: idle auto-unload/reload
winml serve --model resnet-50 --multi           → Phase 3: multi-model slot manager + LRU eviction
```

All phases share one FastAPI app, selected at startup based on args. Routes depend only on the `ModelManager` protocol — Phase 1→3 migration is a drop-in swap.

```
commands/serve.py  ─── CLI entry point (click)
commands/run.py    ─── one-shot inference (--connect to serve or embedded)
serve/
├── app.py             ─── FastAPI app factory + routes (Phase 1-3)
├── cli_api.py         ─── Phase 0 CLI wrapper
├── manager.py         ─── ModelManager protocol, SingleModelManager, ModelSlotManager
├── schema.py          ─── Pydantic request/response models
├── static/index.html  ─── Demo UI dashboard
inference/
├── engine.py          ─── InferenceEngine (HF Pipeline + ORT backend)
├── tasks.py           ─── TASK_REGISTRY: 30 HF tasks + 5 aliases
├── schema_generator.py─── OpenAI/MCP tool definitions
├── types.py           ─── PredictionResult, Prediction
scripts/mcp_server.py  ─── Standalone MCP server for Claude Desktop
```

See [docs/serve-architecture.md](docs/serve-architecture.md) for full architecture diagrams.
See [docs/design-input-schema.md](docs/design-input-schema.md) for the registry-driven input schema design.

### Registry-Driven Input Schema (30 HF Tasks)

All input parsing is driven by `TASK_REGISTRY` — zero hardcoded task branches. Each task defines `user_inputs` (what the user provides) and `PipelineMapping` (how to call the HF pipeline):

| Category | Tasks | CLI shortcut |
|---|---|---|
| Single image (6) | image-classification, object-detection, segmentation, depth, features, image-to-image | `--file` |
| Single text (9) | text-classification, token-clf, fill-mask, text-gen, text2text, features, summarization, translation, TTS | `--text` |
| Single audio (2) | audio-classification, ASR | `--file` |
| Single video (1) | video-classification | `--file` |
| Image + text (4) | VQA, doc-QA, image-text-to-text, image-to-text | `--file --text` |
| Image + json (2) | zero-shot-image-clf, zero-shot-object-det | `--file -I candidate_labels='[...]'` |
| Audio + json (1) | zero-shot-audio-clf | `--file -I candidate_labels='[...]'` |
| Text + text (1) | question-answering | `-I question=... -I context=...` |
| Text + json (2) | zero-shot-clf, table-QA | `--text -I candidate_labels/table='...'` |
| Image pair (1) | keypoint-matching | `-I image_0=@a -I image_1=@b` |
| Image + spatial (1) | mask-generation (SAM) | `--file [-I input_points='...']` |

### `winml run` — Embedded + Connect Mode

```
winml run --model microsoft/resnet-50 --file cat.jpg
```

By default, runs embedded inference (loads model in-process). With `--connect`, probes a running `winml serve` instance first:

1. Probes `localhost:8000/v1/health` (0.5s timeout)
2. If a matching `winml serve` is running → routes request there (zero model load time)
3. Otherwise → falls back to embedded inference

All input combinations (`--file`, `--text`, `--input`, `-P`) are forwarded to the server — file uploads go to `/v1/predict/file` (with extra inputs via the `inputs` form field), text/named inputs go to `/v1/predict` (JSON).

### REST API — `/v1/predict/file` Input Support

The file upload endpoint supports structured form fields beyond `file` and `text`:

| Field | Type | Purpose |
|---|---|---|
| `file` | binary | Media file (image/audio/video), max 20 MB |
| `text` | string | Auto-mapped to sole text field in schema |
| `inputs` | JSON string | Additional named inputs (e.g., `{"candidate_labels": ["cat","dog"]}`) |
| `params` | JSON string | Pipeline parameters (e.g., `{"top_k": 5}`) |

This enables zero-shot tasks and SAM via file upload:
```bash
curl -F "file=@cat.jpg" -F 'inputs={"candidate_labels":["cat","dog"]}' localhost:8000/v1/predict/file
```

Coverage: `/v1/predict` (JSON) covers 30/30 tasks. `/v1/predict/file` covers 17/30 (all tasks with exactly 1 binary input).

### MCP Server

Standalone script (`scripts/mcp_server.py`) that doesn't import `winml.modelkit` to avoid heavy ML dependency timeouts. Queries `/v1/models` at startup and generates one MCP tool per loaded model.

### Other Changes

- **NPU auto-precision**: `int8` → `w8a16` (uint8 weights + uint16 activations)
- **hub_models.json**: expanded catalog with quantization metadata, accuracy `drop_pct` → `delta_display`
- **CLI options**: `ep_option` removed `free_text`, `device_option` added `SUPPORTED_DEVICES_WITH_AUTO`

## Documentation

- [docs/winml-run-task-coverage.md](docs/winml-run-task-coverage.md) — CLI task coverage table + flowchart
- [docs/winml-serve-task-coverage.md](docs/winml-serve-task-coverage.md) — REST API task coverage table + flowchart
- [docs/design-input-schema.md](docs/design-input-schema.md) — Registry-driven input schema design (updated)
- [docs/serve-architecture.md](docs/serve-architecture.md) — Architecture diagrams (updated)

### `winml run --model microsoft/resnet-50 --file cat.jpg`

```
Task:    image-classification
Model:   microsoft/resnet-50
Device:  auto

Results:
   1. tabby                          0.4312
   2. tiger_cat                      0.3876
   3. Egyptian_cat                   0.0911

Latency: 45.2ms
```

### REST API

```bash
# Image classification (file upload)
curl -X POST localhost:8000/v1/predict/file -F "file=@cat.jpg"

# Zero-shot (file + labels)
curl -X POST localhost:8000/v1/predict/file \
  -F "file=@cat.jpg" \
  -F 'inputs={"candidate_labels":["cat","dog","bird"]}'

# Question answering (JSON)
curl -X POST localhost:8000/v1/predict \
  -H "Content-Type: application/json" \
  -d '{"inputs":{"question":"Who?","context":"Tim Cook is CEO."}}'

# Multi-model management
curl -X POST localhost:8000/v1/models \
  -d '{"model_id":"ProsusAI/finbert","task":"text-classification"}'
```


src/winml/modelkit/commands/run.py
```diff
@@ -0,0 +1,314 @@
...
+        pass
+    if value.lower() == "true":
+        return True
+    if value.lower() == "false":
+        return False
+    return value
+
+
+@click.command("run")
+@click.argument("model_path")
```
xieofxie: use model_option from utils.cli

src/winml/modelkit/commands/serve.py
```diff
@@ -0,0 +1,181 @@
...
+import sys
+
+import click
+
+
+logger = logging.getLogger(__name__)
+
+
+@click.command("serve")
+@click.argument("model_path", required=False, default=None)
```
xieofxie: use model_option

src/winml/modelkit/commands/serve.py
```diff
@@ -0,0 +1,181 @@
...
+@click.option("--task", default=None, help="Task type (required for raw .onnx files)")
+@click.option(
```
xieofxie: use device_option

src/winml/modelkit/commands/serve.py
```diff
@@ -0,0 +1,181 @@
...
+
+import click
+
+
+logger = logging.getLogger(__name__)
+
+
+@click.command("serve")
+@click.argument("model_path", required=False, default=None)
+@click.option("--task", default=None, help="Task type (required for raw .onnx files)")
```
xieofxie, line 28: should we add common options for others that are reused accros different commands?

src/winml/modelkit/commands/run.py
```diff
@@ -0,0 +1,314 @@
...
+    "--port",
+    default=_DEFAULT_PORT,
+    type=int,
+    show_default=True,
+    help="Port to check for a running winml serve instance (auto-connect)",
+)
+@click.option(
+    "--no-connect",
+    is_flag=True,
+    default=False,
```
xieofxie, line 497: should run not connect by default? Commands do not depend on each other

src/winml/modelkit/serve/app.py
```diff
@@ -0,0 +1,858 @@
...
+                logger.info("Pre-loading model: %s", model_path)
+                async with mgr.borrow(model_path):
+                    pass
+            else:
+                logger.info("Multi-model server started (empty — load via POST /v1/models)")
+            app.state.manager = mgr
+        else:
+            engine = InferenceEngine()
+            engine.load(model_path, task=task, device=device, ep=ep)
+            app.state.manager = SingleModelManager(engine, idle_timeout_sec=idle_timeout_sec)
```
xieofxie, line 147: should we merge them? i.e. SingleModelManager is ModelSlotManager with memory budget = 0. i.e. It could load one but always get limit exceeded

src/winml/modelkit/commands/run.py
DingmaomaoBJTU: Make sense. Let me change it to default off.

src/winml/modelkit/commands/serve.py
DingmaomaoBJTU: Yes, we should do this. Let me create an issue to track it.

DingmaomaoBJTU: https://github.com/microsoft/WinML-ModelKit/issues/369 Issue created

src/winml/modelkit/serve/app.py
DingmaomaoBJTU: Good suggestion! The two classes do share a lot of API surface, and they already implement a shared `ModelManager` Protocol ([manager.py#L43](https://github.com/microsoft/WinML-ModelKit/blob/5f45dab817512c06cf8f3826612aac48b543e92f/src/winml/modelkit/serve/manager.py#L43)).

However, merging them has some practical friction:
1. `SingleModelManager` accepts a pre-loaded engine and serializes via a simple `asyncio.Lock`; `ModelSlotManager` creates engines on-demand with refcount-based concurrency — different lifecycle models.
2. `SingleModelManager.switch_ep()` has no `ModelSlotManager` equivalent.
3. `memory_budget=0` would actually prevent loading any model at all, since `total_memory >= 0` is always true. We'd need a special sentinel (e.g., `budget=inf, max_slots=1`) which adds conditional branches throughout.

I think the current approach — two focused implementations behind a shared Protocol — gives us the right balance of simplicity and extensibility. If the implementations start diverging in ways that cause bugs, we can revisit merging. WDYT?

xieofxie: yes, please keep separate

src/winml/modelkit/commands/serve.py
xieofxie: should we fix it first? so this pr is aligned

DingmaomaoBJTU: Let's use another pr to address this as it will bring cross component changes

scripts/mcp_server.py
```diff
@@ -0,0 +1,420 @@
...
+# ---------------------------------------------------------------------------
+# Entry point
+# ---------------------------------------------------------------------------
+
+
+def main() -> None:
+    """Parse arguments and run the MCP server."""
+    parser = ArgumentParser(description="ModelKit MCP Server (standalone)")
+    parser.add_argument(
+        "--model-url",
```
xieofxie: name is a little wierd

src/winml/modelkit/commands/run.py
```diff
@@ -0,0 +1,812 @@
...
+    show_default=True,
+    help="Output format",
+)
+@click.option(
+    "--output",
+    "-o",
+    default=None,
+    help="Write output to file instead of stdout",
+)
+@click.option(
```
xieofxie, line 480: maybe add a --host later? to support connect to other servers

src/winml/modelkit/commands/serve.py
```diff
@@ -0,0 +1,175 @@
...
+    type=float,
+    show_default=True,
+    help="Memory budget in MB for multi-model manager",
+)
+@click.option(
+    "--reload",
+    is_flag=True,
+    default=False,
+    hidden=True,
+    help="Dev mode: auto-reload on file changes",
```
xieofxie, line 69: --auto-reload?

src/winml/modelkit/data/hub_models.json
xieofxie, line 2: how is it generated? is there readme?

src/winml/modelkit/inference/tasks.py
```diff
@@ -0,0 +1,401 @@
...
+    mapping: PipelineMapping
+
+
+# ---------------------------------------------------------------------------
+# Task Registry
+# ---------------------------------------------------------------------------
+
+TASK_REGISTRY: dict[str, TaskInputSpec] = {
+    # -- Single image -------------------------------------------------------
+    "image-classification": TaskInputSpec(
```
xieofxie, line 241: if the task changes, will we detect code mismatch automatically?

src/winml/modelkit/serve/app.py
```diff
@@ -0,0 +1,846 @@
...
+_log_handler = _RingHandler()
+_log_handler.setFormatter(logging.Formatter("%(message)s"))
+# Attach to modelkit root logger so all sub-loggers feed into the ring
+logging.getLogger("winml.modelkit").addHandler(_log_handler)
+logging.getLogger("winml.modelkit").setLevel(logging.INFO)
+
+# ---------------------------------------------------------------------------
+# Valid EP shorthands accepted by POST /v1/ep
+# ---------------------------------------------------------------------------
+_VALID_EPS = {"cpu", "dml", "qnn", "openvino", "cuda", "auto"}
```
xieofxie: do we support cuda?

src/winml/modelkit/serve/schema.py
```diff
@@ -0,0 +1,171 @@
...
+
+from pydantic import BaseModel, Field
+
+
+# ---------------------------------------------------------------------------
+# Requests
+# ---------------------------------------------------------------------------
+
+
+class EpSwitchRequest(BaseModel):
```
xieofxie, line 23: what does this do? I thought ep is linked with each model load

pyproject.toml
```diff
@@ -51,6 +55,7 @@ dependencies = [
   "torchmetrics[detection]>=1.0",
   "torchvision>=0.24",
   "transformers>=4.57",
+  "uvicorn[standard]>=0.42.0",
```
xieofxie, line 59: we have so many dependencies.. should we split to different optional dependency packs?

pyproject.toml
```diff
@@ -75,6 +80,7 @@ optional-dependencies.dev = [
   "tqdm>=4.67",
   "types-colorama>=0.4.15.20250801",
 ]
+optional-dependencies.audio = [ "soundfile>=0.13" ]
 optional-dependencies.openvino = [ "openvino>=2023" ]
 optional-dependencies.qnn = [
   "onnxruntime-qnn>=1.24.1; python_version>='3.11'",
```
xieofxie, line 87: conflict with onnxruntime-windowsml

src/winml/modelkit/serve/manager.py
```diff
@@ -0,0 +1,502 @@
...
+
+    @asynccontextmanager
+    def borrow(self, model_id: str, task: str | None = None) -> AsyncIterator[InferenceEngine]:
+        """Context manager that yields a ready InferenceEngine.
+
+        Blocks until the engine is available (respects concurrency limits).
+        ``task`` is a routing hint used by ModelSlotManager when ``model_id``
+        is the sentinel ``"_"``.
+        """
+        ...
```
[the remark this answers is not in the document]
DingmaomaoBJTU: Fixed — added explicit `...` after each Protocol method docstring in `manager.py` to satisfy CodeQL.

src/winml/modelkit/serve/manager.py
```diff
@@ -0,0 +1,502 @@
...
+        try:
+            await asyncio.sleep(self._idle_timeout)
+            async with self._lock:
+                if self._engine.is_loaded:
+                    logger.info(
+                        "SingleModelManager: idle timeout (%.0fs) — unloading session",
+                        self._idle_timeout,
+                    )
+                    self._engine.unload()
+        except asyncio.CancelledError:
```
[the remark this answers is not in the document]
DingmaomaoBJTU: Fixed — moved the explanatory comment to its own line above `pass` so CodeQL recognizes it.

src/winml/modelkit/serve/manager.py
```diff
@@ -0,0 +1,502 @@
...
+    async def _expire_slot(self, model_id: str, delay: float) -> None:
+        try:
+            await asyncio.sleep(delay)
+            async with self._lock:
+                slot = self._slots.get(model_id)
+                if slot and slot.refcount == 0:
+                    logger.info("ModelSlotManager: idle timeout — evicting %s", model_id)
+                    slot.engine.unload()
+                    del self._slots[model_id]
+        except asyncio.CancelledError:
```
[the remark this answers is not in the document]
DingmaomaoBJTU: Same fix applied here.

src/winml/modelkit/commands/run.py
DingmaomaoBJTU: Already using `cli_utils.model_option(required=True)` — see run.py line 416.

src/winml/modelkit/commands/serve.py
DingmaomaoBJTU: Already using `cli_utils.model_option(required=False)` — see serve.py line 27.

src/winml/modelkit/commands/serve.py
DingmaomaoBJTU: Already using `cli_utils.device_option(required=False, default="auto", include_auto=True)` — see serve.py line 29.

src/winml/modelkit/commands/run.py
DingmaomaoBJTU: Done — `--connect` is already `default=False` (opt-in only).

scripts/mcp_server.py
DingmaomaoBJTU: Fixed — renamed `--model-url` to `--server-url`, since it's the serve instance address, not a model identifier.

src/winml/modelkit/commands/serve.py
DingmaomaoBJTU: Fixed — renamed to `--auto-reload`.

pyproject.toml
DingmaomaoBJTU: Let me check with zhipeng & zheng about this

pyproject.toml
DingmaomaoBJTU: Let me start a sperate thread in our channel to discuss about this.

src/winml/modelkit/serve/schema.py
DingmaomaoBJTU: This is for hot-swapping the EP on a running server without restarting. The flow is: `POST /v1/ep {"ep": "dml"}` → `engine.switch_ep()` which does `unload() → reload()` with the new EP. Use case: user loads a model on CPU, then wants to try DML/QNN acceleration without restarting the server. It's a Phase 1 single-model feature — only `SingleModelManager` implements it; multi-model mode doesn't expose it (each model's EP is fixed at load time).

src/winml/modelkit/serve/app.py
DingmaomaoBJTU: Removed. CUDA is not supported — dropped from `_VALID_EPS`, `EpSwitchRequest` description, and the dashboard UI.

DingmaomaoBJTU: cuda removed

src/winml/modelkit/inference/tasks.py
DingmaomaoBJTU: Currently no — `TASK_REGISTRY` is a static dict maintained manually. If HF upstream renames a task or changes its input signature, we wouldn't detect it automatically.

Two practical mitigations already in place:
1. Unregistered tasks still work — they fall back to heuristic parsing (`_parse_heuristic`) and raw tensor inference, so nothing breaks silently.
2. The `test_run_task_combos.py` tests import `TASK_REGISTRY` and exercise each task's schema/coercion, so a removed or renamed task would surface as a test failure.

DingmaomaoBJTU: To add context: these task definitions are very stable upstream — HF hasn't changed existing task input signatures in years. New tasks get added occasionally but existing ones don't break. And even for unregistered tasks, inference still works via heuristic fallback — just without the `--file`/`--text` shortcut convenience. So the practical risk is low.

src/winml/modelkit/data/hub_models.json
DingmaomaoBJTU: No readme yet. This file is generated from our weekly evaluation data — there's an agent on that side that curates the model list based on test results. We'll formalize the release/update process later.

src/winml/modelkit/commands/run.py
DingmaomaoBJTU: Makes sense — currently `--connect` is hardcoded to `localhost:{port}`. Added.

xieofxie (approved):

src/winml/modelkit/serve/app.py
```diff
@@ -0,0 +1,846 @@
...
+                # Collision: inputs vs params
+                collision = set(merged.keys()) & set(pipe_params.keys())
+                if collision:
+                    key = sorted(collision)[0]
+                    raise ValueError(
+                        f"'{key}' specified in both inputs and params. "
+                        "Use inputs for model inputs and params for pipeline parameters."
+                    )
+
+                return await loop.run_in_executor(
```
timenick, line 297: **Concurrency bug**: `ModelSlotManager.borrow()` only maintains a refcount and does not hold an exclusive lock during inference. When two concurrent requests target the same model in multi-model mode, both get the same engine and `engine.predict()` runs concurrently in the shared executor.

This violates the explicit contract at `inference/engine.py:108`:
> Not thread-safe on its own — callers (SingleModelManager, winml run) must ensure exclusive access before calling predict() or switch_ep().

Concrete races under concurrent load:
- `self._request_count += 1` at `engine.py:248` — lost updates
- `self._latency_samples.append(...)` at `engine.py:247` — deque mutation from multiple threads
- HF `Pipeline.__call__` is not documented as thread-safe

`SingleModelManager` is fine because its `asyncio.Lock` is held across `yield` → `run_in_executor`. `ModelSlotManager` needs the same guarantee — e.g. add a per-slot `asyncio.Lock` to `ModelSlot` and acquire it inside `borrow()` before yielding.

This is the Phase 3 core scenario (concurrent requests across models), so it will trigger on day one.

🤖 Generated with Claude Code

tests/unit/serve/test_predict_file_inputs.py
```diff
@@ -0,0 +1,275 @@
...
+
+import base64
+import json
+from contextlib import asynccontextmanager
+from unittest.mock import MagicMock
+
+from fastapi import FastAPI
+from fastapi.testclient import TestClient
+
+from winml.modelkit.inference.tasks import TASK_REGISTRY, InputField
```
timenick: Per CLAUDE.md's Import Rules, test code must import public symbols from the package level, not reach into internal submodules:

```python
# Current — bypasses public API
from winml.modelkit.inference.tasks import TASK_REGISTRY, InputField
from winml.modelkit.inference.types import Prediction, PredictionResult

# Expected
from winml.modelkit.inference import TASK_REGISTRY, InputField, Prediction, PredictionResult
```

All four symbols are already exported from `winml.modelkit.inference.__init__`. The same fix applies to:
- `tests/unit/commands/test_run_spec.py:30`
- `tests/unit/commands/test_run_task_combos.py:42`
- `tests/unit/serve/test_predict_json.py:22-23`

`_`-prefixed imports (`_register_routes`, `_format_text`, etc.) are fine — CLAUDE.md explicitly allows them. Only the public symbols need to go through the package.

🤖 Generated with Claude Code

src/winml/modelkit/commands/run.py
```diff
@@ -0,0 +1,822 @@
...
+                logger.debug("Auto-connected to winml serve at %s", base_url)
+                return resp.json()
+
+            # Route 2: no file → JSON /v1/predict with named inputs
+            #   Coerce raw CLI strings (JSON arrays, numbers, booleans)
+            #   so the server receives properly typed values.
+            inputs: dict[str, Any] = {}
+            inputs.update({k: _parse_heuristic(v) for k, v in raw_inputs.items()})
+            if text is not None:
+                inputs["text"] = text
```
timenick: When `--connect --text "..."` is used against a task whose text field isn't named `text` (e.g. `question-answering` expects `question`/`context`), this path sends `{"inputs": {"text": "..."}}` to the server, which responds with a 400 "Unknown input".

This diverges from the embedded path: `_resolve_shortcuts()` has full schema-driven resolution and raises a clear `--text is ambiguous for this task (has 2 text inputs: question, context)` error before inference. The auto-connect path silently generates an opaque server error.

Two ways to fix:
1. Probe `GET /v1/schema` after the health check and run the same resolution logic as `_resolve_shortcuts` before forming the payload.
2. Forward through `/v1/predict/file` form-encoding consistently (the server's `_build_file_inputs` already does schema-driven name resolution), even without a file — or add a `/v1/predict/text` shortcut on the server.

Option 1 is the smaller change. Given `--connect` targets first-class UX, the silent divergence feels worth fixing.

🤖 Generated with Claude Code

src/winml/modelkit/serve/static/index.html
```diff
@@ -0,0 +1,2592 @@
...
+    const statusColor = m.status === 'ready' ? 'var(--green)' : m.status === 'loading' ? 'var(--yellow)' : 'var(--text2)';
+    return `
+    <div class="model-card" id="mc-${cssId(m.model_id)}">
+      <div class="model-card-header">
+        <span style="width:8px;height:8px;border-radius:50%;background:${statusColor};flex-shrink:0;display:inline-block"></span>
+        <span class="model-name" title="${escHtml(m.model_id)}">${escHtml(m.model_id)}</span>
+        <span class="tag tag-blue" style="font-size:10px">${m.task||'—'}</span>
+        <span class="tag tag-yellow" style="font-size:10px">${m.ep||m.device||'—'}</span>
+        <button class="btn btn-outline btn-sm" onclick="goInferWithModel('${escHtml(m.model_id)}')">▶ Infer</button>
+        <button class="btn btn-outline btn-sm" onclick="runBench('${escHtml(m.model_id)}',this)">⚡ Bench</button>
```
timenick: `escHtml` escapes only `&`, `<`, `>` — it does not escape single quotes or backslashes, so the JS string context in these `onclick` attributes is unsafe:

```html
<button onclick="goInferWithModel('${escHtml(m.model_id)}')">▶ Infer</button>
<button onclick="runBench('${escHtml(m.model_id)}',this)">⚡ Bench</button>
<button onclick="unloadModel('${escHtml(m.model_id)}')">✕</button>
```

A `model_id` containing `'` or `\` breaks out of the JS literal. Since `POST /v1/models` accepts arbitrary `model_id` from the client, this is reachable: an attacker with access to the local API (CORS is `*`) can load a model with a crafted id, then the UI executes JS when rendering.

Same pattern at `_renderHubCard` (line 1407) and several `${m.task||'—'}`, `${m.status}` interpolations on lines 1278, 1287 that aren't escaped at all.

Recommended fix — use data attributes + `addEventListener`, which sidesteps both HTML and JS-string escaping:
```html
<button class="btn-infer" data-model-id="${escHtml(m.model_id)}">▶ Infer</button>
```
```js
container.addEventListener('click', e => {
  if (e.target.classList.contains('btn-infer')) {
    goInferWithModel(e.target.dataset.modelId);
  }
});
```

Severity is low (localhost-only, trusted model sources in practice), but defense-in-depth matters here because the bind address and CORS policy are both permissive.

🤖 Generated with Claude Code

src/winml/modelkit/serve/static/index.html
```diff
@@ -0,0 +1,2735 @@
...
+    bodyHtml = `<pre style="font-size:11px;color:var(--text2);overflow:auto;max-height:200px;background:var(--surface2);padding:8px;border-radius:5px;white-space:pre-wrap;word-break:break-all">${escHtml(JSON.stringify(preds, null, 2))}</pre>`;
+  }
+
+  card.innerHTML = `
+    <div class="result-header">
+      <div class="result-meta">
+        <span class="tag tag-blue">${data.task||'unknown'}</span>
+        <span class="tag tag-green">${data.device||''}</span>
+        ${data.ep?`<span class="tag tag-yellow">${data.ep}</span>`:''}
+        ${data.model_id?`<span style="font-size:11px;color:var(--text2)">${data.model_id}</span>`:''}
```
timenick, line 1318: These four fields are interpolated into `innerHTML` without `escHtml()`, inconsistent with the rest of this file (e.g., `renderForm` / `buildPipelineParams` wrap every interpolation in `escHtml`). In `--multi` mode, `model_id` (and alias/description) are user-controlled via `POST /v1/models`, so an attacker-supplied model id with HTML chars would render unescaped. Low risk on a localhost dev server, but cheap to fix:

```js
<span class="tag tag-blue">${escHtml(data.task||'unknown')}</span>
<span class="tag tag-green">${escHtml(data.device||'')}</span>
${data.ep?`<span class="tag tag-yellow">${escHtml(data.ep)}</span>`:''}
${data.model_id?`<span style="font-size:11px;color:var(--text2)">${escHtml(data.model_id)}</span>`:''}
```

🤖 Generated with Claude Code

src/winml/modelkit/commands/serve.py
```diff
@@ -0,0 +1,175 @@
...
+        import uvicorn
+    except ImportError:
+        click.echo(
+            "Error: uvicorn is required. Install with: pip install uvicorn[standard]",
+            err=True,
+        )
+        sys.exit(1)
+
+    if ctx.obj and ctx.obj.get("debug"):
+        logging.getLogger("modelkit").setLevel(logging.DEBUG)
```
timenick, line 123: Wrong logger name — the rest of the codebase uses `winml.modelkit` as the root logger (e.g., the telemetry module). `logging.getLogger("modelkit")` returns an unrelated, empty logger, so `--debug` has no effect on the serve path.

```python
logging.getLogger("winml.modelkit").setLevel(logging.DEBUG)
```

🤖 Generated with Claude Code

src/winml/modelkit/commands/serve.py
```diff
@@ -0,0 +1,175 @@
...
+        ep=ep,
+    )
+
+    uvicorn.run(
+        inference_app,
+        host=host,
+        port=port,
+        reload=auto_reload,
+        log_level="warning",
+    )
```
timenick, line 175: `uvicorn.run(app_instance, reload=True)` silently disables reload — uvicorn requires an import string (e.g. `"winml.modelkit.serve.app:create_app"`) when `reload=True`. Since `--auto-reload` is `hidden=True` and dev-only this is cosmetic, but whoever finds the flag will think it's broken. Same issue at line 136 (Phase 0 path).

Two options:
- Drop `reload=auto_reload` from both `uvicorn.run` calls (removes the dead flag entirely), or
- When `auto_reload=True`, switch to a factory entry point: `uvicorn.run("winml.modelkit.serve.app:create_app", factory=True, host=host, port=port, reload=True, ...)`.

🤖 Generated with Claude Code

timenick (commented):
## Import rule violations (CLAUDE.md)

Per the CLAUDE.md source-code rule, imports in `src/` must route through the package's `__init__.py` public API. `inference/__init__.py` already exports `BINARY_TYPES`, `TASK_REGISTRY`, and `PredictionResult` in `__all__`, but these spots bypass it and reach into submodules:

- `src/winml/modelkit/commands/run.py` lines 69, 151, 348 — `from ..inference.tasks import BINARY_TYPES` → `from ..inference import BINARY_TYPES`
- `src/winml/modelkit/serve/app.py` line 42 — `from ..inference.types import PredictionResult` → `from ..inference import PredictionResult`
- `src/winml/modelkit/serve/app.py` lines 722, 777, 801 — `from ..inference.tasks import TASK_REGISTRY/BINARY_TYPES` → `from ..inference import ...`
- `src/winml/modelkit/serve/schema_generator.py` line 17 — `from ..inference.tasks import BINARY_TYPES, TASK_REGISTRY` → `from ..inference import ...`

`run.py:610` already uses `from ..inference import InferenceEngine` correctly — the newer spots just need to match that pattern.

🤖 Generated with Claude Code

src/winml/modelkit/serve/static/index.html
DingmaomaoBJTU: Fixed — wrapped all four interpolations with `escHtml()`.

src/winml/modelkit/commands/serve.py
DingmaomaoBJTU: Fixed — changed to `logging.getLogger("winml.modelkit")` to match the rest of the codebase.

src/winml/modelkit/commands/serve.py
DingmaomaoBJTU: Fixed — removed `--auto-reload` flag and `reload=` param from both `uvicorn.run` calls since it silently does nothing with an app instance.

timenick (approved):

PR #357, opened 2026-04-15
kill timeout process

fangyangci:
kill timeout process
add pattern
move case_runner to modelkitArtifacts


src/winml/modelkit/analyze/runtime_checker/runner.py
chinazhangchao: The shutdown() method calls self.executor.shutdown(wait=True), but if a worker is hung, wait=True could block indefinitely. Consider applying the same kill-lingering-processes logic there, or at least adding a timeout.

chinazhangchao (approved):

chinazhangchao (approved):

PR #361, opened 2026-04-16
Remove local run check when optimize.

chinazhangchao:
- Test sam2.1-hiera-large analyze("total_operators": 1081) on qnn device. Time deduction 68%.
91.99s -> 29.37s

- Remove local run on AMD devcie.
- Fix file name case issue.

### DIFF
--- a/src/winml/modelkit/build/common.py
+++ b/src/winml/modelkit/build/common.py
@@ -143,6 +143,7 @@ def _run_analyze_loop(
                 iter_model,
                 ep=ep,
                 device=device,
+                run_unknown_op=False,
                 on_ep_start=on_ep_start,
                 on_node_result=on_node_result,
             )
@@ -188,6 +189,7 @@ def _run_analyze_loop(
             iter_model,
             ep=ep,
             device=device,
+            run_unknown_op=False,
             on_node_result=lambda _: None,
         )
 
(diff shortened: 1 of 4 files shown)


src/winml/modelkit/commands/analyze.py
```diff
@@ -513,6 +513,10 @@ def analyze(
         live: Live | None = None
         ep_counter = 0
 
+        run_unknown_op_for_ep = run_unknown_op
+        if ep == "VitisAIExecutionProvider":
+            run_unknown_op_for_ep = False
```
xieofxie, line 518: add log about why disabling this? So user knows why it is always disabled

src/winml/modelkit/build/common.py
```diff
@@ -143,6 +143,7 @@ def _run_analyze_loop(
                 iter_model,
                 ep=ep,
                 device=device,
+                run_unknown_op=False,
```
DingmaomaoBJTU, line 146: Not sure whether we should disable rerun here - it may impact auto-conf result if user use build cmd directly.

chinazhangchao: Discuss design later.

DingmaomaoBJTU (approved):

PR #365, opened 2026-04-17
Fix unhashable cache key error.

chinazhangchao:
Some cases the cache key is unhashable. For example:

- There's no column files (e.g. AMD), so the conditions is not filtered and there are some `nparray` type values in the conditions which is not hashable.
- Some ops (e.g. CumSum) have `nparray` type values in the conditions. The root cause investigation is still in progress.


DingmaomaoBJTU (approved):

chinazhangchao:
Replace with https://github.com/microsoft/WinML-ModelKit/pull/366

PR #367, opened 2026-04-17
feat(telemetry): Phase 1 — OneCollector library + device ID foundation

timenick:
## Summary

PR 1 of 4 in the ModelKit telemetry rollout. Introduces the transport foundation:

- **`telemetry/library/`** — a custom `opentelemetry-sdk` log exporter that speaks Common Schema 4.0 to the OneCollector endpoint.
- **`telemetry/deviceid/`** — a SHA256-hashed UUID persisted per machine, used as the telemetry `device_id`.

Emission-free by design: no `Telemetry` singleton, no consent prompt, no CLI wiring yet. User-facing behavior is unchanged.

## What's next

- **PR 2** (blocked on Privacy sign-off of prompt text + event schemas): `constants.py` + `utils.py` (exception / PII scrubbing) + `consent.py` (first-run prompt) + `telemetry.py` (singleton).
- **PR 3**: `ActionGroup` auto-instrumentation + `cli.py` integration + `--disable-telemetry` flag.
- **PR 4**: `docs/Privacy.md` + README telemetry section.

PR #371, opened 2026-04-21
feat(telemetry): Phase 2 — core telemetry logic (consent, scrub, singleton)

timenick:
## Summary

Phase 2 of the [telemetry rollout](../blob/main/docs/superpowers/specs/2026-04-17-modelkit-telemetry-design.md). Builds on Phase 1 (`library/` + `deviceid/`, merged in #367) with the ModelKit-specific telemetry layer.

- `constants.py` — empty `INSTRUMENTATION_KEY` placeholder; official builds inject the real iKey before packaging. No runtime / env-var override.
- `utils.py` — exception scrubbing pipeline (path trim → 200-char cap → PII regex for email / GUID / IPv4 / IPv6 / long-token), structured traceback extraction (only `{file, line, function}`), base64 cache encoding, `msvcrt` exclusive file lock.
- `consent.py` — first-run prompt (accept-by-default `[Y/n]`), JSON config at `%USERPROFILE%\.modelkit\config.json` with atomic writes that preserve unrelated keys, CI auto-detection via `CI` / `TF_BUILD` / `GITHUB_ACTIONS` / etc. (fail-closed in non-interactive environments).
- `telemetry.py` — lazy singleton that wires `LoggerProvider` + `Resource`, exposes `log_heartbeat` / `log_action` / `log_error` with per-event allowlist enforcement, `self._logger is None` guards, defensive `try/except` on every public method, idempotent `shutdown`.

Still **not wired to the CLI** — `ActionGroup` auto-instrumentation lands in Phase 3.

## Test results

- `uv run pytest tests/unit/telemetry/` → **105 PASSED** (39 from Phase 1 + 66 new).
- `uv run pytest tests/` → no regressions.
- `uv run python -c "import winml.modelkit.telemetry"` → no side effects (iKey empty → `Telemetry.disabled` stays `True`).
- `uvx ruff check src/winml/modelkit/telemetry/ tests/unit/telemetry/` → clean.

## Demo — captured wire format

A local-only preview harness (`winml.modelkit.telemetry._demo`, not shipped) monkey-patches the iKey, points consent at a temp config, and intercepts `requests.Session.post` so the exporter serializes CS 4.0 envelopes to stdout instead of the network. Output below (device ID + session ID redacted):

<details>
<summary>First-run consent prompt text</summary>

```
ModelKit can collect anonymous usage data to help improve the product.

What is collected:
  - Command name, duration, success/failure
  - Target device/EP (when the command specifies them)
  - OS, architecture, ModelKit version
  - Unhandled exception types, code locations, and scrubbed error
    messages (paths trimmed, length capped, PII patterns scrubbed)

What is never collected:
  - File paths, model contents, command arguments, credentials

Enable telemetry? [Y/n]:
```

</details>

<details>
<summary>ModelKitHeartbeat envelope</summary>

```json
{
  "ver": "4.0",
  "name": "ModelKitHeartbeat",
  "time": "2026-04-21T07:12:41.603Z",
  "iKey": "o:fake-demo-tenant-token",
  "data": {},
  "ext": {
    "device": {
      "localId": "<redacted-sha256>",
      "authId": "EXISTING",
      "deviceClass": "AMD64"
    },
    "os": {
      "name": "Windows",
      "ver": "10.0.26310",
      "release": "10"
    },
    "app": {
      "ver": "0.0.1",
      "sesId": "<redacted-uuid>",
      "initTs": 1776755561.4647715
    }
  }
}
```

</details>

<details>
<summary>ModelKitAction envelope — happy path (build on NPU)</summary>

Extra kwargs outside the allowlist (e.g. `leaked_field`) are dropped before emission.

```json
{
  "ver": "4.0",
  "name": "ModelKitAction",
  "time": "2026-04-21T07:12:41.603Z",
  "iKey": "o:fake-demo-tenant-token",
  "data": {
    "invoked_from": "Interactive",
    "action_name": "build",
    "device": "NPU",
    "ep": "QNNExecutionProvider",
    "duration_ms": 4321,
    "success": true
  },
  "ext": { "...": "same as heartbeat" }
}
```

</details>

<details>
<summary>ModelKitAction envelope — failure path</summary>

```json
{
  "ver": "4.0",
  "name": "ModelKitAction",
  "time": "2026-04-21T07:12:41.603Z",
  "iKey": "o:fake-demo-tenant-token",
  "data": {
    "invoked_from": "Interactive",
    "action_name": "quantize",
    "device": null,
    "ep": null,
    "duration_ms": 87,
    "success": false
  },
  "ext": { "...": "same as heartbeat" }
}
```

</details>

<details>
<summary>ModelKitError envelope — scrubbed message + structured stack</summary>

Original exception was `FileNotFoundError(r"model file not found: C:\Users\Alice\models\mobilenet.onnx")`. Path is outside the package root, so `_trim_path` falls back to basename. Real ModelKit errors typically only hit the path rule; the defensive regex rules are shown separately below.

```json
{
  "ver": "4.0",
  "name": "ModelKitError",
  "time": "2026-04-21T07:12:41.604Z",
  "iKey": "o:fake-demo-tenant-token",
  "data": {
    "exception_type": "FileNotFoundError",
    "exception_message": "model file not found: mobilenet.onnx",
    "exception_stack": [
      {
        "file": "winml/modelkit/telemetry/_demo.py",
        "line": 128,
        "function": "_run_demo"
      }
    ]
  },
  "ext": { "...": "same as heartbeat" }
}
```

</details>

<details>
<summary>PII scrubbing pipeline — rule-by-rule examples</summary>

Pipeline order: per-token path trim → 200-char length cap → PII regex sweep. Each rule in isolation:

| Rule | Input | Output |
|---|---|---|
| Path (inside package) | `C:\Users\Alice\src\winml\modelkit\commands\build.py` | `winml/modelkit/commands/build.py` |
| Path (outside package) | `C:\Users\Alice\models\mobilenet.onnx` | `mobilenet.onnx` |
| Email | `failed to reach alice@example.com` | `failed to reach <scrubbed>` |
| GUID | `job 12345678-1234-5678-1234-567812345678 failed` | `job <scrubbed> failed` |
| IPv4 | `connection to 10.0.0.1 refused` | `connection to <scrubbed> refused` |
| IPv4 (invalid octet — preserved) | `expected version 1.18.999.1` | `expected version 1.18.999.1` |
| IPv6 | `bind 2001:db8::1 failed` | `bind <scrubbed> failed` |
| Long token (has digit) | `Bearer example_token_1234567890abcdefghij` | `Bearer <scrubbed>` |
| Long token (no digit — preserved) | `WinMLImageFeatureExtractionEvaluator` | `WinMLImageFeatureExtractionEvaluator` |

The path rule is what typical ModelKit errors hit. Email / IPv4 / IPv6 / GUID / long-token rules are defensive — they catch PII that might leak through from lower-level libraries (stdlib URL errors, Windows API GUIDs, upstream HTTP clients), not patterns ModelKit itself produces.

</details>


src/winml/modelkit/telemetry/utils.py
```diff
@@ -0,0 +1,205 @@
...
+    #   1) compressed form with "::" (e.g. 2001:db8::1)
+    #   2) full uncompressed form (e.g. 2001:db8:85a3:0:0:8a2e:370:7334)
+    re.compile(
+        r"\b[0-9a-fA-F]{1,4}(?::[0-9a-fA-F]{0,4}){1,7}::[0-9a-fA-F:]*\b"
+        r"|\b(?:[0-9a-fA-F]{1,4}:){2,7}[0-9a-fA-F]{1,4}\b"
+        r"|\b[0-9a-fA-F]{1,4}::[0-9a-fA-F:]*\b"
+    ),
+    # Long opaque token (API keys, JWTs, base64 blobs). Runs last so it
+    # doesn't swallow already-scrubbed markers.
+    re.compile(r"\b[A-Za-z0-9_\-]{24,}\b"),
```
DingmaomaoBJTU: The long-token regex `[A-Za-z0-9_\-]{24,}` will match legitimate Python class and function names that commonly appear in exception messages. For example:

- `WinMLImageFeatureExtractionEvaluator` (37 chars)
- `BatchLogRecordProcessor` (23 chars — just under, but close)
- `WinMLTextClassificationEvaluator` (32 chars)

An error like `TypeError: WinMLImageFeatureExtractionEvaluator.compute() got unexpected keyword argument` would have the class name scrubbed to `<scrubbed>.compute()`, making the error event much less useful for debugging.

Consider tightening the heuristic — e.g., requiring at least one digit or special char (`[A-Za-z0-9_\-]*\d[A-Za-z0-9_\-]*` with length >= 24), or raising the minimum length to 32+.

src/winml/modelkit/telemetry/utils.py
```diff
@@ -0,0 +1,205 @@
...
+_PII_PATTERNS: list[re.Pattern[str]] = [
+    # Email
+    re.compile(r"[a-zA-Z0-9._%+\-]+@[a-zA-Z0-9.\-]+\.[a-zA-Z]{2,}"),
+    # GUID (8-4-4-4-12 hex)
+    re.compile(
+        r"\b[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}"
+        r"-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}\b"
+    ),
+    # IPv4
+    re.compile(r"\b(?:\d{1,3}\.){3}\d{1,3}\b"),
```
DingmaomaoBJTU: The IPv4 pattern `(?:\d{1,3}\.){3}\d{1,3}` also matches version strings like `1.18.0.1`, `3.11.0.0`, `0.0.1.0` — these are common in ONNX Runtime and Python version-related error messages. For an ML toolkit, version numbers in errors are useful diagnostic info that probably shouldn't be scrubbed.

A stricter check could validate octet ranges (0–255) or require a preceding space/punctuation boundary that isn't `.` to avoid matching `x.y.z.w` version patterns.

src/winml/modelkit/telemetry/consent.py
```diff
@@ -0,0 +1,173 @@
...
+
+import json
+import os
+import sys
+import tempfile
+from pathlib import Path
+from typing import Literal
+
+
+_CONFIG_PATH: Path = Path(os.environ.get("USERPROFILE", "")) / ".modelkit" / "config.json"
```
DingmaomaoBJTU: If `USERPROFILE` is not set (unusual on Windows, but possible in containers or minimal service accounts), this resolves to `Path("") / ".modelkit" / "config.json"` — a **relative** path `.modelkit/config.json` that writes into the CWD.

Since WinML is Windows-only and `USERPROFILE` is virtually always set, this is low-risk, but the failure mode is surprising. A defensive option:

```python
_user_home = os.environ.get("USERPROFILE") or os.path.expanduser("~")
_CONFIG_PATH = Path(_user_home) / ".modelkit" / "config.json"
```

`expanduser("~")` falls back to `HOMEDRIVE + HOMEPATH` or `HOME` on Windows, covering more edge cases.

src/winml/modelkit/telemetry/telemetry.py
```diff
@@ -0,0 +1,207 @@
...
+class Telemetry:
+    """Process-wide telemetry singleton. Use :meth:`get_or_init` to access."""
+
+    def __init__(self) -> None:
+        self._logger = None  # set when enabled; None when disabled
+        self._provider = None
+        self._disabled = True  # set to False only after successful init
+        self._init_ts = time.time()
+        self._app_instance_id = str(uuid.uuid4())
+        self._invoked_from = "Script" if not sys.stdin.isatty() else "Interactive"
```
DingmaomaoBJTU, line 82: Minor observation: `_invoked_from = "Script"` is effectively dead data.

`resolve_consent()` returns `"disabled"` when `sys.stdin.isatty()` is `False` (line 2 in its precedence). So whenever `_invoked_from` would be `"Script"`, telemetry is disabled and no events are emitted. The field only ever reaches the wire as `"Interactive"`.

This is fine if the intent is forward-compatibility (Phase 3 may change the consent model to honor stored consent in non-TTY mode), but worth a brief comment so future readers don't wonder whether it's tested.

src/winml/modelkit/telemetry/telemetry.py
```diff
@@ -0,0 +1,207 @@
...
+
+        filtered = _filter_allowlist(event_name, attrs)
+        record = LogRecord(
+            timestamp=time.time_ns(),
+            body=event_name,
+            attributes=filtered,
+        )
+        self._logger.emit(record)
+
+    def shutdown(self) -> None:
```
DingmaomaoBJTU, line 208: No `atexit` handler is registered to call `shutdown()`. If the CLI process exits without an explicit `shutdown()` call, `BatchLogRecordProcessor` may drop pending events that haven't been flushed.

Presuming Phase 3's ActionGroup will own the lifecycle (and call `shutdown` in its `__exit__`), this is fine for now. But until then, any manual `Telemetry.get_or_init()` usage risks losing the last batch. Worth a TODO or a note in the docstring.

DingmaomaoBJTU (commented):
Thorough Phase 2 implementation. Clean separation of concerns (consent, scrubbing, singleton), defensive error handling throughout, and solid test coverage (66 tests). A few items worth discussing below.

timenick:
Thanks for the review — all 5 points addressed across commits ff41e35 + 7288707:

**#3 (long-token regex):** Added `(?=[A-Za-z0-9_\-]*\d)` lookahead so the pattern only matches if there's at least one digit. Pythonic class/function names (no digits) pass through untouched; API keys / JWTs / SHA hashes / base64 blobs still match. Added regression test.

**#4 (IPv4 vs version strings):** Added octet range validation (0-255) — obvious nonsense like `999.888.777.666` now passes through. The 4-part version trade-off (`1.18.0.1`) remains: regex alone can't distinguish, and the spec's "conservative over leaky" philosophy tips toward scrubbing. Documented in a comment. Happy to revisit if Privacy pushes back either way.

**#5 (USERPROFILE missing):** Fall back to `HOMEDRIVE + HOMEPATH` (both Windows-native). ModelKit is Windows-only so avoiding `os.path.expanduser("~")`, which would pull in POSIX `HOME` semantics.

**#6 (`_invoked_from = "Script"` dead path):** Added inline comment explaining it's forward-compat — today consent returns "disabled" for non-TTY, so only "Interactive" reaches the wire. Schema stays stable if Phase 3 relaxes that rule.

**#7 (no `atexit` for shutdown):** Added a note in `Telemetry`'s class docstring that the caller owns lifecycle and Phase 3's `ActionGroup` will own the teardown call. Deferring the actual `atexit` wiring to Phase 3 where it naturally lives.

src/winml/modelkit/telemetry/consent.py
```diff
@@ -0,0 +1,186 @@
+# -------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+# --------------------------------------------------------------------------
+
+r"""Consent decision for ModelKit telemetry.
+
+A first-run interactive prompt collects user consent (default: accept)
+and persists it to ``%USERPROFILE%\.modelkit\config.json``. This module
```
xieofxie, line 9: why we persist consent in file but device id in reg?

src/winml/modelkit/telemetry/consent.py
```diff
@@ -0,0 +1,186 @@
...
+    "CI",
+    "TF_BUILD",
+    "GITHUB_ACTIONS",
+    "JENKINS_URL",
+    "CODEBUILD_BUILD_ID",
+    "BUILDKITE",
+    "SYSTEM_TEAMFOUNDATIONCOLLECTIONURI",
+)
+
+Consent = Literal["enabled", "disabled"]
```
xieofxie, line 76: Should we include a version? So if we update our consent notice, user could re-consent

Or a version list if user install different versions of model kit with different consent documents

src/winml/modelkit/telemetry/consent.py
timenick: Include consent in the config file so users can easily opt in or out of telemetry collection. Device ID is an opaque per-install identifier. Storing the device ID in the registry keeps it hidden and aligns with the convention used by other Microsoft development tools for small per-user machine identifiers.

src/winml/modelkit/telemetry/consent.py
timenick: Good call — "we changed what we collect, users should re-consent" is a real scenario.

Proposal: add a consent_version: int alongside consent. Schema becomes {"telemetry": {"consent": "enabled", "consent_version": 1}}. If stored version < current, _read_stored_consent() returns None and we re-prompt; old installs that have only consent (no version) are treated as v0 and also re-prompted on the next bump. Tiny forward-compatible extension.

On the per-version list for side-by-side installs — I'd skip it. Users don't typically run multiple ModelKit versions side-by-side, and if they did, the newest version's consent notice is the broadest anyway. Over-engineering for a case we haven't seen.

Happy to add consent_version in this PR. 

Please review this proposal. Thanks. @xieofxie

xieofxie: If this is the case, in our notice we should include something like that the consent covers all previous versions of consent so user knows that new consent convers old ones.

Because I don't know if `the newest version's consent notice is the broadest anyway` is always true

timenick: Fair pushback. Simpler framing: _PROMPT_TEXT always describes the full current scope (not a delta), so whatever vN's notice lists is exactly what the user is consenting to. Supersession is implicit — no extra wording needed.

Plan stays minimal: consent_version: int + re-prompt on bump. Let me know if you see a case that still confuses the user.

xieofxie: It makes sense. We need to document this about how we construct _PROMPT_TEXT

src/winml/modelkit/telemetry/consent.py
```diff
@@ -0,0 +1,207 @@
...
+    return drive + path  # empty string if neither is set
+
+
+_CONFIG_PATH: Path = Path(_resolve_user_home()) / ".modelkit" / "config.json"
+
+# Bumped when the consent notice materially changes (new data category,
+# changed scope). Stored records with an older version are treated as
+# unrecorded on read so the user sees the updated notice and re-consents.
+# Records predating this field are grandfathered as the current version.
+_CONSENT_VERSION: int = 1
```
xieofxie, line 49: I feel we had better move _PROMPT_TEXT  up to be near to this

xieofxie (approved):
Trust CC

DingmaomaoBJTU (approved):

PR #374, opened 2026-04-21
Fangyangci/simplify rules zip

fangyangci:
1. Change update_zip to use a snapshot generation model of full baseline plus incremental updates, reducing duplicate data across opsets and shrinking rules_zip size.
2. Unify the update entry point to just two parameters, ep + device, and remove the external version parameter to reduce misuse risk and ensure incremental-chain consistency.
3. Move op_check_results/rules to rules_zip


xieofxie:
could you explain more about baseline and delta?
- For example we have version 12, 13, 14, do 13, 14 both base on 12 or 13 on 12, 14 on 13?
- delta is per test case CRUD?
Thanks

fangyangci:
> could you explain more about baseline and delta?
> 
> * For example we have version 12, 13, 14, do 13, 14 both base on 12 or 13 on 12, 14 on 13?
> * delta is per test case CRUD?
>   Thanks

As each op has different since_version
every version is a diff with the previous version.  13 on 12, 14 on 13...

Load v22= Load v22+21+...+12

delta is NOT test case level.
it is op version level, this PR is try to remove the same op_check_result save in multi version.

xieofxie:
> > could you explain more about baseline and delta?
> > 
> > * For example we have version 12, 13, 14, do 13, 14 both base on 12 or 13 on 12, 14 on 13?
> > * delta is per test case CRUD?
> >   Thanks
> 
> As each op has different since_version every version is a diff with the previous version
> 
> Load v22= Load v22+21+...+12
> 
> delta is NOT test case level. it is op version level, this PR is try to remove the same op_check_result save in multi version.

In this sense, it is more easy to just convert raw result op data json to pandas dataframe json or other format and then load the since version in code? It could be even faster since we only load one needed. No need to combine them

fangyangci:
> > > could you explain more about baseline and delta?
> > > 
> > > * For example we have version 12, 13, 14, do 13, 14 both base on 12 or 13 on 12, 14 on 13?
> > > * delta is per test case CRUD?
> > >   Thanks
> > 
> > 
> > As each op has different since_version every version is a diff with the previous version
> > Load v22= Load v22+21+...+12
> > delta is NOT test case level. it is op version level, this PR is try to remove the same op_check_result save in multi version.
> 
> In this sense, it is more easy to just convert raw result op data json to pandas dataframe json or other format and then load the since version in code? It could be even faster since we only load one needed. No need to combine them

our runtime queries many nodes/operators in a model, current combined snapshots are often better because they reduce file-open overhead and reuse in-memory data.
Any way, I don't think the performance bottleneck is load json file.

timenick:
Could you please update your PR title so it's easier to read? Please refer to other PRs' title format. Thanks.

chinazhangchao (approved):

PR #377, opened 2026-04-22
Fix local run qdq op.

chinazhangchao:
https://github.com/microsoft/WinML-ModelKit/issues/182


DingmaomaoBJTU:
### Code review

Found 1 issue:

1. `logger.warning` for an expected build failure should be `logger.debug` — this reverses a deliberate downgrade from a prior fix. When shape inference is incomplete (common for large QDQ models), the model-build exception fires for many nodes and floods the log. The equivalent handler on `main` uses `logger.debug`.

https://github.com/microsoft/WinML-ModelKit/blob/769a7b60a481dffa976a2ab2cc010475d3315bf8/src/winml/modelkit/analyze/core/runtime_checker_query.py#L1804-L1812

DingmaomaoBJTU (approved):

DingmaomaoBJTU (approved):

PR #379, opened 2026-04-22
Skip loading external weight data during static analysis.

chinazhangchao:
### Problem

Running `winml analyze` on large models with external data (e.g., Qwen3-8B with a 30.5 GB `.data` sidecar) causes the process to consume all available memory and disk, hanging indefinitely. The analyzer called `onnx.load(path, load_external_data=True)`, loading the entire weight file into RAM despite never inspecting weight values.

### Root Cause

The static analyzer only needs graph structure (operator types, shapes, connectivity, and small embedded constants) to perform op-support checks. Three call sites were loading or attempting to access the full weight tensors unnecessarily:

1. `ONNXStaticAnalyzer.analyze()` — explicitly passed `load_external_data=True`
2. `ONNXLoader.load()` — called bare `onnx.load()` which defaults to `load_external_data=True`
3. `RuntimeCheckerQuery` — called `numpy_helper.to_array()` on every initializer and embedded full `TensorProto`s into single-node models

### Changes

| File | Change |
|---|---|
| analyzer.py | `load_external_data=True` → `False` |
| onnx_loader.py | Add explicit `load_external_data=False` |
| runtime_checker_query.py | For external-data initializers: extract shape from `dims` instead of `to_array()`, and emit graph inputs instead of embedding empty tensors in single-node models |

### Impact

- Models with external data (Qwen3-8B, Llama, etc.) now use ~MB instead of ~30+ GB RAM
- No behavior change for models with inline weights (the `data_location != EXTERNAL` path is unchanged)
- 849 unit tests pass, 0 failures

### Performance (Improve 39.65%)
| Model | Operators | Time |
|---|---|---|
| Qwen/Qwen3-8B | 5333 | 57.36s |
| dbmdz/bert-large-cased-finetuned-conll03-english | 2663 | 28.395s |


src/winml/modelkit/analyze/core/runtime_checker_query.py
```diff
@@ -775,8 +775,14 @@ def update_conditions_(
 
         if inp_name in initializers:
             init = initializers[inp_name]
-            arr = numpy_helper.to_array(init)
-            update_conditions_(conditions, input_name, is_variadic, True, arr.shape, arr)
+            # External data initializers (large weights) may not have data loaded;
+            # fall back to shape from dims and value=None so analysis can proceed.
+            if init.data_location == onnx.TensorProto.EXTERNAL and not init.raw_data:
+                shape = tuple(init.dims) if init.dims else None
```
vortex-captain: Maybe
```suggestion
                shape = tuple(init.dims) if init.dims is not None else None
```
is better? Empty list/tuple should be a valid shape on its own (a single-number scalar, although this would be rare for initializers)

src/winml/modelkit/analyze/core/runtime_checker_query.py
```diff
@@ -1204,7 +1210,21 @@ def _build_single_node_model(
...
+                # Skip external data initializers without loaded data;
+                # treat them as graph inputs instead so the model stays valid.
+                if init.data_location == onnx.TensorProto.EXTERNAL and not init.raw_data:
+                    vi = self.valueinfo.get(inp_name)
+                    if vi is not None:
+                        graph_inputs.append(vi)
+                    else:
+                        graph_inputs.append(
+                            onnx.helper.make_tensor_value_info(
+                                inp_name, init.data_type, list(init.dims) or None
```
vortex-captain:
```suggestion
                                inp_name, init.data_type, list(init.dims) if init.dims is not None else None
```
Similar to other comment. In addition,`list(None)` would raise

src/winml/modelkit/analyze/core/runtime_checker_query.py
```diff
@@ -1204,7 +1210,21 @@ def _build_single_node_model(
...
+                            )
```
vortex-captain, line 1495: If we make value info with `shape=None`, this would lead to `_generate_node_inputs` returning an array of shape (2,) for this input, if I understand correctly? Would this be inaccurate in case the init dim is missing?

src/winml/modelkit/analyze/core/runtime_checker_query.py
```diff
@@ -775,8 +775,14 @@ def update_conditions_(
...
-            arr = numpy_helper.to_array(init)
-            update_conditions_(conditions, input_name, is_variadic, True, arr.shape, arr)
+            # External data initializers (large weights) may not have data loaded;
+            # fall back to shape from dims and value=None so analysis can proceed.
+            if init.data_location == onnx.TensorProto.EXTERNAL and not init.raw_data:
+                shape = tuple(init.dims) if init.dims else None
+                update_conditions_(conditions, input_name, is_variadic, True, shape, None)
+            else:
+                arr = numpy_helper.to_array(init)
+                update_conditions_(conditions, input_name, is_variadic, True, arr.shape, arr)
```
DingmaomaoBJTU, line 1051: **Bug risk: `is_constant=True` with `value=None` may mislead runtime checker rules**

Here `is_constant=True` is passed but `value=None`. Downstream checker rules may see `{input_name}_is_constant = True` and then access `{input_name}_value` expecting it to be non-None.

- Best case: a `None` comparison harmlessly returns False — correct but potentially masks a rule match
- Worst case: rule code indexes into `None` → crash

Consider either:
1. Setting `is_constant=False` for external-data initializers (treating them like dynamic inputs for analysis), or
2. Auditing the checker rules to confirm `value=None` is always safe when `is_constant=True`

src/winml/modelkit/analyze/core/runtime_checker_query.py
DingmaomaoBJTU: **Nit:** `not init.raw_data` is sufficient for `EXTERNAL` tensors, but ONNX tensors can also store data in typed fields (`float_data`, `int32_data`, etc.). For external-data tensors these should always be empty, but a brief comment explaining why `raw_data` alone is sufficient here would help future readers.

src/winml/modelkit/analyze/core/runtime_checker_query.py
```diff
@@ -1204,7 +1210,21 @@ def _build_single_node_model(
...
+                if init.data_location == onnx.TensorProto.EXTERNAL and not init.raw_data:
+                    vi = self.valueinfo.get(inp_name)
+                    if vi is not None:
+                        graph_inputs.append(vi)
+                    else:
+                        graph_inputs.append(
+                            onnx.helper.make_tensor_value_info(
+                                inp_name, init.data_type, list(init.dims) or None
+                            )
+                        )
```
DingmaomaoBJTU, line 1496: **Question:** After `_build_single_node_model` converts an external-data initializer into a graph input, does `_generate_node_inputs` (called at line 1374) know to generate feed data for it? If `_generate_node_inputs` only generates data for non-initializer inputs, the local EP check would run with missing inputs. Worth verifying this path works end-to-end.

src/winml/modelkit/analyze/core/runtime_checker_query.py
DingmaomaoBJTU, line 1051: **Missing test coverage:** There are no new tests exercising the external-data handling in `_collect_conditions_for_node` or `_build_single_node_model`. A unit test that creates a model with an `EXTERNAL` data_location initializer (without actual sidecar file) and runs it through the analyzer would lock down this behavior. For example:

```python
init = onnx.numpy_helper.from_array(np.zeros((2, 3), dtype=np.float32), name="W")
init.data_location = onnx.TensorProto.EXTERNAL
init.ClearField("raw_data")
```

DingmaomaoBJTU (changes requested):
Overall good fix for a real and impactful problem — static analysis never needs multi-GB weight tensors, and the three-site fix is comprehensive. A few items to address before merging.

src/winml/modelkit/analyze/core/runtime_checker_query.py
```diff
@@ -633,6 +633,98 @@ def _normalize_type_var_annotation(type_value: str) -> str:
...
+
+        arr = np.memmap(
+            data_path,
+            dtype=np_dtype,
+            mode="r",
+            offset=offset,
+            shape=(numel,),
+            order="C",
+        )
+        return arr.reshape(shape)
```
DingmaomaoBJTU, line 720: **`np.memmap` file handle leak — particularly risky on Windows**

`_try_load_external_initializer_array` returns a `np.memmap` object. The memmap keeps the underlying file descriptor open until it is garbage collected, which is non-deterministic. On Windows (this project's primary platform), an open handle prevents *any* other process — or even the same process — from deleting, renaming, or writing to the sidecar file. If the analyzer runs before an optimization step that needs to rewrite the sidecar, it could fail with a sharing violation.

Since the data is already bounded to 1 MB by `MAX_EXTERNAL_INITIALIZER_BYTES_FOR_QUERY`, copying is cheap. Suggestion:

```python
        result = np.array(arr.reshape(shape))   # copy to regular ndarray
        del arr                                  # release memmap / fd
        return result
```

This also avoids a subtle issue where `arr.reshape(shape)` returns a *view* of the memmap (still backed by the fd), so even `arr` going out of scope isn't enough — both `arr` and the reshaped view must be collected.

DingmaomaoBJTU (commented):
Good work addressing the previous round of feedback — the `is_constant=False` fix, test coverage additions, and the `raw_data` comment all look solid. The three-site fix is correct and comprehensive. A few new items for this round:

**Non-inline note — `_collect_node_tags` / `ALL_INPUTS_CONSTANT`:**
External-data initializers are still in `self.initializers`, so a node whose inputs are all unloaded external weights would be tagged `ALL_INPUTS_CONSTANT` even though weight data is unavailable. This is informational-only and doesn't affect runtime check results, but could confuse future debugging. Consider filtering external-data initializers without loaded data in that check.

src/winml/modelkit/analyze/core/runtime_checker_query.py
```diff
@@ -633,6 +633,98 @@ def _normalize_type_var_annotation(type_value: str) -> str:
...
+def _get_external_tensor_info(tensor: onnx.TensorProto) -> tuple[str | None, int, int | None]:
+    """Extract location, offset, and length metadata for an external tensor."""
+    info = {entry.key: entry.value for entry in tensor.external_data}
+    location = info.get("location")
+    offset = int(info.get("offset", "0"))
+    length = int(info["length"]) if "length" in info else None
+    return location, offset, length
+
+
+def _tensor_proto_dtype_to_np_dtype(tensor_type: int) -> np.dtype[Any]:
```
DingmaomaoBJTU, line 649: **Nit: consider reusing existing dtype utilities**

There's already `dtype_from_tensorproto_enum` used elsewhere in this file. While it returns a string annotation rather than `np.dtype`, a one-liner like:

```python
np_dtype = SupportedONNXType.from_annotation(dtype_from_tensorproto_enum(tensor.data_type)).np_type
```

would avoid introducing a second dtype-conversion path with its own onnx version compatibility shim (`try: from onnx.helper ... except ImportError: from onnx.mapping ...`). Less code to maintain when onnx's API surface changes again.

Not blocking — the current approach works — but worth considering for consistency.

src/winml/modelkit/analyze/core/runtime_checker_query.py
```diff
@@ -927,8 +1021,31 @@ def update_conditions_(
...
+                    update_conditions_(
+                        conditions,
+                        input_name,
+                        is_variadic,
+                        True,
+                        external_arr.shape,
+                        external_arr,
+                    )
+                else:
+                    shape = tuple(init.dims) if init.dims is not None else None
```
DingmaomaoBJTU, line 1047: **Potential issue: `external_arr` as a `np.memmap` fed into condition hashing**

When `external_arr` is successfully loaded (line 709 returns a memmap view), it's passed as `value` to `update_conditions_`. Later in the pipeline, `make_hashable` materializes numpy arrays. This works functionally, but the memmap view keeps the sidecar file handle open until `make_hashable` creates a copy — or longer if the original `external_arr` reference survives in the conditions dict.

This is an additional reason to copy the data eagerly in `_try_load_external_initializer_array` (as noted in the other comment).

src/winml/modelkit/analyze/core/runtime_checker_query.py
chinazhangchao: The existing SupportedONNXType map is narrower than raw ONNX TensorProto coverage, so replacing the helper inline would risk silently dropping dtypes like BFLOAT16 or COMPLEX*.

vortex-captain (approved):

DingmaomaoBJTU (approved):

PR #381, opened 2026-04-23
fix inference bugs, add cache-aware e2e quality gates

DingmaomaoBJTU:
## Summary

- Fix NER `numpy.float32` serialization error in serve path — `_sanitize_numpy()` converts numpy scalars to Python types before pydantic/JSON serialization
- Fix cache-key-prefixed build directory loading — `_find_build_artifacts()` supports both plain (`model.onnx`) and cache-prefixed (`{key}_model.onnx`) layouts, enabling `engine.load()` to load directly from `~/.cache/winml/artifacts/`
- Fix task-mismatch cache loading — `_find_build_artifacts(task=)` filters by manifest task, avoiding wrong ONNX model (e.g. feature-extraction head for text-classification); falls back to HF rebuild when no match found
- Redirect stdout→stderr during `engine.load()` in CLI to prevent build-pipeline prints from contaminating `--format json` output
- Merge `test_run_e2e.py` + `test_run_hub_models_e2e.py` into 3-tier quality gate (Feature 14 + Schema 35 + Inference 35 = 84 tests)
- Add `test_serve_e2e.py` 4-tier quality gate (103 tests)
- Cache-aware e2e: tests use `~/.cache/winml/artifacts/` when available, per-file `timeout(3600)` for e2e
- 14 new unit tests for `_sanitize_numpy`, `_find_build_artifacts`, NER output normalization

All 187 e2e tests passing (run 84 + serve 103).

### DIFF
--- a/src/winml/modelkit/commands/run.py
+++ b/src/winml/modelkit/commands/run.py
@@ -624,8 +624,8 @@ def run(
 
     try:
         # Redirect stdout → stderr during model load so that build-pipeline
-        # prints (from optimum, onnxruntime, etc.) don't contaminate the
-        # structured output when --format json is used.
+        # prints (from optimum, onnxruntime, etc.) don't contaminate
+        # structured output (--format json) or text output parsing.
         with contextlib.redirect_stdout(sys.stderr):
             engine.load(model, task=task, device=device, ep=ep)
     except (OSError, ValueError, RuntimeError) as exc:
--- a/src/winml/modelkit/commands/run.py
+++ b/src/winml/modelkit/commands/run.py
@@ -25,6 +25,7 @@ Usage:
 
 from __future__ import annotations
 
+import contextlib
 import json
 import logging
 import sys
@@ -622,7 +623,11 @@ def run(
         return
 
     try:
-        engine.load(model, task=task, device=device, ep=ep)
+        # Redirect stdout → stderr during model load so that build-pipeline
+        # prints (from optimum, onnxruntime, etc.) don't contaminate the
+        # structured output when --format json is used.
+        with contextlib.redirect_stdout(sys.stderr):
+            engine.load(model, task=task, device=device, ep=ep)
     except (OSError, ValueError, RuntimeError) as exc:
         click.echo(f"Error loading model: {exc}", err=True)
         ctx.exit(3)
--- a/src/winml/modelkit/inference/engine.py
+++ b/src/winml/modelkit/inference/engine.py
@@ -100,6 +100,72 @@ _DECODERS: dict[str, Any] = {
 }
 
 
+def _sanitize_numpy(obj: Any) -> Any:
+    """Recursively convert numpy scalars to Python types for JSON serialization.
+
+    HF pipelines (e.g. NER/TokenClassification) return dicts containing
+    ``numpy.float32`` scores and ``numpy.int64`` offsets that pydantic
+    cannot serialize.  This function converts them to native Python types.
+    """
+    import numpy as np
+
+    if isinstance(obj, dict):
+        return {k: _sanitize_numpy(v) for k, v in obj.items()}
+    if isinstance(obj, list):
+        return [_sanitize_numpy(v) for v in obj]
+    if isinstance(obj, np.floating):
+        return float(obj)
+    if isinstance(obj, np.integer):
+        return int(obj)
+    if isinstance(obj, np.ndarray):
+        return obj.tolist()
+    return obj
+
+
+# ---------------------------------------------------------------------------
+# Build-directory artifact discovery
+# ---------------------------------------------------------------------------
+
+
+def _find_build_artifacts(build_dir: Path, *, task: str | None = None) -> tuple[Path, dict | None]:
+    """Locate model.onnx and build_manifest.json inside a build/cache directory.
+
+    Supports both plain layout (``model.onnx``) and cache-key-prefixed layout
+    (``{cache_key}_model.onnx``) so that ``InferenceEngine.load()`` can load
+    directly from ``~/.cache/winml/artifacts/{slug}/`` without a full rebuild.
+
+    When *task* is provided, only returns artifacts whose manifest ``task``
+    matches.  A cache directory may contain multiple task variants (e.g.
+    ``feat_*`` for feature-extraction and ``txtcls_*`` for text-classification);
+    using the wrong ONNX (different output head) would produce garbage results.
+
+    Returns:
+        ``(onnx_path, manifest_dict)`` — manifest is ``None`` when no
+        manifest file is found.
+
+    Raises:
+        FileNotFoundError: if no matching ``*model.onnx`` is found.
+    """
+    # Try plain layout first (bare build output)
+    plain_onnx = build_dir / "model.onnx"
+    plain_manifest = build_dir / "build_manifest.json"
+    if plain_onnx.exists():
+        manifest = json.loads(plain_manifest.read_text()) if plain_manifest.exists() else None
+        if task is None or manifest is None or manifest.get("task") == task:
+            return plain_onnx, manifest
+
+    # Cache-key-prefixed layout: {cache_key}_model.onnx
+    # Scan all candidates and match by task when specified.
+    for onnx_path in sorted(build_dir.glob("*_model.onnx")):
+        prefix = onnx_path.name.rsplit("_model.onnx", 1)[0]
+        manifest_path = build_dir / f"{prefix}_build_manifest.json"
+        manifest = json.loads(manifest_path.read_text()) if manifest_path.exists() else None
+        if task is None or manifest is None or manifest.get("task") == task:
+            return onnx_path, manifest
+
+    raise FileNotFoundError(f"No model.onnx matching task={task!r} found in {build_dir}")
+
+
 # ---------------------------------------------------------------------------
 # Lightweight schema helpers (no model load required)
 # ---------------------------------------------------------------------------
@@ -233,7 +299,24 @@ class InferenceEngine:
         path = Path(model_path)
 
         if path.is_dir():
-            self._load_from_build_dir(path, task=task, device=device, ep=ep)
+            try:
+                self._load_from_build_dir(path, task=task, device=device, ep=ep)
+            except FileNotFoundError:
+                # No cached ONNX for this task — check if the manifest has a
+                # model_id we can rebuild from (e.g. cache was built for a
+                # different task like feature-extraction but caller wants
+                # text-classification).
+                model_id = self._resolve_model_id_from_dir(path)
+                if model_id:
+                    logger.info(
+                        "No cached ONNX for task=%s in %s — rebuilding from %s",
+                        task,
+                        path,
+                        model_id,
+                    )
+                    self._load_from_hf(model_id, task=task, device=device, ep=ep)
+                else:
+                    raise
         elif path.suffix == ".onnx" and path.exists():
             self._load_from_onnx(path, task=task, device=device, ep=ep)
         else:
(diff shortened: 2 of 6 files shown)


timenick (commented):
Inline comments on 7 issues flagged against `CLAUDE.md` / `tests/CLAUDE.md`. Core fixes (`_sanitize_numpy`, `_find_build_artifacts` task filter, HF rebuild fallback) look correct — findings are mostly test-layout duplication plus a couple of behavior nits.

🤖 Generated with Claude Code

tests/e2e/test_serve_e2e.py
```diff
@@ -0,0 +1,718 @@
...
+
+@pytest.fixture(scope="module")
+def test_image(tmp_path_factory: pytest.TempPathFactory) -> str:
+    """Generate a 224x224 random RGB JPEG (reused across the module)."""
+    import numpy as np
+    from PIL import Image
+
+    d = tmp_path_factory.mktemp("serve_e2e_assets")
+    img_path = d / "test_image.jpg"
+    rng = np.random.RandomState(42)
```
timenick: **Duplicate fixture** — `test_image` is already defined in [tests/e2e/conftest.py:38](tests/e2e/conftest.py#L38) with identical behavior. `tests/CLAUDE.md` says *"Never duplicate fixtures across multiple conftest.py files"*, and a module-local redefinition that shadows a conftest fixture is the same problem in spirit. Removing this block will let the test file pick up the conftest fixture automatically (the `test_image_bytes` / `test_image_b64` fixtures below just need the string path the conftest version already returns).

🤖 Generated with Claude Code

tests/e2e/test_serve_e2e.py
```diff
@@ -0,0 +1,718 @@
...
+
+# ---------------------------------------------------------------------------
+# Cache-aware model resolution (same logic as test_run_e2e.py)
+# ---------------------------------------------------------------------------
+
+
+def _find_cache_dir(model_id: str, task: str | None = None) -> Path | None:
+    from winml.modelkit.cache import get_cache_dir, model_id_to_slug
+    from winml.modelkit.inference.engine import _find_build_artifacts
+
```
timenick: **Duplicated helpers between the two e2e files** — `_HUB_DATA`, `_unique_pairs`, `_PAIRS`, `_pytest_id`, `_find_cache_dir`, `_resolve_model_arg`, `_TEXT_BY_FIELD`, `_SAMPLE_TEXT` are copy-pasted verbatim from [test_run_e2e.py:325-385](tests/e2e/test_run_e2e.py#L325-L385). Single source of truth would be moving them into `tests/e2e/conftest.py` (or a shared `tests/e2e/_hub_matrix.py` helper). Any future change to hub parametrization or cache-lookup logic will otherwise need to be made in two places and can silently drift.

🤖 Generated with Claude Code

tests/e2e/test_run_e2e.py
```diff
@@ -0,0 +1,535 @@
...
+        )
+        assert schema_result.exit_code == 0, (
+            f"--schema failed (exit {schema_result.exit_code}):\n{schema_result.output}"
+        )
+        schema = json.loads(schema_result.output)
+
+        # Step 2: Build inference args
+        input_args = _build_inference_args(schema["inputs"], task, test_image)
+        if input_args is None:
+            pytest.skip(f"Cannot determine inputs for task '{task}' (empty schema, no fallback)")
```
timenick: **Dynamic `pytest.skip` hides coverage gaps** — same pattern appears at [test_serve_e2e.py:595](tests/e2e/test_serve_e2e.py#L595). `CLAUDE.md` §4: *"Never use `pytest.mark.skip` or `xfail` to hide failures. Skips are only acceptable for hardware/EP requirements."* A task whose required inputs cannot be synthesized is silently dropped from the matrix instead of being fixed. Options: (a) extend `_FALLBACK_INPUT_ARGS` / `_build_predict_body` to cover every task in `hub_models.json`; (b) `xfail(strict=True)` with a concrete reason so new matching tasks don't pass by accident; (c) fail loudly so the generator gets fixed.

🤖 Generated with Claude Code

src/winml/modelkit/commands/run.py
```diff
@@ -622,7 +623,11 @@ def run(
         return
 
     try:
-        engine.load(model, task=task, device=device, ep=ep)
+        # Redirect stdout → stderr during model load so that build-pipeline
+        # prints (from optimum, onnxruntime, etc.) don't contaminate the
+        # structured output when --format json is used.
+        with contextlib.redirect_stdout(sys.stderr):
```
timenick, line 629: **Comment and code disagree on scope** — the comment says *"when `--format json` is used"* but the redirect is applied unconditionally, which means `--format text` users also lose build-pipeline progress output from stdout (it's still visible on stderr, but some terminals color/dim stderr). Suggest gating on format:

```python
if output_format == "json":
    with contextlib.redirect_stdout(sys.stderr):
        engine.load(model, task=task, device=device, ep=ep)
else:
    engine.load(model, task=task, device=device, ep=ep)
```

Or, if the unconditional redirect is genuinely desired, drop the "when `--format json` is used" qualifier from the comment.

🤖 Generated with Claude Code

tests/e2e/test_run_e2e.py
```diff
@@ -0,0 +1,535 @@
...
+    decoder = json.JSONDecoder()
+    # Scan forward for the first '{' that starts a valid JSON object
+    for i, ch in enumerate(output):
+        if ch == "{":
+            try:
+                obj, _ = decoder.raw_decode(output, i)
+                if isinstance(obj, dict):
+                    return obj
+            except json.JSONDecodeError:
+                continue
```
timenick, line 103: **May now be dead-weight workaround** — `_extract_json` scans for the first valid top-level `{…}` in output to survive stdout contamination from build-pipeline prints. That contamination is exactly what the new `contextlib.redirect_stdout(sys.stderr)` in [run.py:629](src/winml/modelkit/commands/run.py#L629) is meant to prevent. If the redirect works, a plain `json.loads(result.output)` (as is used at lines 488, 516, etc.) should succeed here too. Worth verifying: either drop `_extract_json` and call `json.loads` directly at line 532, or add a short comment explaining why the redirect isn't sufficient (e.g. Click's runner handling or pytest capture).

🤖 Generated with Claude Code

src/winml/modelkit/inference/engine.py
```diff
@@ -100,6 +100,72 @@ def _decode_image(data: bytes) -> Any:
...
+            return plain_onnx, manifest
+
+    # Cache-key-prefixed layout: {cache_key}_model.onnx
+    # Scan all candidates and match by task when specified.
+    for onnx_path in sorted(build_dir.glob("*_model.onnx")):
+        prefix = onnx_path.name.rsplit("_model.onnx", 1)[0]
+        manifest_path = build_dir / f"{prefix}_build_manifest.json"
+        manifest = json.loads(manifest_path.read_text()) if manifest_path.exists() else None
+        if task is None or manifest is None or manifest.get("task") == task:
+            return onnx_path, manifest
```
timenick: **Glob ordering is non-deterministic by task** — when `task=None` and the cache dir contains multiple task variants (e.g. `feat_*_model.onnx` and `txtcls_*_model.onnx`), `sorted(...)` returns the first alphabetical match. A caller of `engine.load(cache_dir)` without `task` may silently get a feature-extraction head when they wanted text-classification, with no warning. Consider either: (a) raising if `task is None` and multiple variants exist with different manifest tasks, or (b) documenting "when multiple variants exist, pass `task=` explicitly" in the docstring so callers aren't surprised.

🤖 Generated with Claude Code

src/winml/modelkit/inference/engine.py
```diff
@@ -233,7 +299,24 @@ def load(
...
+                # model_id we can rebuild from (e.g. cache was built for a
+                # different task like feature-extraction but caller wants
+                # text-classification).
+                model_id = self._resolve_model_id_from_dir(path)
+                if model_id:
+                    logger.info(
+                        "No cached ONNX for task=%s in %s — rebuilding from %s",
+                        task,
+                        path,
+                        model_id,
```
timenick: **Narrow except only catches `FileNotFoundError`** — `_find_build_artifacts` also reads manifest JSON (line 162), which raises `json.JSONDecodeError` on a corrupt manifest. With the current narrow `except`, a corrupt manifest won't trigger the HF rebuild fallback and won't surface a helpful error either — it bubbles up as a raw `JSONDecodeError`. If the intent of the fallback is "any problem with the cached dir → rebuild from HF when manifest has `model_id`", broaden to `(FileNotFoundError, json.JSONDecodeError, KeyError)` or add explicit handling for the corrupt-manifest case.

🤖 Generated with Claude Code

tests/e2e/test_serve_e2e.py
DingmaomaoBJTU: Fixed in 0387ce4.

tests/e2e/test_serve_e2e.py
DingmaomaoBJTU: Fixed in 0387ce4.

tests/e2e/test_run_e2e.py
DingmaomaoBJTU: Fixed in 0387ce4.

src/winml/modelkit/commands/run.py
DingmaomaoBJTU: Fixed in 0387ce4.

tests/e2e/test_run_e2e.py
DingmaomaoBJTU: Fixed in 0387ce4.

src/winml/modelkit/inference/engine.py
DingmaomaoBJTU: Fixed in 0387ce4.

src/winml/modelkit/inference/engine.py
DingmaomaoBJTU: Fixed in 0387ce4.

tests/e2e/conftest.py
```diff
@@ -14,16 +18,128 @@
...
+
+
+HUB_PAIRS: list[dict[str, str]] = _unique_pairs()
+
+
+def pytest_id(pair: dict[str, str]) -> str:
+    """Readable pytest ID, e.g. ``finbert-text_classification``."""
+    short = pair["model_id"].rsplit("/", 1)[-1]
+    task = pair["task"].replace("-", "_")
+    return f"{short}-{task}"
```
timenick: **🔴 Blocker — `pytest_id` collides with pytest's hook namespace.** Any function defined in `conftest.py` whose name starts with `pytest_` is treated by pytest/pluggy as a plugin hook. `pytest_id` is not a valid hook, so collection fails with:

```
INTERNALERROR> pluggy._manager.PluginValidationError: unknown hook 'pytest_id'
in plugin <module 'tests.e2e.conftest' from '...\tests\e2e\conftest.py'>
```

Reproduces deterministically on my machine:

```
uv run pytest --collect-only tests/e2e/ -m e2e
```

Fix: rename the helper to something that doesn't start with `pytest_` (e.g. `make_pytest_id`, `hub_test_id`) — or move it out of `conftest.py` into a regular helper module such as `tests/e2e/_hub_matrix.py` (conftest scope isn't required since the tests import it explicitly via `from .conftest import …`). Both `test_run_e2e.py` and `test_serve_e2e.py` import it as `from .conftest import pytest_id as _pytest_id` and will need matching updates.

🤖 Generated with Claude Code

timenick (commented):
Re-review of the feedback commit. All 7 previous comments were addressed substantively (helpers moved to conftest, duplicate fixture removed, `xfail` with actionable reasons, docstrings, `_find_build_artifacts` task-ambiguity guard, broadened except). Unit tests pass locally. One new blocker was introduced by moving the helper into `conftest.py` — see inline.

🤖 Generated with Claude Code

tests/e2e/conftest.py
DingmaomaoBJTU: Fixed in f7ff1bc — renamed `pytest_id` → `hub_test_id`.

timenick (approved):
LGTM

PR #382, opened 2026-04-23
Add Marian encoder-decoder support (static cache)

vortex-captain:
## Summary

- **Marian** (`Helsinki-NLP/opus-mt-*`) encoder-decoder support: `MarianEncoderWrapper` / `MarianDecoderWrapper` using `WinMLStaticCache` (multi-dim `index_put_` → ScatterND). The decoder's `OnnxConfig` exposes `cache_position` as the position input; under static cache, stock HF's `position_ids=cache_position` already indexes the sin/cos table correctly, so no embedding patch is applied at runtime. The `PATCHING_SPECS` infra (`_patched_marian_sinusoidal_forward` / `_build_marian_patching_specs`) is kept in the file with its assignment commented, for reference and future sliding-window re-export. Registered composite: `("marian", "translation")` → `WinMLMarianModel`.

- **BART** (`facebook/bart-large-cnn`, `sshleifer/distilbart-cnn-12-6`) encoder-decoder support: `BartEncoderWrapper` / `BartDecoderWrapper` using `WinMLStaticCache`. Same static-cache approach as Marian — `BartLearnedPositionalEmbedding.forward` correctly receives `position_ids=cache_position` from stock HF and applies its `+self.offset` shift, so no embedding patch is applied at runtime. The `_patched_bart_learned_forward` patch spec remains in the file (commented out on the decoder config) for future sliding-window re-export. Registered composite: `("bart", "summarization")` → `WinMLBartModel`.

### `kv_cache` changes
- `WinMLStaticCache.update` uses multi-dim `index_put_` with full `(batch, head, pos)` coord tuples so ONNX export emits ScatterND instead of ScatterElements.
- `WinMLCache.num_layers` now reads `len(self.layers)` after `super().__init__` (HF's `StaticCache.__init__` builds `self.layers` from `config.get_text_config(decoder=True).num_hidden_layers`, i.e. the decoder's layer count). Previously `config.num_hidden_layers` on the outer `BartConfig` maps to `encoder_layers`, which fails for asymmetric distilled variants like `distilbart-cnn-12-6` (encoder_layers=12, decoder_layers=6). No-op for symmetric models.
- `PastKeyValueInputGenerator` reads `normalized_config.head_dim` unconditionally (no try/except fallback). `MarianConfig` / `BartConfig` have no native `head_dim` attr, so `_MarianDecoderNormalizedConfig` / `_BartDecoderNormalizedConfig` — real `NormalizedConfig` subclasses with UPPERCASE class-level attribute mappings (same pattern as stock `NormalizedTextConfig`) — expose `head_dim` as a computed property.

## Verified

| Model | Test | Result |
|---|---|---|
| `Helsinki-NLP/opus-mt-fr-en` | 3 fr→en translations | 3/3 **exact MATCH** vs HF PyTorch `MarianMTModel` |
| `facebook/bart-large-cnn` | 2 CNN-style summarizations | 2/2 **PREFIX_MATCH** vs HF PyTorch `BartForConditionalGeneration` |
| `sshleifer/distilbart-cnn-12-6` (asymmetric: 12 enc / 6 dec) | 2 CNN-style summarizations | 2/2 **PREFIX_MATCH** vs HF PyTorch |

`PREFIX_MATCH` means WinML ONNX and HF PyTorch agree bit-identically for the full shared prefix (e.g. 53 tokens for BART article 1, 40 for article 2) and then diverge at a single EOS-vs-continuation argmax flip near the natural sentence boundary — the expected fingerprint of fp32 drift from fused ops (`gelu_fusion`, `matmul_add_fusion`).

## Commits

1. Add Marian (Helsinki-NLP/opus-mt) encoder-decoder support
2. Add BART (facebook/bart-large-cnn) encoder-decoder support
3. Fix WinMLCache.num_layers for asymmetric encoder-decoder models
4. Derive head_dim in per-model NormalizedConfig instead of kv_cache fallback


xieofxie:
do we need tests? although no one will review them (?)

src/winml/modelkit/models/hf/bart.py
```diff
@@ -0,0 +1,557 @@
...
+        """
+        decoder_input_ids = args[0]
+        encoder_hidden_states = args[1]
+        attention_mask = args[2]
+        decoder_attention_mask = args[3]
+        # position_id = args[4]  # sliding-window cache
+        cache_position = args[4]  # static cache: absolute seq pos, drives mask + learned pos
+        kv_start = 5
+
+        # Build WinMLSlidingWindowCache from input KV tensors.  update() does
```
zhenchaoni: The code is actually using `WinMLStaticCache` while the comment says sliding window cache

src/winml/modelkit/models/hf/bart.py
```diff
@@ -0,0 +1,557 @@
...
+        # Slice+Concat (not index_copy_/ScatterElements) — evicting the N
+        # oldest entries and appending the N new ones at the right.
+        max_cache_len = args[kv_start].size(2)
+        # self_attn_cache = WinMLSlidingWindowCache(self.config, max_cache_len=max_cache_len)
+        self_attn_cache = WinMLStaticCache(self.config, max_cache_len=max_cache_len)
+        self_attn_cache.early_initialization(
```
zhenchaoni, line 304: As we discussed earlier, creating the cache object and then pass to the original model is very common across decoders. I propose we can have a common decoder wrapper class.

```py
# -------------------------------------------------------------------------
# Copyright (c) Microsoft Corporation. All rights reserved.
# Licensed under the MIT License.
"""Base class for decoder export wrappers with a traceable KV cache.

HuggingFace decoders represent the KV cache as a Python ``Cache`` object,
which has no shape at the ONNX graph boundary — a caller cannot supply past
KV through it, and HF cannot return new-token KV out of it as named tensors.
``WinMLDecoderWrapper`` adapts an HF decoder so its KV cache *is* exportable:
past KV become flat per-layer ONNX inputs, and new-token KV become flat
per-layer ONNX outputs.  Transformer math inside HF runs unmodified.

What ``forward`` does during ``torch.onnx.export``:

1. Build a ``WinMLCache`` whose per-layer storage is **aliased** to the ONNX
   past-KV input tensors.  HF then reads/writes the cache as usual, and every
   such op surfaces as a graph edge on the named graph input.
2. Hand the cache to the unmodified HF decoder via its ``past_key_values=``
   argument.
3. Pack ``logits`` plus the per-layer new-token K/V (captured by
   ``WinMLStaticCache.update``) into the ONNX output tuple.

The ONNX input name ordering and decoder layer count are read from the
family's ``OnnxConfig`` at construction time, so per-architecture shape
knowledge lives in one place.  A per-family subclass only has to:

* set ``_HF_MODEL_CLS`` and ``_IO_CONFIG_CLS`` class constants,
* override ``_invoke_hf`` with the HF decoder call,
* (optional) override ``_make_cache`` — call ``super()`` first, then attach
  any trace-time side channel to the returned cache (e.g.,
  ``cache.set_trace_position(...)``).
"""

from __future__ import annotations

from typing import TYPE_CHECKING, Any, ClassVar

import torch
import torch.nn as nn

from ..winml.kv_cache import WinMLStaticCache


if TYPE_CHECKING:
    from optimum.exporters.onnx import OnnxConfig

    from ..winml.kv_cache import WinMLCache


class WinMLDecoderWrapper(nn.Module):
    """Base class for static-KV-cache decoder export wrappers.

    Instance attributes (set by ``from_pretrained``):
        model         — the HF model, called by ``_invoke_hf``
        config        — full HF ``PretrainedConfig``
        onnx_config   — instance of ``_IO_CONFIG_CLS`` — source of truth for
                        ONNX input name ordering and decoder layer count
        num_layers    — derived from ``onnx_config._normalized_config.num_layers``
    """

    _HF_MODEL_CLS: ClassVar[type]
    #: IOConfig class used at export — the sole source of per-architecture
    #: normalisation (input names, sub-config paths, layer count).
    _IO_CONFIG_CLS: ClassVar[type]
    #: Task name passed to the IOConfig constructor.
    _TASK: ClassVar[str] = "text2text-generation"
    #: Cache class for Step 1's aliasing.
    _CACHE_CLS: ClassVar[type] = WinMLStaticCache

    # ---- Instance attrs ----
    model: nn.Module
    config: Any
    onnx_config: OnnxConfig
    num_layers: int

    # =====================================================================
    # Factory
    # =====================================================================

    @classmethod
    def from_pretrained(cls, model_name_or_path: str, **kwargs: Any) -> WinMLDecoderWrapper:
        """Load the HF model and wrap it for export."""
        full = cls._HF_MODEL_CLS.from_pretrained(model_name_or_path, **kwargs)
        self = cls()
        self.model = full
        self.config = full.config
        self.onnx_config = cls._IO_CONFIG_CLS(full.config, task=cls._TASK)
        self.num_layers = self.onnx_config._normalized_config.num_layers
        self.eval()
        return self

    # =====================================================================
    # torch.onnx.export entry point
    # =====================================================================

    def get_export_args(self, inputs: dict[str, torch.Tensor]) -> tuple[torch.Tensor, ...]:
        """Order dict inputs positionally to match the IOConfig's input order."""
        return tuple(inputs.values())

    def forward(self, *args: torch.Tensor) -> tuple[torch.Tensor, ...]:
        """Execute the three-step adapter.  Subclasses override ``_invoke_hf``."""
        inputs = dict(zip(self.onnx_config.inputs.keys(), args))

        # Step 1 — Create cache and alias ONNX inputs onto it.  The cache
        # captures new-token KV for Step 3.  Subclasses may override
        # ``_make_cache`` (calling super) to attach trace-time side channels.
        cache = self._make_cache(inputs)

        # Step 2 — invoke HF.
        logits = self._invoke_hf(cache, inputs)

        # Step 3 — pack new-token KV (captured by the cache) as ONNX outputs.
        result: list[torch.Tensor] = [logits]
        for i in range(self.num_layers):
            k, v = cache.captured[i]
            result.extend([k, v])
        return tuple(result)

    # =====================================================================
    # Shared helpers
    # =====================================================================

    def _make_cache(self, inputs: dict[str, torch.Tensor]) -> WinMLCache:
        """Alias the flat past-KV inputs onto a fresh ``WinMLCache``."""
        # Pick any past-KV input as a shape probe
        # past_i_key/value share the same layout: [batch, heads, max_cache_len, head_dim]
        sample_k = inputs["past_0_key"]
        decoder_config = self.config.get_text_config(decoder=True)
        cache = self._CACHE_CLS(decoder_config, max_cache_len=sample_k.size(2))
        cache.early_initialization(
            batch_size=inputs["decoder_input_ids"].size(0),
            num_heads=sample_k.size(1),
            head_dim=sample_k.size(3),
            dtype=sample_k.dtype,
            device=sample_k.device,
        )

        # Alias each cache slot to the corresponding ONNX input tensor.  After
        # this loop, any read/write HF performs on ``cache.layers[i].keys`` is
        # an op on the named graph input ``past_{i}_key`` — that's how the
        # cache becomes "visible" at the ONNX boundary.
        for i in range(self.num_layers):
            cache.layers[i].keys = inputs[f"past_{i}_key"]
            cache.layers[i].values = inputs[f"past_{i}_value"]
        return cache

    def _invoke_hf(
        self, cache: WinMLCache, inputs: dict[str, torch.Tensor]
    ) -> torch.Tensor:
        """Call the HF decoder with ``past_key_values=<cache>``.  Returns logits."""
        raise NotImplementedError


__all__ = ["WinMLDecoderWrapper"]

```

src/winml/modelkit/models/hf/bart.py
```diff
@@ -0,0 +1,557 @@
...
+    """
+
+    VOCAB_SIZE = "vocab_size"
+    HIDDEN_SIZE = "d_model"
+    NUM_LAYERS = "decoder_layers"
+    NUM_ATTENTION_HEADS = "decoder_attention_heads"
+    MAX_CACHE_LEN = "max_position_embeddings"
+
+    @property
+    def head_dim(self) -> int:
```
zhenchaoni, line 402: NIT: I find `head_dim` is not available in many models, the `blip` and `tocr` model also doesn't have head_dim in this config. Maybe we can put this in a base class.

src/winml/modelkit/models/hf/bart.py
```diff
@@ -0,0 +1,557 @@
...
+        attention_mask: torch.Tensor,
+    ) -> torch.Tensor:
+        """Return encoder last hidden state."""
+        return self.encoder(
+            input_ids=input_ids,
+            attention_mask=attention_mask,
+        ).last_hidden_state
+
+
+class BartDecoderWrapper(nn.Module):
```
zhenchaoni, line 238: I have tested this code with TOCR model as well. It looks good.

zhenchaoni (commented):
Reviewed the bart.py

src/winml/modelkit/models/hf/bart.py
vortex-captain: indeed, we could consider extracting a common parent class for normalized configs, which would be good for a separate PR. Tracked in https://github.com/microsoft/WinML-ModelKit/issues/401

src/winml/modelkit/models/hf/bart.py
```diff
@@ -0,0 +1,557 @@
...
+    Input past KV: full buffer ``[batch, heads, max_decode, d_kv]``.
+    Output present KV: new token only ``[batch, heads, 1, d_kv]``.
+    """
+
+    NORMALIZED_CONFIG_CLASS = _BartDecoderNormalizedConfig
+    DUMMY_INPUT_GENERATOR_CLASSES = (
+        EncoderDecoderInputGenerator,
+        PastKeyValueInputGenerator,
+    )
+    # PATCHING_SPECS = _build_bart_patching_specs()  # sliding only; no-op under static
```
zhenchaoni, line 429: Since we are using static cache, does it mean the bart's patcher is currently not used?

vortex-captain: Yes, the patcher is for sliding window cache only, and kept in the code in case we need to compare 2 caches in the future. (instructions of switching to sliding window cache is in code comments)

src/winml/modelkit/models/hf/bart.py
vortex-captain: fixed

vortex-captain:
> do we need tests? although no one will review them (?)

UTs added

zhenchaoni:
Overall is good to me. I leveraged AI to do a check as well. I think both AI comments can be solved by refactoring the code to use the common `decoder_wrapper`. You can consider doing it in a separate PR.

### 1. Heavy structural duplication between bart.py and marian.py

These two files are ~90% structurally identical:

| Component | bart.py | marian.py | Difference |
|---|---|---|---|
| `*EncoderWrapper` | ✓ | ✓ | class name + outer model class |
| `*DecoderWrapper.forward` | line 281-352 | line 319-392 | **bit-identical body**, just renamed class |
| `*DecoderIOConfig.inputs/outputs` | line 432-456 | line 472-496 | **bit-identical** |
| `_*DecoderNormalizedConfig` | line 388-403 | line 428-443 | **bit-identical** field map |
| `_patched_*_forward` | line 142-184 | line 179-214 | same shape (different embedding type) |
| `_build_*_patching_specs` | line 187-200 | line 217-233 | **bit-identical pattern** |

This is independent of any other PR — even between just bart and marian, a shared `_BartFamilyDecoderWrapperBase` (or even just `_BartFamilyDecoderIOConfigBase` for the inputs/outputs body) would cut ~250-300 LOC of duplicated code. Two parallel files mean every future bug fix has to be applied twice.

**Suggestion:** factor at least the IOConfig `inputs`/`outputs` body and the NormalizedConfig field map into a shared base in either `bart.py` (and `marian.py` imports from there) or a new `_bart_family.py`. The decoder wrapper's `forward` body is harder to share without a true base class, but the IOConfig body and the NormalizedConfig are easy wins.

---

### 2. Positional `args[N]` indexing couples the wrapper to the IOConfig's input order — fragile

In [bart.py:293-313]:
```python
decoder_input_ids = args[0]
encoder_hidden_states = args[1]
attention_mask = args[2]
decoder_attention_mask = args[3]
cache_position = args[4]
kv_start = 5

max_cache_len = args[kv_start].size(2)
...
for i in range(self.num_layers):
    self_attn_cache.layers[i].keys = args[kv_start + i * 2]
    self_attn_cache.layers[i].values = args[kv_start + i * 2 + 1]
```

The contract is implicit: *"the IOConfig's `inputs` property declares names in this exact order; `forward` receives them positionally in the same order; we read `args[N]` by hardcoded index."* Nothing in the wrapper *checks* that the IOConfig matches this expectation — and nothing in the IOConfig knows that the wrapper depends on its order. If anyone reorders or inserts an input in the IOConfig's `inputs` dict (e.g., adds `head_mask`), `forward` silently misreads and produces nonsense ONNX.

**Suggestion:** in `forward`, build a name-keyed dict from the IOConfig:

```python
inputs = dict(zip(self.onnx_config.inputs.keys(), args, strict=True))
decoder_input_ids = inputs["decoder_input_ids"]
# ...
for i in range(self.num_layers):
    self_attn_cache.layers[i].keys = inputs[f"past_{i}_key"]
```

This makes the wrapper explicit about which names it depends on, fails loudly with `strict=True` if the count mismatches, and is robust to IOConfig reorderings.

---

zhenchaoni (approved):

PR #383, opened 2026-04-23
feat: add eval_option support in build config and eval command

ssss141414:
## Summary

Add eval field to WinMLBuildConfig and refactor the eval command to a config-centric flow, enabling evaluation dataset configuration to be embedded in build config files.  
Post-validation, fix one regression in CLI override mapping for output and extract a shared trust-remote-code CLI option across build/config/eval commands.

## Changes

- Refactor eval command to config-centric 3-step flow: build config -> resolve -> evaluate
- Merge precedence: CLI > config file eval > quant/loader > dataclass defaults
- Rename eval_option -> eval in build config
- Move dataset-related fields into dataset config (build_script, label_mapping_file)
- Fix output override mapping (output_path cli_name metadata)
- Extract shared trust_remote_code_option and reuse in build/config/eval

## Test Commands

- resnet-50 CLI:  
  `uv run winml eval -m microsoft/resnet-50 --task image-classification --dataset timm/mini-imagenet --split test --samples 100 --device cpu --output temp/eval_compare/resnet50_cli_100.json`

- resnet-50 Config-file:  
  `uv run winml eval --config temp/eval_compare/config_resnet50.json -m microsoft/resnet-50 --samples 100 --device cpu --output temp/eval_compare/resnet50_config_100.json`

- resnet-50 run_eval.py:  
  `uv run python scripts/e2e_eval/run_eval.py --registry scripts/e2e_eval/testsets/models_with_acc.json --eval-type accuracy --device cpu --timeout 1800 --task image-classification --hf-model microsoft/resnet-50`

- clip-vit-base-patch16 CLI:  
  `uv run winml eval -m openai/clip-vit-base-patch16 --task zero-shot-image-classification --dataset uoft-cs/cifar100 --split test --samples 1000 --column input_column=img --column label_column=fine_label --device npu --output temp/eval_compare/clip_cli_1000.json`

- clip-vit-base-patch16 Config-file:  
  `uv run winml eval --config temp/eval_compare/config_clip_zsic.json -m openai/clip-vit-base-patch16 --output temp/eval_compare/clip_config_1000.json`

- clip-vit-base-patch16 run_eval.py:  
  `uv run python scripts/e2e_eval/run_eval.py --registry scripts/e2e_eval/testsets/models_with_acc.json --eval-type accuracy --device npu --timeout 3600 --task zero-shot-image-classification --hf-model openai/clip-vit-base-patch16`

## Results

| Model | CLI | Config-file | run_eval.py | Status |
|---|---|---|---|---|
| microsoft/resnet-50 | 0.78 | 0.78 | 0.784 | PASS |
| openai/clip-vit-base-patch16 | top1=63.4 / top5=87.9 | top1=63.4 / top5=87.9 | top1=62.5 (ACCURACY_PASS, baseline=62.6, delta=-0.1) | PASS |


src/winml/modelkit/commands/eval.py
```diff
@@ -171,6 +171,57 @@ def eval(
                 samples = build_cfg.quant.samples
             if not cli_utils.is_cli_provided(ctx, "dataset_name"):
                 dataset_name = build_cfg.quant.dataset_name
+        # Apply eval_dataset from config (CLI options take precedence)
+        if build_cfg.eval_dataset:
+            ed = build_cfg.eval_dataset
+            # Run build_script if needed to generate local dataset
+            build_script = ed.get("build_script")
```
zhenchaoni: Better to add a `--trust-remote-code` option. Only when it is true, a remote script can be executed to build the dataset.

Better to rename the "build_script" to "dataset_script" so the meaning is clear, the script is to build a dataset.

src/winml/modelkit/commands/eval.py
```diff
@@ -171,6 +171,57 @@ def eval(
                 samples = build_cfg.quant.samples
             if not cli_utils.is_cli_provided(ctx, "dataset_name"):
                 dataset_name = build_cfg.quant.dataset_name
+        # Apply eval_dataset from config (CLI options take precedence)
+        if build_cfg.eval_dataset:
+            ed = build_cfg.eval_dataset
```
zhenchaoni: Although most of the eval option today is for dataset, it is possible to have other non-dataset options in future. I suggest to rename this to `eval_option`

src/winml/modelkit/commands/eval.py
```diff
@@ -171,6 +171,57 @@ def eval(
...
+            if not cli_utils.is_cli_provided(ctx, "split") and ed.get("split"):
+                split = ed["split"]
+            if not cli_utils.is_cli_provided(ctx, "samples") and ed.get("samples"):
+                samples = ed["samples"]
+            if not column and ed.get("columns_mapping"):
+                column = tuple(f"{k}={v}" for k, v in ed["columns_mapping"].items())
+            if label_mapping is None and ed.get("label_mapping_file"):
+                lm_path = Path(ed["label_mapping_file"])
+                if lm_path.exists():
+                    label_mapping = lm_path
```
zhenchaoni: I would suggest not to put logic in this config default logic. You can just use the `config_file`'s value if the cli option is not provided. The remaining logic can stay just in the below `column_mapping` and `parsed_label_mapping` code. The only responsibility of the config_file is to give a default value when cli option is not provided.

src/winml/modelkit/commands/eval.py
```diff
@@ -171,6 +171,57 @@ def eval(
                 samples = build_cfg.quant.samples
             if not cli_utils.is_cli_provided(ctx, "dataset_name"):
                 dataset_name = build_cfg.quant.dataset_name
+        # Apply eval_dataset from config (CLI options take precedence)
+        if build_cfg.eval_dataset:
+            ed = build_cfg.eval_dataset
+            # Run build_script if needed to generate local dataset
+            build_script = ed.get("build_script")
+            if build_script:
```
zhenchaoni: I would suggest to decouple the config_file and the dataset_script logic. Currently they lives in the same place and easy to get mixed.

You can add a `dataset_script` option, it is for building the dataset for eval. This logic can be decoupled with the config_file logic. The config_file's only responsibility is to provide the value.

After the value is set, then you can `run_dataset_script` to build the dataset.

src/winml/modelkit/commands/eval.py
```diff
@@ -171,6 +171,57 @@ def eval(
...
+                # Use the path from config as cache dir (expand ~),
+                # fallback to ~/.cache/winml/eval_datasets/<script_stem>/
+                raw_path = ed.get("path", "")
+                if raw_path:
+                    cache_dir = Path(raw_path).expanduser()
+                else:
+                    cache_dir = (
+                        Path.home() / ".cache" / "winml" / "eval_datasets" / script_path.stem
+                    )
+                if not (cache_dir / "dataset_info.json").exists():
```
zhenchaoni: Do you think we can just put the cache_dir logic inside the dataset script? In that way, the user just
1. Write a dataset script, this script will be executed by us. 
2. Provide us a dataset path, we just load dataset from it.

The cache control can be inside the dataset script. Then our logic will be much simpler.

src/winml/modelkit/config/build.py
```diff
@@ -126,6 +126,7 @@ class WinMLBuildConfig:
     optim: WinMLOptimizationConfig = field(default_factory=WinMLOptimizationConfig)
     quant: WinMLQuantizationConfig | None = field(default_factory=WinMLQuantizationConfig)
     compile: WinMLCompileConfig | None = field(default_factory=WinMLCompileConfig)
+    eval_dataset: dict | None = None
```
DingmaomaoBJTU: WinMLEvaluationConfig

src/winml/modelkit/config/build.py
DingmaomaoBJTU, line 95: dynamic load

src/winml/modelkit/commands/eval.py
ssss141414: Now read from script output:

```
        script_output = result.stdout.strip()
        if script_output:
            dataset_path = script_output
```

Also update scripts

scripts/e2e_eval/datasets/build_fairface.py
```diff
@@ -95,11 +95,16 @@ def build_dataset(output_dir: Path) -> None:
...
+
 def main() -> None:
     parser = argparse.ArgumentParser(description="Build fairface validation dataset")
-    parser.add_argument("--output", type=Path, required=True, help="Output directory")
+    parser.add_argument("--output", type=Path, default=None, help="Output directory (default: ~/.cache/winml/eval_datasets/build_fairface)")
     args = parser.parse_args()
-    build_dataset(args.output)
+    output_dir = args.output or _DEFAULT_CACHE_DIR
+    build_dataset(output_dir)
+    print(output_dir)
```
DingmaomaoBJTU: could we remove this?

src/winml/modelkit/commands/eval.py
```diff
@@ -171,6 +185,52 @@ def eval(
...
                 dataset_name = build_cfg.quant.dataset_name
+        # Apply eval_option from config (CLI options take precedence)
+        if build_cfg.eval_option:
+            eo = build_cfg.eval_option
+            if not cli_utils.is_cli_provided(ctx, "dataset_path") and eo.dataset.path:
+                dataset_path = eo.dataset.path
+            if not cli_utils.is_cli_provided(ctx, "dataset_name") and eo.dataset.name:
+                dataset_name = eo.dataset.name
+            if not cli_utils.is_cli_provided(ctx, "split"):
+                split = eo.dataset.split
```
DingmaomaoBJTU: `split` 和 `samples` 跟 `dataset_path`/`dataset_name` 的处理方式不一致：后者有 `and eo.dataset.path` 的 guard 来避免用 `None` 覆盖，但 `split` 和 `samples` 因为 `DatasetConfig` 有默认值 (`"validation"` / `100`)，即使 JSON 里没写也会被覆盖进来。

目前恰好 CLI default 跟 `DatasetConfig` default 一致所以没问题，但如果将来有一边改了就会变成 subtle bug。建议要么也加 guard（比如只覆盖 JSON 里显式出现的 key），要么加个注释说明这是 intentional 的。

src/winml/modelkit/commands/eval.py
```diff
@@ -171,6 +185,52 @@ def eval(
...
+        logger.info("Building dataset via %s ...", script_path.name)
+        result = subprocess.run(  # noqa: S603
+            [sys.executable, str(script_path)],
+            capture_output=True,
+            text=True,
+            timeout=300,
+        )
+        if result.returncode != 0:
+            raise click.ClickException(
+                f"Dataset script failed: {result.stderr.strip()[-200:]}"
```
DingmaomaoBJTU: 两个小问题：
1. `result.stderr` 为空时，错误消息变成 `"Dataset script failed: "` 没有有用信息
2. 建议也带上 exit code

```suggestion
                f"Dataset script failed (exit {result.returncode}): {result.stderr.strip()[-200:] or '(no stderr)'}"
```

src/winml/modelkit/commands/eval.py
```diff
@@ -171,6 +185,52 @@ def eval(
...
+            )
+        # Use the last line of stdout as dataset path (earlier lines are log messages)
+        stdout = result.stdout.strip()
+        script_output = stdout.splitlines()[-1].strip() if stdout else ""
+        if script_output:
+            dataset_path = script_output
```
DingmaomaoBJTU: "stdout 最后一行 = dataset path" 的协议比较脆，如果 script 多打了一行空行或 warning 就会拿错路径。建议拿到 `script_output` 后验证一下路径是否存在：

```python
if script_output:
    script_output_path = Path(script_output)
    if not script_output_path.exists():
        raise click.ClickException(
            f"Dataset script output path does not exist: {script_output}"
        )
    dataset_path = script_output
```

src/winml/modelkit/commands/eval.py
```diff
@@ -171,6 +185,52 @@ def eval(
                 samples = build_cfg.quant.samples
             if not cli_utils.is_cli_provided(ctx, "dataset_name"):
                 dataset_name = build_cfg.quant.dataset_name
+        # Apply eval_option from config (CLI options take precedence)
```
DingmaomaoBJTU: 这里 `eval_option` 是在 `quant` 之后 apply 的，所以 `eval_option` 的 `dataset_path` / `samples` 等会覆盖掉上面从 `quant` 拿到的值。优先级链：CLI > eval_option > quant defaults。

建议在注释里把完整的优先级写清楚，方便后续维护者理解 fallback 顺序。

scripts/e2e_eval/datasets/build_fairface.py
DingmaomaoBJTU: 这行 `print` 现在是 load-bearing 的（eval CLI 靠它拿 dataset 路径），但看起来像是 debug print，容易被人误删。建议加个注释说明用途：

```suggestion
    # Print output_dir as last line — the eval CLI reads this to locate the built dataset.
    print(output_dir)
```

其他几个 build script 同理。

src/winml/modelkit/config/build.py
```diff
@@ -126,6 +127,7 @@ class WinMLBuildConfig:
     optim: WinMLOptimizationConfig = field(default_factory=WinMLOptimizationConfig)
     quant: WinMLQuantizationConfig | None = field(default_factory=WinMLQuantizationConfig)
     compile: WinMLCompileConfig | None = field(default_factory=WinMLCompileConfig)
+    eval_option: WinMLEvaluationConfig | None = None
```
DingmaomaoBJTU: Nit: 文件顶部的 Configuration Hierarchy 注释（line 14-19）和 `WinMLBuildConfig` 的 class docstring 里列的字段（loader, export, optim, quant, compile）都没包含 `eval_option`，建议一并更新保持同步。

src/winml/modelkit/commands/eval.py
```diff
@@ -171,6 +185,52 @@ def eval(
...
+        if build_cfg.eval_option:
```
zhenchaoni: I suggest not to call `eval_option`, just `eval`. This is consistent with other options like `quant`.

src/winml/modelkit/commands/eval.py
```diff
@@ -171,6 +185,52 @@ def eval(
...
+            [sys.executable, str(script_path)],
+            capture_output=True,
+            text=True,
+            timeout=300,
+        )
+        if result.returncode != 0:
+            raise click.ClickException(
+                f"Dataset script failed: {result.stderr.strip()[-200:]}"
+            )
+        # Use the last line of stdout as dataset path (earlier lines are log messages)
```
zhenchaoni: I would suggest not to read the dataset script output to know the dataset path. You can just config the local dataset path in the config. As long as the dataset script and the path agree, it will be OK. And both the dataset script and the path is provided by the user.

For example, in models_with_cc.json, we have local path defined like 

```
    "dataset_config": {
      "build_script": "scripts/e2e_eval/datasets/build_fairface.py",
      "path": "~/.cache/winml/eval_datasets/build_fairface",
      "metric": "top1_accuracy",
      "winml_metric_key": "accuracy",
      "columns_mapping": {
        "label_column": "gender"
      }
    }
```

src/winml/modelkit/commands/eval.py
```diff
@@ -171,6 +185,52 @@ def eval(
...
+                samples = eo.dataset.samples
+            if not column and eo.dataset.columns_mapping:
+                column = tuple(f"{k}={v}" for k, v in eo.dataset.columns_mapping.items())
+            if label_mapping is None and eo.label_mapping_file:
+                label_mapping = Path(eo.label_mapping_file)
+            if dataset_script is None and eo.dataset_script:
+                dataset_script = eo.dataset_script
+
+    # Run dataset_script if provided (requires --trust-remote-code)
+    if dataset_script:
```
zhenchaoni: The logic to run dataset_script should be in a separate method.

src/winml/modelkit/commands/eval.py
zhenchaoni: Since we will support config file as input, this part needs to be refactored.
1. DatasetConfig and WinMLEvaluationConfig should be loaded at earlier part of this method. You merge the values from option or the config file
2. Then you apply the "resolve logic", resolve model id, column_mapping, device etc. The resolve **updates** the config object of the first step in place.
3. Then you pass the config object directly to the evaluator method.

In this way, we are around the **config object". We load it, we sets some value to it by resolve logic, we pass it down the stream. So we can have a good encapsulation.


Below is a detail of this comment

---

**`WinMLEvaluationConfig` isn't really being used as the config**

Two problems with the current flow in eval.py:

1. **Not encapsulated.** `WinMLEvaluationConfig` is only constructed at the very end. Until then, everything lives in loose locals (`task`, `samples`, `dataset_path`, `column`, `label_mapping`, `dataset_script`, …).
2. **Lots of hand‑written if/else.** Merging config‑file values into those locals is a ~20‑line `if not is_cli_provided(...)` cascade, and it silently drops fields that weren't wired (e.g. `device`, `output_path`, `shuffle`, `streaming`). Adding a new field means touching three places.

**Suggested flow**

1. Build the `WinMLEvaluationConfig` **at the top** of the command — first from defaults, then overlay the config file, then overlay only CLI‑provided options.
2. Resolution steps mutate the config **in place** (`-m` → `model_id`/`model_path`, parse `--column`, load `--label-mapping`, `resolve_device`, run `build_script`).
3. Pass the single config object to `evaluate(cfg)`.

Sketch:

```python
def eval(ctx, *, model, model_id, dataset_path, dataset_name, task, device,
         samples, split, shuffle, streaming, column, label_mapping,
         output, dataset_script, trust_remote_code, verbose,
         show_schema, config_file):

    # 1. Build the config once, with clear precedence.
    cfg = WinMLEvaluationConfig()                       # defaults
    if config_file is not None:
        cfg = cfg.merge(load_eval_config(config_file))  # config file overrides defaults
    cfg = cfg.merge(eval_config_from_cli(ctx, locals()))  # CLI overrides config file

    if show_schema:
        _print_schema(cfg.task); return

    # 2. Resolve in place — each step mutates cfg, returns nothing.
    _resolve_model(cfg, model, model_id)            # splits -m into model_id / model_path
    _resolve_device(cfg)                            # 'auto' -> concrete device
    _resolve_columns(cfg, column)                   # parses key=value into cfg.dataset.columns_mapping
    _resolve_label_mapping(cfg)                     # loads JSON file referenced by cfg
    _resolve_dataset_script(cfg, trust_remote_code) # runs build_script, sets cfg.dataset.path

    cfg.validate()
    logger.debug("Effective eval config: %s", cfg.to_dict())

    # 3. Hand the single object to the evaluator.
    result = evaluate(cfg)
    _write_and_display(result, cfg.output_path)
```

Where `merge` is a generic "take non‑None / CLI‑provided fields from `other`" — no per‑field if/else. With this, adding a new option is one place: add it to the dataclass, and both the config file and CLI pick it up automatically.

zhenchaoni (changes requested):

scripts/e2e_eval/datasets/build_fairface.py
ssss141414: 我们不采用这种方式啦，会把path传进去给script，详见zhenchao的comments

src/winml/modelkit/commands/eval.py
```diff
@@ -161,113 +175,250 @@ def eval(
...
+    if ds_overrides:
+        overrides["dataset"] = ds_overrides
+
+    if overrides:
+        cfg = merge_config(cfg, overrides)
+
+    return cfg
+
+
+def _resolve_model(cfg: object, model_arg: str | None) -> None:
```
zhenchaoni: You need to resolve conflicts from main. The main branch already has a resolve model logic there. Try to use the main's logic.

src/winml/modelkit/commands/eval.py
```diff
@@ -161,113 +175,250 @@ def eval(
...
-        build_cfg = cli_utils.load_build_config(config_file)
-        if build_cfg.loader and not cli_utils.is_cli_provided(ctx, "task"):
-            task = build_cfg.loader.task
-        if build_cfg.quant:
-            if not cli_utils.is_cli_provided(ctx, "samples"):
-                samples = build_cfg.quant.samples
-            if not cli_utils.is_cli_provided(ctx, "dataset_name"):
-                dataset_name = build_cfg.quant.dataset_name
+        cfg = _merge_from_config_file(cfg, config_file)
+    cfg = _merge_cli_overrides(cfg, ctx, column, label_mapping)
```
zhenchaoni: I think you can have one merge method. It simply merges 3 things, config default <- config file <- cli inline option.

src/winml/modelkit/commands/eval.py
```diff
@@ -161,113 +175,250 @@ def eval(
...
-    if model is not None:
-        p = Path(model)
+
+# ── CLI-to-config field mappings ──
+# Adding a new option only requires one entry here; both config-file and CLI
+# paths pick it up automatically via merge_config.
+_CLI_TO_CONFIG: dict[str, str] = {
+    "model_id": "model_id",
+    "task": "task",
+    "device": "device",
```
zhenchaoni: Below is a suggestion to avoid defining a full list of mapping here. Please consider if you can leverage. Alternatively, you can follow the same pattern as `quantize.py` which adds many `if` clauses.

## Avoid hardcoded CLI-to-config field mappings

The `_CLI_TO_CONFIG` and `_CLI_TO_DATASET` dictionaries duplicate knowledge that already lives in the dataclass definitions. Every time a field is added to `WinMLEvaluationConfig` or `DatasetConfig`, these dicts must be manually updated — easy to forget and drift.

Note: `quantize.py` already has this same tech debt (hardcoded `if` chains for `samples`, `method`, `weight_type`, `activation_type`, etc.). This PR shouldn't propagate the pattern further.

### Why this can be eliminated

Two APIs already give us everything we need:

1. **`dataclasses.fields(cls)`** (stdlib) — returns all field names defined on a dataclass at runtime. This replaces the *values* side of the hardcoded dict. Instead of manually listing `"path"`, `"name"`, `"split"`, `"samples"`, ... we just ask `DatasetConfig` what fields it has.

2. **`ctx.get_parameter_source(param_name)`** (Click 8 API, wrapped by our `is_cli_provided()`) — tells us whether a CLI param was explicitly passed. Combined with `ctx.params` (which maps every CLI param name to its value), this replaces the *keys* side. We don't need to enumerate CLI param names — we iterate `ctx.params` and check each one.

The key insight: for most params, the CLI name already matches the dataclass field name (`split` → `split`, `samples` → `samples`, `shuffle` → `shuffle`, `streaming` → `streaming`). We only need an explicit rename for the few exceptions where names differ.

### Suggestion

Use `dataclasses.field(metadata=...)` to annotate the few fields whose CLI param name differs from the field name. This keeps the mapping on the field that owns it — single source of truth.

**Step 1:** Annotate dataclass fields with `metadata` where CLI names differ:

```python
from dataclasses import dataclass, field

@dataclass
class DatasetConfig:
    path: str | None = field(default=None, metadata={"cli_name": "dataset_path"})
    name: str | None = field(default=None, metadata={"cli_name": "dataset_name"})
    split: str = "validation"       # matches CLI param name — no metadata needed
    samples: int = 100              # matches CLI param name — no metadata needed
    shuffle: bool = True
    streaming: bool = False
```

**Step 2:** Add a generic utility to `cli.py`:

```python
import dataclasses

def collect_cli_overrides(ctx: click.Context, cls: type) -> dict[str, object]:
    """Collect CLI-provided values that match fields on a dataclass.

    Iterates ``ctx.params`` and returns ``{field_name: value}`` for every
    CLI param that was explicitly provided AND maps to a field on ``cls``.

    Name mapping is derived from ``field(metadata={"cli_name": ...})``
    on the dataclass. Fields without ``cli_name`` metadata match by name.

    Args:
        ctx: Click context.
        cls: Target dataclass — its fields define the valid key set.
    """
    # Build reverse map: cli_name -> field_name from metadata
    rename: dict[str, str] = {}
    valid_fields: set[str] = set()
    for f in dataclasses.fields(cls):
        valid_fields.add(f.name)
        cli_name = f.metadata.get("cli_name")
        if cli_name:
            rename[cli_name] = f.name

    overrides: dict[str, object] = {}
    for cli_name, value in ctx.params.items():
        field_name = rename.get(cli_name, cli_name)
        if field_name in valid_fields and is_cli_provided(ctx, cli_name):
            overrides[field_name] = value
    return overrides
```

**Step 3:** Call sites become trivial — no mapping dicts anywhere:

```python
overrides = cli_utils.collect_cli_overrides(ctx, WinMLEvaluationConfig)
ds_overrides = cli_utils.collect_cli_overrides(ctx, DatasetConfig)
```

### Why this is better

- **Single source of truth** — the CLI name mapping lives on the field that owns it, not in a separate dict in the command file
- **Self-documenting** — reading `DatasetConfig` tells you which fields have different CLI names
- **Zero-arg call sites** — callers never need to know about name mismatches
- **Self-maintaining** — adding a new same-named field to the dataclass + CLI option requires zero changes to the merge logic
- **Fields without `metadata`** match by name automatically (the common case — only 2-3 fields need annotation)

As a follow-up, `quantize.py` can also adopt this utility, eliminating its existing hardcoded `if` chain too.

zhenchaoni (changes requested):

src/winml/modelkit/commands/eval.py
```diff
@@ -118,6 +118,18 @@
...
     help="Enable verbose output.",
 )
+@click.option(
+    "--dataset-script",
+    type=str,
+    default=None,
+    help="Path to a Python script that builds the evaluation dataset.",
+)
+@click.option(
+    "--trust-remote-code",
```
xieofxie: add a shared option for this?

ssss141414: Sure. And help message will differ.

zhenchaoni (approved):
Please add unit test for the config build. Ensure the cli option > config file > default

zhenchaoni (approved):

PR #384, opened 2026-04-23
feat: implement image-to-text evaluator with teacher-forcing perplexity

xieofxie:
## Summary

Implements `WinMLImageToTextEvaluator` for the `image-to-text` task, enabling `winml eval` to measure accuracy regression across all `vision-encoder-decoder` models (donut, trocr, vit-gpt2, etc.).

### Metric: image-conditional perplexity under teacher forcing

For each (image, caption) pair, run a single forward pass with `decoder_input_ids = caption[:-1]` and compute cross-entropy loss vs `caption[1:]`. Aggregate as mean NLL + exp(NLL) = perplexity. Comparing PyTorch vs. WinML (ONNX) perplexity surfaces accuracy regressions from quantization or export.

**Why not the HF pipeline?** The `image-to-text` pipeline runs autoregressive generation — slow, non-deterministic (beam search), and doesn't expose per-token logits. Teacher forcing computes the full likelihood in a single pass, making it fast and directly comparable between PyTorch and ONNX.

### End-to-end result (naver-clova-ix/donut-base-finetuned-cord-v2, 100 samples, CPU)

```
nll: 7.8136  |  perplexity: 2474.1393
```

High perplexity is expected — donut is a document-understanding model evaluated out-of-domain on MSCOCO captions. The value is stable and comparable between PyTorch and WinML for regression detection.

## Changed files

| File | Change |
|---|---|
| `src/winml/modelkit/eval/image_to_text_evaluator.py` | New evaluator class |
| `src/winml/modelkit/eval/evaluate.py` | Register evaluator + default dataset (`clip-benchmark/wds_mscoco_captions`) |
| `src/winml/modelkit/eval/__init__.py` | Export `WinMLImageToTextEvaluator` |
| `scripts/e2e_eval/run_pytorch_baseline.py` | Add `"image-to-text": "perplexity"` to `_TASK_HF_METRIC_KEY` |
| `pyproject.toml` | Add `sentencepiece>=0.2` (required by `XLMRobertaTokenizerFast` used by DonutProcessor) |
| `tests/unit/eval/test_image_to_text_evaluator.py` | 27 unit tests |

## Design notes

**ONNX static decoder shape handling:** ONNX models compiled with fixed shapes require `decoder_input_ids` to be exactly the right length. The evaluator reads `model.io_config["input_shapes"]` to detect this and pads/truncates accordingly, excluding pad positions from NLL accumulation.

**Default dataset:** `clip-benchmark/wds_mscoco_captions` — parquet/WebDataset, no `.py` loading script, supports streaming. The originally-considered `nlphuji/flickr30k` raises `RuntimeError: Dataset scripts are no longer supported` on current HuggingFace.

**`pad_token_id` type guard:** `pad_id: int = raw_pad if isinstance(raw_pad, int) else 0` — prevents `TypeError` from `torch.full()` when `pad_token_id` is `None` or absent.

## Tests

27 tests covering: schema, init, column mapping, `_decoder_seq_len` (static vs. dynamic), `_logits` (dict/object), `_score_sample` (happy path, short skip, image/forward failures, pad/truncate for static shapes), `compute` (uniform logits → perplexity == vocab_size, None skipping, list captions, no-valid-samples error), registry entries.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -47,6 +47,7 @@ dependencies = [
   "rich>=14",
   "scikit-learn>=1.7",
   "scipy>=1.15,<1.16",
+  "sentencepiece>=0.2",
   "seqeval>=1.2.2",
   "snakemd>=2",
   "timm>=1.0.22",
--- a/scripts/e2e_eval/run_pytorch_baseline.py
+++ b/scripts/e2e_eval/run_pytorch_baseline.py
@@ -65,6 +65,7 @@ _TASK_HF_METRIC_KEY: dict[str, str] = {
     "sentence-similarity": "cosine_spearman",
     "image-feature-extraction": "knn_top1_accuracy",
     "fill-mask": "pseudo_perplexity",
+    "image-to-text": "perplexity",
 }
 
 
@@ -76,8 +77,8 @@ _TASK_HF_METRIC_KEY: dict[str, str] = {
 def _load_pytorch_model(model_id: str, task: str, device_str: str):
     """Load a native PyTorch model with the task-appropriate AutoModel class."""
     import torch
-
     from transformers import AutoConfig
+
     from winml.modelkit.loader.task import resolve_task_and_model_class
 
     config = AutoConfig.from_pretrained(model_id)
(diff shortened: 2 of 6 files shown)


xieofxie:
dup of zhenchao's work

PR #385, opened 2026-04-23
update Resize, ReduceMean

xieofxie:


src/winml/modelkit/pattern/op_input_gen/resize_input_generator.py
```diff
@@ -150,6 +150,21 @@ def get_input_and_infinite_attribute_combinations(self) -> list[dict[str, InputC
                 }
             )
 
+            # p1 model - segmentation with sizes (pytorch_half_pixel + linear + sizes)
+            combinations.append(
+                {
+                    "coordinate_transformation_mode": "pytorch_half_pixel",
```
xieofxie, line 156: There are many cases.. Will create a pr to see if we could enumerate all combinations of them

vortex-captain (approved):

PR #373, opened 2026-04-21
refactor: rename EP short names to snake_case and AITK to Foundry Toolkit

timenick:
Closes #372.

## Summary

- Rename EP short names for consistency (snake_case, match ORT symbol casing)
  - `directml` → `dml`
  - `tensorrt` / `tensorrt_rtx` → `nv_tensorrt_rtx`
- Rename `WinMLCompileConfig.for_tensorrt` → `for_nv_tensorrt_rtx` (+ dispatch table + tests)
- Rename pytest markers: `directml` → `dml`, `tensorrt` → `nv_tensorrt_rtx`
- Drop dead classic `TensorrtExecutionProvider` references (not supported)
- Fix bug: docstring cited non-existent `DirectMLExecutionProvider` (ORT symbol is `DmlExecutionProvider`)
- Prose corrections: `DirectML` → `Dml` and `TensorRT` → `NvTensorRTRTX` in EP-list contexts
- Rebrand: `"group": "AITK"` → `"group": "Foundry Toolkit"` in testset JSONs, build script, README, and viewer HTML
- Drop `onnxruntime-directml` references — DML is shipped inside `onnxruntime-windowsml` on Windows

## Kept intentionally

- `DirectML` stays when referring to the Microsoft product/API (e.g., "Factory for DirectML compilation" docstring)
- Pip package literals (real PyPI names) untouched where they are the actual install target

PR #392, opened 2026-04-24
fix: allow more EP in static analyzer, infer device automatically

### LINKED ISSUE #391: Static analyzer blocks non-QNN/OpenVINO/VitisAI EPs (e.g. DML)
## Problem

`winml perf -m microsoft/resnet-50 --device gpu --ep dml` fails because the static analyzer's `--ep` option only accepts three EPs (QNN, OpenVINO, VitisAI). Any other EP like DML, CPU, TensorRT, or MIGraphX is rejected at the CLI level, blocking downstream commands that depend on the analyzer.

Additionally:
- `--device` defaults to NPU even for EPs where that makes no sense (e.g. DML → GPU)
- No user-friendly output when an EP has no rule data — the summary renders empty
- Noisy warnings for benign conditions (empty node names, no pattern matches)

## Expected Behavior

- `--ep` should accept all known ORT execution providers (full names and aliases)
- Device should be auto-inferred from the EP when not specified (DML → GPU, QNN → NPU)
- Invalid EP+device combos should produce actionable errors (e.g. "DmlExecutionProvider only supports GPU.")
- EPs without rule data should be gracefully skipped with a clear message

## Reproduction

```bash
winml analyze --model model.onnx --ep dml
# Error: Invalid value for '--ep': 'dml' is not one of ...

winml perf -m microsoft/resnet-50 --device gpu --ep dml
# Blocked at static analyzer step
```

DingmaomaoBJTU:
## Summary

Fixes #391

- **Expand `--ep` choices** — `ALL_EP_NAMES` now derived from `sysinfo`'s EP→device map + aliases; adds `cpu`, `dml`, `tensorrt`, `migraphx` while keeping `click.Choice` validation
- **Auto-infer `--device`** — when not specified, probes rule data first, then falls back to `sysinfo`'s `_EP_DEVICE_MAP` (DML→GPU, QNN→NPU, CPU→CPU, OpenVINO→NPU)
- **Validate EP+device combos** — `--ep dml --device CPU` → `ERROR: DmlExecutionProvider only supports GPU.`; only validates when user explicitly passes `--device`
- **Graceful skip for EPs without rule data** — pattern detection still runs; summary shows clear message instead of empty output
- **Demote noisy warnings to info** — empty node names, no pattern matches, empty results
- **Display resolved target** — header shows `🎯 Target: DmlExecutionProvider on GPU`

## Changes

| File | What |
|---|---|
| `utils/constants.py` | `SUPPORTED_EPS` from `sysinfo`; new aliases (cpu, dml, tensorrt, migraphx) |
| `utils/cli.py` | `--ep` uses expanded `ALL_EP_NAMES` via `click.Choice` |
| `analyze/utils/ep_utils.py` | New `get_devices_with_rule_data()`, `has_rule_data_for_ep()` |
| `analyze/analyzer.py` | Skip EPs without rule data; auto-infer device |
| `commands/analyze.py` | CLI-side device resolution, EP+device validation, Target display, empty-result UX |
| `pattern/base.py`, `pattern_extractor.py`, `output_aggregator.py` | warning → info |
| 3 test files | 88 tests (32 new) |

### DIFF
--- a/src/winml/modelkit/analyze/utils/ep_utils.py
+++ b/src/winml/modelkit/analyze/utils/ep_utils.py
@@ -54,8 +54,9 @@ def infer_ihv_from_ep_name(ep_name: str) -> IHVType:
     if "openvino" in ep_lower or "intel" in ep_lower:
         return IHVType.INTEL
 
-    # VitisAI / AMD / ACE (AMD)
-    if "amd" in ep_lower or "quark" in ep_lower or "vitis" in ep_lower or "ace" in ep_lower:
+    # VitisAI / MIGraphX / AMD / ACE (AMD)
+    amd_keywords = ("amd", "quark", "vitis", "ace", "migraphx")
+    if any(kw in ep_lower for kw in amd_keywords):
         return IHVType.AMD
 
     raise ValueError(
--- a/src/winml/modelkit/analyze/utils/ep_utils.py
+++ b/src/winml/modelkit/analyze/utils/ep_utils.py
@@ -62,3 +62,52 @@ def infer_ihv_from_ep_name(ep_name: str) -> IHVType:
         f"Unknown execution provider: {ep_name}. "
         "Supported: QNNExecutionProvider, OpenVINOExecutionProvider, VitisAIExecutionProvider"
     )
+
+
+def get_devices_with_rule_data(ep_name: str) -> list[str]:
+    """Return all devices supported by an EP.
+
+    First probes rule zip search directories for files matching
+    ``{ep_name}_{device}_*.zip``.  If no rule data is found, falls
+    back to the EP→device mapping from :func:`sysinfo.get_ep_device_map`.
+
+    Args:
+        ep_name: Full execution provider name (e.g., ``"QNNExecutionProvider"``).
+
+    Returns:
+        List of device strings (e.g., ``["NPU", "GPU"]``), empty if
+        the EP is completely unknown.
+    """
+    from winml.modelkit.sysinfo.device import get_ep_device_map
+
+    devices = [d for d in ("NPU", "GPU", "CPU") if has_rule_data_for_ep(ep_name, d)]
+    if devices:
+        return devices
+    # Fallback: derive from the authoritative EP→device mapping
+    device_str = get_ep_device_map().get(ep_name, "")
+    return [d.upper() for d in device_str.split("/") if d]
+
+
+def has_rule_data_for_ep(ep_name: str, device: str) -> bool:
+    """Check whether runtime check rule data exists for a given EP and device.
+
+    Probes the rule zip search directories for any zip file matching the
+    naming convention ``{ep_name}_{device}_*.zip``.  This is a fast
+    filesystem check — no zip contents are read.
+
+    Args:
+        ep_name: Full execution provider name (e.g., ``"QNNExecutionProvider"``).
+        device: Device type (e.g., ``"NPU"``, ``"GPU"``, ``"CPU"``).
+
+    Returns:
+        ``True`` if at least one rule zip exists for this EP + device pair.
+    """
+    from .rule_loader import get_runtime_rules_search_dirs
+
+    prefix = f"{ep_name}_{device.upper()}_"
+    for search_dir in get_runtime_rules_search_dirs():
+        if not search_dir.is_dir():
+            continue
+        if any(search_dir.glob(f"{prefix}*.zip")):
+            return True
+    return False
--- a/src/winml/modelkit/analyze/utils/ep_utils.py
+++ b/src/winml/modelkit/analyze/utils/ep_utils.py
@@ -79,9 +79,16 @@ def get_devices_with_rule_data(ep_name: str) -> list[str]:
         List of device strings (e.g., ``["NPU", "GPU"]``), empty if
         the EP is completely unknown.
     """
-    from winml.modelkit.sysinfo.device import get_ep_device_map
+    from ...sysinfo.device import get_ep_device_map
 
-    devices = [d for d in ("NPU", "GPU", "CPU") if has_rule_data_for_ep(ep_name, d)]
+    # Priority order: NPU > GPU > CPU (first match used as default device)
+    known_devices = {d.upper() for v in get_ep_device_map().values() for d in v.split("/") if d}
+    priority = ["NPU", "GPU", "CPU"]
+    probe_order = [d for d in priority if d in known_devices]
+    # Append any devices not in the priority list
+    probe_order.extend(d for d in sorted(known_devices) if d not in priority)
+
+    devices = [d for d in probe_order if has_rule_data_for_ep(ep_name, d)]
     if devices:
         return devices
     # Fallback: derive from the authoritative EP→device mapping
--- a/src/winml/modelkit/commands/analyze.py
+++ b/src/winml/modelkit/commands/analyze.py
@@ -405,9 +405,7 @@ def _render_analysis_summary(
     required=False, optional_message="If not specified, analyzes all supported EPs"
 )
 @cli_utils.device_option(
-    required=False,
-    optional_message="If not specified, inferred from EP rule data",
-    default=None,
+    required=False, optional_message="If not specified, uses NPU as default", default="NPU"
 )
 @cli_utils.verbosity_options
 @cli_utils.build_config_option
@@ -504,22 +502,10 @@ def analyze(
         )
 
         ep_normalized = normalize_ep_name(ep)
-        device_user_specified = device is not None
-
-        # Resolve device: explicit > inferred from EP support > None
-        if device is None and ep_normalized:
-            available = get_devices_with_rule_data(ep_normalized)
-            if available:
-                device = available[0]
-                logger.info(
-                    "Device not specified — resolved to %s for %s",
-                    device,
-                    ep_normalized,
-                )
 
         # Validate only when the user explicitly specified --device
         if (
-            device_user_specified
+            cli_utils.is_cli_provided(ctx, "device")
             and ep_normalized
             and device
             and not has_rule_data_for_ep(ep_normalized, device)
--- a/src/winml/modelkit/commands/analyze.py
+++ b/src/winml/modelkit/commands/analyze.py
@@ -299,6 +299,9 @@ def _render_analysis_summary(
     results: list,
     ep_instance_counts: dict[str, dict[str, dict[str, int]]],
     ep_patterns: dict[str, dict[str, dict]] | None = None,
+    *,
+    ep: str | None = None,
+    device: str | None = None,
 ) -> None:
     """Render the Analysis Summary section after pattern detection.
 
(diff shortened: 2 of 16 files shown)


src/winml/modelkit/commands/analyze.py
```diff
@@ -384,7 +405,9 @@ def _render_analysis_summary(
     required=False, optional_message="If not specified, analyzes all supported EPs"
 )
 @cli_utils.device_option(
-    required=False, optional_message="If not specified, uses NPU as default", default="NPU"
+    required=False,
+    optional_message="If not specified, inferred from EP rule data",
```
xieofxie: "inferred from EP rule data" not very useful for user.. Even I don't understand it..

xieofxie: I feel NPU is fine since analyze is for a fixed target. Caller should know what to check

src/winml/modelkit/utils/constants.py
```diff
@@ -21,6 +23,10 @@
     "ov": "OpenVINOExecutionProvider",
     "vitisai": "VitisAIExecutionProvider",
     "vitis": "VitisAIExecutionProvider",
+    "cpu": "CPUExecutionProvider",
+    "dml": "DmlExecutionProvider",
+    "tensorrt": "NvTensorRTRTXExecutionProvider",
```
xieofxie: how about trtrtx? because tensorrt is also an EP

src/winml/modelkit/commands/analyze.py
xieofxie: Or if we want to give user single command experience, by default, we check
- all support ep + device combinations
- all ep + device combinations

src/winml/modelkit/utils/constants.py
timenick: https://github.com/microsoft/WinML-ModelKit/pull/373 according to this PR, it should be nv_tensorrt_rtx

src/winml/modelkit/utils/constants.py
```diff
@@ -7,12 +7,14 @@
...
 
-# Supported execution providers
-SUPPORTED_EPS = [
-    "QNNExecutionProvider",
-    "OpenVINOExecutionProvider",
-    "VitisAIExecutionProvider",
-]
+# Supported execution providers — derived from sysinfo's authoritative EP→device map.
+def _get_supported_eps() -> list[str]:
+    from winml.modelkit.sysinfo.device import get_ep_device_map
```
timenick: **Absolute import in source code** — `CLAUDE.md` § Import Rules: *"Source code (`src/`): Must use relative imports through `__init__.py`, never absolute `from winml.modelkit.*` paths."* This should be `from ..sysinfo.device import get_ep_device_map` (or `from ..sysinfo import get_ep_device_map` if you re-export it from `sysinfo/__init__.py`).

Same violation at [src/winml/modelkit/analyze/utils/ep_utils.py:81](src/winml/modelkit/analyze/utils/ep_utils.py#L81) — that one needs three dots: `from ...sysinfo.device import get_ep_device_map`.

🤖 Generated with Claude Code

src/winml/modelkit/analyze/utils/ep_utils.py
```diff
@@ -62,3 +62,52 @@ def infer_ihv_from_ep_name(ep_name: str) -> IHVType:
...
+
+    Returns:
+        ``True`` if at least one rule zip exists for this EP + device pair.
+    """
+    from .rule_loader import get_runtime_rules_search_dirs
+
+    prefix = f"{ep_name}_{device.upper()}_"
+    for search_dir in get_runtime_rules_search_dirs():
+        if not search_dir.is_dir():
+            continue
```
timenick, line 118: **New public functions not exported in `__init__.py`** — `has_rule_data_for_ep` and `get_devices_with_rule_data` are public (no `_` prefix) but [analyze/utils/__init__.py](src/winml/modelkit/analyze/utils/__init__.py) wasn't updated.

`CLAUDE.md` § Import Rules: *"When adding new public classes/functions, add them to both the import and `__all__`."*

This is also the reason [tests/unit/analyze/test_has_rule_data.py:13-16](tests/unit/analyze/test_has_rule_data.py#L13-L16) reaches into `...ep_utils` directly. Once exported, the test can do `from winml.modelkit.analyze.utils import get_devices_with_rule_data, has_rule_data_for_ep`.

🤖 Generated with Claude Code

src/winml/modelkit/analyze/utils/ep_utils.py
```diff
@@ -62,3 +62,52 @@ def infer_ihv_from_ep_name(ep_name: str) -> IHVType:
...
+    Args:
+        ep_name: Full execution provider name (e.g., ``"QNNExecutionProvider"``).
+
+    Returns:
+        List of device strings (e.g., ``["NPU", "GPU"]``), empty if
+        the EP is completely unknown.
+    """
+    from winml.modelkit.sysinfo.device import get_ep_device_map
+
+    devices = [d for d in ("NPU", "GPU", "CPU") if has_rule_data_for_ep(ep_name, d)]
```
timenick: **Hardcoded device tuple** — `("NPU", "GPU", "CPU")` is the universe of devices for the rule-data probe, but the docstring above says the function falls back to `get_ep_device_map()` as the *authoritative* source of EP→device. Pulling the device universe from the same source keeps the two paths aligned and means new device types (DSP, FPGA, etc.) added to `_EP_DEVICE_MAP` are picked up automatically:

```python
from winml.modelkit.sysinfo.device import get_ep_device_map  # adjust to relative
_KNOWN_DEVICES = sorted({d.upper() for v in get_ep_device_map().values() for d in v.split("/") if d})
```

or compute on each call if startup cost is a concern.

🤖 Generated with Claude Code

timenick (commented):
Review against `CLAUDE.md`. Tests pass, functional logic is sound. Findings are mostly project-rule violations (imports, `__init__.py` exports) plus a misleading display fallback and one weak test assertion.

🤖 Generated with Claude Code

src/winml/modelkit/commands/analyze.py
```diff
@@ -475,10 +498,51 @@ def analyze(
...
+            else:
+                logger.error(
+                    "%s has no rule data for %s.",
+                    ep_normalized,
+                    device,
+                )
+            sys.exit(2)
 
+        ep_label = ep_normalized or "all EPs"
+        device_label = device or "NPU"
```
timenick, line 529: **Misleading device label** — `device_label = device or "NPU"` displays `NPU` whenever device resolution fails to land on a value, even though the analyzer may not actually use NPU. Two paths hit this:

1. User runs `winml analyze --ep dml` and DML has no rule data → `device` stays `None`, but the header prints `🎯 Target: DmlExecutionProvider on NPU` (DML is GPU-only).
2. User omits `--ep` → `ep_normalized` is `None`, the device-resolution branch is skipped (`if device is None and ep_normalized:`), so `device` stays `None` and the header prints `on NPU` even though `analyze_from_proto` will pick a device per-EP later.

Suggest `device or "auto"` or `device or "(auto-resolved per EP)"` so the display tracks the actual resolution behavior.

🤖 Generated with Claude Code

tests/unit/analyze/test_static_analyzer_cli.py
```diff
@@ -620,3 +633,136 @@ def test_analyze_called_with_correct_parameters(
...
+        mock_instance.analyze.return_value = mock_analyzer_result
+        mock_analyzer_class.return_value = mock_instance
+
+        # dml without --device should not exit 2 on validation
+        result = runner.invoke(
+            analyze,
+            ["--model", str(model_file), "--ep", "dml"],
+        )
+        # Should proceed to analysis (not fail on validation)
+        assert result.exit_code != 2 or mock_instance.analyze.called
```
timenick: **Assertion is too permissive** — `assert result.exit_code != 2 or mock_instance.analyze.called` passes whenever `exit_code` is anything other than 2, including `exit_code=1` with `analyze` never called. The intended invariant is *both* (CLI proceeded past validation **and** reached the analyzer), so use `and`:

```python
assert result.exit_code != 2 and mock_instance.analyze.called
```

(Also worth confirming the success exit code — `assert result.exit_code == 0` would be even tighter.)

🤖 Generated with Claude Code

src/winml/modelkit/commands/analyze.py
DingmaomaoBJTU: Fixed — reverted to `default="NPU"` with original help text. Also switched to `is_cli_provided(ctx, "device")` for validation gating and removed the dead device-inference code path.

src/winml/modelkit/utils/constants.py
DingmaomaoBJTU: Good call — renamed to `nv_tensorrt_rtx` to match the codebase convention from PR #373.

DingmaomaoBJTU: Done — changed to `nv_tensorrt_rtx`.

src/winml/modelkit/utils/constants.py
DingmaomaoBJTU: Fixed both — `constants.py` now uses `from ..sysinfo.device`, `ep_utils.py` uses `from ...sysinfo.device`.

src/winml/modelkit/analyze/utils/ep_utils.py
DingmaomaoBJTU: Fixed — added `get_devices_with_rule_data` and `has_rule_data_for_ep` to both the import and `__all__` in `analyze/utils/__init__.py`.

src/winml/modelkit/analyze/utils/ep_utils.py
DingmaomaoBJTU: Fixed — now derives the device universe from `get_ep_device_map().values()` with a priority ordering (NPU > GPU > CPU) so new device types are picked up automatically.

src/winml/modelkit/commands/analyze.py
DingmaomaoBJTU: This is no longer an issue — with `default="NPU"` restored, `device` is never `None`, so `device or "NPU"` always evaluates to the actual device value.

tests/unit/analyze/test_static_analyzer_cli.py
DingmaomaoBJTU: Fixed — tightened to `assert result.exit_code == 0` and `assert mock_instance.analyze.called`.

tests/unit/analyze/test_has_rule_data.py
```diff
@@ -0,0 +1,181 @@
...
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+import pytest
+
+from winml.modelkit.analyze.utils.ep_utils import (
+    get_devices_with_rule_data,
+    has_rule_data_for_ep,
+    infer_ihv_from_ep_name,
```
timenick: **Test still reaches into the submodule** — now that both functions are exported in `analyze/utils/__init__.py`, this should import from the package level per `CLAUDE.md` § Import Rules: *"Test code: Never reach into internal submodules for symbols exported by `__init__.py`."*

```python
from winml.modelkit.analyze.utils import (
    get_devices_with_rule_data,
    has_rule_data_for_ep,
)
```

🤖 Generated with Claude Code

timenick (commented):
Follow-up to my earlier comment about exporting `has_rule_data_for_ep` / `get_devices_with_rule_data` in `analyze/utils/__init__.py`. The exports were added correctly, but one test still reaches into the submodule.

🤖 Generated with Claude Code

tests/unit/analyze/test_has_rule_data.py
DingmaomaoBJTU: Fixed — now imports from `winml.modelkit.analyze.utils` (package level) instead of reaching into `ep_utils` submodule.

timenick (approved):

PR #393, opened 2026-04-24
e2e_eval: dynamic tasks, priority/group cascade, --priority list

xieofxie:
### `build_registry.py`
- **Fetch pipeline tasks dynamically** from HuggingFace `/api/tasks` (47 tasks) instead of the 21-entry hand-written `NLP_TASKS + CV_TASKS` lists.
- **3-level cascade for `priority` and `group`** when classifying entries:
  1. **Curated** — `models_curated.json` overrides win verbatim (group, priority both honored — previously `priority` was forced to `P0`).
  2. **Existing** — if `(hf_id, task)` was in the previous `models_all.json`, inherit that row's `priority`/`group`. Lets hand-edits to `models_all.json` survive rebuilds.
  3. **New-rule defaults** — `microsoft/`-prefixed → `(microsoft, P1)`; everything else → `(Top200, P2)`. (`P2` is a new tier; `optimum_supported` no longer drives priority.)
- **Phase 1.5 preservation**: existing `(hf_id, task)` rows that aren't re-picked by Phase 1 are kept verbatim except `downloads` + `optimum_supported` are refreshed. Lets the registry stay stable across runs and supports tasks HF's API doesn't expose (e.g. `masked-lm`).
- **Per-task `order` field**: every entry now carries `order: 1..N` ranked by downloads descending so consumers don't need to re-sort.

### `run_eval.py`
- `--priority` now accepts **multiple values** — e.g. `--priority P0 P1` runs the union. Backward-compatible with single value.
- `filter_registry()` widened: `priority: str | Sequence[str] | None`.

### Curated list
- Add 4 ISV models to `models_curated.json`.

### Regenerated `models_all.json` (439 entries, sorted by `(hf_id, task)`)
- Priority: `P0: 22 | P2: 202 | P3: 215`
- Groups: `Top200: 398 | Foundry Toolkit: 16 | microsoft: 15 | Benchmark: 6 | ISV: 4`
- Optimum-supported: `311/439`


scripts/e2e_eval/testsets/models_curated.json
```diff
@@ -20,5 +20,10 @@
   {"hf_id": "facebook/detr-resnet-50", "task": "feature-extraction", "model_type": "detr", "group": "Benchmark", "priority": "P0"},
   {"hf_id": "facebook/sam2.1-hiera-small", "task": "mask-generation", "model_type": "sam2_video", "group": "Benchmark", "priority": "P0"},
   {"hf_id": "Salesforce/blip-image-captioning-base", "task": "image-to-text", "model_type": "blip", "group": "Benchmark", "priority": "P0"},
-  {"hf_id": "ai-forever/Real-ESRGAN", "task": "", "model_type": "unknown", "group": "Benchmark", "priority": "P0"}
+  {"hf_id": "ai-forever/Real-ESRGAN", "task": "", "model_type": "unknown", "group": "Benchmark", "priority": "P0"},
+
+  {"hf_id": "microsoft/trocr-large-printed", "group": "ISV", "priority": "P2"},
```
xieofxie: should we mark them as P1 although not officially requested?

DingmaomaoBJTU: Make sense

scripts/e2e_eval/run_eval.py
```diff
@@ -1020,7 +1020,13 @@ def parse_args() -> argparse.Namespace:
         ),
     )
     parser.add_argument("--task", help="Filter by HF task")
-    parser.add_argument("--priority", choices=["P0", "P1", "P2"], help="Filter by priority")
+    parser.add_argument(
+        "--priority",
+        nargs="+",
+        choices=["P0", "P1", "P2", "P3"],
```
DingmaomaoBJTU, line 1026: Add a default value ["P0", "P1", "P2"] so it won't break e2e pipeline?

DingmaomaoBJTU (approved):

PR #399, opened 2026-04-27
feat(telemetry): Phase 3 — CLI auto-instrumentation, persistent cache, .winml paths

timenick:
## Summary

PR 3 of 4 of the ModelKit telemetry rollout. Three concerns landed in
one PR after follow-up discussion:

### Phase 3 proper — wire telemetry into the CLI

- `ActionGroup` (subclass of `click.Group`) wraps each registered
  subcommand's `invoke` to emit `ModelKitHeartbeat` (once per CLI
  invocation), `ModelKitAction` on completion, and `ModelKitError` on
  unhandled exception. No per-command decoration is required.
- `LazyGroup` in `cli.py` now extends `ActionGroup`, so `winml` is both
  lazy-loading and auto-instrumented without any startup-time regression.
- `cli.py` registers `Telemetry.shutdown` via `ctx.call_on_close` to flush
  the `BatchLogRecordProcessor` before exit, guarded so it never builds
  a fresh singleton on the way out.
- Heartbeat emission lives inside the wrapped subcommand `invoke` so
  `winml --help`, `winml --version`, and `winml <cmd> --help` short-circuit
  inside Click's parameter parsing and never materialize the Telemetry
  singleton.
- No `--disable-telemetry` flag, no env-var override, no `winml telemetry`
  subcommand. Consent is via editing the config file directly.

### Close Phase 2 cache-wiring gap

Phase 2 added base64+json encoding helpers and an `msvcrt` file lock to
`utils.py` but never wired them up. The result: a network blip during a
CLI invocation lost the entire batch, and the design's "no telemetry data
is lost due to transient network issues" guarantee was unmet.

- New `telemetry/_cache.py` with `_PersistentCache` (append-or-drain,
  locked) at `%USERPROFILE%\.winml	elemetry\modelkit.cache`. Honors
  `MODELKIT_TELEMETRY_CACHE_DIR` env-var override (developer-facing only).
- `OneCollectorLogExporter` now drains the cache on first export
  (best-effort, single shot per process), persists the current batch on
  POST failure, and re-caches cached envelopes that fail recovery.
- `Telemetry` clears the cache when init resolves to disabled (consent
  declined / empty iKey / init crash) so a disabled session never
  resends events the user has since opted out of.
- `tests/unit/telemetry/test_cache_integration.py` locks in the
  end-to-end design contract — exists specifically to catch the kind of
  wiring gap that slipped Phase 2 review.

### Unify user-config paths under `.winml`

Phase 2 shipped consent at `%USERPROFILE%\.modelkit\config.json`. With
the cache also living under `%USERPROFILE%\.winml	elemetry\`, both
move under `.winml` for a single inspectable directory. No migration
shim — Phase 2 hasn't shipped to end users.

### Official build pipeline

- `.pipelines/modelkit-official-build.yml`: build sdist BEFORE iKey
  injection (sdist stays clean for PyPI), inject key into
  `constants.py` from the `INSTRUMENTATION_KEY` secret variable, build
  wheel AFTER injection (wheel carries the key), then verify both:
  wheel contains a non-empty key, sdist contains the empty placeholder.

## Notes

- Plan called for `cls=ActionGroup` directly on the main group; aligned
  with the existing `LazyGroup` (added in MVP v2 for `winml --help`
  startup speed) by making `LazyGroup` extend `ActionGroup`.
- Heartbeat originally specified inside `ActionGroup.invoke`; moved
  into the wrapped subcommand `invoke` because `ctx.invoked_subcommand`
  is set by `click.Group.invoke` *after* it's called, and putting
  heartbeat at the subcommand entry point also correctly skips
  `cli build --help` paths.
- Cache wiring is a Phase 2 follow-up that was never assigned to a
  later phase; surfaced during this PR's review and pulled in here
  rather than left as a known gap.


DingmaomaoBJTU:
### Code review

Found 2 issues:

1. `sys.exit()` failures are logged as `success=True`. In `wrapped_invoke`, the `except Exception` block does not catch `SystemExit` (which inherits from `BaseException`, not `Exception`). Commands that call `sys.exit(1)` — such as `analyze.py` which uses `sys.exit(1)` through `sys.exit(4)` — will have their `ModelKitAction` event emitted with `success=True` in the `finally` block, because the `success = False` assignment is never reached.

https://github.com/microsoft/WinML-ModelKit/blob/49b69e538c4c15956dd34aeeef17e0bee8534163/src/winml/modelkit/telemetry/click_group.py#L74-L90

2. `_reset_singleton` fixture is duplicated across two `conftest.py` files with nearly identical implementations. [`tests/CLAUDE.md`](https://github.com/microsoft/WinML-ModelKit/blob/49b69e538c4c15956dd34aeeef17e0bee8534163/tests/CLAUDE.md#L17) says "Never: Duplicate fixtures across multiple `conftest.py` files" and "Put shared fixtures in the narrowest `conftest.py` that covers all consumers." The fixture should live in a single `conftest.py` at a common ancestor (e.g. `tests/conftest.py`).

https://github.com/microsoft/WinML-ModelKit/blob/49b69e538c4c15956dd34aeeef17e0bee8534163/tests/unit/telemetry/conftest.py#L15-L34

https://github.com/microsoft/WinML-ModelKit/blob/49b69e538c4c15956dd34aeeef17e0bee8534163/tests/integration/telemetry/conftest.py#L18-L35

DingmaomaoBJTU (approved):

timenick:
> ### Code review
> Found 2 issues:
> 
> 1. `sys.exit()` failures are logged as `success=True`. In `wrapped_invoke`, the `except Exception` block does not catch `SystemExit` (which inherits from `BaseException`, not `Exception`). Commands that call `sys.exit(1)` — such as `analyze.py` which uses `sys.exit(1)` through `sys.exit(4)` — will have their `ModelKitAction` event emitted with `success=True` in the `finally` block, because the `success = False` assignment is never reached.
> 
> https://github.com/microsoft/WinML-ModelKit/blob/49b69e538c4c15956dd34aeeef17e0bee8534163/src/winml/modelkit/telemetry/click_group.py#L74-L90
> 
> 2. `_reset_singleton` fixture is duplicated across two `conftest.py` files with nearly identical implementations. [`tests/CLAUDE.md`](https://github.com/microsoft/WinML-ModelKit/blob/49b69e538c4c15956dd34aeeef17e0bee8534163/tests/CLAUDE.md#L17) says "Never: Duplicate fixtures across multiple `conftest.py` files" and "Put shared fixtures in the narrowest `conftest.py` that covers all consumers." The fixture should live in a single `conftest.py` at a common ancestor (e.g. `tests/conftest.py`).
> 
> https://github.com/microsoft/WinML-ModelKit/blob/49b69e538c4c15956dd34aeeef17e0bee8534163/tests/unit/telemetry/conftest.py#L15-L34
> 
> https://github.com/microsoft/WinML-ModelKit/blob/49b69e538c4c15956dd34aeeef17e0bee8534163/tests/integration/telemetry/conftest.py#L18-L35
> 
> 

All addressed

PR #403, opened 2026-04-27
fix(pipeline): twine check fails on Metadata-Version 2.4 + non-package files

timenick:
## Summary
- Narrow the `twine check` glob from `$(ob_outputDirectory)\*` to `*.whl` + `*.tar.gz` so rule zips that are staged into the same output directory aren't validated as Python packages.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -3,12 +3,8 @@ name: Lint
 on:
   pull_request:
     branches: [main]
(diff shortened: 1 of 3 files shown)


xieofxie:
or we use a specific version? no need to always upgrade

xieofxie (approved):

xieofxie (approved):

PR #404, opened 2026-04-27
fix: thread ep parameter through to WinMLSession

### LINKED ISSUE #402: Bug: --ep flag is silently dropped, never reaches WinMLSession
## Summary

When `--ep` is passed via CLI (e.g., `winml perf --ep dml`), the value is correctly parsed and forwarded to `WinMLAutoModel.from_pretrained(ep=...)`, but **silently dropped** before reaching `WinMLSession`. This causes `ModelCompiler` to receive session options with no explicit EP registered, falling back to policy-based selection which `ModelCompiler` does not support — resulting in an empty provider type string and a runtime crash.

## Error

```
TransformerMemcpyImpl::IsNodeCompatibleWithProvider node_provider != nullptr was false.
Unable to get provider associated with provider type
```

The provider type is an **empty string** because no EP was explicitly registered.

## Root Cause

The `ep` parameter is dropped at three points in the call chain:

1. **`WinMLAutoModel.from_pretrained()`** ([`src/winml/modelkit/models/auto.py:424-428`](https://github.com/microsoft/WinML-ModelKit/blob/main/src/winml/modelkit/models/auto.py#L424-L428)) — constructs `winml_class(...)` without passing `ep`:
   ```python
   model = winml_class(
       onnx_path=onnx_path,
       config=hf_config,
       device=device,
       # ep is NOT passed here
   )
   ```

2. **`WinMLPreTrainedModel.__init__()`** ([`src/winml/modelkit/models/winml/base.py:62-68`](https://github.com/microsoft/WinML-ModelKit/blob/main/src/winml/modelkit/models/winml/base.py#L62-L68)) — signature does not accept `ep` parameter at all.

3. **`WinMLPreTrainedModel.__init__()`** ([`src/winml/modelkit/models/winml/base.py:85-89`](https://github.com/microsoft/WinML-ModelKit/blob/main/src/winml/modelkit/models/winml/base.py#L85-L89)) — initializes `WinMLSession(...)` without `ep`:
   ```python
   self._session = WinMLSession(
       onnx_path=self._onnx_path,
       device=device,
       session_options=session_options,
       # ep is NOT forwarded here
   )
   ```

`WinMLSession.__init__` already accepts `ep` and handles it correctly — the problem is purely that the upstream callers never pass it through.

## Fix

Thread `ep` through the call chain:

1. Add `ep: str | None = None` to `WinMLPreTrainedModel.__init__()` signature
2. Forward `ep=ep` to `WinMLSession(...)` in `WinMLPreTrainedModel.__init__()`
3. Pass `ep=kwargs.get("ep")` when constructing `winml_class(...)` in `WinMLAutoModel.from_pretrained()`

DingmaomaoBJTU:
## Summary

- The `--ep` flag was silently dropped in the model construction path: `WinMLAutoModel` received it but never forwarded it to `WinMLPreTrainedModel` → `WinMLSession`
- Added `ep` parameter to `WinMLPreTrainedModel.__init__()` and forwarded it to `WinMLSession`
- Passed `ep` at all three `winml_class(...)` construction sites in `WinMLAutoModel` (`from_pretrained`, `from_onnx` skip-build path, `from_onnx` build path)

### Default-path behavior change

Previously, when no `--ep` was set, `_build_session_options` always fell through to `set_provider_selection_policy(...)` (PREFER_NPU/GPU/CPU) and let ORT pick the EP. This PR also changes that path: for non-CPU devices, it now first calls `_find_ep_for_device` and, if a match is found, uses `add_provider_for_devices` directly — bypassing the device policy. This is intentional: it gives us explicit control over which physical device is used (matching what we already do for the `--ep` path) and avoids WinML EP registration issues that the `providers=` string-based path cannot handle.

Fixes #402


src/winml/modelkit/session/session.py
```diff
@@ -287,10 +287,24 @@ def compile(self) -> None:
...
+                providers = self._resolve_providers(target_device)
+            else:
+                # EP configured via add_provider_for_devices — don't override
+                providers = None
             with _suppress_native_output(compile_log):
-                session = ort.InferenceSession(str(model_path), sess_options=sess_options)
+                session = ort.InferenceSession(
+                    str(model_path),
+                    sess_options=sess_options,
+                    providers=providers,
```
xieofxie: do we support it? I don't think so. For winml. EP should be added to session options

DingmaomaoBJTU: You're right. WinML-registered EPs like QNN don't support `providers=` — they must go through `add_provider_for_devices`.

The `providers=` path here was a fallback for built-in EPs (e.g., DML), but `_resolve_providers` didn't distinguish between built-in and WinML-registered EPs. If the QNN device lookup failed in `_build_session_options`, we'd fall through to policy mode and `_resolve_providers` would still return `["QNNExecutionProvider", ...]` — which wouldn't work via `providers=` anyway.

Fixed by removing `_resolve_providers` entirely and always passing only `sess_options`. EP is either configured via `add_provider_for_devices` or left to ORT's device policy in the fallback case.

src/winml/modelkit/session/session.py
```diff
@@ -452,14 +467,46 @@ def _build_session_options(self, device: str) -> ort.SessionOptions:
         return opts
 
     @staticmethod
-    def _find_ep_device(ep_name: str) -> Any:
-        """Find an OrtEpDevice matching the given EP name.
+    def _find_ep_device(ep_name: str, device: str | None = None) -> Any:
```
timenick: **Unused parameter — and it contradicts the only caller.**

The new `device: str | None = None` parameter (and the `DEVICE_TO_DEVICE_TYPE` import / filter logic at lines 482-488) is dead code: the only caller at line 439 calls `self._find_ep_device(target_name)` without passing `device`.

The added comment in the caller actively justifies *not* filtering by device:

> `# Don't filter by device type — trust the user's --ep choice (e.g., QNN reports as NPU in get_ep_devices but can target GPU)`

So the new parameter contradicts its own intended caller. Recommend removing the `device` parameter and the `DEVICE_TO_DEVICE_TYPE` import to keep `_find_ep_device` simple — or pass `device` from the call site and remove the contradictory comment. One or the other.

src/winml/modelkit/session/session.py
```diff
@@ -452,14 +467,46 @@ def _build_session_options(self, device: str) -> ort.SessionOptions:
...
+        """
+        from ..utils.constants import DEVICE_TO_DEVICE_TYPE
+
+        device_type = DEVICE_TO_DEVICE_TYPE.get(device.upper())
+        if device_type is None:
+            return None
         for ep_dev in ort.get_ep_devices():
-            if ep_dev.ep_name == ep_name:
+            if ep_dev.device.type == device_type:
                 return ep_dev
```
timenick, line 509: **Non-deterministic for multi-EP devices.**

This returns the *first* EP whose `device.type` matches. After this PR, with `QNNExecutionProvider` mapping to both `npu` and `gpu` and `DmlExecutionProvider` also being a GPU EP, on a Snapdragon system with `device="gpu"` the choice between QNN (Adreno) and DML depends entirely on the order `ort.get_ep_devices()` returns them — which is registry-order dependent and not part of any documented contract.

The previous code path used `DEVICE_POLICY_MAP` (`PREFER_GPU` etc.) and let ORT do priority-aware selection. The new path drops that.

Suggest either: (a) define an explicit per-device EP priority list so the selection is deterministic, or (b) at minimum, document the registry-order dependency in the docstring so a future reader doesn't assume there's a priority being applied.

src/winml/modelkit/session/session.py
```diff
@@ -416,33 +422,42 @@ def _build_session_options(self, device: str) -> ort.SessionOptions:
...
                 logger.warning(
-                    "EP '%s' (%s) not found in available devices; falling back to policy",
+                    "EP '%s' (%s) not found in available devices",
                     self._ep,
                     target_name,
                 )
 
-        # Policy-based selection (default path)
+        # No explicit EP — discover available EP for this device type
+        if not self._ep and device.lower() != "cpu":
```
timenick, line 454: **Behavior change beyond the PR title.**

The PR title is *"thread ep parameter through to WinMLSession"*, but this branch also changes the **default** path (no `--ep` set): previously the function always fell to `set_provider_selection_policy(...)` (PREFER_NPU/GPU/CPU) and let ORT pick. Now, when `device != "cpu"`, it first calls `_find_ep_for_device` and uses `add_provider_for_devices` directly, bypassing the device policy entirely.

The PR description doesn't mention this. It may be intentional (and fixes the same root cause as the `--ep` thread-through), but it's worth either splitting into a separate commit with its own justification or updating the PR description to call it out — otherwise a future bisect will land on the wrong commit when somebody hits an EP-selection regression.

src/winml/modelkit/session/session.py
```diff
@@ -416,33 +422,42 @@ def _build_session_options(self, device: str) -> ort.SessionOptions:
...
+        EPs for the target device type. Falls back to policy-based
+        selection only as a last resort.
 
         Note: Returns a **fresh** SessionOptions when using explicit EP to
         avoid "already registered" errors from repeated calls.
         """
         # Explicit EP targeting: create fresh opts to avoid double-registration
+        # Don't filter by device type — trust the user's --ep choice
+        # (e.g., QNN reports as NPU in get_ep_devices but can target GPU)
         if self._ep and self._ep != "cpu":
```
timenick, line 438: Nit: docstring says *"When `self._ep` is set, uses `add_provider_for_devices`..."*, but the actual guard is `self._ep and self._ep != "cpu"`. The `"cpu"` exclusion is undocumented. A short note ("`"cpu"` falls through to policy-based selection") would close the gap.

timenick (commented):
Code review — 4 issues found. The core fix (threading `ep` from `WinMLAutoModel` → `WinMLPreTrainedModel` → `WinMLSession`) looks correct. Comments below are on the supporting changes in `session.py`.

src/winml/modelkit/session/session.py
DingmaomaoBJTU: Good catch. Removed the `device` parameter entirely in cba9189 — it was dead code. The caller intentionally skips device-type filtering (comment: *"QNN reports as NPU in get\_ep\_devices but can target GPU"*), so the parameter contradicted its own call site. `_find_ep_device` now just matches on EP name.

src/winml/modelkit/session/session.py
DingmaomaoBJTU: Fair point. Added a note to the `_find_ep_for_device` docstring in cba9189 documenting the registry-order dependency and that callers needing a specific EP should use `self._ep` to bypass discovery entirely. Defining a full priority list would be a larger change and the existing `--ep` flag already gives users deterministic control, so opting for option (b) for now.

src/winml/modelkit/session/session.py
DingmaomaoBJTU: Good call-out. Updated the PR description to explicitly document the default-path change: for non-CPU devices without `--ep`, `_build_session_options` now uses `add_provider_for_devices` when a matching EP is found, bypassing the old `set_provider_selection_policy` path. The motivation is the same as the `--ep` change — the `providers=` string path doesn't work with WinML-registered EPs.

src/winml/modelkit/session/session.py
DingmaomaoBJTU: Fixed in cba9189 — the docstring now states that `"cpu"` is excluded from the `add_provider_for_devices` path and falls through to policy-based selection.

timenick (approved):

PR #405, opened 2026-04-27
fix: DML/GPU build crash — skip compile config for EPs with enable_ep_context=False

### LINKED ISSUE #396: Build fails for DML/GPU: compiled.onnx missing when enable_ep_context=false
## Problem

`winml build` fails with `FileNotFoundError` when targeting GPU/DML because the compile stage does not produce an output file, but the build pipeline unconditionally assumes one exists.

```
✅ Compile                                         14.1s
   📦 Artifact: ..._compiled.onnx  (0.0 MB)
   ❌ Build failed: [Errno 2] No such file or directory: '..._compiled.onnx'
```

## Reproduction

```bash
# Via e2e eval
uv run python scripts/e2e_eval/run_eval.py --hf-model AdamCodd/vit-base-nsfw-detector --device gpu

# Or directly
uv run winml perf -m AdamCodd/vit-base-nsfw-detector --device gpu
```

NPU works fine. CPU works fine. Only GPU (DML) fails.

## Root Cause

The failure chain:

1. `--device gpu` → `winml config` generates `execution_provider: "dml"` with `enable_ep_context: false`
2. In `compiler/stages/compile.py:91`, `_finalize_output` is only called when `enable_ep_context` is true — so for DML, no output file is ever written
3. `compile_result.output_path` stays `None`, `compile_result.success` is `True`
4. In `commands/build.py:964`, `current_path = compiled_path` is set **unconditionally** — even though `compiled_path` does not exist on disk
5. The finalize step tries to copy `current_path` (a non-existent file) → `FileNotFoundError`

## Suggested Fix

**Immediate (build.py):** Only update `current_path` if the compiled file actually exists:

```python
# commands/build.py:964
# Before:
current_path = compiled_path

# After:
if compiled_path.exists():
    current_path = compiled_path
```

This ensures DML compile (which only validates, no EPContext output) falls through gracefully — `current_path` stays at `quantized.onnx`, finalize copies that to `model.onnx`, and perf runs normally.

**Config layer (optional):** When `enable_ep_context=false`, consider setting `compile: null` in the generated config to skip the compile stage entirely, since it produces no artifact.

## Key Files

| File | Line | Issue |
|------|------|-------|
| `commands/build.py` | 964 | `current_path = compiled_path` unconditional |
| `compiler/stages/compile.py` | 91 | `_finalize_output` gated on `enable_ep_context` |
| `compiler/stages/compile.py` | 222-224 | EPContext not found → `add_warning` instead of `add_error` |

DingmaomaoBJTU:
## Summary

Fixes #396

`winml build` crashes with `FileNotFoundError` when targeting **GPU/DML** because the compile stage produces no EPContext file, but the pipeline unconditionally advances `current_path` to the non-existent `compiled_path`.

## Root Cause

**Why DML hits this but CPU doesn't:**

CPU is protected at config generation time: `_DEVICE_TO_PROVIDER["cpu"] = None` → `WinMLCompileConfig.for_provider(None)` returns `None` → `config.compile = None` → compile stage is skipped entirely by the first guard in `_run_compile_stage`.

DML goes through a different path: `_DEVICE_TO_PROVIDER["gpu"] = "dml"` → `WinMLCompileConfig.for_dml()` returns a config with `enable_ep_context=False`. This produced a **non-None** compile config, so the stage was entered. But `enable_ep_context=False` means no EPContext file is ever written. The old code then unconditionally set `current_path = compiled_path` on a file that doesn't exist.

## Fix

Two changes:

### 1. Fix at config level — `WinMLCompileConfig.for_provider()` returns `None` for non-EPContext EPs

Rather than checking `enable_ep_context` inside `build.py`, the fix teaches `for_provider()` that EPs with `enable_ep_context=False` have no offline compile step and should return `None`:

```python
factory = factories.get(provider)
if factory:
    config = factory()
    # EPs that don't produce EPContext have no offline compile step
    if not config.ep_config.enable_ep_context:
        return None
    return config
```

This naturally covers DML, CPU, CUDA, NvTensorRTRTX, VitisAI, and MIGraphX without hardcoding EP names. The logic is derived from the `enable_ep_context` flag already set in each factory. Unknown/custom EPs (generic fallback) are not subject to this rule.

### 2. Raise explicitly when EPContext-producing EP reports success but output is missing

Replaces the silent `compiled_path.exists()` fallback with an explicit `RuntimeError`, so real failures (e.g. QNN) surface immediately:

```python
if not compiled_path.exists():
    raise RuntimeError(
        f"Compile reported success but output not found: {compiled_path}"
    )
```

### 3. `--no-compile` defaults to `True`

Changed from `is_flag=True, default=False` to `--no-compile/--compile` pair with `default=True`, so compile is disabled by default in the CLI.


xieofxie (approved):

xieofxie:
but I do think DML also supports compile, bug?

DingmaomaoBJTU:
Good catch - DML does support compile, but in our current flow it is JIT-style (usually no persisted EPContext file). See: https://github.com/microsoft/WinML-ModelKit/blob/77d553b75846178e94e0e1dca0d5d0ec623cfc7e/src/winml/modelkit/compiler/configs.py#L177 So the bug fixed here is the path assumption: _run_compile_stage switched to compiled_path even when no file was produced. Now it only switches when the file actually exists.

vortex-captain (approved):

tezheng (approved):

PR #410, opened 2026-04-28
fix: multiple analyze/compile bugs and noisy warnings

DingmaomaoBJTU:
## Summary

Fixes several bugs encountered during model analysis and compilation, plus silences noisy warnings that polluted user-facing output.

### 1. Fix `PatternMatcher` crash on ONNX validation failure

`onnx.checker.check_model()` is strict — many valid, functional models fail it (custom ops, EPContext nodes, etc.). This was aborting the entire pattern matching.

**Before:**
```
PatternMatchingValidator: Pattern matching failed - Model validation failed during pattern matching initialization
```

**After:** Validation failure is logged at `debug` level. Pattern matching proceeds normally via graph topology.

---

### 2. Fix `PatternMatcher` crash on missing external data

When models are loaded with `load_external_data=False` (as the analyzer does), `numpy_helper.to_array()` crashes on initializers whose data is in external files.

**Before:**
```
ERROR: Analysis failed: Data of TensorProto (tensor name: convnext.embeddings.patch_embeddings.weight) should be stored in model_opt.onnx.data, but it doesn't exist
```

**After:** Missing external data is gracefully skipped with a debug log. Pattern matching works via graph topology; only value-based constraint checks are skipped for those tensors.

---

### 3. Fix `winml compile -o` option missing

There was no way to specify an output file name for `winml compile`.

**Before:**
```
$ winml compile -m model.onnx --ep qnn -o output.onnx
Error: No such option: -o
```

**After:** Added `--output`/`-o` option. `-o` (file path) takes precedence over `--output-dir` (directory).

---

### 4. Fix compile ignoring user-specified output filename

Even after adding `-o`, `_finalize_output()` always generated the filename as `{stem}_{device}_ctx.onnx`.

**Before:**
```
$ winml compile -m model.onnx --ep qnn -o model_compiled.onnx
Output: model_opt_int8_qnn_ctx.onnx    # -o ignored
```

**After:**
```
$ winml compile -m model.onnx --ep qnn -o model_compiled.onnx
Output: model_compiled.onnx            # -o respected
```

The `.bin` companion file also follows the user-specified stem.

---

### 5. Fix `pattern_to_id` mismatch in `default.json` alternatives

Gelu and Gemm alternatives all had `pattern_to_id` set to the **source** pattern's ID (e.g., `"SUBGRAPH/GeluPattern"`), instead of the **target** pattern's ID. This caused the information engine to fail to look up alternatives.

**Before:**
```
WARNING  Action target pattern 'SUBGRAPH/SingleGeluPattern' not found in alternatives for 'SUBGRAPH/GeluPattern'
WARNING  Action target pattern 'SUBGRAPH/ReshapeGemmReshapePattern' not found in alternatives for 'SUBGRAPH/GemmPattern'
```

**After:** Each alternative now has a distinct `pattern_to_id` matching its target class (e.g., `"SUBGRAPH/SingleGeluPattern"`, `"SUBGRAPH/ReshapeGemmReshapePattern"`). Consistent with how LayerNorm and ReshapeTransposeReshape alternatives were already defined.

---

### 6. Downgrade noisy warnings to debug

Multiple warnings in `doc_constraint_checker.py` and `runtime_checker_query.py` fired for expected/benign conditions (unmapped ops, missing rule files, optional inputs, shape inference failures). These are now `debug` level to keep user output clean.

**Before:**
```
WARNING  Mapping failed for Erf: ONNX operator
WARNING  Negative rule file not found: VitisAIExecutionProvider_NPU_ai.onnx_opset17_negative_rules.json
WARNING  No constraints found for operator 'Erf' with dtype category ...
```

**After:** Clean output. Details available with `--verbose`.

## Tests added

| Test file | Tests | Covers |
|---|---|---|
| `test_pattern_matcher_robustness.py` | 2 | Invalid model non-fatal, missing external data graceful |
| `test_compile_command.py` | 3 | `-o` in help, `-o` passes file path, `-o` precedence over `--output-dir` |
| `test_compiler_stages.py` | 2 | `_finalize_output` respects user file path, `.bin` uses user stem |
| `test_unified_pattern_config.py` | 1 | Gemm alternatives have distinct `pattern_to_id` values |
| `test_make_hashable.py` | 7 | ndarray → tuple, set/dict usage, float replacement, nesting, empty |

### DIFF
--- a/src/winml/modelkit/pattern/base.py
+++ b/src/winml/modelkit/pattern/base.py
@@ -1257,10 +1257,7 @@ class PatternMatcher:
             try:
                 check_onnx_model(self.model)
             except onnx.checker.ValidationError as e:
-                raise InvalidPatternMatcherModelError(
-                    f"Model failed ONNX validation: {e}",
-                    error_tag=_MODEL_TAG_INVALID_PATTERN_MATCHER_MODEL,
-                ) from e
+                logger.debug("Model failed ONNX checker validation (non-fatal): %s", e)
             # Warn about nodes with empty names; they get auto-generated names
             # (node_{idx}) in _build_lookups and are added to all lookup structures,
             # so pattern matching proceeds normally via tensor connectivity.
@@ -1320,7 +1317,15 @@ class PatternMatcher:
         for initializer in self.graph.initializer:
             self.producer_lookup.setdefault(initializer.name, (initializer.name, 0, "Initializer"))
             if initializer.name:
-                self.tensor_values[initializer.name] = numpy_helper.to_array(initializer)
+                try:
+                    self.tensor_values[initializer.name] = numpy_helper.to_array(initializer)
+                except Exception:
+                    # External data not loaded — skip tensor value extraction.
+                    # Pattern matching still works via graph topology; only
+                    # value-based constraint checks will be skipped for this tensor.
+                    logger.debug(
+                        "Skipping tensor value for '%s': external data not loaded", initializer.name
+                    )
 
         for node_idx, node in enumerate(self.graph.node):
             node_name = node.name or f"node_{node_idx}"
@@ -1335,9 +1340,15 @@ class PatternMatcher:
                         tensor_proto = attr.t
                         for output_name in node.output:
                             if output_name:
-                                self.tensor_values[output_name] = numpy_helper.to_array(
-                                    tensor_proto
-                                )
+                                try:
+                                    self.tensor_values[output_name] = numpy_helper.to_array(
+                                        tensor_proto
+                                    )
+                                except Exception:
+                                    logger.debug(
+                                        "Skipping constant '%s': external data not loaded",
+                                        output_name,
+                                    )
                                 # Add shape and type for constant
                                 self.tensor_shapes[output_name] = tuple(tensor_proto.dims)
                                 self.tensor_types[output_name] = self._elem_type_to_str(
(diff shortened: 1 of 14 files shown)


src/winml/modelkit/pattern/base.py
```diff
@@ -1320,7 +1317,15 @@ def _build_lookups(self) -> None:
         for initializer in self.graph.initializer:
             self.producer_lookup.setdefault(initializer.name, (initializer.name, 0, "Initializer"))
             if initializer.name:
-                self.tensor_values[initializer.name] = numpy_helper.to_array(initializer)
```
vortex-captain, line 1323: would it be feasible to load the missing value by partially reading `.onnx.data` (possibly in a lazy manner)? I'm not sure whether a skipped value would impact the matching and rewrite later on. e.g. we could use the newly added util `_try_load_external_initializer_array`.

src/winml/modelkit/pattern/base.py
```diff
@@ -1335,9 +1340,15 @@ def _build_lookups(self) -> None:
...
-                                self.tensor_values[output_name] = numpy_helper.to_array(
-                                    tensor_proto
-                                )
+                                try:
+                                    self.tensor_values[output_name] = numpy_helper.to_array(
+                                        tensor_proto
+                                    )
+                                except Exception:
+                                    logger.debug(
+                                        "Skipping constant '%s': external data not loaded",
```
vortex-captain: Any real world case which is failing at this point? constant values would always be in the `.onnx` file if I understand correctly

fangyangci (approved):

src/winml/modelkit/pattern/base.py
DingmaomaoBJTU: You're right — Constant node data is stored as inline attributes in the graph, not in external data files. This path is unlikely to be triggered in practice. The try/except here is purely defensive.

vortex-captain (approved):

vortex-captain (approved):

PR #411, opened 2026-04-28
Refactor analyze runtime rules: migrate to Parquet and streamline checker tooling

fangyangci:
## Summary
This PR migrates the analyze runtime-rule pipeline from ZIP materialization to a Parquet-native flow and improves runtime checker tooling, result processing, and observability.

## What changed
- Migrated runtime rule querying/loading to Parquet-backed paths and caches.
- Updated rule download/loading scripts and CI pipeline wiring for Parquet artifacts.
- Improved runtime result processing with per-operator output, deduplication, and conflict CSV generation.
- Added conflict-file based case_index filtering for targeted reruns.
- Centralized MODELKIT_TIMING_LOG parsing and structured timing logging into a shared utility.
- Removed deprecated ZIP expansion tooling and obsolete related tests.

## Why
- Simplifies the runtime-rule toolchain by removing ZIP expansion steps.
- Improves maintainability and consistency across analyze components.
- Makes diagnostics and targeted reruns easier in CI and local triage.

## Compatibility notes
- Legacy ZIP expansion utilities were removed.
- Internal workflows should use the Parquet-based rule path.

## Measured impact
- Rule data footprint reduced from ~20 GB to ~2.2 GB (about 89% smaller).
- Rule ZIP artifact reduced from ~250 MB to ~25 MB (about 90% smaller).
- For most model SA runs, end-to-end analyze time improved by around 50%.

Examples:
- winml analyze -m bert-base-multilingual-cased.onnx: 18s -> 8s (about 56% faster).
- winml analyze -m resnet-50.onnx: 35s -> 18s (about 49% faster).

These gains come from the Parquet-native rule pipeline and related runtime-checker streamlining in this PR.

related PR:  https://github.com/gim-home/ModelKitArtifacts/pull/125


DingmaomaoBJTU:
### Code review

Found 1 issue:

1. `KeyError` when building `query_signature` for nodes with infinite-property columns — `_build_table_filter_conditions` skips any column present in `infinite_properties` (does not add it to the returned dict), but the subsequent `query_signature` construction unconditionally accesses every entry in `op_columns` via `parquet_filter_conditions[col]`. For any node whose condition tree includes an infinite-property column, this raises a `KeyError` at runtime.

https://github.com/microsoft/WinML-ModelKit/blob/022eaf1ee2fde350b5294bbd978c5af32cd77d2e/src/winml/modelkit/analyze/core/runtime_checker_query.py#L1795-L1805

The fix is to use `.get(col)` (or exclude infinite-property columns from `op_columns` before building the signature) so the cache key is consistent with what the filter actually covers.

fangyangci:
> Thanks for the review. This is a good catch to call out.

For the current pipeline, all Parquet rule files are generated by the latest 
esult_processor, which already excludes infinite_properties from condition columns. Under that assumption, this KeyError path should not be hit in normal production artifacts.

That said, we kept a small defensive fix to build query_signature from available filter keys only. This keeps behavior safe if mixed/legacy/non-standard artifacts are introduced, and it aligns the cache key with the actual filter columns.

Performance-wise, this change is negligible (same order of work over a tiny column set, no additional I/O, no extra table scan), so there is effectively no runtime impact.

Implemented in commit 8254d00 with regression tests in 	est_runtime_checker_query_helpers.py.

src/winml/modelkit/analyze/core/node_checkers/base.py
```diff
@@ -63,7 +63,7 @@ def check(
         opset_version: int,
         pattern_match: PatternMatchResult,
         alternatives: list[PatternAlternative],
-        **kwargs: dict[str, Any],
+        **kwargs: Any,
```
xieofxie, line 66: any preference for Any not dict?

src/winml/modelkit/analyze/core/runtime_checker_query.py
```diff
@@ -820,7 +633,7 @@ def get_query_conditions_for_node(
                 conditions[column_name] = None
                 conditions[f"{column_name}_is_none"] = True
             else:
-                logger.debug(
+                logger.warning(
```
xieofxie, line 641: will it be too many?

scripts/download_rules.py
```diff
@@ -2,14 +2,9 @@
...
 # --------------------------------------------------------------------------
-"""Download runtime check rule zips from gim-home/ModelKitArtifacts.
+"""Download runtime check rule parquet files from gim-home/ModelKitArtifacts.
 
-For Microsoft internal use only. Requires gh CLI authenticated with an account
-that has access to the gim-home org.
-
-External contributors should instead download rule zips from the latest
-WinML-ModelKit GitHub release; see
-src/winml/modelkit/analyze/rules/runtime_check_rules/README.md.
```
timenick, line 12: Why are we removing these words?

scripts/download_rules.py
```diff
@@ -46,12 +41,7 @@ def _get_clone_url(account: str | None = None) -> str:
             "ERROR: gh account is required.\n"
             "Specify via --account or GH_ACCOUNT env var:\n"
             "  uv run python scripts/download_rules.py --account <account>\n"
-            "  GH_ACCOUNT=<account> uv run python scripts/download_rules.py\n"
-            "\n"
-            "This script is for Microsoft internal use (gim-home org access required).\n"
-            "External contributors: download rule zips from the latest GitHub release:\n"
-            "  gh release download --repo microsoft/WinML-ModelKit --pattern '*.zip' \\\n"
```
timenick, line 53: same here

src/winml/modelkit/analyze/utils/rule_loader.py
```diff
@@ -7,6 +7,7 @@
 import json
```
DingmaomaoBJTU, line 7: **Missing: ZIP-detection warning leads to silent data loss**

When a user updates the code but has not re-downloaded rules in Parquet format, `resolve_rule_parquet_path` falls through all search directories and returns a non-existent path. Downstream code logs a warning and treats every operator as unknown—**the analysis completes without error but produces completely wrong results, with no indication that the data is stale**.

Suggested fix: add a one-time check in `get_runtime_rules_search_dirs` or `resolve_rule_parquet_path` that looks for legacy `*.zip` files in each candidate directory. If ZIPs exist but no Parquet files are found, emit a `logging.warning` (or `print` to stderr) with an actionable message, e.g.:

```
WARNING: Legacy rule ZIP files found in <dir> but no Parquet files detected.
Rule data format changed in this release. Please re-download:
  uv run python scripts/download_rules.py --account <account>
Runtime check results will be incomplete until rules are updated.
```

This turns a silent wrong-answer failure into a loud, actionable one.

src/winml/modelkit/analyze/rules/runtime_check_rules/README.md
```diff
@@ -1,88 +1,53 @@
...
 
-The zip files are **not tracked by git**. They are hosted in a separate repo.
+Files are **not tracked by git** and are expected to come from `ModelKitArtifacts`.
 
 ## Setup
 
-### Option 1: Download from the latest GitHub release (for external contributors)
+### Option 1: Download script (recommended)
 
-Rule zips are published as individual assets on the **latest** [WinML-ModelKit release](https://github.com/microsoft/WinML-ModelKit/releases/latest). No special access required.
```
DingmaomaoBJTU, line 11: **External contributor download path removed without replacement**

The previous README had a dedicated Option 1 for external contributors (no gim-home access):

```bash
gh release download --repo microsoft/WinML-ModelKit --pattern '*.zip' \
  --dir src/winml/modelkit/analyze/rules/runtime_check_rules
```

This PR removes that option entirely. The new README only documents `download_rules.py --account <gim-home_account>` (internal only) and manual copy from the gim-home repo (also internal only).

Two questions need answering before merge:

1. **Will Parquet rule files be published as GitHub release assets?** If yes, restore a public download option (updating the pattern to `*.parquet` or `*.tar.gz`). If no, state explicitly that `winml analyze` runtime checks require internal access, so external contributors know what to expect rather than silently getting incomplete results.

2. **If no public release path**, the "What happens if parquet rules are missing" section should note that external contributors will always see incomplete runtime check results—this is expected, not a setup error.

src/winml/modelkit/analyze/rules/runtime_check_rules/README.md
```diff
@@ -1,88 +1,53 @@
...
-
-### Option 4: Use external rules directory via environment variable
-
-Set `MODELKIT_RULES_DIR` to one or more directories containing runtime rule zip files.
-
-Important: relative paths are resolved from `src/winml/modelkit/analyze/utils/` (the directory of `rule_loader.py`), not from your current terminal working directory.
+### Option 2: Manual copy
 
-- Windows (PowerShell, user-level absolute path): `[Environment]::SetEnvironmentVariable("MODELKIT_RULES_DIR", "C:\*path*\rules_zip", "User")`
-- Windows (PowerShell, user-level repo-relative path): `[Environment]::SetEnvironmentVariable("MODELKIT_RULES_DIR", "..\..\..\..\..\..\ModelKitArtifacts\rules_zip", "User")`
```
DingmaomaoBJTU, line 46: **No migration notice for existing users**

The README describes setup for new users but gives no guidance to existing users (both wheel and repo) who need to migrate from ZIP to Parquet. Without this, the typical upgrade path is:

1. User does `git pull` / `pip install --upgrade`
2. Their `runtime_check_rules/` still contains the old ZIP files (git-untracked, so not removed automatically)
3. `winml analyze` runs, finds no Parquet files, silently treats all ops as unknown
4. User gets wrong results with no indication anything is wrong

Suggest adding a short **Migration** section to the README:

```markdown
## Migrating from ZIP-based rules (pre-v0.x)

Rule data format changed from ZIP to Parquet. Existing ZIP files are not used
and can be deleted. To update:

1. Delete old ZIP files from this directory (optional but recommended to free ~250 MB)
2. Re-download rules in Parquet format:
   uv run python scripts/download_rules.py --account <account>
```

This is a one-time action per environment and prevents the silent wrong-results scenario.

src/winml/modelkit/analyze/rules/runtime_check_rules/README.md
timenick, line 11: An update is needed for external users, not a removal.

src/winml/modelkit/analyze/core/runtime_checker_query.py
fangyangci: Thanks for calling this out.

This log is emitted only when all of the following are true:

the attribute is missing on the node,
the schema marks it as required, and
the schema does not provide a default value.

So it **will not** trigger when:

the attribute is present on the node,
the attribute is optional, or
the attribute is missing but has a schema default (we auto-fill it).

Given that condition set, if it does trigger, it indicates a real model/schema inconsistency (missing required information), which is why we intentionally keep it at warning level rather than debug.

src/winml/modelkit/analyze/core/node_checkers/base.py
fangyangci: Good question. Here we intentionally use Any for kwargs, not dict.

In Python type hints, the annotation on kwargs applies to each keyword argument value, not to the kwargs container itself.
So kwargs: Any means each passed keyword value can be any type, which is what this extension point needs.
If we wrote kwargs: dict[str, Any], it would mean every keyword value must itself be a dict, which is incorrect and too restrictive for current usage (for example ep_name is a string).

So the Any annotation here is intentional and semantically correct.

scripts/download_rules.py
fangyangci: Good catch. We did not intend to remove guidance for external contributors. We restored the wording so both internal and external contributor paths are clear.

scripts/download_rules.py
fangyangci: Agreed, same issue here. We restored the error message guidance as well, so it clearly states the script is for internal use and points external contributors to the public rule guidance.

src/winml/modelkit/analyze/utils/rule_loader.py
fangyangci: Thanks for raising this. In normal usage, there should not be a ZIP migration issue because rule parquets are installed automatically with each wheel installation. We also added fail-fast behavior: if no parquet files are found in any configured search directory, winml analyze exits with code 2 and prompts the user to reinstall or manually provide rule files, instead of returning a seemingly normal result.

src/winml/modelkit/analyze/rules/runtime_check_rules/README.md
fangyangci: Good point. We updated the README for end users to clarify that manual download is typically unnecessary because wheel installations already include parquet rule files. If files are missing, the README now points users to the recovery path.

src/winml/modelkit/analyze/rules/runtime_check_rules/README.md
fangyangci: Agree on reducing upgrade confusion. Since wheel installation already delivers parquet rules, we removed the unnecessary ZIP-to-parquet migration guidance and replaced it with actionable recovery guidance for missing parquet files.

src/winml/modelkit/analyze/rules/runtime_check_rules/README.md
fangyangci: Agreed. We should update guidance for external users, not remove it. Contributor-facing guidance is preserved in the script text, while the README remains user-focused on default install behavior and recovery steps.

.pipelines/modelkit-release-github.yml
```diff
@@ -88,10 +88,10 @@ extends:
               $outDir = "$(ob_outputDirectory)"
               New-Item -ItemType Directory -Path $outDir -Force | Out-Null
               $wheels = Get-ChildItem "$artifactDir" -Filter "*.whl" -Recurse
-              $zips = Get-ChildItem "$artifactDir" -Filter "*.zip" -Recurse
-              Write-Host "Found $($wheels.Count) wheel(s) and $($zips.Count) zip(s)"
+              $sdists = Get-ChildItem "$artifactDir" -Filter "*.tar.gz" -Recurse
```
timenick: we should release parquet file to github release, not tar.gz

src/winml/modelkit/analyze/rules/runtime_check_rules/README.md
```diff
@@ -1,88 +1,58 @@
...
-This directory contains zip files with runtime check rules (negative rules and tables) used by the static analyzer. Each zip corresponds to a specific `{EP}_{Device}_{Domain}_opset{N}` combination.
+This directory contains parquet runtime rule artifacts used by the analyzer.
 
-The zip files are **not tracked by git**. They are hosted in a separate repo.
+Files are **not tracked by git** and are expected to come from `ModelKitArtifacts`.
 
 ## Setup
 
-### Option 1: Download from the latest GitHub release (for external contributors)
+No manual download is required for normal `pip`/wheel installs. Runtime rule
```
timenick: We need to offer users the option to build from source code. In this situation, the user will need the following command to set up.

`gh release download --repo microsoft/WinML-ModelKit --pattern '*.parquet' --dir src/winml/modelkit/analyze/rules/runtime_check_rules`

src/winml/modelkit/analyze/rules/runtime_check_rules/README.md
```diff
@@ -1,88 +1,71 @@
...
+This directory contains parquet runtime rule artifacts used by the analyzer.
 
-The zip files are **not tracked by git**. They are hosted in a separate repo.
+Files are **not tracked by git**. They are bundled in wheel installs and can
+also be fetched from release assets or `ModelKitArtifacts` for source builds.
 
 ## Setup
 
-### Option 1: Download from the latest GitHub release (for external contributors)
+No manual download is required for normal `pip`/wheel installs. Runtime rule
```
timenick: remove this?

fangyangci: I think this explanation is reasonable because it clearly separates source-build users from wheel-install users and prevents unnecessary manual downloads.
I think keeping it also gives external contributors a clear path to get parquet rule assets and reduces setup confusion.

src/winml/modelkit/analyze/core/runtime_checker_query.py
```diff
@@ -2063,510 +1877,338 @@ def _maybe_save_failed_node_result(
...
-                exception_type,
-                node.op_type,
-                node.name,
-                str(e),
-            )
-            return PatternRuntime(
-                pattern_id=get_pattern_id(is_qdq),
+            return _finish(
+                PatternRuntime(
+                pattern_id=pattern_id,
```
timenick: Indentation here is misleading: the arguments to `PatternRuntime(` are at the same column as the opening paren rather than nested inside it. Code is valid Python but visually suggests the args belong to `_finish(...)`. `ruff format` would fix this — please rerun the formatter on this block (`pattern_id=`, `result=`, `alternatives=`, `pattern_match=` should each get one extra level of indent).

🤖 Generated with Claude Code

src/winml/modelkit/analyze/core/runtime_checker_query.py
```diff
@@ -128,404 +117,217 @@ def _sanitize_df(df: pd.DataFrame) -> pd.DataFrame:
...
 
-    def values(self):  # type: ignore[override]
-        self._ensure_loaded()
-        return super().values()
+            _PARQUET_TABLE_GLOBAL_CACHE_COND.wait()
 
-    def items(self):  # type: ignore[override]
-        self._ensure_loaded()
-        return super().items()
+    table_df = _read_and_sanitize_parquet_table(parquet_path)
```
timenick: If anything between the `is_loading=True` write (line 311) and the `is_loading=False` write here raises, the cache entry is stranded as `is_loading=True` and any concurrent waiter blocks forever on `_PARQUET_TABLE_GLOBAL_CACHE_COND.wait()`. `_read_and_sanitize_parquet_table` currently swallows exceptions, but the contract is fragile.

Consider a `try / finally` around the load so the entry is always transitioned (with `table_df=None` on failure) and waiters always get notified:

```python
try:
    table_df = _read_and_sanitize_parquet_table(parquet_path)
finally:
    with _PARQUET_TABLE_GLOBAL_CACHE_COND:
        _PARQUET_TABLE_GLOBAL_CACHE[global_cache_key] = _ParquetTableCacheEntry(
            is_loading=False,
            table_df=table_df if 'table_df' in locals() else None,
        )
        _PARQUET_TABLE_GLOBAL_CACHE_COND.notify_all()
```

🤖 Generated with Claude Code

src/winml/modelkit/analyze/utils/model_utils.py
```diff
@@ -29,11 +34,90 @@
...
+
+    Rule condition columns can contain mixed Python object shapes that pyarrow
+    cannot serialize reliably (for example tuple entries with mixed str/int).
+    This encoder normalizes values to a stable, type-tagged JSON string.
+    """
+    normalized = _normalize_for_parquet_encoding(make_hashable(value))
+    return json.dumps(normalized, sort_keys=True, separators=(",", ":"), ensure_ascii=False)
+
+
+def build_rule_row_index_from_encoded_conditions(encoded_conditions: Mapping[str, str]) -> str:
```
timenick: `build_rule_row_index_from_encoded_conditions` and `build_rule_row_index_from_conditions` (added below) are public new functions but have no callers anywhere in this PR or the rest of the repo. Per CLAUDE.md ("Don't add abstractions beyond what the task requires" / no design for hypothetical future requirements), please drop both unless a follow-up will use them.

🤖 Generated with Claude Code

src/winml/modelkit/analyze/utils/__init__.py
```diff
@@ -18,7 +22,8 @@
     "get_runtime_rules_search_dirs",
     "has_rule_data_for_ep",
     "infer_ihv_from_ep_name",
+    "load_case_indices_from_conflict_file",
     "match_pattern_with_wildcards",
-    "resolve_rule_zip_path",
+    "resolve_rule_parquet_path",
     "validate_json_schema",
 ]
```
timenick, line 31: `encode_rule_condition_value_for_parquet` is a new public symbol in `model_utils.py`, but it isn't re-exported here. Per CLAUDE.md ("Every symbol that external code needs must be exported in the package's `__init__.py`") and the import rules (test code must use absolute imports from package level), please add it to both the import block and `__all__`. The new test in `test_runtime_checker_query_parquet.py` currently has to reach into `utils.model_utils` directly because of this.

🤖 Generated with Claude Code

tests/unit/analyze/runtime_checker/test_ep_checker.py
```diff
@@ -0,0 +1,34 @@
+# -------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+# --------------------------------------------------------------------------
+
+import onnxruntime as ort
+
```
timenick, line 7: Per CLAUDE.md test-import rules, tests should use absolute imports from package level and not reach into internal submodules for non-private symbols. `EPChecker` is not exported from `winml.modelkit.analyze.runtime_checker.__init__.py`. Either export it there and import via `from winml.modelkit.analyze.runtime_checker import EPChecker`, or — if `EPChecker` is intentionally internal — keep the test focused on private helpers only.

🤖 Generated with Claude Code

src/winml/modelkit/analyze/utils/ep_utils.py
```diff
@@ -59,17 +63,23 @@ def infer_ihv_from_ep_name(ep_name: str) -> IHVType:
     if any(kw in ep_lower for kw in amd_keywords):
         return IHVType.AMD
 
+    # NVIDIA / TensorRT RTX
+    nvidia_keywords = ("nvidia", "nvtensorrt", "tensorrt", "rtx")
```
timenick, line 69: Heads-up: `nvidia_keywords = ("nvidia", "nvtensorrt", "tensorrt", "rtx")` is fairly permissive — anything with `"rtx"` or `"tensorrt"` substring will silently classify as NVIDIA. It runs as the last fallback before `raise ValueError`, which is safer than running first, but it does eat any future EP name that happens to contain those tokens. Worth a brief comment that callers should still pass canonical names, or tightening to require a `nvidia` / `nvtensorrt` prefix match.

🤖 Generated with Claude Code

src/winml/modelkit/analyze/core/runtime_checker.py
```diff
@@ -245,11 +267,27 @@ def subgraph_support(
 
         logger.info("Checking subgraph-level runtime support for %d patterns", len(patterns))
```
timenick: Minor: now that pattern-level rules have been removed and `RuntimeCheckerQuery.run_for_subgraph` always falls through to `_run_for_subgraph_per_node`, this log line ("Checking subgraph-level runtime support for %d patterns") is no longer accurate — every pattern is decomposed and checked per-node. Consider rewording to avoid implying there is a separate subgraph-level lookup path.

🤖 Generated with Claude Code

timenick (commented):
Automated review against project standards. 7 items, mostly minor; #1 and #2 are the most worth addressing before merge.

🤖 Generated with Claude Code

src/winml/modelkit/analyze/rules/runtime_check_rules/README.md
fangyangci: removed

timenick (approved):

PR #412, opened 2026-04-28
fix(pattern): skip ONNX validation when external data is unloaded

xieofxie:
## Summary

- Fixes a regression from #379 where models loaded with `load_external_data=False` caused `PatternMatcher` to log `WARNING Model validation failed for pattern matching` and skip all pattern detection entirely.
- Root cause: `check_onnx_model` called `onnx.checker.check_model` on a proto whose tensors had `data_location=EXTERNAL` but no `raw_data`; the checker couldn't find the sidecar `.data` file and raised `ValidationError`.
- Adds `skip_if_unloaded_external_data: bool = False` to `check_onnx_model`. Default is `False` so existing callers are unaffected; `PatternMatcher` opts in explicitly.
- `_build_lookups` skips `numpy_helper.to_array` for unloaded external initializers and records their names in `_external_unloaded_names`.
- `_check_constant_constraints` skips the value comparison for those names so topology-based pattern matching (Gelu, LayerNorm, etc.) still works without tensor data.
- Adds `has_unloaded_external_data` utility exported from `winml.modelkit.onnx`.
- New `tests/unit/onnx/test_onnx_utils.py` and `TestPatternMatchingWithUnloadedExternalData`.

### DIFF
--- a/src/winml/modelkit/onnx/__init__.py
+++ b/src/winml/modelkit/onnx/__init__.py
@@ -20,7 +20,7 @@ from .io import InputTensorSpec, OutputTensorSpec, generate_inputs_from_onnx, ge
 from .metadata import capture_metadata, restore_metadata
 from .persistence import cleanup_onnx, load_onnx, save_onnx
 from .shape import infer_onnx_shapes, infer_shapes
-from .utils import EXTERNAL_DATA_THRESHOLD, check_onnx_model, get_model_size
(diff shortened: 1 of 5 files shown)


xieofxie:
See warnings like ` WARNING  PatternMatchingValidator: Pattern matching failed - Model validation failed during pattern matching initialization  `

xieofxie:
https://github.com/microsoft/WinML-ModelKit/pull/410/changes#diff-da3e2286a2625431d6c74f354f5db3470402b22816d4fc9197268d1c2ff44d23

PR #425, opened 2026-04-30
fix(perf): add --no-compile flag and fix inference wrapper EP

DingmaomaoBJTU:
## Summary

- **EP warning fix** (`auto.py`): `from_pretrained` was passing `ep=resolved_ep` (the compile-time EP derived from `--device`, e.g. `"qnn"`) to the inference wrapper. This caused `WinMLSession` to enter explicit-EP mode and log `WARNING: EP 'qnn' (QNNExecutionProvider) not found` on machines without QNN hardware. Fixed to pass `ep=kwargs.get("ep")` (user's explicit `--ep` only).

- **`--no-compile/--compile` flag** (`perf.py`): adds a toggle matching the `build` command convention — default `True` (skip compile). The override is applied *after* `generate_hf_build_config` returns because the device policy in STEP 4.5 always re-sets `compile`, so an override dict passed earlier would be silently overwritten.

- **Tests**: 8 new cases in `test_perf_cli.py` (flag presence in help, kwarg forwarding for both `from_pretrained` and `from_onnx`, CLI toggle); new file `test_from_pretrained_kwargs.py` with 6 cases covering the EP fix and `no_compile` behaviour.

### DIFF
--- a/src/winml/modelkit/commands/perf.py
+++ b/src/winml/modelkit/commands/perf.py
@@ -76,6 +76,7 @@ class BenchmarkConfig:
     batch_size: int = 1
     output_path: Path | None = None
     no_quantize: bool = False
+    no_compile: bool = True
     rebuild: bool = False
     ignore_cache: bool = False
     monitor: bool = False
@@ -344,6 +345,7 @@ class PerfBenchmark:
             "use_cache": use_cache,
             "force_rebuild": force_rebuild,
             "shape_config": self.config.shape_config,
+            "no_compile": self.config.no_compile,
         }
 
         if is_onnx:
@@ -500,6 +502,7 @@ def _perf_modules(
     warmup: int,
     batch_size: int,
     no_quantize: bool,
+    no_compile: bool,
     output: Path | None,
     verbose: bool,
     console: Console,
@@ -519,6 +522,7 @@ def _perf_modules(
         warmup: Number of warmup iterations.
         batch_size: Batch size for input generation.
         no_quantize: If True, skip quantization and compilation.
+        no_compile: If True, skip EP compilation (no EPContext generation).
         output: Output JSON path, or None for auto-generated path.
         verbose: If True, log exceptions at DEBUG level.
         console: Rich console for output.
@@ -582,6 +586,8 @@ def _perf_modules(
         # Skip quant/compile for faster iteration when requested
         if no_quantize:
             cfg.quant = None
+            cfg.compile = None
+        elif no_compile:
             cfg.compile = None
 
         with tempfile.TemporaryDirectory() as tmpdir:
@@ -1090,6 +1096,12 @@ def _run_onnx_benchmark(
     help="Skip quantization during model build",
 )
 @click.option(
+    "--no-compile/--compile",
+    "no_compile",
+    default=True,
+    help="Skip EP compilation during model build (no EPContext generation). Default: skip.",
+)
+@click.option(
     "--rebuild",
     is_flag=True,
     default=False,
@@ -1151,6 +1163,7 @@ def perf(
     batch_size: int,
     shape_config_path: Path | None,
     no_quantize: bool,
+    no_compile: bool,
     rebuild: bool,
     ignore_cache: bool,
     module_class: str | None,
@@ -1246,6 +1259,7 @@ def perf(
             warmup=warmup,
             batch_size=batch_size,
             no_quantize=no_quantize,
+            no_compile=no_compile,
             output=output,
             verbose=verbose,
             console=console,
@@ -1288,6 +1302,7 @@ def perf(
         batch_size=batch_size,
         output_path=output,
         no_quantize=no_quantize,
+        no_compile=no_compile,
         rebuild=rebuild,
         ignore_cache=ignore_cache,
         monitor=monitor,
--- a/src/winml/modelkit/models/auto.py
+++ b/src/winml/modelkit/models/auto.py
@@ -363,6 +363,11 @@ class WinMLAutoModel:
             ep=kwargs.get("ep"),
         )
 
+        # Apply stage overrides AFTER policy resolution (policy runs inside
+        # generate_hf_build_config and would overwrite compile/quant otherwise).
+        if kwargs.get("no_compile"):
+            build_config.compile = None
+
         resolved_task = build_config.loader.task
         logger.debug("Generated config with task: %s", resolved_task)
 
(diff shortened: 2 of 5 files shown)


tests/unit/models/auto/test_from_pretrained_kwargs.py
```diff
@@ -0,0 +1,190 @@
...
+# - Module-level imports: patch via winml.modelkit.models.auto.<name>
+# - Local (in-function) imports: patch via the source module
+_AUTO = "winml.modelkit.models.auto"
+_GENERATE_HF_BUILD_CONFIG = "winml.modelkit.config.generate_hf_build_config"
+_BUILD_HF_MODEL = "winml.modelkit.build.build_hf_model"
+
+
+def _make_build_config(compile_provider: str | None = "qnn"):
+    """Return a minimal WinMLBuildConfig with controllable compile field."""
+    from winml.modelkit.compiler.configs import EPConfig, WinMLCompileConfig
```
timenick: **Import rule violation.** `EPConfig` and `WinMLCompileConfig` are exported from `winml.modelkit.compiler.__init__` (see `__all__`). Per CLAUDE.md, test code must use absolute imports from package level and must not reach into internal submodules for symbols exported by `__init__.py`.

```python
from winml.modelkit.compiler import EPConfig, WinMLCompileConfig
```

🤖 Generated with Claude Code

tests/unit/commands/test_perf_cli.py
```diff
@@ -328,3 +328,143 @@ def test_onnx_load_model_passes_ep(self, tmp_path: Path) -> None:
...
+                [
+                    "-m",
+                    "microsoft/resnet-50",
+                    "--compile",
+                    "-o",
+                    str(tmp_path / "out.json"),
+                ],
+                obj={},
+            )
+
```
timenick, line 468: **Vacuous assertion.** If `mock_fp` is never called (e.g., the CLI errors out, routing changes, or a guard short-circuits before `from_pretrained`), the `if mock_fp.called:` guard makes the test pass silently without verifying anything. This hides regressions in exactly the path you're trying to test.

Replace with an unconditional assertion:

```python
mock_fp.assert_called_once()
assert mock_fp.call_args.kwargs.get("no_compile") is False
```

🤖 Generated with Claude Code

tests/unit/commands/test_perf_cli.py
```diff
@@ -328,3 +328,143 @@ def test_onnx_load_model_passes_ep(self, tmp_path: Path) -> None:
...
+        ):
+            mock_cls.return_value.run.return_value = MagicMock()
+            runner.invoke(
+                perf,
+                ["-m", str(onnx_file), "-o", str(tmp_path / "out.json")],
+                obj={},
+            )
+
+        # The ONNX path uses _run_onnx_benchmark, not PerfBenchmark, so verify
+        # via BenchmarkConfig default instead.
```
timenick: **Dead-code test / duplicate of `test_no_compile_default_is_true`.** The `runner.invoke` block sets up patches and runs the CLI, but the comment admits the assertion doesn't use any of that — it falls back to `BenchmarkConfig().no_compile is True`, which is exactly what `test_no_compile_default_is_true` already covers a few lines above.

Either drop the `runner.invoke` setup entirely, or rewrite the test to verify the actual CLI default flow on the HF path (mock `WinMLAutoModel.from_pretrained` and assert `kwargs.get("no_compile") is True`), mirroring the structure of `test_cli_compile_flag_sets_no_compile_false`.

🤖 Generated with Claude Code

src/winml/modelkit/commands/perf.py
```diff
@@ -583,6 +587,8 @@ def _perf_modules(
         if no_quantize:
             cfg.quant = None
             cfg.compile = None
+        elif no_compile:
+            cfg.compile = None
```
timenick, line 591: **Untested branch.** This new `elif no_compile: cfg.compile = None` path inside `_perf_modules` (modules-mode benchmark) has no direct test coverage. The new tests in `test_perf_cli.py` exercise `BenchmarkConfig._load_model` and `from_onnx` kwarg forwarding, but none of them hit this branch.

Please add a unit test that exercises `_perf_modules` with `no_quantize=False, no_compile=True` and asserts `cfg.compile` is cleared before the build call.

🤖 Generated with Claude Code

src/winml/modelkit/models/auto.py
```diff
@@ -363,6 +363,11 @@ def from_pretrained(
             ep=kwargs.get("ep"),
         )
 
+        # Apply stage overrides AFTER policy resolution (policy runs inside
+        # generate_hf_build_config and would overwrite compile/quant otherwise).
+        if kwargs.get("no_compile"):
+            build_config.compile = None
```
timenick, line 369: **Note (not blocking this PR).** This workaround is needed because `generate_hf_build_config`'s STEP 4.5 (config/build.py:600-602) unconditionally re-sets `parent_config.compile = WinMLCompileConfig.for_provider(...)` after the override merge. The same root cause means a user passing `config=WinMLBuildConfig(compile=None)` as a Tier-1 override will *also* have their `compile=None` silently overwritten — they'd need this same `kwargs.get("no_compile")` escape hatch to actually disable compilation.

Worth filing a follow-up to make STEP 4.5 respect explicit Tier-1 `compile`/`quant` overrides instead of clobbering them. The current PR fix is fine as a tactical solution.

🤖 Generated with Claude Code

tests/unit/models/auto/test_from_pretrained_kwargs.py
```diff
@@ -0,0 +1,190 @@
...
+            return_value=MagicMock(final_onnx_path="model.onnx"),
+        ),
+        patch(f"{_AUTO}.get_winml_class", return_value=winml_class),
+    ]
+
+
+# ---------------------------------------------------------------------------
+# EP fix tests
+# ---------------------------------------------------------------------------
+
```
timenick, line 80: **Style (optional).** The `with (patches[0], patches[1], ..., patches[6]):` pattern requires every call site to track the list length — adding or removing a patch in `_base_patches` silently breaks the contract. `contextlib.ExitStack` would handle this cleanly:

```python
with contextlib.ExitStack() as stack:
    for p in _base_patches(build_cfg, winml_class=mock_winml_class):
        stack.enter_context(p)
    WinMLAutoModel.from_pretrained("some/model", **extra_kwargs)
```

Not blocking — just easier to maintain.

🤖 Generated with Claude Code

timenick (commented):
Review of fix(perf): add --no-compile flag and fix inference wrapper EP. The two fixes are sound — passing `kwargs.get("ep")` to the wrapper is correct, and clearing `compile` after `generate_hf_build_config` is the right tactical workaround given STEP 4.5 unconditionally overwrites `parent_config.compile`. A few issues in the tests and a follow-up note below.

🤖 Generated with Claude Code

PR #426, opened 2026-04-30
Sam2 add defaut task config.

chinazhangchao:


DingmaomaoBJTU:
### Design feedback: `MODEL_TASK_DEFAULTS` introduces a second parallel registry

`sam.py` now expresses the same fact — "sam2's canonical export task is mask-generation" — in two places:

```python
# 1. MODEL_TASK_DEFAULTS
MODEL_TASK_DEFAULTS = {"sam2": "mask-generation"}

# 2. MODEL_CLASS_MAPPING (already implies this via its primary entry)
MODEL_CLASS_MAPPING = {
    ("sam2", "mask-generation"): SAM2MaskGeneration,  # same information
    ("sam2", "feature-extraction"): Sam2VisionEncoder,
}
```

The value of `MODEL_TASK_DEFAULTS["sam2"]` and the key of `MODEL_CLASS_MAPPING`'s primary entry are the same string. Adding a new model now requires keeping two tables in sync, with no enforcement — it's easy to define `MODEL_TASK_DEFAULTS["newmodel"] = "task-a"` while forgetting to add `("newmodel", "task-a")` to `MODEL_CLASS_MAPPING`, causing a silent wrong-class fallback.

**Alternative: encode the default task as a `None`-sentinel entry in `MODEL_CLASS_MAPPING`**

```python
# sam.py — one table instead of two
MODEL_CLASS_MAPPING = {
    ("sam2", None): "mask-generation",         # None = default task for auto-detection
    ("sam2", "mask-generation"): SAM2MaskGeneration,
    ("sam2", "feature-extraction"): Sam2VisionEncoder,
    ...
}
```

`_get_custom_model_class` reads the sentinel before the class lookup:

```python
def _get_custom_model_class(model_type: str, task: str) -> tuple[str, type | None]:
    default_task = MODEL_CLASS_MAPPING.get((model_type_normalized, None))
    resolved_task = default_task or task
    return resolved_task, MODEL_CLASS_MAPPING.get((model_type_normalized, resolved_task))
```

Benefits:
- Single table; the sentinel entry structurally enforces that a matching class entry must exist
- `task.py` no longer imports `MODEL_TASK_DEFAULTS` — all model-specific data flows through the existing `MODEL_CLASS_MAPPING` path
- Adding a new model still only requires editing `hf/newmodel.py`

The only cost is that `MODEL_CLASS_MAPPING`'s value type becomes `str | type` (sentinel entries hold a task string, others hold a class). Worth considering as a follow-up if this pattern is going to extend to other model families.

src/winml/modelkit/loader/task.py
```diff
@@ -257,6 +257,43 @@ def _detect_task_and_class_from_config(config: PretrainedConfig) -> tuple[str, t
...
+                f"Add the corresponding (model_type, task) entry."
+            )
+        if default_task != task:
+            logger.info(
+                "Overriding auto-detected task %r with model-type default %r for %s",
+                task,
+                default_task,
+                model_type_normalized,
+            )
+            task = default_task
```
vortex-captain: Would it be slightly clearer to `return default_task, default_class` here?

vortex-captain (approved):

PR #437, opened 2026-04-30
feat(eval): add depth-estimation evaluator (#326)

### LINKED ISSUE #326: feat: support depth-estimation model evaluation
## Summary

Implement evaluation support for depth-estimation models (Depth Anything, DepthPro, ZoeDepth).

## Context

Depth estimation predicts per-pixel distance from camera to scene objects. The task is already recognized in our loader (`loader/task.py`, abbrev `depth`) and we have ONNX export configs for three model families (`depth_anything.py`, `depth_pro.py`, `zoedepth.py`), but there is **no evaluator** registered in `_EVALUATOR_REGISTRY` (`eval/evaluate.py`) and no default dataset or metric configured.

Standard depth evaluation metrics are well-established in the literature (AbsRel, RMSE, delta thresholds), making this a good candidate for structured evaluation.

### Representative models

| Model | HuggingFace ID |
|-------|---------------|
| Depth Anything V2 | `depth-anything/Depth-Anything-V2-Small-hf` |
| DepthPro | `apple/DepthPro-hf` |
| ZoeDepth | `Intel/zoedepth-nyu-kitti` |

## Current State

- Existing evaluator base class: `eval/base_evaluator.py`
- Evaluator registry: `eval/evaluate.py:31-40` — no entry for `depth-estimation`
- Dataset mapping: `datasets/__init__.py:38-51` — no entry for `depth-estimation`
- ONNX configs exist: `models/hf/depth_anything.py`, `models/hf/depth_pro.py`, `models/hf/zoedepth.py`
- No depth-specific metric in `eval/metrics/`
- Reference patterns: `WinMLImageSegmentationEvaluator` (closest per-pixel evaluator)

## Desired State

`winml eval --task depth-estimation` works end-to-end, producing standard depth metrics for any supported depth model.

Usage:
```bash
uv run winml eval \
    -m <model_path_or_id> \
    --task depth-estimation \
    --dataset sayakpaul/nyu_depth_v2 \
    --split validation \
    --samples 100
```

## Acceptance Criteria

- [ ] New `WinMLDepthEstimationEvaluator` class inheriting from `WinMLEvaluator`
- [ ] Register `"depth-estimation"` in `_EVALUATOR_REGISTRY`
- [ ] Implement `schema_info()` returning expected dataset columns (image, depth_map)
- [ ] New `DepthMetric` class in `eval/metrics/` computing standard depth metrics:
  - AbsRel (absolute relative error)
  - RMSE (root mean squared error)
  - δ1 threshold accuracy (% of pixels where max(pred/gt, gt/pred) < 1.25)
- [ ] Add default dataset config in `_DEFAULT_DATASETS` (e.g., `sayakpaul/nyu_depth_v2`, split `validation`)
- [ ] Add `"depth-estimation"` to `TASK_DATASET_MAPPING` in `datasets/__init__.py`
- [ ] Handle depth map alignment: pipeline output may need resizing/interpolation to match ground truth resolution
- [ ] Unit tests covering evaluator instantiation, schema validation, and metric computation
- [ ] Integration test verifying end-to-end evaluation on at least one model
- [ ] Tests are passing (`uv run pytest tests/`)

## Technical Notes

- The HF `depth-estimation` pipeline returns `{"predicted_depth": PIL.Image, "depth": PIL.Image}`. The evaluator needs to convert predictions and ground truth to comparable tensors before computing metrics.
- Depth maps may require scale-and-shift alignment (median scaling) since monocular depth models produce relative depth. Decide whether to evaluate relative or absolute depth and document the choice.
- `WinMLImageSegmentationEvaluator` is the closest reference for per-pixel evaluation structure (loading images, preparing predictions vs references, computing per-pixel metrics).
- Consider using `torchmetrics` for depth metrics to stay consistent with existing metric implementations (`MeanIoUMetric`, `MAPMetric`).

## Related Files

- `src/winml/modelkit/eval/base_evaluator.py` — base class to inherit from
- `src/winml/modelkit/eval/evaluate.py:31-40` — evaluator registry to update
- `src/winml/modelkit/eval/image_segmentation_evaluator.py` — closest per-pixel reference evaluator
- `src/winml/modelkit/eval/metrics/mean_iou.py` — reference metric implementation pattern
- `src/winml/modelkit/datasets/__init__.py:38-51` — dataset mapping to update
- `src/winml/modelkit/models/hf/depth_anything.py` — existing Depth Anything ONNX config
- `src/winml/modelkit/models/hf/depth_pro.py` — existing DepthPro ONNX config
- `src/winml/modelkit/models/hf/zoedepth.py` — existing ZoeDepth ONNX config
- `tests/unit/eval/test_image_segmentation_evaluator.py` — test pattern reference

jeon185:
Resolves #326.

Adds `WinMLDepthEstimationEvaluator` and `DepthMetric` (Absolute Relative error, RMSE, delta-1) following the NYU/KITTI evaluation protocol. HuggingFace `evaluate` doesn't ship a depth-estimation evaluator, so the metric loop is implemented manually.

### Background

Depth-estimation models fall into a few groups, and the same input image gives wildly different prediction scales depending on which group the model belongs to.

- Metric-depth models (ZoeDepth, DepthPro) predict depth in meters directly.
- Relative-depth models (Depth-Anything, Marigold) predict depth up to an unknown scale and shift.
- Disparity models (DPT, MiDaS) predict `1 / depth` (inverse depth) up to scale and shift.

Comparing predictions against the NYU ground truth therefore requires (1) optionally inverting disparity into depth and (2) aligning the prediction to the ground truth before computing metrics. This is what AbsRel/RMSE/delta-1 benchmarks in the literature do, and what this PR adds as user-selectable options.

### Options

Two `columns_mapping` keys, both overridable via `--column`, and both visible in `winml eval --schema --task depth-estimation`.

`align` controls how each prediction is rescaled against the ground truth depth map before metrics are computed:

- `affine` (default): per-image least-squares fit of `pred_aligned = s * pred + t`, where `s` is a scalar scale and `t` is a scalar shift, solved on the valid pixels (those passing the depth range mask). Suitable for relative-depth and disparity models.
- `median`: scale-only alignment, `pred_aligned = (median(gt) / median(pred)) * pred`. No shift. Cheaper but less accurate when the model has a non-zero offset.
- `none`: use the prediction as-is. Suitable for metric-depth models that already output meters.

`depth_kind` indicates what the model outputs:

- `depth` (default): prediction is interpreted as depth.
- `disparity`: prediction is interpreted as inverse depth, so it is inverted (`pred := 1 / pred`) before alignment. Needed for DPT/MiDaS-style outputs.

The depth range used for the valid-pixel mask is also overridable: `min_depth` (default 1e-3, NYU convention) and `max_depth` (default 10.0 meters, NYU convention). Only pixels with `min_depth <= gt <= max_depth` contribute to the metrics.

### Default dataset and testset

Default dataset is `sayakpaul/nyu_depth_v2`. All 11 depth-estimation entries from `models_all.json` are added to `models_with_acc.json`, with per-model overrides only where the defaults don't match the model family:

- `Intel/zoedepth-nyu-kitti` and `apple/DepthPro-hf` set `align=none` (metric-depth).
- `Intel/dpt-hybrid-midas` and `Intel/dpt-large` set `depth_kind=disparity`.
- The remaining 7 entries (Depth-Anything family, Marigold, etc.) rely on the defaults (`align=affine`, `depth_kind=depth`).

### Tests

Unit tests cover the new evaluator and the metric, including the affine-fit path and the disparity inversion path. The slow/network integration test runs the full pipeline end-to-end on Depth-Anything V2, ZoeDepth, and DPT.


zhenchaoni (commented):
# PR #437 — Reproducible Failures Found During Review

Branch tested: `feat/eval-depth-estimation` @ `2ca7c6ac` (merged with `main`).
Run from repo root. Both failures block any depth-estimation model from producing a metric end-to-end.

---

## Issue 1 — Quantization calibration breaks the static ONNX input shape

### Repro

```powershell
uv run python scripts/e2e_eval/run_eval.py `
    --registry scripts/e2e_eval/testsets/models_with_acc.json `
    --eval-type both `
    --device npu `
    --hf-model depth-anything/Depth-Anything-V2-Small-hf `
    --timeout 1800
```

### Observed (~78s, before any metric)

```
[1/1] depth-anything/Depth-Anything-V2-Small-hf / depth-estimation  (P2, Top200)
  [FAIL (COMPILE_FAIL)] 78.6s  acc=SKIP/build_failed
...
❌ Quantize
onnxruntime.capi.onnxruntime_pybind11_state.InvalidArgument:
[ONNXRuntimeError] : 2 : INVALID_ARGUMENT : Got invalid dimensions for input: pixel_values
 index: 3 Got: 364 Expected: 518
```

Full trace saved by the run at:
`eval_results/<date>/models/depth-anything__Depth-Anything-V2-Small-hf__depth-estimation/eval_result.json`

### Root cause

The evaluator's `prepare_pipeline` (in `src/winml/modelkit/eval/depth_estimation_evaluator.py`) patches the Depth-Anything image processor:

```python
pipe.image_processor.size = {"height": h, "width": w}
pipe.image_processor.keep_aspect_ratio = False
pipe.image_processor.do_pad = False
```

The **quantization calibration data reader** uses the same dataset + image processor but **without** that patch, so it produces `(364, 518)` pixel values for a static `(1, 3, 518, 518)` ONNX → calibration session crashes.

### Suggested fix

Factor the processor override into a shared helper (e.g. `force_static_image_processor(processor, h, w)`) and call it from both:

1. `WinMLDepthEstimationEvaluator.prepare_pipeline`
2. The depth-estimation calibration data reader path (so the quant pipeline emits the same `(518, 518)` tensors the model was exported for).

---

## Issue 2 — Default depth-estimation dataset cannot be loaded

### Repro (shortest — exact call path the evaluator uses)

```powershell
uv run python -c "from datasets import load_dataset; load_dataset('sayakpaul/nyu_depth_v2', split='validation', streaming=False)"
```

### Or end-to-end via `winml eval` (no NPU, no quant needed)

```powershell
uv run winml export -m depth-anything/Depth-Anything-V2-Small-hf -o temp/da_small.onnx

uv run winml eval `
    -m temp/da_small.onnx `
    --model-id depth-anything/Depth-Anything-V2-Small-hf `
    --task depth-estimation `
    --samples 10 `
    --device cpu
```

### Observed (~16s, inside `prepare_data`)

```
File "src/winml/modelkit/eval/base_evaluator.py", line 103, in prepare_data
    dataset = load_dataset(ds.path, name=ds.name, split=ds.split, streaming=ds.streaming)
...
RuntimeError: Dataset scripts are no longer supported, but found nyu_depth_v2.py
Error: Evaluation failed: Dataset scripts are no longer supported, but found nyu_depth_v2.py
```

### Root cause

`_DEFAULT_DATASETS["depth-estimation"]` in `src/winml/modelkit/eval/evaluate.py` points to `sayakpaul/nyu_depth_v2`, which is a script-based dataset. The current `datasets` library (`trust_remote_code` removed) refuses to execute it. The dataset is only loadable via `revision='refs/convert/parquet'`.

Verified that the parquet revision works and contains float32 depth maps in meters (min 0, max ~6.6 m), so the data itself is fine — only the loading path is broken.

### Suggested fix (pick one)

1. **Add a `revision` field to `DatasetConfig`** and set it to `"refs/convert/parquet"` in both `_DEFAULT_DATASETS["depth-estimation"]` and in all 11 depth-estimation entries in `scripts/e2e_eval/testsets/models_with_acc.json`.
2. **Switch the default dataset** to a parquet-native NYU v2 mirror (e.g. one without a `.py` loader script).
3. **Use the `build_script` pattern** already in use elsewhere (see `scripts/e2e_eval/datasets/build_indonlu_posp.py`) to pre-materialize the dataset to `~/.cache/winml/eval_datasets/`.

Option 1 is the smallest diff and keeps the existing dataset reference intact.

---

## Additional note — test coverage gap

The integration test in `tests/integration/eval/test_depth_estimation.py` would have caught Issue 2 (it calls the same `load_dataset` path against the same `sayakpaul/nyu_depth_v2`), but it's gated by `@pytest.mark.slow` and `@pytest.mark.network` and is not part of default CI. Recommend either:

- Un-gating a fast variant (e.g. 1 sample, 1 small model) for CI, or
- Adding a unit test that asserts the default dataset for each task resolves to something `load_dataset` can actually open with the current `datasets` library.

jeon185:
Thanks for the detailed review. Both points are fixed.
1) Calibration now applies the same static image-processor settings as export/eval, so the shape mismatch is gone.
2) The NYU dataset revision (refs/convert/parquet) is now passed consistently through config, CLI, eval, and calibration paths.
I re-ran the local export + eval pipeline (CPU/NPU) and it passes.

src/winml/modelkit/eval/depth_estimation_evaluator.py
```diff
@@ -0,0 +1,207 @@
...
+            )
+        if isinstance(predicted, torch.Tensor):
+            predicted = predicted.detach().cpu().numpy()
+        return np.asarray(predicted, dtype=np.float32).squeeze()
+
+    def _validate_schema(self, dataset: Dataset) -> None:
+        """Check dataset has required image and depth columns."""
+        col_names = list(dataset.column_names)
+        if self._input_col not in col_names:
+            raise ValueError(
```
zhenchaoni: You can throw `DatasetValidationError` now. We have refined the user message for invalid dataset

src/winml/modelkit/eval/depth_estimation_evaluator.py
```diff
@@ -0,0 +1,207 @@
...
+loop manually against ``DepthMetric`` (AbsRel, RMSE, delta1).
+"""
+
+from __future__ import annotations
+
+import logging
+from typing import TYPE_CHECKING, Any
+
+import numpy as np
+import torch
```
zhenchaoni: In recent testing, we find that import large packages at the top can cause the command to spend a long time to print the first message in console. We are refining it. So I suggest you to import the torch, transformers or .metrics later when you want to use them.

src/winml/modelkit/eval/depth_estimation_evaluator.py
```diff
@@ -0,0 +1,207 @@
...
+    from .config import WinMLEvaluationConfig
+
+logger = logging.getLogger(__name__)
+
+
+class WinMLDepthEstimationEvaluator(WinMLEvaluator):
+    """Evaluator for monocular depth estimation."""
+
+    @classmethod
+    def schema_info(cls) -> list:
```
zhenchaoni: As we have moved the schema_info to the eval_utils, this is not needed.

We put the schema_info to the eval_utils to make `wml eval --schema` can respond within 1 second.

src/winml/modelkit/eval/depth_estimation_evaluator.py
```diff
@@ -0,0 +1,207 @@
...
+            ),
+        ]
+
+    def __init__(
+        self,
+        config: WinMLEvaluationConfig,
+        model: WinMLPreTrainedModel,
+    ) -> None:
+        mapping = config.dataset.columns_mapping
+        self._input_col = mapping.get("input_column", "image")
```
zhenchaoni: You can read the default value from the schema info in the eval_utils. So you don't need to repeat the default value in 2 places.

src/winml/modelkit/eval/base_evaluator.py
```diff
@@ -98,11 +98,11 @@ def prepare_data(self) -> Dataset:
                     name=ds.name,
                     split=ds.split,
                     streaming=ds.streaming,
+                    revision=ds.revision,
                 )
         except Exception as e:
             raise DatasetValidationError(
-                f"Failed to load dataset '{ds.path}' "
-                f"(name={ds.name!r}, split='{ds.split}'): {e}",
+                f"Failed to load dataset '{ds.path}' (name={ds.name!r}, split='{ds.split}'): {e}",
```
zhenchaoni: I find that there are many diffs which only adjust the format of the code. Could you help to undo that part so that the code can be easier to review？

src/winml/modelkit/eval/config.py
```diff
@@ -14,6 +14,17 @@
 from ..utils.constants import EPNameOrAlias
 
 
+@dataclass
+class SchemaColumn:
```
zhenchaoni: This is not needed anymore. We have put the schema column to the eval_utils

src/winml/modelkit/datasets/depth_estimation.py
```diff
@@ -0,0 +1,220 @@
...
+# Using the same dataset family in calibration and evaluation keeps behavior
+# consistent when users rely on the built-in defaults.
+DEFAULT_DEPTH_ESTIMATION_DATASET = "sayakpaul/nyu_depth_v2"
+DEFAULT_DEPTH_ESTIMATION_SPLIT = "validation"
+# Use the parquet mirror revision so the dataset can be loaded reliably
+# through the standard HuggingFace datasets API.
+DEFAULT_DEPTH_ESTIMATION_REVISION = "refs/convert/parquet"
+
+
+class DepthEstimationDataset(ImageDataset):
```
zhenchaoni, line 38: The dataset class is not used for evaluation, it is used for quantization. Is it feasible to use image dataset to quantize?

zhenchaoni (changes requested):

src/winml/modelkit/eval/depth_estimation_evaluator.py
jeon185: Done. Removed the override entirely and registered the depth schema in `TASK_SCHEMAS`, so `prepare_data` now validates the dataset and raises `DatasetValidationError` with the refined message.

src/winml/modelkit/eval/depth_estimation_evaluator.py
jeon185: Done. Moved torch, numpy, tqdm and `.metrics` to lazy imports inside the methods that actually use them. Only `logging` and type-checking imports stay at module level.

src/winml/modelkit/eval/depth_estimation_evaluator.py
jeon185: Removed the `schema_info()` classmethod. The depth-estimation schema is now registered directly in `TASK_SCHEMAS` (in `eval_utils`) alongside the other tasks.

src/winml/modelkit/eval/depth_estimation_evaluator.py
jeon185: Done. `__init__` now reads the per-task defaults via `get_default(task, key)` from `eval_utils`, so the defaults live only in `TASK_SCHEMAS`.

src/winml/modelkit/eval/base_evaluator.py
jeon185: Reverted the format-only changes. The only diff left in `base_evaluator.py` is forwarding `revision=ds.revision` to `load_dataset()`.

src/winml/modelkit/eval/config.py
jeon185: Removed the `SchemaColumn` dataclass. The only remaining change in `config.py` is the new `revision` field on `DatasetConfig` so users can pin a dataset revision from the eval config.

src/winml/modelkit/datasets/depth_estimation.py
jeon185: Done. `DepthEstimationDataset` now reuses `ImageDataset._load_and_sample()` and only keeps what calibration genuinely needs: the NYU default + processor overrides (`size` / `keep_aspect_ratio` / `do_pad`) derived from the model's static ONNX `io_config`. A plain `ImageDataset` won't work because it requires a `ClassLabel` column and does not match the fixed input shape required by depth models like Depth-Anything-V2.

DingmaomaoBJTU (commented):
**Overall: Well-structured feature addition.** This PR adds comprehensive depth-estimation support across the full stack (dataset, evaluator, metrics, model wrapper, ONNX config, eval scripts). The code follows established patterns from other tasks (image-segmentation, object-detection) and includes both unit and integration tests.

**Highlights:**
- Clean metric implementation following Eigen et al. protocol with proper alignment strategies
- Good use of the parquet mirror revision for reliable dataset loading
- Proper lazy-loading patterns in __init__.py files

**A few suggestions:**

1. **_revision attribute safety** (depth_estimation.py): The _load_and_sample call uses getattr(self, '_revision', None). Since _revision is only set inside _get_default_dataset() (called conditionally when _dataset_name is None), consider adding _revision: str | None = None as a class-level attribute to make the contract explicit.

2. **Affine alignment pixel filtering** (metrics/depth.py): In the affine branch, after pred_v = pred_v * scale + shift, pixels <= min_depth are dropped — changing the effective pixel count vs. median/none branches. This is standard (Eigen/MiDaS eval does this), but a brief inline comment would help future maintainers.

3. **Pipeline input type safety** (depth_estimation_evaluator.py): self.pipe(image) in the compute loop assumes the dataset yields PIL images. Consider a defensive check or clear error message if the column yields tensors instead.

jeon185:
Thanks for the review! All three suggestions are addressed:

1. **_revision safety** — added `_revision: str | None = None` as a class-level attribute in `depth_estimation.py`, so the contract is explicit instead of relying on the conditional set in `_get_default_dataset()`.
2. **Affine pixel filtering** — added an inline comment in `metrics/depth.py` noting the post-alignment re-filter follows Eigen/MiDaS eval and that affine may use fewer pixels than the median/none branches.
3. **Pipeline input type safety** — added `_validate_image_input()` that raises a clear, task-specific `TypeError` when the column yields tensors/NumPy instead of PIL images, plus a unit test covering it.

zhenchaoni (approved):

src/winml/modelkit/datasets/depth_estimation.py
jeon185: Yes, this is intentional. The class only supplies representative samples for PTQ calibration, not ground-truth labels, so an image-only dataset is sufficient. _derive_overrides() forces the processor to emit pixel_values matching the exported ONNX input shape (square, keep_aspect_ratio=False, do_pad=False), so the calibration tensors line up exactly with what the quantized model expects. Confirmed working end-to-end on both depth models.

PR #440, opened 2026-05-06
fix(perf): forward --ep/--device through perf --module path and narrow EP discovery

xieofxie:
## Summary

Two related fixes for `--ep` / `--device` not flowing all the way through:

### `commands/perf.py` — `--module` path dropped CLI flags
`_perf_modules` did not accept `device`, `ep`, or `precision`, so `winml perf -m bert-base-uncased --module BertAttention --device npu --ep qnn` ignored those flags and ran on default device/EP. Plumbed them through to `generate_hf_build_config`, `build_hf_model`, and `WinMLSession`.

### `session/session.py` — explicit-EP path ignored device
`_find_ep_device(ep_name)` and `_find_ep_for_device(device)` were two methods, and the explicit-EP branch only matched by EP name. On systems where one EP exposes multiple `OrtEpDevice` entries (e.g. QNN-on-NPU and QNN-on-GPU), `--ep qnn --device gpu` could return QNN-on-NPU. Merged into a single `_find_ep_device(ep_name=None, device=None)` with AND'd filters so both intents are honored.

## Tests

- `tests/unit/commands/test_perf_module.py::TestPerfModuleParameterForwarding` — invokes the `--module` CLI with `--device npu --ep qnn`, mocks the four boundary calls, asserts each receives the expected kwargs. Verified to fail on the unfixed code (`KeyError: 'device'`).
- `tests/unit/session/test_winml_session.py::TestFindEpDevice` — 5 cases covering ep_name only / device only / both-must-match / no-match / `device="auto"` no-op. 4 of 5 fail on the old API.

### DIFF
--- a/src/winml/modelkit/sysinfo/device.py
+++ b/src/winml/modelkit/sysinfo/device.py
@@ -188,6 +188,11 @@ def resolve_device(device: str = "auto") -> tuple[str, list[str]]:
         for dev in available_devices:
             compatible_eps = _DEVICE_EP_MAP.get(dev, [])
             if any(ep in available_eps for ep in compatible_eps):
+                logger.info(
+                    "Auto-selected device '%s' with compatible EPs: %s for auto device",
+                    dev,
+                    sorted(ep for ep in compatible_eps if ep in available_eps),
+                )
                 return dev, available_devices
         # Fallback: CPU is always valid
         return "cpu", available_devices
--- a/src/winml/modelkit/commands/perf.py
+++ b/src/winml/modelkit/commands/perf.py
@@ -504,6 +504,9 @@ def _perf_modules(
     verbose: bool,
     console: Console,
     monitor: bool = False,
+    device: str = "auto",
+    ep: str | None = None,
+    precision: str = "auto",
 ) -> None:
     """Run per-module build and benchmark for matching submodules.
 
@@ -523,13 +526,20 @@ def _perf_modules(
         verbose: If True, log exceptions at DEBUG level.
         console: Rich console for output.
         monitor: If True, wrap each per-module benchmark with HWMonitor.
+        device: Target device policy ("auto", "cpu", "gpu", "npu").
+        ep: Explicit execution provider (e.g., "qnn", "dml"). Overrides
+            device-to-provider mapping when set.
+        precision: Precision mode passed through to the build stage.
     """
     import json as json_mod
     import tempfile
 
     from ..build import build_hf_model
     from ..config import generate_hf_build_config
+    from ..sysinfo import resolve_device
     from .build import _instantiate_parent_model
+
+    resolved_device, _ = resolve_device(device=device)
 
     console.print(f"[dim]Generating module configs for {module_class}...[/dim]")
 
@@ -538,6 +548,9 @@ def _perf_modules(
             model_id=hf_model,
             task=task,
             module=module_class,
+            device=resolved_device,
+            precision=precision,
+            ep=ep,
         )
     except Exception as e:
         console.print(f"[red]Error generating module configs: {e}[/red]")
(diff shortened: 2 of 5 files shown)


DingmaomaoBJTU:
### Code review

Found 2 issues:

1. Explicit-EP path AND-filters by device type, reversing the intentional design from PR #404. QNN is registered as `OrtHardwareDeviceType.NPU` in ORT's EP registry even when targeting GPU workloads. With this change, `--ep qnn --device gpu` finds no match (NPU ≠ GPU) and silently falls back to policy-based selection — precisely the failure mode the PR #404 comment "Don't filter by device type — trust the user's --ep choice (e.g., QNN reports as NPU in get_ep_devices but can target GPU)" was protecting against. The PR's motivating case (`--ep qnn --device cpu`) works because QNN-on-CPU gets a distinct `OrtEpDevice` entry, but that assumption may not hold for GPU.

https://github.com/microsoft/WinML-ModelKit/blob/256c2ade1dcf48160a7b841a3b3e55d2d5dd4481/src/winml/modelkit/session/session.py#L439-L444

2. `device="auto"` bypasses the device filter in `_find_ep_device`, returning the first registered EP unconditionally. The old `_find_ep_for_device` returned `None` when `DEVICE_TO_DEVICE_TYPE.get(device.upper())` was `None` (which includes `"auto"`, since it is not a key in the dict), triggering policy-based EP selection. The new unified function's guard (`if device else None`) only catches `None`/empty string, not the string `"auto"`. When `WinMLSession` is constructed with the default `device="auto"` and `self._ep` is unset, the no-EP discovery branch at line 456 calls `_find_ep_device(device="auto")`, which now returns the first entry in `ort.get_ep_devices()` instead of falling through to policy.

https://github.com/microsoft/WinML-ModelKit/blob/256c2ade1dcf48160a7b841a3b3e55d2d5dd4481/src/winml/modelkit/session/session.py#L499-L508

xieofxie:
> ### Code review
> Found 2 issues:
> 
> 1. Explicit-EP path AND-filters by device type, reversing the intentional design from PR [fix: thread ep parameter through to WinMLSession #404](https://github.com/microsoft/WinML-ModelKit/pull/404). QNN is registered as `OrtHardwareDeviceType.NPU` in ORT's EP registry even when targeting GPU workloads. With this change, `--ep qnn --device gpu` finds no match (NPU ≠ GPU) and silently falls back to policy-based selection — precisely the failure mode the PR [fix: thread ep parameter through to WinMLSession #404](https://github.com/microsoft/WinML-ModelKit/pull/404) comment "Don't filter by device type — trust the user's --ep choice (e.g., QNN reports as NPU in get_ep_devices but can target GPU)" was protecting against. The PR's motivating case (`--ep qnn --device cpu`) works because QNN-on-CPU gets a distinct `OrtEpDevice` entry, but that assumption may not hold for GPU.
> 
> https://github.com/microsoft/WinML-ModelKit/blob/256c2ade1dcf48160a7b841a3b3e55d2d5dd4481/src/winml/modelkit/session/session.py#L439-L444
> 
> 2. `device="auto"` bypasses the device filter in `_find_ep_device`, returning the first registered EP unconditionally. The old `_find_ep_for_device` returned `None` when `DEVICE_TO_DEVICE_TYPE.get(device.upper())` was `None` (which includes `"auto"`, since it is not a key in the dict), triggering policy-based EP selection. The new unified function's guard (`if device else None`) only catches `None`/empty string, not the string `"auto"`. When `WinMLSession` is constructed with the default `device="auto"` and `self._ep` is unset, the no-EP discovery branch at line 456 calls `_find_ep_device(device="auto")`, which now returns the first entry in `ort.get_ep_devices()` instead of falling through to policy.
> 
> https://github.com/microsoft/WinML-ModelKit/blob/256c2ade1dcf48160a7b841a3b3e55d2d5dd4481/src/winml/modelkit/session/session.py#L499-L508
> 
> 

Fix the 2nd one

The 1st one is hallucination. QNN ep will always be registered as two devices (for one dll), so `--ep qnn --device gpu`  will work

src/winml/modelkit/sysinfo/device.py
```diff
@@ -188,6 +188,11 @@ def resolve_device(device: str = "auto") -> tuple[str, list[str]]:
         for dev in available_devices:
             compatible_eps = _DEVICE_EP_MAP.get(dev, [])
             if any(ep in available_eps for ep in compatible_eps):
+                logger.info(
```
xieofxie, line 191: The resolve device is used in perf, so auto handling actually not work in session - but keep as it is.

This is also a bug in perf - resolve device doesn't consider ep, so `--device auto --ep cpu` will resolve to qnn

Fix in following pr

DingmaomaoBJTU (approved):

PR #456, opened 2026-05-07
fix(telemetry): OneCollector iKey wire format and 4xx handling

timenick:
## Summary

Three telemetry fixes that came out of debugging a persistent
`InvalidTenantToken` 401 from OneCollector, plus a small consent-prompt
rebrand.

### 1. Wire-format bug (the actual cause of the 401)

The exporter was embedding the **full** instrumentation key in both the
`x-apikey` HTTP header AND the envelope's `iKey` field. OneCollector
expects the envelope `iKey` in `o:<tenant_token>` form (just the prefix
before the first `-`); sending the full ikey there gets the request
rejected with `Collector-Error: Invalid Tenant Token`. This silently
401-ed every event since telemetry shipped — clients don't notice
because all telemetry errors are swallowed.

`serialization._envelope_ikey()` now derives `o:<tenant_token>` from the
full ikey; the exporter calls it once at `__init__` and embeds the
result in every envelope. `x-apikey` still carries the full ikey.

### 2. `kill-tokens` directive ignored

When OneCollector decides our tenant is misbehaving (e.g. unknown to
the tenant directory), it returns `kill-tokens: o:<token>:all` plus
`kill-duration: N` telling the client to back off for N seconds. We
were ignoring both, retrying every batch and every process startup —
exactly the noisy-client behavior the directive exists to suppress.

The exporter now parses both headers, records the kill window, and
short-circuits exports during it. Envelopes for the kill window are
**dropped**, not cached — caching them just guarantees forever-failure
on the next startup within the window.

### 3. DEBUG log on non-2xx was unactionable

Previously logged only the status code. Now also includes the
`Collector-Error` response header and a 200-byte body excerpt — the
two pieces of info OneCollector uses to communicate the actual
rejection reason. Without these, diagnosing a tenant or wire-format
issue required a live probe round-trip.

### 4. Consent prompt rebrand

`ModelKit` → `WinML CLI` in the user-visible consent text.
`_CONSENT_VERSION` is unchanged because no scope changed.

## Validation

Verified end-to-end against the live OneCollector ingest with a real
ikey: with the wire-format fix, a single Heartbeat envelope returns
`status: 200, body: {"acc":1}`. New unit tests cover the helper,
kill-tokens parsing, kill-window enforcement, log content, and a
regression guard that the full ikey never appears in the envelope
`iKey` field.


DingmaomaoBJTU:
### Code review

Found 1 issue:

1. `exporter` fixture leaks a `requests.Session` — it uses `return` instead of `yield` + `shutdown()`, so the session's connection pool is never closed. The same pattern was flagged in PR #367 and fixed for `test_factory.py`; the new fixture here has the same problem. `OneCollectorLogExporter.__init__` creates a `requests.Session`, and `shutdown()` exists specifically to close it.

https://github.com/microsoft/WinML-ModelKit/blob/dd1a4f981c143751c10d3808e7b8f515b7fec25f/tests/unit/telemetry/library/test_exporter.py#L40-L53

Fix: change to `yield` + teardown:
```python
@pytest.fixture
def exporter(tmp_path):
    cache = _PersistentCache(path=tmp_path / "modelkit.cache")
    exp = OneCollectorLogExporter(
        ikey="abc-def",
        endpoint="https://example.invalid/OneCollector/1.0/",
        cache=cache,
    )
    yield exp
    exp.shutdown()
```

src/winml/modelkit/telemetry/library/exporter.py
```diff
@@ -94,20 +125,29 @@ def export(self, batch: Sequence[ReadableLogRecord]) -> LogRecordExportResult:
...
+        # backend is still down. If the failure is because we just got
+        # killed, drop the cached batch instead of looping it forever.
         if not self._cache_flushed:
             self._cache_flushed = True
             cached = self._cache.drain()
-            if cached and not self._post_envelopes(cached):
+            if cached and not self._post_envelopes(cached) and not self._is_killed():
                 self._cache.append(cached)
 
         if not self._post_envelopes(envelopes):
```
DingmaomaoBJTU, line 143: If the cache-flush POST just triggered a kill (inside `_post_envelopes(cached)`), this second POST still fires in the same call — the `_is_killed()` guard at the top of `export()` only fires on the next invocation. Consider adding an early-return here:
```python
if self._is_killed():
    return LogRecordExportResult.FAILURE
```
before calling `_post_envelopes(envelopes)`.

src/winml/modelkit/telemetry/library/exporter.py
```diff
@@ -47,11 +51,26 @@ def __init__(
         if not endpoint:
             raise ValueError("endpoint must be non-empty")
         self._ikey = ikey
```
DingmaomaoBJTU: `self._ikey` is stored here but never read — `x-apikey` uses the raw `ikey` parameter from the constructor scope, and envelopes use `self._envelope_ikey`. This assignment looks like leftover from the refactor and can be removed.

src/winml/modelkit/telemetry/consent.py
```diff
@@ -48,12 +48,12 @@ def _default_config_path() -> Path | None:
 _CONSENT_VERSION: int = 1
 
 _PROMPT_TEXT = """\
-ModelKit can collect anonymous usage data to help improve the product.
+WinML CLI can collect anonymous usage data to help improve the product.
```
DingmaomaoBJTU, line 51: `_PROMPT_TEXT` now says "WinML CLI" but the module docstring still says "ModelKit telemetry" — worth updating for consistency.

src/winml/modelkit/telemetry/library/exporter.py
timenick: Done in 076ac87 — added an `_is_killed()` early-return between cache flush and new-batch POST. Picked SUCCESS (not FAILURE) for consistency with the top-of-export guard: both signal "tenant told us to stop, don't retry, drop these envelopes". Added a regression test that seeds the cache, mocks a kill response on the cache-flush POST, and asserts the second POST never fires.

🤖 Generated with Claude Code

src/winml/modelkit/telemetry/library/exporter.py
timenick: Done in 076ac87 — removed. Confirmed via grep that `self._ikey` was set once and never read; the full ikey is captured via the constructor parameter for the `x-apikey` header, and envelopes use `self._envelope_ikey`. Leftover from the refactor as you suspected.

🤖 Generated with Claude Code

src/winml/modelkit/telemetry/consent.py
timenick: Done in 076ac87 — updated to "WinML CLI telemetry" to match the user-visible prompt.

🤖 Generated with Claude Code

DingmaomaoBJTU (approved):

PR #458, opened 2026-05-07
fix QDQ node count in analyze

xieofxie:
before:
image
after:
image


DingmaomaoBJTU:
### Code review

Found 1 issue:

1. Import bypasses the package's public API — `QDQ_SUFFIX` is imported directly from the internal submodule `core/runtime_checker_query`, but it is not exported via `analyze/__init__.py`. [CLAUDE.md](https://github.com/microsoft/WinML-ModelKit/blob/main/CLAUDE.md) says: _"Relative imports through `__init__.py`: `from ..onnx import ONNXDomain` (not `from winml.modelkit.onnx.domains import ONNXDomain`)"_ and _"Every symbol that external code needs must be exported in the package's `__init__.py`"_. The fix is to either export `QDQ_SUFFIX` from `src/winml/modelkit/analyze/__init__.py` and import as `from ..analyze import QDQ_SUFFIX`, or define the constant locally in `analyze.py`.

https://github.com/microsoft/WinML-ModelKit/blob/0c643f241fbb636537cc809d2ae03c4c6510085d/src/winml/modelkit/commands/analyze.py#L26-L28

DingmaomaoBJTU (approved):

PR #459, opened 2026-05-07
feat(sysinfo): add ep filter to resolve_device

xieofxie:
## Summary

Add an optional ``ep`` keyword to ``resolve_device``. When set, both ``available_devices`` and ``available_eps`` are narrowed:

- **``available_devices``** is filtered to the device types the EP can target (per ``_EP_DEVICE_MAP``). E.g., ``ep=\"qnn\"`` keeps only ``[\"npu\", \"gpu\"]``; ``ep=\"dml\"`` keeps only ``[\"gpu\"]``.
- **``available_eps``** is filtered to just that EP, intersected with what is actually available on the system.

This lets callers like ``winml perf --ep qnn --device auto`` auto-resolve to a sensible device type without considering EPs that won't actually be used.

The new parameter is keyword-only and defaults to ``None``, so existing callers are unchanged.

## Tests

New ``TestResolveDeviceWithEp`` class covers:
- ``ep=\"qnn\"`` filters available_devices to ``[\"npu\", \"gpu\"]``
- ``ep=\"qnn\"`` on a GPU-only system auto-picks ``gpu``
- ``ep=\"dml\"`` filters to ``[\"gpu\"]`` only
- ``ep=\"vitisai\"`` when not actually available → empty available_eps + warning + auto fallback to cpu
- ``ep=\"tpu\"`` (unknown) raises ``ValueError``
- ``ep=\"QNN\"`` (case-insensitive) works
- ``ep=\"qnn\" + device=\"cpu\"`` returns ``\"cpu\"`` per user override but with the existing no-compatible-EP warning

Full unit suite: 4643 passed, no regressions.


xieofxie:
With this change, `--ep cpu` will work correct. If without, it will be resolved to `qnn + npu` for device auto

xieofxie:
zheng's pr will address this

DingmaomaoBJTU (approved):

DingmaomaoBJTU (approved):

PR #464, opened 2026-05-07
Support ep option for eval command

zhenchaoni:
1. Add --ep option to `winml eval` command. This allows the user to specify the ep to use.
2. Fix the evaluate baseline cache which is mistakenly cleared by an earlier commit.

### DIFF
--- a/scripts/e2e_eval/utils/reporter.py
+++ b/scripts/e2e_eval/utils/reporter.py
@@ -61,11 +61,10 @@ def build_eval_result(
             "error": perf_proc.get("error_summary", ""),
         }
 
-    return {
+    result = {
         "model": entry.hf_id,
         "task": entry.task,
         "device": device,
-        "ep": ep,
         "model_type": entry.model_type,
         "group": entry.group,
         "priority": entry.priority,
@@ -76,6 +75,10 @@ def build_eval_result(
         "perf": perf_section,
         "accuracy": accuracy_result,
     }
+    # Optional fields: only include when explicitly provided by the user.
+    if ep is not None:
+        result["ep"] = ep
+    return result
 
 
 # ---------------------------------------------------------------------------
--- a/scripts/e2e_eval/utils/reporter.py
+++ b/scripts/e2e_eval/utils/reporter.py
@@ -37,12 +37,15 @@ def build_eval_result(
     device: str,
     eval_types_run: list[str],
     accuracy_result: dict | None = None,
+    ep: str | None = None,
 ) -> dict:
     """Build a unified eval_result dict (facts only, no derived fields).
 
     perf_proc is the raw subprocess result from run_model(), or None when
     eval_types_run is ["accuracy"] (accuracy-only mode, perf phase skipped).
     accuracy_result is the accuracy sub-section dict (or None if not run).
+    ep is the explicit execution provider (e.g., "qnn", "dml"), or None when
+    not specified (device-to-provider mapping was used).
     """
     perf_section: dict | None = None
     if perf_proc is not None:
@@ -62,6 +65,7 @@ def build_eval_result(
         "model": entry.hf_id,
         "task": entry.task,
         "device": device,
+        "ep": ep,
         "model_type": entry.model_type,
         "group": entry.group,
         "priority": entry.priority,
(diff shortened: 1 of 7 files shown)


src/winml/modelkit/commands/eval.py
```diff
@@ -67,6 +67,15 @@
     show_default=True,
     help="Device to run on. 'auto' detects the best available device.",
 )
+@click.option(
```
xieofxie: we have a shared ep option @cli_utils.ep_option

DingmaomaoBJTU: Other commands (`run.py`, `serve.py`, `analyze.py`) all use the shared `@cli_utils.ep_option(required=False)` decorator.

The shared version uses `click.Choice(ALL_EP_NAMES, case_sensitive=False)`, which gives you:
- Input validation at CLI parse time — invalid values fail immediately with a helpful error message
- Consistent help text including all aliases (`qnn`, `ov/openvino`, `vitis/vitisai`)

With the current `type=str`, passing `--ep foo` will silently proceed and only fail deep in the eval stack.

Suggest replacing with:
```python
@cli_utils.ep_option(required=False)
```

scripts/e2e_eval/utils/reporter.py
```diff
@@ -62,6 +65,7 @@ def build_eval_result(
         "model": entry.hf_id,
         "task": entry.task,
         "device": device,
+        "ep": ep,
```
DingmaomaoBJTU: `"ep": ep` is always written to the result JSON even when `--ep` is not specified, resulting in `"ep": null` in every eval output.

Other optional fields (`task`, `model_path`) in `WinMLEvaluationConfig.to_dict()` are only written when non-`None`. Suggest being consistent:
```python
if ep is not None:
    result["ep"] = ep
```

scripts/e2e_eval/utils/reporter.py
```diff
@@ -340,6 +344,8 @@ def generate_html_report(
             {
                 "hf_id": hf_id,
                 "task": task,
+                "device": r.get("device", ""),
```
DingmaomaoBJTU, line 350: `"device"`, `"ep"`, and `"metric"` (~line 369) are added to the row dicts here, but the JavaScript `renderTable` function does not render any of them — they are absent from both the table header and the row template.

If the intent is to display these values in the HTML report, the JS template needs to be updated too. If they are only for downstream JSON consumers, a comment explaining the intent would help.

tests/unit/eval/test_eval.py
```diff
@@ -836,6 +836,7 @@ def test_load_model_from_pretrained(self):
             "test/model",
             task="image-classification",
             device="cpu",
+            ep=None,
```
DingmaomaoBJTU, line 1033: This only adds `ep=None` to an existing test call — no new test coverage for the `--ep` feature itself. Consider adding tests for:
- `--ep` being parsed and passed through the eval pipeline
- `ep` being read from `build_cfg.compile.ep_config.provider` when not provided on the CLI
- `ep` appearing correctly in the result JSON

scripts/e2e_eval/utils/reporter.py
xieofxie: should we use english?

src/winml/modelkit/commands/eval.py
zhenchaoni: fixed

zhenchaoni: fixed

scripts/e2e_eval/utils/reporter.py
zhenchaoni: fixed

scripts/e2e_eval/utils/reporter.py
zhenchaoni: Recording the data for future report refinement, since we will have per ep data. Similar case for recording the metric values. Adding new field in the data has almost no side effect for existing report. As the code is straightforward, I decide not to add an additional comment here.

tests/unit/eval/test_eval.py
zhenchaoni: fixed, add more unit test

DingmaomaoBJTU (approved):

xieofxie (approved):

PR #467, opened 2026-05-07
fix(perf): poll the right adapter under --device gpu/--monitor (#445)

### LINKED ISSUE #445: winml perf --monitor ignores --device gpu: GPU utilization never shown
## Description

`winml perf --monitor --device gpu` shows only NPU and CPU utilization — GPU utilization is never displayed. The `--monitor` flag is hardcoded to discover and poll the NPU (Compute engine) regardless of the `--device` value passed by the user.

## Steps to Reproduce

```bash
winml perf -m ProsusAI/finbert --device gpu --monitor --iterations 100
```

## Expected Behavior

The live hardware monitor shows GPU utilization (and GPU memory) when `--device gpu` is specified. The monitored device tracks the device being benchmarked.

## Actual Behavior

GPU utilization is absent from the live chart. Only NPU and CPU are shown. The `--monitor` help text even documents this explicitly: *"Show live **NPU** utilization chart during benchmark"* — GPU was never wired up.

## Root Cause

The issue spans three layers:

**1. `PdhPoller.start()`** ([_pdh.py:352](src/winml/modelkit/session/monitor/_pdh.py)) always calls `discover_npu_luid()` and passes `engine_type="Compute"` to `build_npu_query`. No GPU LUID is ever discovered or queried.

**2. `HWMonitor`** ([hw_monitor.py](src/winml/modelkit/session/monitor/hw_monitor.py)) exposes only NPU-labelled properties (`mean_utilization_pct`, `peak_utilization_pct`, etc.) and `to_dict()` only emits a `"npu"` key — there is no `"gpu"` equivalent.

**3. `_run_benchmark_monitored()`** ([perf.py:384](src/winml/modelkit/commands/perf.py)) instantiates `HWMonitor` with no device hint — the `--device` flag value is never passed to the monitor.

**The fix infrastructure already exists:** `build_adapter_query()` ([_pdh.py:224](src/winml/modelkit/session/monitor/_pdh.py)) is fully generic and accepts any `engine_type` (e.g. `"3D"` for GPU). It just needs to be wired up for GPU.

## Proposed Fix (outline)

1. Add `discover_gpu_luid()` to `sysinfo/pdh_adapters.py` (parallel to `discover_npu_luid()`) that selects the discrete GPU adapter using the `3D` engine type.
2. Update `PdhPoller` to accept a `device` parameter (`"npu"` | `"gpu"` | `"auto"`) and call the appropriate discover + `build_adapter_query(engine_type=...)` path.
3. Update `HWMonitor` to expose device-neutral property names (or parallel `gpu_*` properties) and emit the correct key in `to_dict()`.
4. Pass `--device` into `HWMonitor` from `_run_benchmark_monitored()` in `perf.py`.
5. Fix the `--monitor` help text: *"Show live hardware utilization chart for the benchmarked device (NPU, GPU, or CPU)"*.

## Environment

- OS: Windows 11
- Device under test: GPU (DirectML / DML EP)
- ModelKit version: latest main

xieofxie:
Fixes #445 — `winml perf --monitor --device gpu` now shows GPU utilization instead of silently falling back to NPU.

## Summary

- **Adapter routing.** `PdhPoller` / `HWMonitor` now accept `device=` (`npu` | `gpu` | `cpu` | `auto`) and `ep_name=`. Per-device queries go through `build_npu_query` (Compute engine) or `build_gpu_query` (3D engine).
- **ORT-first LUID resolution.** Added `resolve_adapter_luid(device_kind, ep_name=None)` in `sysinfo/pdh_adapters.py`. It walks `ort.get_ep_devices()`, filters by EP name + device type, reads `OrtHardwareDevice.metadata["LUID"]`, and converts the decimal LUID to the PDH `0xHHHHHHHH_0xHHHHHHHH` form via `(v >> 32, v & 0xFFFFFFFF)`. ORT-resolved LUIDs are sanity-checked against `enumerate_adapters()` — if PDH doesn't enumerate the LUID (some drivers register the adapter only for autoEP), we skip it and fall through to the PDH-fingerprint helpers (`discover_npu_luid`, `discover_gpu_luid`) instead of raising later in `build_adapter_query`. Same fallback applies when ORT isn't installed, when `get_ep_devices()` raises, when no matching ep_device exists, when LUID metadata is missing, or when an individual ep_device entry is malformed (each is wrapped in `try/except` so one bad entry can't break resolution).
- **Graceful build-query degradation.** `PdhPoller.start()` wraps `build_npu_query` / `build_gpu_query` in a local `try/except ValueError`. If a resolved LUID still proves unusable, it clears `device_kind` / `adapter_luid` and degrades to CPU/RAM-only — CPU/RAM counters still register and the benchmark continues.
- **`perf.py` plumbing.** `_run_benchmark_monitored` (HF), `_run_onnx_benchmark` (ONNX direct), and `_perf_modules` (per-module) all pass the resolved device + normalized EP name into `HWMonitor`. `--monitor` help text updated to *"Show live hardware utilization chart for the benchmarked device (NPU, GPU, or CPU)"*.
- **`HWMonitor.to_dict()` schema.** Always emits an `npu` block (zeros when monitoring GPU — back-compat with existing consumers/tests). When `device_kind == "gpu"` a parallel `gpu` block is added. New top-level keys `device_kind` and `adapter_luid`.
- **Live chart labels follow what's actually polled.** Centralised `adapter_label(device_kind)` helper in `hw_monitor.py`. `LiveMonitorDisplay` accepts `device_kind=` (threaded through from `_run_monitored_loop` as `hw.device_kind`); `HWLiveDisplay` derives the label dynamically from `self._hw.device_kind`. So `auto` resolving to GPU now labels "GPU", and CPU-only mode no longer falsely says "NPU".
- **CPU-only display drops the adapter row entirely.** When no adapter is polled (`--device cpu` or `auto` resolving to nothing), `LiveMonitorDisplay` / `HWLiveDisplay` skip the green plot, drop the swatch from the chart legend, and collapse the status row to `CPU: ... | Sys Mem: ...` (no adapter cell, no Device Mem cell). The post-benchmark `display_console_report` in `perf.py` likewise prints CPU/RAM only — no zeroed `NPU: 0.0% avg, 0.0% peak`.

## Verification

- Targeted suite: `tests/unit/session/test_ep_monitor.py` (84 passed, +15 new) plus `test_perf_cli.py` and `test_perf_module.py` (26 passed).
- Full unit suite green: 4665 passed, 73 skipped, 2 xfailed.
- `ruff check` clean on every modified file.
- Smoke-checked on a Snapdragon X laptop after `winml.register_execution_providers(ort=True)`:
  ```
  GPU (no ep_name)  -> 0x00000000_0x000121CF   # Adreno X1-85 via DML and QNN
  GPU (ep=DML)      -> 0x00000000_0x000121CF
  GPU (ep=QNN)      -> 0x00000000_0x000121CF
  NPU (ep=QNN)      -> 0x00000000_0x00012C89   # authoritative, not just fingerprinted
  NPU (no ep_name)  -> 0x00000000_0x00012C89
  ```


src/winml/modelkit/commands/_live_chart.py
```diff
@@ -174,13 +201,16 @@ def _render_chart(
         from rich.console import Group
         from rich.text import Text
 
-        # Rich-colored title line with legend swatches
-        if has_cpu:
```
xieofxie, line 178: I think it should always have cpu?

src/winml/modelkit/session/monitor/hw_monitor.py
```diff
@@ -140,23 +191,23 @@ def to_dict(self) -> dict[str, Any]:
                 "used_mb": round(self._pdh.ram_used_mb, 2),
                 "peak_mb": round(self._pdh.peak_ram_used_mb, 2),
             },
-            "npu": {
-                "mean_pct": round(self._pdh.mean_utilization_pct, 2),
-                "peak_pct": round(self._pdh.peak_utilization_pct, 2),
-                "sample_count": self._pdh.utilization_sample_count,
-            },
+            # "npu" key is always present; populated when device_kind=="npu".
+            "npu": adapter_block if kind == "npu" else zero_block,
```
timenick: The `npu` block being unconditionally present with zeros when `device_kind == "gpu"` is a back-compat shim that's likely to bite downstream consumers. Anything that aggregates `result["npu"]["mean_pct"]` for monitoring will silently report 0 on a GPU-monitored run — looking like "NPU is idle" rather than "NPU was not the monitored device". `device_kind` disambiguates only if every consumer remembers to check it; the new `display_console_report` path already shows that's easy to forget.

Suggest either dropping the zero `npu` block when `kind == "gpu"` (with a one-release deprecation notice), or at minimum adding a `# DEPRECATED:` marker on this line so the shim has a clear sunset path. The current docstring framing ("always present for backward compatibility") reads as a permanent contract.

🤖 Generated with Claude Code

src/winml/modelkit/commands/_live_chart.py
```diff
@@ -38,11 +38,26 @@ def __init__(
         chart_width: int = 80,
         chart_height: int = 15,
         poll_interval_ms: int = 100,
+        device_kind: str | None = None,
```
timenick, line 43: Function-local imports usually exist to break a circular dependency, but this one doesn't appear to — `_live_chart.py` and `hw_monitor.py` aren't part of the same import cycle. Hoist to the module-level imports at the top of the file, or add a comment explaining why deferral is intentional (e.g., heavy import cost).

🤖 Generated with Claude Code

src/winml/modelkit/session/monitor/_pdh.py
```diff
@@ -449,14 +516,57 @@ def _poll_loop(self) -> None:
...
+    def device_kind(self) -> str | None:
+        """Resolved adapter kind: ``"npu"``, ``"gpu"``, or None when only CPU/RAM."""
+        return self._device_kind
+
     @property
     def npu_luid(self) -> str | None:
-        """NPU LUID string, or None if not discovered."""
-        return self._npu_luid
+        """NPU LUID string, or None if not monitoring an NPU."""
+        return self._adapter_luid if self._device_kind == "npu" else None
```
timenick: This silently changes the semantics of `npu_luid` — it used to return "the LUID we resolved" but now returns `None` when monitoring GPU even though `adapter_luid` has a value. Existing callers that read `poller.npu_luid` (or `to_dict()["npu_luid"]`, which threads through this same property) to mean "the LUID being polled" will now get `None` on `--device gpu` runs.

If the intent is to deprecate this in favor of `adapter_luid`, mark it explicitly (`# DEPRECATED: use adapter_luid`) so consumers have a migration signal. Otherwise consider keeping the prior semantics ("the LUID we resolved, regardless of kind") since the property name doesn't carry strong "NPU only" connotations in caller code today.

🤖 Generated with Claude Code

src/winml/modelkit/session/monitor/hw_monitor.py
```diff
@@ -127,9 +162,25 @@ def is_available(cls) -> bool:
         return sys.platform == "win32"
 
     def to_dict(self) -> dict[str, Any]:
```
xieofxie, line 164: no such `HWMonitor still only exposes is_npu_available() (inherited via _pdh)` thing?

src/winml/modelkit/session/monitor/hw_monitor.py
xieofxie: Sure, just cleaned up

src/winml/modelkit/session/monitor/_pdh.py
xieofxie: just removed

src/winml/modelkit/commands/_live_chart.py
xieofxie: moved

timenick (approved):

PR #475, opened 2026-05-08
feat(catalog): rename hub → catalog, add size/type/EP columns

### LINKED ISSUE #413: Bug:Rename WinML Hub to WinML Catalog + fix model list color styling to map to type + Model size
## Summary

Two related UX improvements for WinML Hub:

1. **Rename WinML Hub → WinML Catalog** — update all references in UI labels, page titles, navigation, and documentation strings.

2. **Model list color styling: map full item to type, not just highlighted portion** — currently color styling in the model list only applies to the highlighted/selected portion. It should apply based on model typing across the full item display, so the color mapping reflects the model type consistently regardless of selection state.

## Motivation

- "Catalog" better reflects the browsing/discovery nature of the feature and aligns with naming conventions used elsewhere in the ecosystem.
- The partial color styling creates visual inconsistency — users expect type-based color coding to be a persistent property of the item, not a state-dependent decoration.

## Acceptance Criteria

- [ ] All UI surfaces that say "Hub" updated to "Catalog" (nav, titles, breadcrumbs, strings)
- [ ] Model list items use type-based color styling for the full item, not only the highlighted state
- [ ] No regression in model list filtering or selection behavior

DingmaomaoBJTU:
## Summary

Closes #413

- **Rename `winml hub` → `winml catalog`** — updates CLI command, module (`hub.py` → `catalog.py`), test file, README, and serve UI (`index.html`)
- **Size column** — shows parameter count (e.g. `110M`, `1.3B`) from `num_parameters` field in catalog JSON
- **Model Type column with type-based color coding** — deterministic palette via character-sum hash; same architecture always gets the same color
- **`--ep` / `--device` filter flags** — filter models by execution provider or device; when exactly one is given, a contextual 5th column appears:
  - `--ep ov` → `Devices` column shows `CPU / GPU / NPU` (same for all rows)
  - `--device NPU` → `EPs` column shows per-model strings like `OV / QNN / VitisAI`
- **Hint line after table** — `Verified picks from Hugging Face. Try any model — run winml inspect <model-id> to check compatibility first.`
- **Updated `hub_models.json`** — 57 models with `supported_eps` list schema; DML/MLAS are always-on (not listed)
- **66 unit tests**, all passing

## Screenshots

**`winml catalog`**

screenshot_default

**`winml catalog --ep ov`**

screenshot_ep_ov

**`winml catalog --device NPU`**

screenshot_device_npu


tests/unit/commands/test_catalog.py
```diff
@@ -0,0 +1,581 @@
...
+    assert len(colors) > 1
+
+
+# ---------------------------------------------------------------------------
+# CLI integration tests via CliRunner
+# ---------------------------------------------------------------------------
+
+
+def test_hub_default_shows_table(runner, patched_catalog):
+    result = runner.invoke(catalog, ["--output", "/dev/null"])
```
timenick: `/dev/null` doesn't exist on Windows — this writes a file literally named `dev/null` in the current working directory. Per CLAUDE.md: *"Tests should not use `/dev/null` (Windows-incompatible), use `tmp_path` instead"*. Other tests in this file already do this (e.g. `test_hub_filter_model_type` uses `tmp_path / "out.json"`).

🤖 Generated with Claude Code

tests/unit/commands/test_catalog.py
```diff
@@ -0,0 +1,581 @@
...
+    out = tmp_path / "out.json"
+    result = runner.invoke(catalog, ["--task", "fill-mask", "--output", str(out)])
+    assert result.exit_code == 0
+    data = json.loads(out.read_text())
+    assert len(data) == 1
+    assert data[0]["model_id"] == "google-bert/bert-base-uncased"
+
+
+def test_hub_no_match_shows_message(runner, patched_catalog):
+    result = runner.invoke(catalog, ["--model-type", "llama", "--output", "/dev/null"])
```
timenick: Same `/dev/null` issue as above — please use `tmp_path / "out.json"` for cross-platform compatibility.

🤖 Generated with Claude Code

tests/unit/commands/test_catalog.py
```diff
@@ -0,0 +1,581 @@
...
+    colors = {_type_color(t) for t in ["bert", "vit", "swin", "clip", "detr", "roberta"]}
+    assert len(colors) > 1
+
+
+# ---------------------------------------------------------------------------
+# CLI integration tests via CliRunner
+# ---------------------------------------------------------------------------
+
+
+def test_hub_default_shows_table(runner, patched_catalog):
```
timenick: The hub→catalog rename left nine test functions in this file still using the `test_hub_*` prefix:

- `test_hub_default_shows_table`
- `test_hub_table_shows_size_column`
- `test_hub_saves_json_file`
- `test_hub_filter_model_type`
- `test_hub_filter_task`
- `test_hub_no_match_shows_message`
- `test_hub_saves_path_shown_in_output`
- `test_hub_catalog_load_error`
- `test_hub_real_catalog_loads`

Please rename to `test_catalog_*` for consistency with the command rename.

🤖 Generated with Claude Code

src/winml/modelkit/commands/catalog.py
```diff
@@ -0,0 +1,457 @@
...
+        models: Model list to filter.
+        device: Device string from CLI (case-insensitive), or ``None`` to skip.
+
+    Returns:
+        Filtered model list.
+    """
+    if device is None:
+        return models
+    cat_keys = _DEVICE_TO_CAT_KEYS.get(device.upper(), frozenset())
+    if not cat_keys:
```
timenick: This conflates "device unknown to the mapping" with "device covered by an always-on EP" — both fall through to `return models`. So `_filter_by_device(models, "FPGA")` returns all models instead of an empty list. Currently masked by Click's `Choice(SUPPORTED_DEVICES)` validation at the CLI layer, but the function is buggy in isolation and inconsistent with `_filter_by_ep`, which correctly returns `[]` for unknown EPs.

Suggested fix:
```python
if device.upper() not in _DEVICE_TO_CAT_KEYS:
    return []
cat_keys = _DEVICE_TO_CAT_KEYS[device.upper()]
if not cat_keys:
    return models  # always-on EP covers this device
return [m for m in models if any(k in (m.get("supported_eps") or []) for k in cat_keys)]
```

🤖 Generated with Claude Code

tests/unit/commands/test_catalog.py
```diff
@@ -0,0 +1,581 @@
...
+
+def test_make_ep_col_fn_for_ep_cpu_devices():
+    _, fn = _make_ep_col_fn_for_ep("CPUExecutionProvider")
+    assert fn({}) == "CPU"
+
+
+def test_make_ep_col_fn_for_ep_same_value_for_all_models():
+    """--ep mode: every model row shows the same device string."""
+    _, fn = _make_ep_col_fn_for_ep("QNNExecutionProvider")
+    assert fn(MINIMAL_CATALOG["models"][0]) == fn(MINIMAL_CATALOG["models"][1])
```
timenick: This assertion is tautological — the cell function returned by `_make_ep_col_fn_for_ep` ignores its argument by design (`def cell_fn(m): return devices_str`). Asserting `fn(m0) == fn(m1)` only confirms the function doesn't crash with two different inputs. Consider asserting against the expected device string directly (e.g. `assert fn(MINIMAL_CATALOG["models"][0]) == "GPU / NPU"`) — though that's already covered by `test_make_ep_col_fn_for_ep_qnn_devices`, so this test may simply be removable.

🤖 Generated with Claude Code

timenick (commented):
Review of PR #475 (hub → catalog rename + new columns/filters).

Nice work overall — the data-schema simplification and per-device contextual column logic are clean. A few items to address before merge:

- **Two `/dev/null` usages in tests** violate CLAUDE.md's Windows-compatibility rule (inline comments below).
- **Nine test functions still carry the `test_hub_*` prefix** after the rename (inline comment below).
- **`_filter_by_device` returns all models for unknown device strings** — masked by Click's `Choice` validation at the CLI today, but buggy in isolation and inconsistent with `_filter_by_ep` (inline comment below).
- **Tautological test** in `test_make_ep_col_fn_for_ep_same_value_for_all_models` (inline comment below).
- **Naming consistency follow-up (not in this diff):** the HTTP endpoint `/v1/hub` and the docstring at `src/winml/modelkit/serve/app.py:16` still say "WinML Hub". The `index.html` UI label was updated to "ModelKit Catalog" in this PR — please confirm whether keeping `/v1/hub` is intentional (API stability) or a follow-up rename.

🤖 Generated with Claude Code

src/winml/modelkit/commands/catalog.py
```diff
@@ -0,0 +1,457 @@
...
+    if task:
+        result = [m for m in result if m["task"].lower() == task.lower()]
+    return result
+
+
+# EPs always supported by every catalog model (no supported_eps entry needed)
+_ALWAYS_ON_EPS: frozenset[str] = frozenset({"DmlExecutionProvider", "CPUExecutionProvider"})
+
+# Maps optional EP full name → catalog supported_eps key
+_OPTIONAL_EP_TO_CAT_KEY: dict[str, str] = {
```
xieofxie: why we add another mapping? we already have many ep name list

src/winml/modelkit/data/hub_models.json
```diff
@@ -2,2603 +2,609 @@
...
-        "QNN": {
-          "verdict": "PASS",
-          "delta_display": "0.3%"
-        },
-        "OV": {
-          "verdict": "PASS",
-          "delta_display": "0.3%"
-        }
-      }
+      "supported_eps": [
```
xieofxie: we should use a ep-device tuple

src/winml/modelkit/commands/catalog.py
```diff
@@ -0,0 +1,436 @@
...
+    result = models
+    if model_type:
+        result = [m for m in result if m["model_type"].lower() == model_type.lower()]
+    if task:
+        result = [m for m in result if m["task"].lower() == task.lower()]
+    return result
+
+
+# Short display labels for all EPs, in canonical column order
+_EP_SHORT_LABEL: dict[str, str] = {
```
xieofxie: should we place all related logic code in one place?

tests/unit/commands/test_catalog.py
DingmaomaoBJTU: Fixed — both occurrences now use `tmp_path / "out.json"` for cross-platform compatibility.

tests/unit/commands/test_catalog.py
DingmaomaoBJTU: Fixed — same as above, now uses `tmp_path / "out.json"`.

tests/unit/commands/test_catalog.py
DingmaomaoBJTU: Fixed — all nine `test_hub_*` functions renamed to `test_catalog_*`.

src/winml/modelkit/commands/catalog.py
DingmaomaoBJTU: Fixed — added an explicit `device_upper not in {"CPU", "GPU", "NPU"}` guard that returns `[]` for unknown device strings, consistent with `_filter_by_ep`.

tests/unit/commands/test_catalog.py
DingmaomaoBJTU: Fixed — removed the tautological test. Replaced with tests that assert the actual cell values returned by `_make_ep_col_fn_for_ep` against known model data.

src/winml/modelkit/data/hub_models.json
DingmaomaoBJTU: Changed to a `{ep_name: [devices]}` dict format, e.g. `{"QNNExecutionProvider": ["GPU", "NPU"]}`. More compact than a list of tuples and self-describing — no separate lookup table needed to map EPs to devices.

src/winml/modelkit/commands/catalog.py
DingmaomaoBJTU: Done — `_EP_SHORT_LABEL` is now removed

src/winml/modelkit/commands/catalog.py
DingmaomaoBJTU: Further simplified: `supported_eps` keys are now canonical EP aliases (`cpu`, `dml`, `qnn`, `ov`, `vitisai`) matching `EP_ALIASES` in `constants.py`. `_EP_SHORT_LABEL` is keyed by the same aliases, so there is only one naming scheme across CLI input, JSON data, and display labels.

src/winml/modelkit/data/hub_models.json
DingmaomaoBJTU: Further updated: keys are now EP aliases rather than full names, e.g. `{"qnn": ["GPU", "NPU"], "ov": ["CPU", "GPU", "NPU"]}`. More compact and directly matches what users type in `--ep qnn`. Filter functions call `normalize_ep_name()` on the JSON key to handle both alias and full-name lookups.

timenick (approved):
This change might affect the serve command, as it is now hidden. It's acceptable for now.

xieofxie (approved):

PR #477, opened 2026-05-08
fix(build): optional config, autoconf status display, EP compatibility fixes, and perf improvements

DingmaomaoBJTU:
### `winml build` UX improvements
- `-c/--config` is now optional: omit it to auto-generate the build config from `-m <model>` directly (one-step workflow; `-m` becomes required when `-c` is omitted)
- `WinMLBuildConfig` gains a top-level `auto` field (default `True`); the build pipeline sets it to `False` after the optimize + autoconf loop and saves it into `winml_build_config.json` as part of the build output — this marks the config as "already resolved" so CI/CD pipelines or subsequent `winml build -c` runs can reuse it without re-triggering the analyzer
- Setup header now shows `(autoconf on)` / `(autoconf off)` next to the config name; summary section prints the saved config path after the final artifact

### Performance
- Skip PyTorch weight loading in `from_pretrained()` when the ONNX artifact is already cached — use `AutoConfig` instead (~1 s vs ~60 s on a warm cache)

### EP compatibility workaround
- Add `_suppress_ep_registration_stderr()` to silence the WinApp SDK 2.0 / WinML 1.8 API version mismatch warning printed to native stderr on every run:
  ```
  The requested API version [24] is not available, only API versions [1, 23] are supported.
  Current ORT Version is: 1.23.5
  ```
  Root cause: SDK 2.0 EP DLLs target ORT API v24; installed WinML runtime is still v1.8 (ORT 1.23.5, max API v23). No functional impact — ORT falls back cleanly. Workaround uses `SetStdHandle` (Win32) + `os.dup2` (CRT) because the DLL writes via `GetStdHandle(STD_ERROR_HANDLE)`, bypassing Python's stderr. Remove once WinML runtime upgrades to 2.0.

#### `winml perf --ep openvino --device cpu` not working
`winml perf -m <hf-model> --ep openvino --device cpu` was broken. Root cause: when loading a HF model ID, `perf` explicitly triggered EPContext compilation via `generate_hf_build_config()`, which called `WinMLCompileConfig.for_provider(compile_provider)` without forwarding `device`. As a result, `for_openvino()` set no `device_type` in `provider_options`, so OpenVINO compiled a GPU/NPU blob instead of a CPU-specific one. When `WinMLSession` later tried to load that blob on CPU, it raised `INVALID_GRAPH`.

Two fixes:
1. `for_provider()` / `for_openvino()` / `for_qnn()` / `for_vitisai()` now all accept and forward `device` → `provider_options["device_type"]`, so the compiled EPContext is device-specific and cache keys don't collide across devices
2. `winml perf` now defaults to `--no-compile` (skip EPContext compilation during benchmarking); use `--compile` to opt in. Perf is for benchmarking existing artifacts, not building new ones — compilation belongs in `winml build`


src/winml/modelkit/session/session.py
```diff
@@ -155,6 +198,7 @@ def _init_winml_eps_once(cls) -> None:
         try:
             registry = WinMLEPRegistry.get_instance()
             if registry.winml_available:
+                # with _suppress_ep_registration_stderr():
```
timenick: This call site is commented out, so `_suppress_ep_registration_stderr` is never invoked — the function as added is dead code. The PR description says this workaround silences the "API version [24] not available" warning printed at registration time, but with this line commented out the warning will still print on every run.

Either drop the `# ` so the workaround actually runs, or remove the function and the WORKAROUND comment block above to avoid implying functionality that isn't wired up.

🤖 Generated with Claude Code

tests/unit/session/test_winml_session.py
```diff
@@ -773,3 +773,208 @@ def test_ep_none_and_device_auto_returns_none(self) -> None:
...
+
+        from winml.modelkit.session.session import _suppress_native_output
+
+        trap = tmp_path / "stderr_trap.bin"
+        trap_fd = os.open(str(trap), os.O_WRONLY | os.O_CREAT | os.O_TRUNC)
+        old_stderr = os.dup(2)
+        os.dup2(trap_fd, 2)
+        os.close(trap_fd)
+        try:
+            with _suppress_native_output(suppress_stderr=True):
```
timenick, line 792: `_suppress_native_output(log_path=None)` does not accept a `suppress_stderr` parameter — its docstring explicitly says "stderr is left untouched." All three tests below (`test_stderr_suppressed_when_flag_set`, `test_stderr_not_suppressed_by_default`, `test_stderr_captured_to_log_file`) will fail with `TypeError: got an unexpected keyword argument 'suppress_stderr'`.

This looks like the test counterpart to the commented-out workaround in `session.py`: the tests were written against an intended API (`_suppress_native_output(..., suppress_stderr=True)`) that was never added. Either extend `_suppress_native_output` to actually accept and honor `suppress_stderr`, or retarget these tests at `_suppress_ep_registration_stderr` (which is what `test_win32_std_error_handle_redirected_and_restored` already does).

🤖 Generated with Claude Code

src/winml/modelkit/compiler/configs.py
```diff
@@ -116,9 +122,29 @@ def for_provider(cls, provider: str | None) -> WinMLCompileConfig | None:
         return cls(ep_config=EPConfig(provider=provider, enable_ep_context=False))
 
     @classmethod
-    def for_qnn(cls) -> WinMLCompileConfig:
-        """Factory for QNN compilation."""
-        return cls(ep_config=EPConfig(provider="qnn"))
+    def for_qnn(cls, quantize: bool | None = None, device: str | None = None) -> WinMLCompileConfig:
```
timenick: `quantize` is added as a brand-new parameter and immediately deprecated with a `DeprecationWarning` (same in `for_openvino` and `for_vitisai` below). I grepped — no caller in this repo passes `quantize=` to any of these three factories. Deprecating a parameter on the same commit that introduces it adds dead API surface that future readers will have to evaluate.

If there's an external/private caller that motivates this shim, please call that out in the docstring or PR description. Otherwise, dropping the `quantize` parameter (and its warning) from all three factories keeps the API surface honest.

🤖 Generated with Claude Code

src/winml/modelkit/compiler/stages/compile.py
```diff
@@ -219,6 +226,12 @@ def _finalize_output(self, context: CompileContext, model_path: Path, output_dir
                 src_ctx_path = pattern
                 break
 
+        # Glob fallback: catch any device variant (e.g. stem_gpu_ctx.onnx)
+        if src_ctx_path is None:
+            candidates = sorted(model_path.parent.glob(f"{model_path.stem}_*_ctx.onnx"))
```
timenick: `sorted(...)[0]` picks the alphabetically first match, not the one matching the current device. If the work dir contains EPContext outputs from earlier compiles for different devices (e.g. both `<stem>_cpu_ctx.onnx` and `<stem>_npu_ctx.onnx`), this fallback will silently pick `_cpu_ctx.onnx` for an NPU run.

The explicit-pattern loop above already covers the in-spec case (`<stem>_<device>_ctx.onnx` and `<stem>_ctx.onnx`). When that fails and we hit this fallback, it usually means the file name encodes a *different* device — picking arbitrarily is the wrong recovery. Two options that don't widen the bug surface:

1. Drop the glob fallback entirely and let the existing `"EPContext model not found"` warning fire — making the failure mode explicit.
2. Keep the glob but raise (or warn loudly) when `len(candidates) > 1`, so a re-used work dir doesn't silently route the wrong artifact.

🤖 Generated with Claude Code

src/winml/modelkit/build/common.py
```diff
@@ -98,6 +102,10 @@ def run_optimize_analyze_loop(
         analyze_iterations, analyze_black_nodes, analyze_details = 0, 0, {}
 
     elapsed = time.monotonic() - t0
+
+    # Mark as resolved so the saved config skips the analyzer on next build
+    config.auto = False
```
timenick: `run_optimize_analyze_loop` now mutates its `config` argument as a side effect (sets `config.auto = False`). The function name suggests it only runs the optimize/analyze loop and returns timings; callers passing a shared `WinMLBuildConfig` instance will see it modified.

Two small alternatives that keep the contract clearer:

- Move the `config.auto = False` assignment to the caller in `build.py`, right after `run_optimize_analyze_loop` returns and before `to_dict()`-serialization. The mutation belongs to "we just resolved this config," not to "we ran the loop."
- Or, at minimum, add a one-line note at the top of the function docstring that this call mutates `config.auto`. A future reader who only reads the signature won't expect it.

🤖 Generated with Claude Code

src/winml/modelkit/commands/build.py
```diff
@@ -274,9 +275,8 @@ def _build_modules(
     "--no-compile/--compile",
     "no_compile",
     default=None,
-    help="Override compilation from config. --no-compile forces skip; "
-    "--compile forces enable (config must have a compile section). "
-    "Default: inherit from config file.",
+    help="Override compilation. --compile forces enable (config must have a compile section). "
+    "Default: compilation is disabled unless --compile is passed.",
```
xieofxie: disabled even config has compile section?

src/winml/modelkit/build/common.py
xieofxie: Agreeded. Or we return a new  config and caller determines should resave new config or not

src/winml/modelkit/session/session.py
DingmaomaoBJTU: Fixed: re-enabled the call in cea1394e. The `# with` was accidentally left commented out — the workaround is now active again.

DingmaomaoBJTU: Fixed: re-enabled the call in cea1394e. The `# with` was accidentally left commented out — the workaround is now active again.

tests/unit/session/test_winml_session.py
DingmaomaoBJTU: Fixed in 847f4f6f: added `suppress_stderr: bool = False` to `_suppress_native_output`. When `True`, fd 2 is also redirected to the same log/devnull destination as fd 1.

src/winml/modelkit/compiler/configs.py
DingmaomaoBJTU: Fixed in cea1394e: dropped `quantize` from all three factories (`for_qnn`, `for_openvino`, `for_vitisai`) along with the DeprecationWarning shims and unused `import warnings`.

src/winml/modelkit/compiler/stages/compile.py
DingmaomaoBJTU: Fixed in cea1394e: dropped the glob fallback entirely. The existing "EPContext model not found" warning now fires, making the failure explicit.

src/winml/modelkit/build/common.py
DingmaomaoBJTU: Fixed in cea1394e: removed the mutation from `run_optimize_analyze_loop` and moved `config.auto = False` to the call site in `build.py` right after the loop returns, where the "we just resolved this config" intent is clear.

src/winml/modelkit/commands/build.py
DingmaomaoBJTU: Fixed in cea1394e: updated the help text to "inherit from config; when auto-generating config (no -c), compilation is off unless --compile is passed" — distinguishing the config-file path (inherits compile section) from the auto-generate path.

src/winml/modelkit/build/common.py
DingmaomaoBJTU: Fixed in cea1394e: moved to caller — `config.auto = False` now sits in `build.py` immediately after `run_optimize_analyze_loop` returns. The function no longer mutates its argument.

src/winml/modelkit/config/build.py
xieofxie: if we don't fix resolve_device, we should remove all uses of it because it didn't resolve auto considering ep. So it could be totally wrong like `device = auto, ep = cpu` on a npu machine

src/winml/modelkit/models/auto.py
```diff
@@ -427,7 +429,7 @@ def from_pretrained(
             onnx_path=onnx_path,
             config=hf_config,  # HF PretrainedConfig for pipeline compatibility
             device=device,  # pass user's original device string; WinMLSession handles "auto"
-            ep=resolved_ep,
+            ep=kwargs.get("ep"),
```
xieofxie: why since ep=resolved_ep is used above

DingmaomaoBJTU: Good catch — fixed in fe0bbe72. The `kwargs.get("ep")` was a mistake introduced during a rebase; `resolved_ep` is the right value (it's what was actually used to compile the model, so it must also be what the inference wrapper receives).

src/winml/modelkit/config/build.py
DingmaomaoBJTU: Addressed. `resolve_device` now accepts `ep` (added in the main merge at 11ccc692 via `sysinfo/device.py`). All call sites in `config/build.py` already pass `ep=ep` — the two stray calls in `perf.py` that did not were removed in the merge commit.

src/winml/modelkit/session/session.py
```diff
@@ -281,31 +337,55 @@ def compile(self) -> None:
...
-                    "onnx_path": str(self._onnx_path),
-                    "error": str(e),
-                },
-                suggestion=self._get_compile_suggestion(target_device, e),
-            ) from e
+        except Exception as ep_err:
+            # When an explicit EP fails on CPU (e.g. OpenVINO-CPU cannot
+            # deserialize the model), fall back to PREFER_CPU policy so
+            # CPUExecutionProvider handles the model instead.
+            if self._ep and target_device == "cpu":
```
timenick: The guard here is `self._ep and target_device == "cpu"` — so this silent fallback triggers for **any** explicit EP that fails on CPU, not just the OpenVINO-CPU NPU-blob case the PR description targets:

- `--ep openvino --device cpu` failing → falls back (intended)
- `--ep qnn --device cpu` failing for any reason → also falls back
- `--ep dml --device cpu` failing → also falls back

In all of these the user explicitly asked for a specific EP and gets `CPUExecutionProvider` instead. The only signal is `logger.warning(...)`, which doesn't surface to the CLI by default — `session.compile()` returns successfully and `session.state == COMPILED`. The unit test `TestOpenVINOCpuFallback` covers the OpenVINO case but doesn't constrain the breadth.

Two narrowing options:

1. Narrow the trigger to the documented bug: `if self._ep == "openvino" and target_device == "cpu":`. If other EPs hit similar issues later, broaden case-by-case with their own tests.
2. Keep the broad fallback but echo the substitution to the user (e.g. `console.print("⚠️ Requested EP '{ep}' unavailable on CPU, falling back to CPUExecutionProvider")`) so silently swapping the EP doesn't surprise anyone reading the perf/build output.

🤖 Generated with Claude Code

DingmaomaoBJTU: Agreed — removed the fallback entirely. When an explicit EP fails, it should raise CompilationError immediately so the user sees the real error rather than silently switching to CPUExecutionProvider.

xieofxie (approved):

PR #570, opened 2026-05-09
fix(build): inherit --compile from config, add --no-compile override

### LINKED ISSUE #434: fix: honor --compile from config in build, add --no-compile override
## Description

After the internal decision to make \`--compile\` off by default, users who pass \`--compile\` during \`config\` (to produce an EP-specific binary for performance tuning and deployment on fixed hardware) should not need to re-specify it again during \`build\`. Currently the flag is silently ignored across commands, causing confusion.

## Proposed Behavior

- If \`--compile\` is specified in \`config\`, it is honored and **persists as the default for subsequent \`build\` invocations** — users do not need to repeat it.
- If a user wants to skip the compilation step in \`build\` after having set \`--compile\` in \`config\`, they can override with \`--no-compile\`.
- If neither flag is passed to \`build\`, the compile behavior is inherited from what was set in \`config\`.

## Summary of Flag Semantics

| Scenario | Result |
|---|---|
| \`config --compile\` → \`build\` | Compilation runs (inherited) |
| \`config --compile\` → \`build --no-compile\` | Compilation skipped (explicit override) |
| \`config\` (no flag) → \`build\` | Compilation skipped (default) |
| \`config\` (no flag) → \`build --compile\` | Compilation runs (explicit opt-in) |

## Source

Reported via user feedback from Panther Lake testing; design direction confirmed by Qiong Wu and Bruno Sonnino.

DingmaomaoBJTU:
## Summary

- `build`'s `--no-compile/--compile` flag was `default=True`, so the compile section was always cleared from the config file at build time — even when `config --compile` had already saved it
- Changed the flag to a tristate (`default=None`): no flag = inherit from config, `--no-compile` = force skip, `--compile` = force enable (raises a clear error if config has no compile section)

Closes #434


DingmaomaoBJTU:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

src/winml/modelkit/commands/build.py
```diff
@@ -69,13 +69,18 @@ def _load_config(
     Args:
         config_file: Path to JSON config file.
         no_quant: If True, set config.quant = None (skip quantization).
-        no_compile: If True, set config.compile = None (skip compilation).
+        no_compile: Tristate compile override.
```
xieofxie: Tristate ?

DingmaomaoBJTU: Good catch — replaced 'Tristate' with `bool | None` and added explicit per-value descriptions. Pushed in 394f303.

xieofxie (approved):

PR #571, opened 2026-05-09
fix(session): suppress VerifyEachNodeIsAssignedToAnEp warning for explicit EP

DingmaomaoBJTU:
## Problem

`winml perf -m microsoft/resnet-50 --device npu --iterations 100` emits a yellow ORT warning:

```
[W:onnxruntime:, session_state.cc:1316 onnxruntime::VerifyEachNodeIsAssignedToAnEp] Some nodes were not
assigned to the preferred execution providers which may or may not have an negative impact on performance.
```

### Root cause

Two issues introduced in #404 and #440:

**Issue 1 — no-ep discovery path used `add_provider_for_devices`.**
When no `--ep` flag is given, `ef2db0b` added an EP discovery step that called `add_provider_for_devices([matched])` with a single EP device. Any node that couldn't run on that EP (e.g. shape ops on NPU) fell back to CPU without CPU being a registered provider, triggering `VerifyEachNodeIsAssignedToAnEp`.

**Issue 2 — explicit EP path registers only one provider.**
When `--device npu` is used, the build config always sets `resolved_ep="qnn"`, so the session always hits the explicit EP path. `6b8b004` correctly narrowed `_find_ep_device` to return QNN-on-NPU (instead of QNN-on-CPU as before). But with only QNN-on-NPU registered, shape ops have no registered home when they fall back to CPU → warning fires.

## Fix

1. **Remove no-ep discovery block.** Without `--ep`, fall straight to `set_provider_selection_policy(PREFER_*)`. This is the correct abstraction when the user's intent is "prefer this device", not "use this exact EP".

2. **Register CPU as explicit fallback in the explicit EP path.** After registering the primary EP via `add_provider_for_devices`, also register `CPUExecutionProvider`. Shape-related ops assigned to CPU now have a valid registered provider, silencing the warning without touching log severity levels.

## Note

TODO: re-verify once zheng's PR is merged — upstream ORT changes may affect how `add_provider_for_devices` interacts with CPU fallback registration.

### DIFF
--- a/src/winml/modelkit/session/session.py
+++ b/src/winml/modelkit/session/session.py
@@ -423,11 +423,9 @@ class WinMLSession:
 
         When ``self._ep`` is set (and not ``"cpu"``), uses
         ``add_provider_for_devices`` to explicitly bind that EP.
-        ``"cpu"`` falls through to policy-based selection so ORT handles
-        CPU-only inference without any EP registration.
-        When ``self._ep`` is not set, queries ``get_ep_devices()`` to
-        discover an available EP for the target device type. Falls back to
-        policy-based selection only as a last resort.
+        Otherwise uses policy-based selection via DEVICE_POLICY_MAP so ORT
+        handles provider assignment (including CPU fallback for unsupported
+        ops) without triggering ``VerifyEachNodeIsAssignedToAnEp`` warnings.
 
         Note: Returns a **fresh** SessionOptions when using explicit EP to
         avoid "already registered" errors from repeated calls.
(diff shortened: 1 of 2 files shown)


src/winml/modelkit/session/session.py
xieofxie, line 138: use definition from constant.py

src/winml/modelkit/session/session.py
```diff
@@ -445,6 +443,12 @@ def _build_session_options(self, device: str) -> ort.SessionOptions:
 
                     opts = ort.SessionOptions()
                     opts.add_provider_for_devices([matched], self._provider_options)
+                    # Register CPU as explicit fallback so shape-related ops
+                    # have a registered home and VerifyEachNodeIsAssignedToAnEp
+                    # does not warn about unassigned nodes.
+                    cpu_dev = self._find_ep_device(ep_name="CPUExecutionProvider", device="cpu")
```
xieofxie, line 449: maybe only change this line only

PR #586, opened 2026-05-12
fix(perf): per-module benchmark works for ResNet (and similar HF models)

### LINKED ISSUE #192: wmk perf --module: module path construction fails for ResNet (AttributeError: ResNetModel has no attribute 'resnet')
## Summary

\`wmk perf -m microsoft/resnet-50 --module ResNetStage\` crashes with \`AttributeError: ResNetModel has no attribute 'resnet'\` due to a mismatch between the model class used during config generation and the class used when re-instantiating the parent model.

## Repro

\`\`\`bash
wmk perf -m microsoft/resnet-50 --module ResNetStage --iterations 50
\`\`\`

**Error:**
\`\`\`
AttributeError: ResNetModel has no attribute 'resnet'
  File "src/winml/modelkit/commands/perf.py", line 596, in _perf_modules
    submodule = parent_model.get_submodule(module_path)
\`\`\`

## Root Cause

Config generation uses \`model_id="microsoft/resnet-50"\` which resolves to \`ResNetForImageClassification\`. torchinfo traces this model and produces \`module_path = "resnet.encoder.stages.0"\` (with the top-level \`.resnet\` wrapper).

When re-instantiating the parent model in \`_perf_modules\` (\`perf.py:571-575\`), only \`model_type="resnet"\` is passed — \`task\` is \`None\` because submodule configs intentionally omit it (see \`test_module_build_e2e.py:37\`). \`_instantiate_parent_model("resnet", task=None)\` resolves to \`ResNetModel\` (base class) via \`get_supported_tasks("resnet")[0]\`, which has no \`.resnet\` attribute, causing \`get_submodule()\` to fail.

## Fix

Pass \`cfg.loader.task\` (from the parent model's original config) into \`_instantiate_parent_model()\` at \`perf.py:575\` to ensure both phases use the same task-specific model class.

## Acceptance Criteria

- [ ] \`wmk perf -m microsoft/resnet-50 --module ResNetStage --iterations 50\` completes without error
- [ ] Per-module perf works for at least ResNet and BERT model families
- [ ] Test added covering the re-instantiation path with task preservation

xieofxie:
## Summary

Fixes `winml perf --module <ClassName>` on HuggingFace models like `microsoft/resnet-50`, which previously failed with `AttributeError: <Parent> has no attribute '<...>'` and, once that was past, `forward() missing 1 required positional argument`.

Two root causes:

1. **Wrong parent class instantiated.** `_perf_modules` passed the CLI `task` (typically `None`) to `_instantiate_parent_model`. When the model_type's first supported task differs from the architectures-derived task used by `generate_hf_build_config`, the parent class mismatches the recorded `module_path`. For `microsoft/resnet-50`, build configs use `ResNetForImageClassification` (paths prefixed with `resnet.`), but perf was instantiating bare `ResNetModel` → `AttributeError: ResNetModel has no attribute 'resnet'`. Fix: re-resolve the parent task from the model_id via `resolve_loader_config`, matching the path used to compute `module_path`.

2. **Wrong input tensor names.** `_build_submodule_config` hardcoded input names as `input_0`, `input_1`, …, but PyTorch submodules expect specific positional/keyword names (e.g., `ResNetStage.forward(input)`, `ResNetBottleNeckLayer.forward(hidden_state)`, `ResNetEmbeddings.forward(pixel_values)`). Calling `submodule(input_0=tensor)` raises `forward() missing 1 required positional argument`. Fix: propagate forward-arg names through `SubmoduleInfo.input_names` (sourced from the existing hook-capture data, with a `inspect.signature` fallback) and use them when building `InputTensorSpec`. Falls back to `input_{i}` only when names can't be discovered.

End-to-end on a CPU box:

```
$ winml perf -m microsoft/resnet-50 --module ResNetStage --ep cpu --device cpu
                Per-Module Perf: ResNetStage
| Module Path             | Mean (ms) | P90 (ms) | Min  | Max  |
| resnet.encoder.stages.0 |      4.16 |     5.22 | 3.30 | 5.22 |
| resnet.encoder.stages.1 |     14.04 |    24.09 | 8.69 | 24.09 |
| resnet.encoder.stages.2 |      7.41 |     8.65 | 4.93 | 8.65 |
| resnet.encoder.stages.3 |      5.20 |     7.09 | 3.59 | 7.09 |
```

Closes https://github.com/microsoft/WinML-ModelKit/issues/192

### DIFF
--- a/src/winml/modelkit/commands/perf.py
+++ b/src/winml/modelkit/commands/perf.py
@@ -576,12 +576,23 @@ def _perf_modules(
 
     console.print(f"[dim]Found {len(module_configs)} {module_class} instances[/dim]")
 
-    # Instantiate parent with init weights (no pretrained download)
+    # Instantiate parent with init weights (no pretrained download).
+    # Submodule configs intentionally drop `loader.task`, so re-resolve the
+    # parent task from the model_id — the same path `generate_hf_build_config`
+    # used to compute module_path. Without this, models whose architectures
+    # field differs from the first supported task for their model_type (e.g.
+    # microsoft/resnet-50: architectures=ResNetForImageClassification but
+    # supported_tasks[0]=feature-extraction) instantiate the wrong parent
+    # class and get_submodule() raises AttributeError.
     model_type = module_configs[0].loader.model_type
     if not model_type:
         console.print("[red]Error: module configs missing model_type[/red]")
         sys.exit(3)
-    parent_model = _instantiate_parent_model(model_type, task=task)
+
+    from ..loader import resolve_loader_config
+
+    parent_loader_cfg, _, _ = resolve_loader_config(model_id=hf_model, task=task)
+    parent_model = _instantiate_parent_model(model_type, task=parent_loader_cfg.task)
 
     all_results: list[dict[str, Any]] = []
     for i, cfg in enumerate(module_configs):
--- a/src/winml/modelkit/commands/perf.py
+++ b/src/winml/modelkit/commands/perf.py
@@ -578,11 +578,10 @@ def _perf_modules(
     # Instantiate parent with init weights (no pretrained download).
     # Submodule configs intentionally drop `loader.task`, so re-resolve the
     # parent task from the model_id — the same path `generate_hf_build_config`
-    # used to compute module_path. Without this, models whose architectures
-    # field differs from the first supported task for their model_type (e.g.
-    # microsoft/resnet-50: architectures=ResNetForImageClassification but
-    # supported_tasks[0]=feature-extraction) instantiate the wrong parent
-    # class and get_submodule() raises AttributeError.
+    # used to compute module_path. Without this, models whose `architectures`
+    # field maps to a different task than `get_supported_tasks(model_type)[0]`
+    # instantiate the wrong parent class and `get_submodule()` raises
+    # AttributeError.
     model_type = module_configs[0].loader.model_type
     if not model_type:
         raise click.ClickException("module configs missing model_type")
--- a/src/winml/modelkit/config/build.py
+++ b/src/winml/modelkit/config/build.py
@@ -259,6 +259,9 @@ class SubmoduleInfo:
         output_shapes: Shape of each output tensor (e.g., [[1,16,64]])
         input_dtypes: Dtype of each input tensor (e.g., ["float32", "float32"])
         output_dtypes: Dtype of each output tensor (e.g., ["float32"])
+        input_names: Forward-arg names for each input (e.g., ["hidden_state"]
+            or ["pixel_values"]). Empty when hook capture didn't run; callers
+            then fall back to generic ``input_{i}`` names.
     """
 
     class_name: str
@@ -267,6 +270,7 @@ class SubmoduleInfo:
     output_shapes: list[list[int]]
     input_dtypes: list[str]
     output_dtypes: list[str]
+    input_names: list[str] = field(default_factory=list)
 
 
 # =============================================================================
@@ -814,10 +818,20 @@ def _build_submodule_config(
         - Inherited optim/compile from parent
         - Quant with task=None, model_name=None (RandomDataset fallback)
     """
-    # Build InputTensorSpec for EACH input tensor (not just the first)
+
+    # Build InputTensorSpec for EACH input tensor (not just the first).
+    # Use the submodule's actual forward-arg names so build_hf_model can
+    # call submodule(**kwargs) correctly (e.g., ResNetStage expects `input`,
+    # ResNetBottleNeckLayer expects `hidden_state`). Fall back to generic
+    # input_{i} only when names were not discovered.
+    def _input_name(i: int) -> str:
+        if i < len(sub_info.input_names) and sub_info.input_names[i]:
+            return sub_info.input_names[i]
+        return f"input_{i}"
+
     input_tensors = [
         InputTensorSpec(
-            name=f"input_{i}",
+            name=_input_name(i),
             shape=tuple(shape),
             dtype=sub_info.input_dtypes[i] if i < len(sub_info.input_dtypes) else None,
         )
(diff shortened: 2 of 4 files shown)


DingmaomaoBJTU:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

src/winml/modelkit/config/build.py
```diff
@@ -267,6 +270,7 @@ class SubmoduleInfo:
     output_shapes: list[list[int]]
     input_dtypes: list[str]
     output_dtypes: list[str]
+    input_names: list[str] = field(default_factory=list)
```
timenick, line 274: The new `input_names` field plus the `_input_name` fallback in `_build_submodule_config` and the `inspect.signature` path in `_find_submodules_by_class` aren't covered by unit tests. `TestBuildSubmoduleConfig` in `tests/unit/config/test_build.py` is the natural home — worth adding three small cases:

- `input_names=["hidden_state"]` propagates to `InputTensorSpec.name`
- `input_names=[]` (or shorter than `input_shapes`, or contains `""`) falls back to `input_{i}`
- A `_find_submodules_by_class` case where `hook_data` is empty exercises the new `inspect.signature` branch

Root cause #2 in the PR description is currently only validated by the manual `winml perf -m microsoft/resnet-50 ...` run.

🤖 Generated with Claude Code

src/winml/modelkit/commands/perf.py
```diff
@@ -576,12 +576,23 @@ def _perf_modules(
 
     console.print(f"[dim]Found {len(module_configs)} {module_class} instances[/dim]")
 
-    # Instantiate parent with init weights (no pretrained download)
+    # Instantiate parent with init weights (no pretrained download).
+    # Submodule configs intentionally drop `loader.task`, so re-resolve the
+    # parent task from the model_id — the same path `generate_hf_build_config`
+    # used to compute module_path. Without this, models whose architectures
+    # field differs from the first supported task for their model_type (e.g.
+    # microsoft/resnet-50: architectures=ResNetForImageClassification but
```
timenick: Per CLAUDE.md style, comments shouldn't pin themselves to specific PR-context models (this rots if HF renames `ResNetForImageClassification` or reorders supported tasks). Consider generalizing to something like "models whose `architectures` field maps to a different task than `get_supported_tasks(model_type)[0]`" — the invariant survives without naming a moving target.

🤖 Generated with Claude Code

src/winml/modelkit/config/build.py
```diff
@@ -814,10 +818,20 @@ def _build_submodule_config(
         - Inherited optim/compile from parent
         - Quant with task=None, model_name=None (RandomDataset fallback)
     """
-    # Build InputTensorSpec for EACH input tensor (not just the first)
+
+    # Build InputTensorSpec for EACH input tensor (not just the first).
+    # Use the submodule's actual forward-arg names so build_hf_model can
+    # call submodule(**kwargs) correctly (e.g., ResNetStage expects `input`,
```
timenick: Same nit as the perf.py comment: naming `ResNetStage` / `ResNetBottleNeckLayer` is helpful at review time but reads like PR-description content. Generalizing to "submodule forward args may be positional (e.g. `input`) or keyword (e.g. `hidden_state`)" keeps the rationale architecture-agnostic.

🤖 Generated with Claude Code

src/winml/modelkit/config/build.py
```diff
@@ -1102,6 +1118,18 @@ def _find_submodules_by_class(
             layer_input_dtypes = [param_dtype] * len(layer_input_shapes)
             layer_output_dtypes = [param_dtype] * len(layer_output_shapes)
 
+            # Without hook data, derive names from the forward signature so
+            # build_hf_model can invoke the submodule with the correct kwargs.
+            import inspect as _inspect
```
timenick: Nit: `import inspect as _inspect` inside the function — there's no name collision (the sibling `winml.modelkit.inspect` package is only referenced via `from ..inspect.module_io_capture import ...`, never `import inspect`), and `inspect` is a stdlib module that's cheap to import at module top. A normal top-level `import inspect` would be more conventional.

🤖 Generated with Claude Code

src/winml/modelkit/commands/perf.py
```diff
@@ -576,12 +576,23 @@ def _perf_modules(
...
+    # class and get_submodule() raises AttributeError.
     model_type = module_configs[0].loader.model_type
     if not model_type:
         console.print("[red]Error: module configs missing model_type[/red]")
         sys.exit(3)
-    parent_model = _instantiate_parent_model(model_type, task=task)
+
+    from ..loader import resolve_loader_config
+
+    parent_loader_cfg, _, _ = resolve_loader_config(model_id=hf_model, task=task)
```
timenick, line 591: Non-blocking observation: `generate_hf_build_config` (line 556) already calls `resolve_loader_config(model_id, task=...)` internally; this is a second call for the same `(model_id, task)`. `AutoConfig.from_pretrained` is in-process cached so the cost is negligible, but a cleaner long-term shape would be to surface the resolved task on the returned `WinMLBuildConfig`-list (or on `SubmoduleInfo`) so the parent task is recoverable without re-resolution. Flagging only — out of scope for this fix.

🤖 Generated with Claude Code

src/winml/modelkit/commands/perf.py
xieofxie: Generalized — the comment now describes the invariant ( field mapping to a different task than ) without naming a specific model. 854e55e8

src/winml/modelkit/config/build.py
xieofxie: Generalized — comment now describes the rule ("submodule forward args may be positional (e.g. `input`) or keyword (e.g. `hidden_state`)") instead of naming ResNet classes. 854e55e8

src/winml/modelkit/config/build.py
xieofxie: Hoisted to top-level `import inspect` — no collision with `..inspect` since that's always imported via `from`. 854e55e8

src/winml/modelkit/config/build.py
xieofxie: Added three tests in `tests/unit/config/test_build.py` (854e55e8):

- `TestBuildSubmoduleConfig.test_input_names_propagate` — `input_names=["hidden_state"]` flows to `InputTensorSpec.name`
- `TestBuildSubmoduleConfig.test_input_names_fallback` — covers empty list, list shorter than `input_shapes`, and empty-string entries; all fall back to `input_{i}`
- `TestFindSubmodulesByClass.test_signature_fallback_when_hook_data_empty` — patches `capture_module_io` to return `{}` and verifies `inspect.signature` derives `["hidden_state"]` for a custom submodule

All 124 tests in `test_build.py` pass.

src/winml/modelkit/commands/perf.py
xieofxie: (Correcting the previous reply — backticks got eaten by the shell.) Generalized — the comment now reads:

> Without this, models whose `architectures` field maps to a different task than `get_supported_tasks(model_type)[0]` instantiate the wrong parent class and `get_submodule()` raises AttributeError.

854e55e8

timenick (approved):

PR #592, opened 2026-05-12
fix: force utf-8 stdout/stderr on package import

vortex-captain:
## Summary

- On Windows shells started with a charmap codec (most commonly `PYTHONIOENCODING=cp1252`), any emoji or non-Latin-1 character emitted via `print`, `rich.Console`, or logging raised `UnicodeEncodeError: 'charmap' codec can't encode character ...` — affecting CLI banners, progress output, and analyzer reports.
- `winml/modelkit/__init__.py` now reconfigures `sys.stdout`/`sys.stderr` to utf-8 on import (guarded with `hasattr(..., "reconfigure")` so wrapped streams under pytest capture/pipelines are left alone).
- Added `tests/unit/test_init_encoding.py`: a subprocess UT that sets `PYTHONIOENCODING=cp1252`, runs `import winml.modelkit; print('\U0001f680')`, and asserts the child exits cleanly. The test fails on `main` with the exact charmap error and passes with this fix.

### DIFF
--- a/src/winml/modelkit/__init__.py
+++ b/src/winml/modelkit/__init__.py
@@ -29,8 +29,16 @@ Usage:
 """
 
 import logging
+import sys
 from importlib.metadata import PackageNotFoundError, version
 
+
+# Force utf-8 stdout/stderr so emoji and Unicode output (rich console, logs,
+# CLI banners) does not raise UnicodeEncodeError on Windows shells that start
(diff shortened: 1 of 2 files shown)


xieofxie (approved):

vortex-captain:
fixes https://github.com/microsoft/WinML-ModelKit/issues/535

PR #599, opened 2026-05-12
feat(eval): SA eval pipeline with per-stage perf, quantize step, and workflow HTML report

DingmaomaoBJTU:
## Summary

- **SA eval pipeline rewrite**: uses `winml config` + `winml build` for model generation (export, optimize, quantize, compile), replacing the old subprocess-based export + Python optimize approach
- **4-stage perf benchmarking**: `winml perf` runs on export, optimized, quantized, and compiled ONNX models; each perf column shows gain% vs the previous stage
- **SA false alarm detection**: compares SA predictions against EPContext ground truth to detect UNSUPPORTED and PARTIAL false alarms (ops SA predicted non-supported but EP actually handled)
- **Non-NPU support**: automatically skips quantize for non-NPU devices; adds `stage_compile_post` to compile optimized.onnx directly when build skips compile
- **Per-stage node counts**: reports graph node count for each pipeline stage (export → graph_optimized → optimized → quantized → compiled)
- **`--registry` flag**: accepts model registry JSON (same format as `run_eval.py`), with `--priority`, `--task`, `--group`, `--model-type` filters (default: P0+P1+P2)
- **EP alias support**: `--ep` accepts both short alias (qnn, dml, cpu) and full ORT name (QNNExecutionProvider)
- **`--report-only` flag**: regenerates the HTML from existing per-model `sa_eval_result.json` files without re-running any eval stages
- **`--cleanup` fix**: now removes all non-JSON files including extensionless ONNX external data and recurses subdirectories
- **`gen_eval_model.py`**: new standalone script to batch-export HuggingFace models to ONNX via `winml export`, with identical config/task parameters as `run_eval.py`

### Review comment fixes
- Reverted `quantize.py` changes (deferred to PR #608)
- Fixed redundant `import json` in `sa_report.py` (CodeQL)
- Replaced all `wmk` references with `winml` in comments/docstrings
- Added `_resolve_ep_arg()` for fail-fast on unknown EPs
- Fixed `_final_perf` using `or` which treats `0.0` as falsy
- Guard `get_optimization_config` when `ep_found=False`

## Usage

```bash
# Full eval on NPU with QNN
uv run python scripts/e2e_eval/run_sa_eval.py --registry scripts/e2e_eval/testsets/models_all.json \
  --ep qnn --device npu --use-cache --cleanup

# Single model on GPU (quantize auto-skipped)
uv run python scripts/e2e_eval/run_sa_eval.py --model google/vit-base-patch16-224 --ep qnn --device gpu

# Regenerate report only
uv run python scripts/e2e_eval/run_sa_eval.py --report-only --output-dir sa_eval_results/2026-06-08

# Batch export models
uv run python scripts/e2e_eval/gen_eval_model.py --model microsoft/resnet-50
uv run python scripts/e2e_eval/gen_eval_model.py --priority P0 --task image-classification
```


src/winml/modelkit/commands/quantize.py
```diff
@@ -98,6 +98,12 @@
     default=None,
     help="Task for calibration dataset selection (e.g., 'image-classification').",
 )
+@click.option(
+    "--model-name",
```
xieofxie, line 102: I suggest use model-id? although conflicts with the parameter inside function, it have benefits
- align with help
- eval.py already uses it

Maybe we should update function to align cases like this should use model-id

src/winml/modelkit/commands/quantize.py
xieofxie, line 42: could use cli_utils

src/winml/modelkit/commands/quantize.py
timenick, line 102: The new flag on `winml quantize` is named `--model-name`, but the rest of the CLI surface uses `--model-id` for the HuggingFace model identifier (e.g. `winml eval --model-id google/vit-base-patch16-224`). The semantic role is identical — load tokenizer/processor for an HF model id — so the names should match. CLI flags are public API; this is hard to change after release.

Suggest renaming to `--model-id` for consistency with `winml eval`.

🤖 Generated with Claude Code

scripts/e2e_eval/run_sa_eval.py
```diff
@@ -108,6 +108,104 @@ def run_wmk_export(hf_id: str, task: str, output: Path) -> tuple[int, str]:
...
+        "--device",
+        device.lower(),
+        "--iterations",
+        str(iterations),
+        "--warmup",
+        str(warmup),
+        "--output",
+        str(output_json),
+    ]
+    ep_arg = _EP_TO_PERF_ARG.get(ep, ep.lower() if ep else None) if ep else None
```
timenick:
```python
ep_arg = _EP_TO_PERF_ARG.get(ep, ep.lower() if ep else None) if ep else None
```

Two issues:
- The outer `if ep else None` makes the inner `if ep else None` dead code — it can't be reached when `ep` is falsy.
- The `ep.lower()` fallback is wrong: if `ep` is an unmapped EP name (typo, or a new EP added to ORT but not `_EP_TO_PERF_ARG`), it passes `"someexecutionprovider"` to `winml perf --ep`, which will error out with a confusing message. Cleaner to fail-fast on unknown EPs:

```python
if ep is None:
    ep_arg = None
elif ep in _EP_TO_PERF_ARG:
    ep_arg = _EP_TO_PERF_ARG[ep]
else:
    raise ValueError(f"Unknown EP {ep!r}; add to _EP_TO_PERF_ARG.")
```

Loud failure beats silently shipping a garbage `--ep` value to the perf subprocess.

🤖 Generated with Claude Code

scripts/e2e_eval/run_sa_eval.py
```diff
@@ -108,6 +108,104 @@ def run_wmk_export(hf_id: str, task: str, output: Path) -> tuple[int, str]:
     return result.returncode, (result.stderr or "").strip()[-500:]
 
 
+# Map full EP names to the short form accepted by `wmk perf --ep`
+_EP_TO_PERF_ARG: dict[str, str] = {
```
timenick, line 114: The mapping hardcodes ORT-symbol → short-form pairs. The `winml perf --ep` command has its own list of accepted short-form values. If anyone adds an EP (or renames a short form) in one place, the other side silently breaks. Possible mitigations:
- Import the canonical mapping from the source where it's defined (if there is one), or
- Add a smoke test in `tests/cli/` that verifies every value in `_EP_TO_PERF_ARG.values()` is a valid `winml perf --ep` choice.

Per the project's EP canonical names: ORT symbol names look correct (`QNNExecutionProvider`, `DmlExecutionProvider`, `NvTensorRTRTXExecutionProvider`). No naming-convention violations.

🤖 Generated with Claude Code

scripts/e2e_eval/run_sa_eval.py
```diff
@@ -108,6 +108,104 @@ def run_wmk_export(hf_id: str, task: str, output: Path) -> tuple[int, str]:
     return result.returncode, (result.stderr or "").strip()[-500:]
 
 
+# Map full EP names to the short form accepted by `wmk perf --ep`
```
timenick: Comments and docstrings throughout this file refer to a `wmk` CLI that doesn't exist in this project — the actual subprocess invocations all use `python -m winml.modelkit.cli`, the entry point in `pyproject.toml` is `winml`, and there are zero other `wmk` references in the repo.

Affected locations in this PR:
- L13 module docstring: `Stage 1: wmk export + Python optimize_onnx (default)`
- L85-86 function name + docstring: `def run_wmk_export(...)` / `"""Run wmk export via subprocess..."""`
- L111 comment: `Map full EP names to the short form accepted by` **`wmk perf --ep`**
- L133 docstring: `Run wmk perf on onnx_path`
- L235 caller of `run_wmk_export`
- L388 docstring: `Run wmk compile ...`
- L474 docstring: `Runs` **`wmk quantize`**`...`

Suggest a search-and-replace `wmk` → `winml` for comments and docstrings, and renaming the helper `run_wmk_export` → `run_winml_export` (the new `run_winml_perf` already follows the right convention). Anyone copying a command out of a docstring should be able to actually run it.

🤖 Generated with Claude Code

src/winml/modelkit/commands/quantize.py
zhenchaoni, line 102: In the `quantize` command e2e test, I also need to add this option to perform e2e testing. Are you OK to undo this part in your PR and my PR will cover this.

https://github.com/microsoft/WinML-ModelKit/pull/608

DingmaomaoBJTU: Reverted quantize.py changes in this PR. PR #608 will cover this.

src/winml/modelkit/commands/quantize.py
DingmaomaoBJTU: Reverted quantize.py changes in this PR. PR #608 will cover this.

src/winml/modelkit/commands/quantize.py
DingmaomaoBJTU: Reverted quantize.py changes in this PR. PR #608 will cover this.

DingmaomaoBJTU: Reverted quantize.py changes in this PR. PR #608 will cover this.

scripts/e2e_eval/sa_report.py
```diff
@@ -609,5 +723,26 @@
...
 }
 .header-stats { display: flex; gap: 20px; font-size: 13px; color: var(--text2); }
-.header-stats span { font-weight: 600; color: var(--accent); }
-"""
+.header-stats span { font-weight: 600; color: var(--accent); }"""
+
+
+if __name__ == "__main__":
+    import argparse
+    import json
```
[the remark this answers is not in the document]
DingmaomaoBJTU: Fixed — removed redundant `import json` in `__main__` block.

scripts/e2e_eval/run_sa_eval.py
DingmaomaoBJTU: Fixed — added `_resolve_ep_arg()` helper that raises `ValueError` on unknown EPs. All `_EP_TO_PERF_ARG.get()` calls replaced.

scripts/e2e_eval/run_sa_eval.py
DingmaomaoBJTU: Added `_resolve_ep_arg()` that fails fast on unknown EPs. Smoke test suggestion noted for follow-up.

scripts/e2e_eval/run_sa_eval.py
DingmaomaoBJTU: Fixed — all `wmk` references replaced with `winml` in comments and docstrings.

DingmaomaoBJTU (commented):
Thorough review of the SA eval pipeline. Found several correctness bugs and edge-case gaps — details in the inline comments below.

scripts/e2e_eval/gen_eval_model.py
```diff
@@ -0,0 +1,383 @@
...
+    t_err = threading.Thread(target=_read, args=(proc.stderr, stderr_chunks))
+    t_out.start()
+    t_err.start()
+
+    try:
+        proc.wait(timeout=timeout)
+    except subprocess.TimeoutExpired:
+        timed_out = True
+        proc.kill()
+        proc.wait(timeout=10)
```
DingmaomaoBJTU: **Unhandled `TimeoutExpired` on second `proc.wait()`.**

After `proc.kill()`, the follow-up `proc.wait(timeout=10)` can also raise `subprocess.TimeoutExpired` if the OS takes longer than 10 s to reap the killed process (rare but possible on a loaded machine). This propagates unhandled all the way up, aborting the entire eval run with a stack trace rather than recording a timeout result.

```python
try:
    proc.wait(timeout=10)
except subprocess.TimeoutExpired:
    pass  # Already killed; ignore if still not dead
```

scripts/e2e_eval/gen_eval_model.py
```diff
@@ -0,0 +1,383 @@
...
+    slug = _make_slug(entry.hf_id, entry.task)
+    model_dir = output_dir / slug
+    model_dir.mkdir(parents=True, exist_ok=True)
+
+    output_path = model_dir / "export.onnx"
+
+    # Skip if already exported
+    if output_path.exists() and output_path.stat().st_size > 0:
+        safe_print(f"  [cached] {output_path}")
+        return {"success": True, "onnx_path": str(output_path), "elapsed": 0}
```
DingmaomaoBJTU, line 148: **Cache-skip logic is broken for composite models.**

The early-return guard (lines 142–144) checks `output_path.exists()`, where `output_path = model_dir / 'export.onnx'`. But composite (encoder-decoder) models never write `export.onnx` — they write `export_{label}.onnx` files. So the skip-if-cached check never fires for composite models; they are re-exported on every run regardless of whether prior artifacts exist.

Fix: check for the presence of the expected `export_{label}.onnx` outputs using the same sub-config glob logic used during export.

scripts/e2e_eval/run_sa_eval.py
```diff
@@ -82,324 +83,551 @@ def is_cached(path: Path) -> bool:
...
+    JSON result files and perf logs are preserved so --report-only and
+    --use-cache still work for the JSON-driven stages.
+    """
+    freed = 0
+    for f in model_dir.iterdir():
+        if f.is_file() and f.suffix != ".json":
+            size = f.stat().st_size
+            f.unlink()
+            freed += size
+    safe_print(f"  [cleanup] Freed {freed / 1024**2:.1f} MB of artifacts")
```
DingmaomaoBJTU, line 256: **`cleanup_onnx_artifacts` has two silent failure modes.**

1. **External data subdirectories are not cleaned.** Large models (Whisper, LLaMA, etc.) store weight tensors as external data files, sometimes in a sub-directory. `model_dir.iterdir()` only visits the *top-level* directory, so those sub-directories and their contents are left behind, defeating the purpose of `--cleanup`.

2. **`f.unlink()` can raise `PermissionError`/`OSError`** (e.g. a file still open by a subprocess that hasn't fully exited). The exception propagates unhandled, aborting cleanup mid-way with `freed` underreported and subsequent models not cleaned.

Suggested fix:
```python
for f in model_dir.rglob('*'):  # recurse into sub-dirs
    if f.is_file() and f.suffix != '.json':
        try:
            size = f.stat().st_size
            f.unlink()
            freed += size
        except OSError:
            pass
```

scripts/e2e_eval/run_sa_eval.py
```diff
@@ -82,324 +83,551 @@ def is_cached(path: Path) -> bool:
...
+    Returns None on success, or a skip reason string on failure.
     """
-    from winml.modelkit.optim import optimize_onnx
+    config_path = model_dir / "build_config.json"
+    export_path = model_dir / "export.onnx"
+    optimized_path = model_dir / "optimized.onnx"
+
+    # Skip entire build if key artifacts already exist and cache is enabled
+    if use_cache and is_cached(export_path) and is_cached(optimized_path):
+        safe_print("  [Build] Using cached artifacts (export.onnx, optimized.onnx)")
```
DingmaomaoBJTU: **Incomplete cache validation in the build stage.**

The guard at lines 270–271 checks only `export_path` and `optimized_path`. If a previous run was interrupted after those two files were produced but before `quantized.onnx` or `compiled.onnx` were written, `use_cache=True` will treat the build as complete. Downstream stages that depend on the missing artifacts will silently skip, resulting in a partial result that looks like a full one.

When `run_quantize=True`, also gate on `is_cached(model_dir / 'quantized.onnx')`. When `run_compile=True`, also gate on `is_cached(model_dir / 'compiled.onnx')`.

scripts/e2e_eval/run_sa_eval.py
```diff
@@ -82,324 +83,551 @@ def is_cached(path: Path) -> bool:
...
+    """Run SA with information on export.onnx (pre-optimization state).
 
-    Uses optimize_onnx Python API directly with WinMLOptimizationConfig kwargs.
-    Returns path to sa_optimized.onnx, or None on failure.
+    Returns (classifications, info_items) or None on failure.
     """
-    from winml.modelkit.optim import optimize_onnx
-
-    sa_opt_path = model_dir / "sa_optimized.onnx"
+    graph_opt_path = model_dir / "graph_optimized.onnx"
```
DingmaomaoBJTU, line 513: **Docstring says `export.onnx` but the function analyzes `graph_optimized.onnx`.**

Line 478: `"""Run SA with information on export.onnx (pre-optimization state).`

The function body (line 486) assigns `graph_opt_path = model_dir / 'graph_optimized.onnx'` and that is the actual input to SA. The docstring is incorrect and will mislead readers trying to understand the pipeline ordering. Update it to say `graph_optimized.onnx`.

scripts/e2e_eval/run_sa_eval.py
```diff
@@ -691,17 +1113,126 @@ def main() -> None:
...
+    if args.report_only:
+        models_dir = output_dir / "models"
+        result_files = sorted(models_dir.glob("*/sa_eval_result.json"))
+        if not result_files:
+            safe_print(f"[ERROR] No sa_eval_result.json files found under {models_dir}")
+            sys.exit(1)
+        all_results = [json.loads(f.read_text(encoding="utf-8")) for f in result_files]
+        safe_print(f"Collected {len(all_results)} model results from disk.")
+        report = build_aggregate_report(
+            all_results, args.models_file, ep=args.ep, device=args.device
```
DingmaomaoBJTU: **`--report-only` ignores `--registry` when building the aggregate report.**

`build_aggregate_report` is called with `args.models_file` (the legacy JSON path), not `args.registry`. When a user passes `--registry models_all.json --report-only`, the report's `input` metadata field will still show the default `models_with_acc.json` path, making the report misleading.

The same bug exists at line 1327 in the normal (non-report-only) code path.

Fix both call sites:
```python
models_source = args.registry if args.registry else args.models_file
report = build_aggregate_report(all_results, models_source, ep=args.ep, device=args.device)
```

scripts/e2e_eval/sa_comparison.py
```diff
@@ -78,6 +80,18 @@ def run_sa_with_info(
...
+        import logging
+
+        logging.getLogger(__name__).warning(
+            "SA produced no results for EP=%s — no runtime rule data available. "
+            "Returning empty classifications.",
+            ep,
+        )
+
     # Get optimization config from SA recommendations
     optim_config = dict(result.get_optimization_config(ep))
```
DingmaomaoBJTU: **`get_optimization_config` is called even when `ep_found=False`, potentially raising or returning wrong data.**

When the EP has no rule data, SA returns no results for it and `ep_found` stays `False`. The warning is logged, but `result.get_optimization_config(ep)` is still called unconditionally on the next line. Depending on the SA API implementation this may raise `KeyError`/`AttributeError`, or silently return an empty/default config that gets written to the result JSON as if SA actually ran — masking the no-data condition from downstream consumers.

Guard the call:
```python
optim_config = dict(result.get_optimization_config(ep)) if ep_found else {}
```

scripts/e2e_eval/gen_eval_model.py
DingmaomaoBJTU: Fixed in 3998eee1.

scripts/e2e_eval/gen_eval_model.py
DingmaomaoBJTU: Fixed in 3998eee1.

scripts/e2e_eval/run_sa_eval.py
DingmaomaoBJTU: Fixed in 3998eee1.

scripts/e2e_eval/run_sa_eval.py
DingmaomaoBJTU: Fixed in 3998eee1.

scripts/e2e_eval/run_sa_eval.py
DingmaomaoBJTU: Fixed in 3998eee1.

scripts/e2e_eval/run_sa_eval.py
DingmaomaoBJTU: Fixed in 3998eee1.

scripts/e2e_eval/sa_comparison.py
DingmaomaoBJTU: Fixed in 3998eee1.

scripts/e2e_eval/sa_report.py
```diff
@@ -126,15 +189,31 @@ def generate_sa_html_report(report_data: dict, output_path: Path) -> None:
...
-    n_regressed = effectiveness.get("models_regressed", 0)
-    avg_pre_unknown = pre_opt.get("avg_unknown_count", 0)
-    avg_post_unknown = post_opt.get("avg_unknown_count", 0)
     delta_cls = "c-good" if avg_delta > 0 else "c-muted"
-    regressed_cls = "c-bad" if n_regressed > 0 else "c-muted"
+
+    # Perf gain summary stats (from viewer_data)
+    # "final" perf = compiled if available, else quantized
+    def _final_perf(d: dict) -> float | None:
+        return d.get("perf_compiled_mean") or d.get("perf_quantized_mean")
```
timenick: `a or b` treats `0` and `0.0` as falsy, so if `perf_compiled_mean` is ever reported as `0.0` (defensively unrealistic for inference latency, but the perf JSON does come from an external subprocess), this silently falls through to `perf_quantized_mean` and mixes stages in the “Avg Perf Gain” card. Safer:

```python
compiled = d.get("perf_compiled_mean")
return compiled if compiled is not None else d.get("perf_quantized_mean")
```

scripts/e2e_eval/sa_report.py
```diff
@@ -126,15 +189,31 @@ def generate_sa_html_report(report_data: dict, output_path: Path) -> None:
...
+        1 for d in viewer_data if d["perf_exported_mean"] is None and _final_perf(d) is not None
+    )
+    gain_vals = [
+        (d["perf_exported_mean"] - _final_perf(d)) / d["perf_exported_mean"] * 100
+        for d in viewer_data
+        if d["perf_exported_mean"] is not None and _final_perf(d) is not None
+    ]
+    avg_perf_gain = sum(gain_vals) / len(gain_vals) if gain_vals else None
+    n_with_gain = sum(1 for g in gain_vals if g > 1)
+    gain_cls = "c-good" if avg_perf_gain and avg_perf_gain > 1 else "c-muted"
```
timenick: A negative `avg_perf_gain` (regression on average) renders in `c-muted` (gray) rather than red, because the branch only goes to `c-good` when `> 1` and to `c-muted` otherwise. Users skimming the summary card may miss that the average is a regression. Suggestion:

```python
if avg_perf_gain is None:
    gain_cls = "c-muted"
elif avg_perf_gain > 1:
    gain_cls = "c-good"
elif avg_perf_gain < -1:
    gain_cls = "c-bad"
else:
    gain_cls = "c-muted"
```

scripts/e2e_eval/run_sa_eval.py
```diff
@@ -412,12 +664,23 @@ def evaluate_model(
...
 ) -> dict | None:
-    """Run the 4+1 stage SA eval pipeline for a single model."""
+    """Run the winml build + SA analysis pipeline for a single model."""
     hf_id = model_entry["hf_id"]
     task = model_entry.get("task", "")
     model_type = model_entry.get("model_type", "")
 
+    # Non-NPU devices: skip quantize entirely
+    if device.upper() != "NPU":
+        run_quantize = False
```
timenick, line 686: This silently overrides the user’s `--no-quantize` / `--quantize-precision` flags when `--device` is non-NPU. A user who explicitly passes `--device DML --quantize-precision int8` will see no quantize stage and no log line explaining why. Consider a one-line `safe_print` so the override is at least visible in stdout, e.g.:

```python
if device.upper() != "NPU" and run_quantize:
    safe_print(f"  [info] device={device} — skipping quantize (NPU-only)")
    run_quantize = False
```

timenick (commented):
Three minor observations on the new pipeline — all low-severity.

xieofxie:
🟠 **`stage_build` cache logic — misleading log + likely incomplete rebuild**

`scripts/e2e_eval/run_sa_eval.py:682-690`:
```python
if use_cache and is_cached(export_path) and is_cached(optimized_path):
    if run_quantize and not is_cached(model_dir / "quantized.onnx"):
        safe_print("  [Build] Cache incomplete (missing quantized.onnx), rebuilding...")
    elif run_compile and not is_cached(model_dir / "compiled.onnx"):
        safe_print("  [Build] Cache incomplete (missing compiled.onnx), rebuilding...")
    else:
        ...
        return None
```

The "rebuilding..." log falls through to the build path, but `--rebuild` is only passed when `not use_cache` (line 731). So we tell the user we're rebuilding while in fact calling `winml build` without `--rebuild` — the behavior depends on whatever `winml build` does when artifacts exist. Either pass `--rebuild` here, or change the log to say "resuming build for missing artifacts."

timenick (approved):

xieofxie (approved):

PR #600, opened 2026-05-12
zip rules to release

fangyangci:


timenick (approved):

timenick:
How about adding a version number to the zip files, like rules-0.0.2.zip?

PR #601, opened 2026-05-12
fix(perf): improve --module UX (reject .onnx, clarify help, suggest classes on typo)

### LINKED ISSUE #195: wmk perf --module: --module flag expects class name not module path; unclear error when wrong format used
## Summary

\`wmk perf --module\` expects a **PyTorch module class name** (e.g., \`ResNetStage\`, \`BertAttention\`), not a dotted module path. When an incorrect format is passed, the error message does not guide the user toward the correct usage. Additionally, CLIP's \`CLIPEncoderLayer\` fails because its \`forward()\` requires \`attention_mask\` and \`causal_attention_mask\` arguments that the random input generator does not supply.

## Repro

\`\`\`bash
# Wrong: module path (fails silently or with confusing error)
wmk perf -m openai/clip-vit-base-patch32 --module vision_model

# Correct: class name
wmk perf -m microsoft/resnet-50 --module ResNetStage

# Still fails: CLIP class name (forward signature issue)
wmk perf -m openai/clip-vit-base-patch32 --module CLIPEncoderLayer
# Error: CLIPEncoderLayer.forward() missing attention_mask and causal_attention_mask
\`\`\`

## Issues

1. **Documentation gap**: \`wmk perf --help\` states \`HF module class name\` but the distinction from module path is not obvious to users
2. **CLIP forward signature**: \`CLIPEncoderLayer.forward()\` requires extra keyword arguments (\`attention_mask\`, \`causal_attention_mask\`) that \`generate_random_inputs()\` does not generate, causing all 12 CLIP layer benchmarks to fail

## Acceptance Criteria

- [ ] \`wmk perf --module\` help text clearly states class name is required with an example
- [ ] Passing a dotted path (e.g., \`encoder.layers.0\`) produces a clear error: "did you mean class name?"
- [ ] Random input generator handles attention mask shapes for CLIP/transformer modules, or a \`--module-kwargs\` escape hatch is provided

### LINKED ISSUE #553: [winml perf] [P1] `--module NoSuchClass` exits 3 with wrong error (blames the model file)
**Severity**: P1  
**Command**: `winml perf`

---
**Category**: Error Messages — the error message must point at the actual problem.

**Repro**:
```
uv run winml perf -m temp\cli-audit\resnet.onnx --iterations 3 --warmup 1 --module NoSuchClass
```

**Actual**:
```
Generating module configs for NoSuchClass...
Error generating module configs: It looks like the config file at 'temp\cli-audit\resnet.onnx' is not a valid JSON file.
```
Exit 3 (a code emitted nowhere else in the CLI).

**Expected**: Validate `--module` against the registered class list at entry:
```
Error: Module 'NoSuchClass' not found. Valid: WinMLModelForImageClassification, WinMLModelForCausalLM, ...
```
Exit 2.

**Why it matters**: The error blames `resnet.onnx` (which is a perfectly valid ONNX) when the actual problem is the unknown class name. Also introduces a 4th distinct exit code in the CLI (0/1/2/3) with no documentation.

xieofxie:
## Summary

Three related fixes to make `winml perf --module CLASSNAME` self-explanatory when things go wrong.

### 1. Reject `--module` on `.onnx` paths (Closes #553)

Previously, `winml perf -m foo.onnx --module SomeClass` failed with:

```
Generating module configs for SomeClass...
Error: Error generating module configs: It looks like the config file at 'foo.onnx' is not a valid JSON file.
```

The message blamed the model file — but the user's `.onnx` was fine; the real problem is that `--module` requires walking a live `nn.Module` graph (via torchinfo), which a frozen ONNX file can't provide. The CLI now rejects this combination upfront:

```
$ winml perf -m foo.onnx --module SomeClass --ep cpu --device cpu
Error: --module is not supported for ONNX files. Submodule benchmarking requires a HuggingFace model ID (e.g., 'microsoft/resnet-50'), not 'foo.onnx'.
$ echo $?
2
```

### 2. Clarify `--module` help text (Closes #195, doc-gap criterion)

The previous help text said `HF module class name (e.g., 'BertAttention')` — which doesn't make the distinction from a dotted module path obvious. Updated to:

```
--module TEXT   PyTorch module class name (NOT a dotted module path) for
                per-module benchmarking. Every instance of the class in
                the model is built and benchmarked separately. Example:
                '--module BertAttention' (correct), not
                '--module encoder.layer.0.attention' (a path, will not match).
```

### 3. Suggest valid class names on typo (#195, suggestion criterion)

When `--module` doesn't match anything, the CLI used to say only `No modules matching 'FooBar' found`. It now lists what's actually in the model and offers a close-match suggestion:

```
$ winml perf -m microsoft/resnet-50 --module ResNetStag
Error: No modules matching 'ResNetStag' found.
Did you mean: ResNetStage?
Available module class names in this model:
  BatchNorm2d
  Conv2d
  Linear
  ResNetBottleNeckLayer
  ResNetStage
  ...
```

Implementation: `_find_submodules_by_class` now raises a new `SubmoduleClassNotFoundError(class_name, available_classes)` carrying the executed class names discovered by torchinfo. `_perf_modules` catches it and renders a `click.UsageError` with `difflib.get_close_matches` suggestions plus the full sorted list.

## Out of scope

The CLIP forward-signature criterion of #195 (`CLIPEncoderLayer.forward()` requires `attention_mask` and `causal_attention_mask` kwargs that the random input generator doesn't supply) is a separate fix and tracked there.

## Tests

- `test_module_with_onnx_path_rejected` — guards #553 regression.
- `test_module_flag_in_help` — guards `--module` help-text presence.
- `test_module_no_match_lists_available_classes` — CLI-level: `SubmoduleClassNotFoundError` surfaces the available class names plus a `Did you mean…?` line.
- `TestFindSubmodulesByClass.test_no_match_raises_with_available_classes` — build-layer: the exception carries the right class names, sorted.

All affected suites pass.

### DIFF
--- a/tests/unit/config/test_build.py
+++ b/tests/unit/config/test_build.py
@@ -861,12 +861,12 @@ class TestBuildSubmoduleConfig:
 
 
 # =============================================================================
-# TestFindSubmodulesByClass - exercises the signature-fallback branch
+# TestFindSubmodulesByClass - signature-fallback and no-match paths
 # =============================================================================
 
 
 class TestFindSubmodulesByClass:
-    """Tests for _find_submodules_by_class signature-fallback branch."""
+    """Tests for _find_submodules_by_class branches."""
 
     def test_signature_fallback_when_hook_data_empty(self) -> None:
         """Empty hook_data triggers inspect.signature fallback for input_names."""
@@ -907,15 +907,6 @@ class TestFindSubmodulesByClass:
 
         assert len(results) == 1
         assert results[0].input_names == ["hidden_state"]
-
-
-# =============================================================================
-# TestFindSubmodulesByClass - no-match path surfaces available class names
-# =============================================================================
-
-
-class TestFindSubmodulesByClass:
-    """Tests for _find_submodules_by_class error reporting."""
 
     def test_no_match_raises_with_available_classes(self) -> None:
         """Wrong class name raises SubmoduleClassNotFoundError listing real classes."""
--- a/tests/unit/config/test_build.py
+++ b/tests/unit/config/test_build.py
@@ -24,12 +24,12 @@ import winml.modelkit.models  # noqa: F401
 from winml.modelkit.commands.config import config as config_command
 from winml.modelkit.compiler import EPConfig, WinMLCompileConfig
 from winml.modelkit.config import (
+    SubmoduleClassNotFoundError,
     WinMLBuildConfig,
     generate_build_config,
     generate_onnx_build_config,
 )
 from winml.modelkit.config.build import (
-    SubmoduleClassNotFoundError,
     SubmoduleInfo,
     _build_submodule_config,
     resolve_quant_compile_config,
--- a/tests/unit/config/test_build.py
+++ b/tests/unit/config/test_build.py
@@ -29,6 +29,7 @@ from winml.modelkit.config import (
     generate_onnx_build_config,
 )
 from winml.modelkit.config.build import (
+    SubmoduleClassNotFoundError,
     SubmoduleInfo,
     _build_submodule_config,
     resolve_quant_compile_config,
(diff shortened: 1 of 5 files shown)


xieofxie:
```
winml perf -m openai/clip-vit-base-patch32 --module CLIPEncoderLayer2 --device cpu --ep cpu
Generating module configs for CLIPEncoderLayer2...
Usage: winml perf [OPTIONS]
Try 'winml perf --help' for help.

Error: No modules matching 'CLIPEncoderLayer2' found.
Did you mean: CLIPEncoderLayer, CLIPEncoder?
Available module class names in this model:
  CLIPAttention
  CLIPEncoder
  CLIPEncoderLayer
  CLIPMLP
  CLIPTextEmbeddings
  CLIPTextModelWithProjection
  CLIPTextTransformer
  Embedding
  LayerNorm
  Linear
  QuickGELUActivation
```

tests/unit/config/test_build.py
```diff
@@ -29,6 +29,7 @@
     generate_onnx_build_config,
 )
 from winml.modelkit.config.build import (
+    SubmoduleClassNotFoundError,
```
timenick: `SubmoduleClassNotFoundError` was added to `winml.modelkit.config.__init__.py`'s `__all__` in this PR — it's now public API. Per CLAUDE.md's import rules, test code shouldn't reach into internal submodules for symbols exported by `__init__.py` (the `_`-prefixed exception doesn't apply here).

Suggest hoisting it into the existing public import block:

```python
from winml.modelkit.config import (
    WinMLBuildConfig,
    SubmoduleClassNotFoundError,  # ← here
    generate_build_config,
    generate_onnx_build_config,
)
from winml.modelkit.config.build import (
    SubmoduleInfo,
    _build_submodule_config,
    resolve_quant_compile_config,
)
```

The sister test file in this PR (`tests/unit/commands/test_perf_module.py`) already imports it correctly via `from winml.modelkit.config import SubmoduleClassNotFoundError` — this would keep the two consistent.

🤖 Generated with Claude Code

xieofxie: Hoisted — `SubmoduleClassNotFoundError` now comes from the public `winml.modelkit.config` import in test_build.py, matching test_perf_module.py. 8e335750

timenick (approved):

timenick (approved):

PR #603, opened 2026-05-12
feat(build): optional -c/--config, auto field, autoconf UX (#477)

DingmaomaoBJTU:
## Summary

Port `winml build` UX improvements from #477:

- **Optional `-c/--config`**: omit it to auto-generate the build config from `-m <model>` directly (one-step workflow; `-m` becomes required when `-c` is omitted)
- **`WinMLBuildConfig.auto` field** (default `True`): the build pipeline sets it to `False` after the optimize + autoconf loop and saves it into `winml_build_config.json` — marks the config as "already resolved" so CI/CD pipelines or subsequent `winml build -c` runs skip the analyzer on reuse
- **Setup header** now shows `(autoconf on)` / `(autoconf off)` next to the config name
- **Summary section** prints the saved config path after the final artifact

## Changed files

| File | Change |
|------|--------|
| `src/winml/modelkit/config/build.py` | Add `auto: bool = True` to `WinMLBuildConfig`; `to_dict()` omits it when `True`, writes `"auto": false` when `False`; `generate_cache_key()` excludes it |
| `src/winml/modelkit/build/common.py` | `run_optimize_analyze_loop()` respects `auto=False` (skips autoconf) and sets `config.auto = False` after the loop |
| `src/winml/modelkit/commands/build.py` | `-c` made optional; auto-generate config path; pass `auto` to `print_setup`; pass config path to `print_final` |
| `src/winml/modelkit/utils/console.py` | `print_setup()` shows autoconf status; `print_final()` prints saved config path |
| `tests/cli/test_main.py` | `TestBuildCommandArgs`: arg-validation tests for optional `-c` |
| `tests/unit/config/test_build.py` | `TestWinMLBuildConfigAutoField`: roundtrip and serialization tests for `auto` field |


xieofxie:
tests need fixed

src/winml/modelkit/commands/build.py
```diff
@@ -234,15 +224,17 @@ def _build_modules(
     "--config",
     "config_file",
     type=click.Path(exists=True),
-    required=True,
-    help="WinMLBuildConfig JSON file (from winml config)",
+    required=False,
+    default=None,
+    help="WinMLBuildConfig JSON file. If omitted, config is auto-generated from -m.",
 )
 @click.option(
```
xieofxie, line 235: should we use cli_utils?

src/winml/modelkit/commands/build.py
```diff
@@ -399,12 +394,26 @@ def build(
...
+                config_file,
+                no_quant=no_quant,
+                no_compile=no_compile,
+            )
+        else:
+            if not model_id:
+                raise click.UsageError("-m/--model is required when -c is not provided.")
+            from ..config import generate_build_config
+
+            config_or_configs = generate_build_config(
```
xieofxie, line 415: interesting that we do not have a flag in this function but have to generate everything and set to None afterwards..

src/winml/modelkit/commands/build.py
```diff
@@ -437,7 +446,7 @@ def build(
             print_setup(
                 console,
                 model=model_id or "random-init",
-                config=Path(config_file).name,
+                config=Path(config_file).name if config_file else "(auto)",
```
xieofxie: make `(auto)` a const

src/winml/modelkit/config/build.py
```diff
@@ -165,16 +166,22 @@ def from_dict(cls, config_dict: dict) -> WinMLBuildConfig:
...
-            "quant": self.quant.to_dict() if self.quant is not None else None,
-            "compile": self.compile.to_dict() if self.compile is not None else None,
-        }
+        result: dict = {}
+        if not self.auto:
+            result["auto"] = False
+        result.update(
+            {
+                "export": self.export.to_dict() if self.export is not None else None,
+                "optim": self.optim.to_dict(),
```
xieofxie: should we also support non-optim?

src/winml/modelkit/cli.py
xieofxie, line 42: forget to merge main? https://github.com/microsoft/WinML-ModelKit/pull/572/changes

PR #606, opened 2026-05-13
fix(perf): write default JSON report under ~/.cache/winml/perf

### LINKED ISSUE #551: [winml perf] [P1] Writes results to CWD by default (50+ stranded `*_perf.json` files in repo root)
**Severity**: P1  
**Command**: `winml perf`

---
**Category**: Output File Safety — never write to CWD by default.

**Repro**:
```
cd c:\Users\zhenni\repos\wmk
uv run winml perf -m temp\cli-audit\resnet.onnx --iterations 5
Get-ChildItem *_perf.json | Measure-Object
```

**Actual**: After previous runs, `Get-ChildItem c:\Users\zhenni\repos\wmk -Filter *_perf.json` returns **50+** stranded `<modelname>_perf.json` files in the repo root.

**Expected**: Either no file written unless `-o` given, or default to `~/.cache/winml/perf/<timestamp>/<modelname>.json` and tell the user where it went on success.

**Why it matters**: Pollutes the user's working directory (or their git repo). Easy to commit a perf file by accident. Easy for the user to lose track of which file is from which run.

xieofxie:
## Summary

`winml perf` used to write `*_perf.json` to CWD when `-o` was not given, leaving 50+ stranded files in the user's working directory (often a git repo). Default output is now:

```
~/.cache/winml/perf/<model_slug>/<timestamp>.json
```

With `--module CLASSNAME`, the module name nests one level deeper:

```
~/.cache/winml/perf/<model_slug>/<module_class>/<timestamp>.json
```

The CLI still prints `Results saved to: <path>` so the file is easy to find.

## Why this layout

It mirrors the layout already in use for cached build artifacts, e.g.:

```
C:\Users\<user>\.cache\winml\artifacts\BAAI_bge-base-en-v1.5\...
```

So both sides of the pipeline keep per-model outputs grouped under a stable user-level cache, instead of one side polluting CWD and the other living in `~/.cache`.

## Changes

- `generate_output_path(model_id, *, module_class=None)` returns the cache-dir path with a `YYYYMMDD-HHMMSS` timestamp.
- `_perf_modules` drops its inline `{slug}_{module_class}_perf.json` and calls `generate_output_path(..., module_class=...)` — single-model and per-module paths now share one default-location policy.
- `--output` help text shows the resolved native path on the current host (on Windows: `C:\Users\<user>\.cache\winml\perf`), instead of a POSIX `~/` placeholder.

## Tests

- `TestPerfOutputPath` in `tests/unit/commands/test_perf_cli.py` updated: asserts parent dir is `~/.cache/winml/perf/<slug>`, filename matches `\d{8}-\d{6}\.json`. Adds `test_module_class_adds_subdir` and `test_path_is_under_user_home`.
- `test_module_default_output_includes_class_name` in `tests/unit/commands/test_perf_module.py` updated for the new nested layout.

All 28 affected tests pass; ruff clean.

Closes #551

### DIFF
--- a/src/winml/modelkit/commands/perf.py
+++ b/src/winml/modelkit/commands/perf.py
@@ -723,8 +723,7 @@ def _perf_modules(
 
     # Write JSON report
     if output is None:
-        slug = hf_model.replace("/", "_").replace("\\", "_")
-        output = Path(f"{slug}_{module_class}_perf.json")
+        output = generate_output_path(hf_model, module_class=module_class)
 
     module_report = {
         "model_id": hf_model,
@@ -838,18 +837,28 @@ def write_json_report(result: BenchmarkResult, output_path: Path) -> None:
         json.dump(result.to_dict(), f, indent=2)
 
 
-def generate_output_path(model_id: str) -> Path:
-    """Generate default output path from model ID.
+def generate_output_path(model_id: str, *, module_class: str | None = None) -> Path:
+    r"""Generate default output path under the user's cache directory.
 
-    For ONNX files: uses the file stem (e.g., "model.onnx" -> "model_perf.json").
-    For HF model IDs: slugifies org/name
-    (e.g., "microsoft/resnet-50" -> "microsoft_resnet-50_perf.json").
+    Returns ``~/.cache/winml/perf/<slug>[/<module_class>]/<timestamp>.json``
+    so repeated runs accumulate under a stable per-model directory without
+    polluting CWD (see #551). The timestamp is generated at call time using
+    local time, format ``YYYYMMDD-HHMMSS``.
+
+    For ONNX inputs, the file stem is used as the slug
+    (e.g., ``model.onnx`` -> ``model``). For HF model IDs, ``/`` and ``\``
+    are replaced with ``_`` (e.g., ``microsoft/resnet-50`` ->
+    ``microsoft_resnet-50``).
     """
     p = Path(model_id)
-    if p.suffix.lower() == ".onnx":
-        return Path(f"{p.stem}_perf.json")
-    slug = model_id.replace("/", "_").replace("\\", "_")
-    return Path(f"{slug}_perf.json")
+    slug = p.stem if p.suffix.lower() == ".onnx" else model_id.replace("/", "_").replace("\\", "_")
+
+    out_dir = Path.home() / ".cache" / "winml" / "perf" / slug
+    if module_class:
+        out_dir = out_dir / module_class
+
+    timestamp = datetime.now().strftime("%Y%m%d-%H%M%S")
+    return out_dir / f"{timestamp}.json"
 
 
 # =============================================================================
@@ -1099,7 +1108,11 @@ def _run_onnx_benchmark(
     "-o",
     type=click.Path(path_type=Path),
     default=None,
-    help="Output JSON file path. Defaults to '{model_slug}_perf.json'",
+    help=(
+        f"Output JSON file path. Defaults to a timestamped file under "
+        f"'{Path.home() / '.cache' / 'winml' / 'perf'}', "
+        f"namespaced by model slug (and module class when --module is set)."
+    ),
 )
 @click.option(
     "--batch-size",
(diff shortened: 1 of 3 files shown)


tests/unit/commands/test_perf_module.py
```diff
@@ -62,21 +62,17 @@ def test_module_no_match_exits_nonzero(self) -> None:
...
-        assert "bert-base-uncased" in str(path)
-
-        # The module-mode output path (from _perf_modules) is
-        # {slug}_{module_class}_perf.json -- tested via the inline logic
-        # in _perf_modules. Here we verify the format difference:
-        from pathlib import Path
-
-        slug = "bert-base-uncased"
-        module_class = "BertAttention"
-        module_path = Path(f"{slug}_{module_class}_perf.json")
```
xieofxie, line 77: pervious test is totally AI

src/winml/modelkit/commands/perf.py
```diff
@@ -1099,7 +1108,11 @@ def _run_onnx_benchmark(
     "-o",
     type=click.Path(path_type=Path),
     default=None,
-    help="Output JSON file path. Defaults to '{model_slug}_perf.json'",
+    help=(
+        f"Output JSON file path. Defaults to a timestamped file under "
+        f"'{Path.home() / '.cache' / 'winml' / 'perf'}', "
```
timenick: The `help=` f-string interpolates `Path.home()` at decorator-evaluation time (module import), so the rendered help text contains the *current process user's* home directory — e.g. `'C:\Users\<builder>\.cache\winml\perf'`. A couple of side effects worth considering:

- `winml perf --help` shows different text per user
- Anything that captures `--help` output for docs / golden tests embeds the doc builder's username
- Snapshot tests over help text become host-specific

The PR description treats this as intentional vs. a `~/` placeholder, but a literal `'~/.cache/winml/perf'` in the help string is still informative and stays user-agnostic. Worth a quick sanity check on which trade-off you prefer.

🤖 Generated with Claude Code

tests/unit/commands/test_perf_cli.py
```diff
@@ -100,36 +100,75 @@ def test_valid_device_choices(self, runner: CliRunner, device: str) -> None:
...
+
+    _TIMESTAMP_RE = r"^\d{8}-\d{6}\.json$"
+
+    def _cache_root(self):
+        from pathlib import Path
+
+        return Path.home() / ".cache" / "winml" / "perf"
 
     def test_hf_model_path(self) -> None:
+        import re
```
timenick: `from winml.modelkit.commands.perf import generate_output_path`, `import re`, and `from pathlib import Path` are repeated inside each new test method and inside the `_cache_root` helper. The file already imports siblings from the same module at the top (line 19: `from winml.modelkit.commands.perf import BenchmarkConfig, PerfBenchmark, perf`).

Suggest hoisting `generate_output_path`, `re`, and `Path` to the module-level imports — keeps the test methods focused on assertions and matches the file's existing style.

🤖 Generated with Claude Code

src/winml/modelkit/commands/perf.py
xieofxie: Good catch — reverted to a literal `'~/.cache/winml/perf/<model_slug>[/<module_class>]/<timestamp>.json'` template. The single `[/...]` segment also lets one line cover both single-model and per-module layouts without the parenthetical. 6c288889

tests/unit/commands/test_perf_cli.py
xieofxie: Hoisted `generate_output_path`, `re`, and `Path` to the module-level imports; dropped the `TYPE_CHECKING`-only `Path` (it was only there for fixture annotations under `from __future__ import annotations`, which are lazily evaluated). The new tests now just contain assertions. 6c288889

tests/unit/commands/test_perf_module.py
xieofxie: Fair — the deleted version was constructing a Path locally just to assert string membership, which didn't exercise the real code path. The replacement calls `generate_output_path(model_id, module_class=...)` so the test actually checks the nesting behavior produced by the function under test.

timenick (approved):

PR #607, opened 2026-05-13
test(optimize): add e2e tests + show (Default: enabled/disabled) in --help

### LINKED ISSUE #500: Identify functional scenarios and implement E2E tests for `winml optimize`
### Description
Identify the functionality list for `winml optimize` and implement E2E pytest cases covering it. Tests live in [tests/e2e](../tree/main/tests/e2e).

Coverage must include:

### Deliverables

### Acceptance Criteria
- `uv run pytest tests/e2e/ -k optimize` is green.

### LINKED ISSUE #501: Feature-owner self-check for `winml optimize`
### Description
Walk through the self-check below for `winml optimize` and record the result. These items cover defects that automated audits cannot detect.

For each item: check it with linked evidence (log excerpt, test run, commit SHA), or mark **N/A** with a one-line reason.

### Checklist
- [ ] Validated the functionality of the command matches all defined behaviors in the PRD / design spec. Link the spec. Validated both good path and bad path across all valid option combinations.
- [ ] No flag is silently ignored or silently changed during execution.
- [ ] Validated functional correctness across all supported EPs.
- [ ] In a fresh Python venv with the package newly installed, the command runs successfully. Attach terminal log.

### Acceptance Criteria
All items above are checked with evidence, or explicitly marked **N/A** with a reason. Every defect surfaced is filed as a separate issue and linked here.

vortex-captain:
Fixes #500. Fixes #501.

## Summary

Adds end-to-end coverage for `winml optimize` in `tests/e2e/test_optimize_e2e.py` (14 cases). Uses synthetic ONNX models built in-test via `onnx.helper` — no network, no HuggingFace download.

Also surfaces each capability's default value in `winml optimize --help` by appending `(Default: enabled)` / `(Default: disabled)` to bool-capability help text (and `(Default: <value>)` for int / choice capabilities, none of which are currently registered). The default is read from `cap.default` on each registered capability — no hardcoded list.

## Test cases

Each goal reduces to "the option(s) under test are *effective*"; for multi-option cases the goal also fixes the expected precedence. Success cases require exit 0 + the optimized ONNX file at the expected path; failure cases require non-zero exit + no ONNX written.

| Test | CLI | Goal |
|---|---|---|
| `TestOptimizeDoc::test_help` | `winml optimize --help` | `--help` is effective: help renders and the new `(Default: enabled)` / `(Default: disabled)` annotations are both present. |
| `TestOptimizeDoc::test_list_capabilities` | `winml optimize --list-capabilities` | `--list-capabilities` is effective: header `"Available optimization flags"` plus the lone default-enabled cap (`--disable-constant-folding`) appears. |
| `TestOptimizeDoc::test_list_rewrites` | `winml optimize --list-rewrites` | `--list-rewrites` is effective: rewrite-capabilities listing renders. |
| `TestOptimizeConstFolding::test_minimal` | `winml optimize -m cst_folding.onnx` | Minimal required args are effective; `constant-folding` is on by default, so the `ConstantOfShape` node is folded away. |
| `TestOptimizeConstFolding::test_missing_model_fails` | `winml optimize` | `-m` is genuinely required: omitting it fails cleanly and no `*.onnx` is leaked. |
| `TestOptimizeConstFolding::test_custom_output` | `... -o explicit.onnx` | `-o` is effective: output is at the custom path AND the default `{stem}_opt.onnx` is **not** also produced. |
| `TestOptimizeConstFolding::test_verbose` | `... -v` | `-v` is effective and does not break the optimize pipeline. |
| `TestOptimizeConstFolding::test_enable_constant_folding` | `... --enable-constant-folding` | Explicit enable is effective: `ConstantOfShape` is folded away. |
| `TestOptimizeConstFolding::test_disable_constant_folding` | `... --disable-constant-folding` | Explicit disable is effective: `ConstantOfShape` survives. |
| `TestOptimizeConstFolding::test_config_disables_folding` | `... --config {"constant-folding": false}` | `--config` is effective: a config file can disable a default-enabled capability. |
| `TestOptimizeConstFolding::test_cli_overrides_config` | `... --config {"constant-folding": false} --enable-constant-folding` | Both args effective with **CLI > config** precedence: `--enable-*` overrides the config file's `false`. |
| `TestOptimizeGeluFusion::test_minimal` | `winml optimize -m expanded_gelu.onnx` | Minimal required args are effective; `gelu-fusion` is off by default, so the expanded GELU pattern (`Div → Erf → Add → Mul → Mul`) is preserved. |
| `TestOptimizeGeluFusion::test_enable_gelu_fusion` | `... --enable-gelu-fusion` | Explicit enable is effective: the expanded pattern collapses into a single `Gelu` op (no `Erf`). |
| `TestOptimizeGeluFusion::test_disable_gelu_fusion` | `... --disable-gelu-fusion` | Explicit disable is effective (matches the default): pattern preserved, `Erf` survives. |

## Help-text change

`commands/optimize.py::capability_options` formerly used `help=cap.description` for every capability flag, which produced help entries like:

```
--enable-constant-folding / --disable-constant-folding
                                Fold subgraphs whose inputs are all constants
```

Without a default annotation, a user reading `--help` couldn't tell which capabilities are already on. The change derives a human-readable suffix from `cap.default` and appends it to the description:

```
--enable-constant-folding / --disable-constant-folding
                                Fold subgraphs whose inputs are all constants
                                (Default: enabled)
```

For `BoolCapability` the suffix is `(Default: enabled)` or `(Default: disabled)`; for `IntCapability` and `ChoiceCapability` it's `(Default: <value>)`. No new state is introduced — the same `cap.default` is what the command already uses downstream when seeding `final_config` (see `commands/optimize.py:369-370`).

`TestOptimizeDoc::test_help` asserts that both `Default: enabled` and `Default: disabled` substrings appear in the rendered help, so this docs change is locked down by the new e2e suite.

### DIFF
--- a/tests/e2e/test_optimize_e2e.py
+++ b/tests/e2e/test_optimize_e2e.py
@@ -1,0 +1,307 @@
+# -------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+# --------------------------------------------------------------------------
+"""E2E tests for the ``winml optimize`` CLI command.
+
+These tests use synthetic ONNX models built in-test via ``onnx.helper`` — no
+network, no HuggingFace download.
+
+Success criteria for any successful invocation:
+    * Command exits with code 0.
+    * The requested (or default) ONNX file exists at the expected path.
+    * Test-specific extra invariants (per-case).
+
+Failure criteria for any failing invocation:
+    * Command exits with a non-zero code.
+    * No ONNX file is left where success would have placed it.
+
+Markers:
+    e2e: End-to-end test invoking the CLI through CliRunner.
+"""
+
+from __future__ import annotations
+
+import json
+import math
+from typing import TYPE_CHECKING
+
+import numpy as np
+import onnx
+import pytest
+from click.testing import CliRunner
+from onnx import TensorProto, helper, numpy_helper
+
+from winml.modelkit.commands.optimize import optimize
+
+
+if TYPE_CHECKING:
+    from pathlib import Path
+
+
+pytestmark = [pytest.mark.e2e, pytest.mark.timeout(120)]
+
+
+# ---------------------------------------------------------------------------
+# CLI helpers (DRY)
+# ---------------------------------------------------------------------------
+
+
+def _invoke(args: list[str], *, catch: bool = False):
+    """Invoke the optimize CLI with a fresh runner and standard ctx.obj."""
+    runner = CliRunner()
+    return runner.invoke(optimize, args, obj={"debug": False}, catch_exceptions=catch)
+
+
+def _assert_succeeds(args: list[str], output_path: Path) -> onnx.ModelProto:
+    """Assert exit==0, output ONNX exists, return the loaded optimized model."""
+    result = _invoke(args)
+    assert result.exit_code == 0, f"optimize failed (exit {result.exit_code}):\n{result.output}"
+    assert output_path.exists(), f"optimized ONNX not found at {output_path}"
+    return onnx.load(str(output_path))
+
+
+def _count_op(model: onnx.ModelProto, op_type: str) -> int:
+    return sum(1 for n in model.graph.node if n.op_type == op_type)
+
+
+def _ops(model: onnx.ModelProto) -> list[str]:
+    return [n.op_type for n in model.graph.node]
+
+
+def _default_opt_path(input_path: Path) -> Path:
+    """The path the CLI writes to when ``-o`` is omitted."""
+    return input_path.parent / f"{input_path.stem}_opt.onnx"
+
+
+def _save(model: onnx.ModelProto, path: Path) -> Path:
+    onnx.save(model, str(path))
+    return path
+
+
+def _write_json(path: Path, payload: dict) -> Path:
+    path.write_text(json.dumps(payload), encoding="utf-8")
+    return path
+
+
+# ---------------------------------------------------------------------------
+# Synthetic ONNX builders + fixtures
+# ---------------------------------------------------------------------------
+
+
+def _build_const_folding_model() -> onnx.ModelProto:
+    """``out = x + ConstantOfShape(shape=[2,3], value=7.0)``.
+
+    Per the ONNX spec, ConstantOfShape takes a single 1-D int64 input whose
+    *values* are the desired output dimensions, and a ``value`` attribute
+    (a length-1 1-D or 0-D tensor) holding the scalar fill value. Here we
+    ask for a 2x3 float32 tensor filled with 7.0. The shape input is a
+    constant initializer, so constant-folding collapses the node into a
+    literal initializer.
+    """
+    target_shape = [2, 3]
+    fill_value = 7.0
+
+    x_info = helper.make_tensor_value_info("x", TensorProto.FLOAT, target_shape)
+    out_info = helper.make_tensor_value_info("out", TensorProto.FLOAT, target_shape)
+
+    # 1-D int64 tensor whose values describe the desired output shape.
+    shape_init = helper.make_tensor(
+        "shape_init",
+        TensorProto.INT64,
+        [len(target_shape)],
+        target_shape,
+    )
+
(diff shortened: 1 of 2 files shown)


tests/e2e/test_optimize_e2e.py
```diff
@@ -0,0 +1,307 @@
...
+
+
+# ---------------------------------------------------------------------------
+# CLI helpers (DRY)
+# ---------------------------------------------------------------------------
+
+
+def _invoke(args: list[str], *, catch: bool = False):
+    """Invoke the optimize CLI with a fresh runner and standard ctx.obj."""
+    runner = CliRunner()
```
timenick, line 51: `tests/e2e/conftest.py:133-137` already provides a shared `runner` fixture for the e2e suite (and its docstring at conftest.py:11 lists `runner` as one of the shared fixtures). Per CLAUDE.md test guidance, new tests should reuse the shared fixture rather than duplicate it.

Two ways to align:

```python
# Option A — accept the fixture explicitly
def _invoke(runner, args, *, catch=False):
    return runner.invoke(optimize, args, obj={"debug": False}, catch_exceptions=catch)

def test_help(self, runner):
    result = _invoke(runner, ["--help"])
```

```python
# Option B — drop the helper, inline
def test_help(self, runner):
    result = runner.invoke(optimize, ["--help"], obj={"debug": False})
```

Either avoids creating a fresh `CliRunner` per call and matches the rest of the e2e suite.

🤖 Generated with Claude Code

tests/e2e/test_optimize_e2e.py
```diff
@@ -0,0 +1,307 @@
...
+            f"expected ConstantOfShape to be folded by default; ops: {_ops(model)}"
+        )
+
+    def test_missing_model_fails(self, tmp_path: Path):
+        # No -m supplied: command must reject and write no ONNX anywhere.
+        result = _invoke([], catch=True)
+        assert result.exit_code != 0, (
+            f"expected failure for missing -m, got exit=0:\n{result.output}"
+        )
+        leftover = list(tmp_path.glob("*.onnx"))
```
timenick: This check is a no-op: `tmp_path` is a fresh, empty fixture-managed dir, and `winml optimize` (invoked here with no `-m`) has no way to discover it. The glob will always return `[]` regardless of CLI behavior, so the assertion never actually verifies the "no ONNX leaked" intent.

Two options:
- Drop the leftover check — the `exit_code != 0` assertion already covers the user-visible behavior.
- Or actually exercise the leak path with `monkeypatch.chdir(tmp_path)` before invoking, so the CLI's CWD == the dir you're inspecting:

  ```python
  def test_missing_model_fails(self, tmp_path, monkeypatch):
      monkeypatch.chdir(tmp_path)
      result = _invoke([], catch=True)
      assert result.exit_code != 0
      assert not list(tmp_path.glob("*.onnx"))
  ```

The second is what the comment seems to intend ("no `*.onnx` is leaked"). Minor — feel free to skip if you'd rather keep the test simple.

🤖 Generated with Claude Code

tests/e2e/test_optimize_e2e.py
vortex-captain: Existing test files in `e2e/` are following pattern `runner = CliRunner()` instead of using the fixture, we should keep the convention here.

tests/e2e/test_optimize_e2e.py
vortex-captain: changed to ` assert not _default_opt_path(cst_folding_path).exists()` for clarity

tests/e2e/test_optimize_e2e.py
timenick: You’re correct. If the fixture does nothing, shall we delete it or we enforce to use it?  Any thoughts?

vortex-captain: It seems ok to either use a shared runner or a new one in every test case, and our e2e codebase currently has both (test_run_e2e.py uses the shared one). There might be potential issues with the shared runner if we plan to run the cases in parallel in the future though.

timenick (approved):

Issue #563, opened 2026-05-09
[cross-cutting] [P2] `--model` semantic drift across 10 commands


zhenchaoni:
**Severity**: P2  
**Command**: `cross-cutting`

---
**Category**: Cross-Command Flag Consistency — same flag should mean the same thing everywhere.

**Repro**:
```
uv run winml analyze --help | Select-String '\-\-model'
uv run winml build --help   | Select-String '\-\-model'
uv run winml compile --help | Select-String '\-\-model'
uv run winml config --help  | Select-String '\-\-model'
uv run winml eval --help    | Select-String '\-\-model|\-\-model-id'
uv run winml export --help  | Select-String '\-\-model'
uv run winml hub --help     | Select-String '\-\-model'
uv run winml inspect --help | Select-String '\-\-model'
uv run winml optimize --help| Select-String '\-\-model'
uv run winml perf --help    | Select-String '\-\-model'
uv run winml quantize --help| Select-String '\-\-model'
```

**Actual**:
- 4 different Python types (`str`, `Path`, `Tuple[str, ...]`, `List[str]`).
- 3 different parameter names: `model`, `model_id`, `hf_model`.
- Inconsistent acceptance: some accept HF ids only, some accept `.onnx` only, some accept either, some accept directories of HF snapshots.
- `eval` uniquely uses `multiple=True` plus a separate `--model-id` flag.

**Expected** — propose this canonical pattern:

```
-m, --model VALUE         # required for commands that need a model; lowercase, singular
```

- Accepts three concrete kinds, auto-detected at entry:
  1. **HF id** — matches `^[A-Za-z0-9._-]+/[A-Za-z0-9._-]+$`. Fetch from Hub.
  2. **Local file** — exists on disk and ends in `.onnx` / `.pt` / `.safetensors`.
  3. **Local directory** — exists on disk and contains `config.json` (HF snapshot layout).
- One internal Python type `ModelRef` with `kind: Literal["hf", "file", "dir"]` discriminator.
- One shared decorator `@model_option(required=True, multiple=False)` applied to every command.
- Three concrete error messages
- `eval`'s multi-role syntax (`-m generator=foo -m discriminator=bar`) becomes a separate decorator `@multi_role_model_option`, used only by commands that need it.

**Why it matters**: Users learn one flag and expect it to behave the same way everywhere. Type drift means automation has to special-case every command.

DingmaomaoBJTU:
**Triage summary**

- **Component**: Other
[...]

PR #608, opened 2026-05-13
Quantize: Implement functional e2e test cases and fix issues found during test

### LINKED ISSUE #504: Identify functional scenarios and implement E2E tests for `winml quantize`
### Description
Identify the functionality list for `winml quantize` and implement E2E pytest cases covering it. Tests live in [tests/e2e](../tree/main/tests/e2e).

Coverage must include:

### Deliverables

### Acceptance Criteria
- `uv run pytest tests/e2e/ -k quantize` is green.

zhenchaoni:
Fixes #504

### Issue 1
When user specify `--precision int16` and config.json but no `weight_type` or`activation_type`, the quantize is using uint8 as the weight and activation.

This is because the config resolution logic uses the default value of the data classes instead of the user specified value in CLI. This issue is fixed by only loading the user specified values in the config json file.

### Issue 2
The `winml quantize` command doesn't take model id, causing the calibration always using RandomDataset. The fix is to add the option.

This doesn't impact `build` command, because build command calls the internal `quantize_onnx` python method directly which has model id parameter.

### Issue 3
When user provides a non-exist output directory, the command fails. The fix is to create that directory.

**Add e2e test**
---

## Precision routing

| Command | Input | Assertion |
|---|---|---|
| `winml quantize -m <tiny> --samples 4` | tiny FP32 | weight zp dtype = `UINT8`; input metadata survives |
| `winml quantize -m <tiny> --precision int8 --samples 4` | tiny | weight zp dtype = `UINT8` (preset → uint8/uint8) |
| `winml quantize -m <tiny> --precision int16 --samples 4` | tiny | weight zp dtype = `INT16` (skip ORT inference) |
| `winml quantize -m <tiny> --precision w8a16 --samples 4` | tiny | activation zp dtype = `UINT16` (skip ORT inference) |
| `winml quantize -m <tiny> --precision int8 --weight-type int8 --activation-type uint8 --samples 4` | tiny | weight zp dtype = `INT8` (explicit beats preset) |
| `winml quantize -m <tiny> --precision fp16 --samples 4` | tiny | weight zp dtype = `UINT8` (documented silent fallback) |

## Calibration method

| Command | Input | Assertion |
|---|---|---|
| `winml quantize -m <tiny> --method minmax --samples 4` | tiny | shared structural checks |
| `winml quantize -m <tiny> --method entropy --samples 4` | tiny | shared structural checks |
| `winml quantize -m <tiny> --method percentile --samples 4` | tiny | shared structural checks |

## Quant options

| Command | Input | Assertion |
|---|---|---|
| `winml quantize -m <tiny> --per-channel --samples 4` | tiny | weight DQ scale init has > 1 element (per-channel vector) |
| `winml quantize -m <tiny> --symmetric --weight-type int8 --samples 4` | tiny | every weight zp value == 0 |

## Per-task calibration datasets (one per `TASK_DATASET_MAPPING` class)

Each row constructs the real `DatasetCalibrationReader` instance for the given task using the
HuggingFace preprocessor of the supplied model.

| Command | Input | Assertion |
|---|---|---|
| `winml quantize -m <tiny> --task random --samples 4` | tiny ONNX (no model_id) | shared structural checks — `RandomDataset` path |
| `winml quantize -m <onnx_imgcls> --task image-classification --model-id microsoft/resnet-50 --samples 4` | exported ResNet-50 ONNX + ResNet `AutoImageProcessor` | shared structural checks — `ImageDataset` path |
| `winml quantize -m <onnx_txtcls> --task text-classification --model-id Intel/bert-base-uncased-mrpc --samples 4` | exported BERT-MRPC ONNX + BERT `AutoTokenizer` | shared structural checks — `TextDataset` path |
| `winml quantize -m <onnx_objdet> --task object-detection --model-id hustvl/yolos-small --samples 4` | exported YOLOS ONNX + YOLOS image processor | shared structural checks — `ObjectDetectionDataset` path |
| `winml quantize -m <onnx_imgseg> --task image-segmentation --model-id nvidia/segformer-b0-finetuned-ade-512-512 --samples 4` | exported SegFormer ONNX + SegFormer image processor | shared structural checks — `ImageSegmentationDataset` path |
| `winml quantize -m <tiny> --task automatic-speech-recognition --samples 4` | tiny ONNX, unsupported task | CLI output contains `falling back to RandomDataset` |

## Output behavior

| Command | Input | Assertion |
|---|---|---|
| `winml quantize -m <tiny> --samples 4` | tiny | output exists at `<tiny.parent>/<tiny.stem>_qdq.onnx` |
| `winml quantize -m <tiny> -o <tmp>/out/custom.onnx --samples 4` | tiny | file exists at exact `-o` path |
| `winml quantize -m <tiny> -o <tmp>/missing/nested/custom.onnx --samples 4` | tiny, missing parent dir | command auto-creates parent dirs; file exists (regression: previously crashed with `FileNotFoundError` from `os.chdir`) |
| `winml quantize -m <tiny_ext>.onnx -o <tmp>/out/quant_ext.onnx --samples 4` | tiny + external data sidecar | both `quant_ext.onnx` and `quant_ext.onnx.data` exist |

## Build-config precedence (CLI vs config file)

| Command | Input | Assertion |
|---|---|---|
| `winml quantize -m <tiny> --config bc.json --samples 4` | tiny + `bc.json={"quant":{"samples":50,"calibration_method":"entropy"}}` | stdout shows `Samples: 4` (CLI wins) and `Method: entropy` (config used) |
| `winml quantize -m <tiny> --config bc.json --precision int16 --samples 4` | tiny + `bc.json={"quant":{}}` | weight zp dtype = `INT16` (regression: explicit `--precision` beats empty config) |

## Build-config key absorption sweep

For each `quant.*` key, assert it is consumed when CLI omits it. Verified by structural
inspection of the produced model (not stdout).

| Command | Input | Assertion |
|---|---|---|
| `winml quantize -m <tiny> --config bc.json --samples 4` | `bc.json={"quant":{"weight_type":"int8"}}` | weight DQ zp dtype = `INT8` |
| `winml quantize -m <tiny> --config bc.json --samples 4` | `bc.json={"quant":{"per_channel":true}}` | weight DQ scale init has > 1 element |
| `winml quantize -m <tiny> --config bc.json --samples 4` | `bc.json={"quant":{"symmetric":true,"weight_type":"int8"}}` | every weight zp value == 0 |
| `winml quantize -m <tiny> --config bc.json --samples 4` | `bc.json={"quant":{"task":"automatic-speech-recognition"}}` | CLI output contains `falling back to RandomDataset` (config task flowed to dataset selection) |

## Verbose

| Command | Input | Assertion |
|---|---|---|
| `winml quantize -m <tiny> -o <out> --samples 4 -v` vs same without `-v` | tiny | verbose output strictly longer than default |

## Errors

| Command | Input | Assertion |
|---|---|---|
| `winml quantize` | — | exit ≠ 0; stderr matches `Missing option .*--model` |
| `winml quantize -m <tmp>/nope.onnx` | non-existent path | exit ≠ 0; stderr matches `does not exist` |
| `winml quantize -m <tiny> --method gaussian` | tiny | exit ≠ 0; stderr matches `Invalid value for '--method'` |
| `winml quantize -m <tiny> --weight-type float8` | tiny | exit ≠ 0; stderr matches `Invalid value for '--weight-type'` |
| `winml quantize -m <bad>.onnx --samples 4` | random bytes | exit ≠ 0; stderr contains `Quantization failed` AND a parse-related substring (parse/protobuf/decode/load/invalid) |

---

**Total: 6 + 3 + 2 + 6 + 4 + 2 + 4 + 1 + 5 = 33 cases.** All passing.

### DIFF
--- a/src/winml/modelkit/commands/quantize.py
+++ b/src/winml/modelkit/commands/quantize.py
@@ -19,6 +19,7 @@ Examples:
 
 from __future__ import annotations
 
+import json
 import logging
 from pathlib import Path
 
@@ -99,6 +100,13 @@ console = Console()
     help="Task for calibration dataset selection (e.g., 'image-classification').",
 )
 @click.option(
+    "--model-id",
+    type=str,
+    default=None,
+    help="HuggingFace model ID (e.g., 'microsoft/resnet-50'). When provided "
+    "with --task, enables task-aware calibration datasets using the model's preprocessor.",
+)
+@click.option(
     "--verbose",
     "-v",
     is_flag=True,
@@ -119,6 +127,7 @@ def quantize(
     per_channel: bool,
     symmetric: bool,
     task: str | None,
+    model_id: str | None,
     verbose: bool,
     config_file: Path | None,
 ) -> None:
@@ -151,23 +160,26 @@ def quantize(
 
     configure_logging(verbose=verbose)
 
-    # Apply build config defaults (CLI explicit options take precedence)
+    # Apply build config defaults (CLI explicit options take precedence).
+    # Only read the JSON for what explicitly specified in config file.
     if config_file is not None:
-        build_cfg = cli_utils.load_build_config(config_file)
-        if build_cfg.quant:
-            qc = build_cfg.quant
-            if not cli_utils.is_cli_provided(ctx, "samples"):
-                samples = qc.samples
-            if not cli_utils.is_cli_provided(ctx, "method"):
-                method = qc.calibration_method
-            if not cli_utils.is_cli_provided(ctx, "weight_type"):
-                weight_type = qc.weight_type
-            if not cli_utils.is_cli_provided(ctx, "activation_type"):
-                activation_type = qc.activation_type
-            if not cli_utils.is_cli_provided(ctx, "per_channel"):
-                per_channel = qc.per_channel
-            if not cli_utils.is_cli_provided(ctx, "symmetric"):
-                symmetric = qc.symmetric
+        qc = (json.loads(config_file.read_text()) or {}).get("quant") or {}
+        if not cli_utils.is_cli_provided(ctx, "samples") and "samples" in qc:
+            samples = qc["samples"]
+        if not cli_utils.is_cli_provided(ctx, "method") and "calibration_method" in qc:
+            method = qc["calibration_method"]
+        if not cli_utils.is_cli_provided(ctx, "weight_type") and "weight_type" in qc:
+            weight_type = qc["weight_type"]
+        if not cli_utils.is_cli_provided(ctx, "activation_type") and "activation_type" in qc:
+            activation_type = qc["activation_type"]
+        if not cli_utils.is_cli_provided(ctx, "per_channel") and "per_channel" in qc:
+            per_channel = qc["per_channel"]
+        if not cli_utils.is_cli_provided(ctx, "symmetric") and "symmetric" in qc:
+            symmetric = qc["symmetric"]
+        if not cli_utils.is_cli_provided(ctx, "task") and "task" in qc:
+            task = qc["task"]
+        if not cli_utils.is_cli_provided(ctx, "model_id") and "model_name" in qc:
+            model_id = qc["model_name"]
 
     # Import quantizer (late import to speed up CLI)
     from ..quant import WinMLQuantizationConfig, quantize_onnx
@@ -200,6 +212,7 @@ def quantize(
         per_channel=per_channel,
         symmetric=symmetric,
         task=task,
+        model_name=model_id,
     )
 
     # Display dataset info from config
--- a/src/winml/modelkit/commands/quantize.py
+++ b/src/winml/modelkit/commands/quantize.py
@@ -100,10 +100,10 @@ console = Console()
     help="Task for calibration dataset selection (e.g., 'image-classification').",
 )
 @click.option(
-    "--model-id",
+    "--model-name",
     type=str,
     default=None,
-    help="HuggingFace model ID (e.g., 'microsoft/resnet-50'). When provided "
+    help="HuggingFace model name (e.g., 'microsoft/resnet-50'). When provided "
     "with --task, enables task-aware calibration datasets using the model's preprocessor.",
 )
 @click.option(
@@ -127,7 +127,7 @@ def quantize(
     per_channel: bool,
     symmetric: bool,
     task: str | None,
-    model_id: str | None,
+    model_name: str | None,
     verbose: bool,
     config_file: Path | None,
 ) -> None:
@@ -163,6 +163,7 @@ def quantize(
     # Apply build config defaults (CLI explicit options take precedence).
     # Only read the JSON for what explicitly specified in config file.
     if config_file is not None:
+        cli_utils.load_build_config(config_file)  # validate; raises click.UsageError
         qc = (json.loads(config_file.read_text()) or {}).get("quant") or {}
         if not cli_utils.is_cli_provided(ctx, "samples") and "samples" in qc:
             samples = qc["samples"]
@@ -178,8 +179,8 @@ def quantize(
             symmetric = qc["symmetric"]
         if not cli_utils.is_cli_provided(ctx, "task") and "task" in qc:
             task = qc["task"]
-        if not cli_utils.is_cli_provided(ctx, "model_id") and "model_name" in qc:
-            model_id = qc["model_name"]
+        if not cli_utils.is_cli_provided(ctx, "model_name") and "model_name" in qc:
+            model_name = qc["model_name"]
 
     # Import quantizer (late import to speed up CLI)
     from ..quant import WinMLQuantizationConfig, quantize_onnx
@@ -192,6 +193,7 @@ def quantize(
     # Determine output path
     if output is None:
         output = model.parent / f"{model.stem}_qdq.onnx"
+    output.parent.mkdir(parents=True, exist_ok=True)
 
     # Show info
     console.print(f"[bold blue]Input:[/bold blue] {model}")
@@ -212,7 +214,7 @@ def quantize(
         per_channel=per_channel,
         symmetric=symmetric,
         task=task,
-        model_name=model_id,
+        model_name=model_name,
     )
 
     # Display dataset info from config
--- a/src/winml/modelkit/quant/quantizer.py
+++ b/src/winml/modelkit/quant/quantizer.py
@@ -175,8 +175,6 @@ def quantize_onnx(
         # resolution.  output_path.parent is guaranteed to exist (caller mkdir).
         abs_model_input = str(Path(model_path).resolve())
         abs_model_output = str(Path(output_path).resolve())
-        # Ensure output parent exists; subsequent os.chdir requires it.
-        output_path.parent.mkdir(parents=True, exist_ok=True)
         # Remove stale output artifacts from a previous build.  ORT/onnx refuse
         # to overwrite an existing external-data sidecar (e.g. quantized.onnx.data),
         # raising FileExistsError, so we proactively clear them here.
--- a/src/winml/modelkit/quant/quantizer.py
+++ b/src/winml/modelkit/quant/quantizer.py
@@ -175,6 +175,8 @@ def quantize_onnx(
         # resolution.  output_path.parent is guaranteed to exist (caller mkdir).
         abs_model_input = str(Path(model_path).resolve())
         abs_model_output = str(Path(output_path).resolve())
+        # Ensure output parent exists; subsequent os.chdir requires it.
+        output_path.parent.mkdir(parents=True, exist_ok=True)
         # Remove stale output artifacts from a previous build.  ORT/onnx refuse
         # to overwrite an existing external-data sidecar (e.g. quantized.onnx.data),
         # raising FileExistsError, so we proactively clear them here.
--- a/tests/e2e/test_quantize_e2e.py
+++ b/tests/e2e/test_quantize_e2e.py
@@ -1,0 +1,825 @@
+# -------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+# --------------------------------------------------------------------------
+"""E2E functional tests for the `winml quantize` CLI command.
+
+Tests invoke ``winml quantize`` via ``CliRunner`` against a tiny hand-built
+FP32 ONNX, asserting exit code + structural properties of the output graph.
+
+Shared structural assertions live in ``_assert_quantized_output``.
+"""
+
+from __future__ import annotations
+
+import json
+from pathlib import Path
+from typing import TYPE_CHECKING
+
+import numpy as np
+import onnx
+import onnxruntime as ort
+import pytest
+from onnx import TensorProto, helper, numpy_helper
+
+from winml.modelkit.commands.quantize import quantize as quantize_cmd
+
+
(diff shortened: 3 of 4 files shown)


src/winml/modelkit/commands/quantize.py
```diff
@@ -98,6 +99,13 @@
     default=None,
     help="Task for calibration dataset selection (e.g., 'image-classification').",
 )
+@click.option(
```
xieofxie, line 102: dup part with https://github.com/microsoft/WinML-ModelKit/pull/599/changes#diff-c10bc310bdc414a9e62dca34cde81af47852d967cdf962f2ca3f887fa5b00318

src/winml/modelkit/commands/quantize.py
```diff
@@ -151,23 +160,26 @@ def quantize(
...
-                method = qc.calibration_method
-            if not cli_utils.is_cli_provided(ctx, "weight_type"):
-                weight_type = qc.weight_type
-            if not cli_utils.is_cli_provided(ctx, "activation_type"):
-                activation_type = qc.activation_type
-            if not cli_utils.is_cli_provided(ctx, "per_channel"):
-                per_channel = qc.per_channel
-            if not cli_utils.is_cli_provided(ctx, "symmetric"):
-                symmetric = qc.symmetric
+        qc = (json.loads(config_file.read_text()) or {}).get("quant") or {}
```
timenick, line 167: **Error-handling regression** — bypassing `cli_utils.load_build_config()` drops three error paths the helper provides:

1. Malformed JSON → was `click.UsageError("Invalid JSON in build config: …")`, now a raw `json.JSONDecodeError` traceback bubbles to the user.
2. Empty file → was `click.UsageError("Config file is empty: …")`, now `JSONDecodeError`.
3. Non-object top-level JSON (e.g. a list/string) → was `click.UsageError("Build config must be a JSON object, got …")`, now `AttributeError` on `.get("quant")`.

Additionally, `qc["samples"]` etc. assume `quant` is a dict; `{"quant": "x"}` would raise `TypeError: string indices must be integers` at `"samples" in qc` is fine but the subsequent lookups would still error in other configs.

Consider keeping `load_build_config()` for parse/validation and only sourcing the precedence check from the raw dict — e.g. add a helper that returns `(parsed_config, raw_quant_dict)` so error messages stay user-friendly.

🤖 Generated with Claude Code

timenick, line 167: **Same precedence bug exists in other commands** — the root cause is `WinMLQuantizationConfig.from_dict()` (and siblings) filling missing JSON keys with dataclass defaults, which the precedence block then can't distinguish from user-set values. The same pattern lives in:

- `commands/compile.py:152`
- `commands/eval.py:255`
- `commands/export.py:202`
- `commands/perf.py:1222`
- `commands/analyze.py:504`

The narrower fix is fine for this PR, but a unified fix (either making `from_dict()` sparse, or returning the raw dict from `load_build_config()`) would close the bug class instead of one instance. Worth at least a follow-up issue so the same regression doesn't surface in other commands.

🤖 Generated with Claude Code

src/winml/modelkit/commands/quantize.py
```diff
@@ -98,6 +99,13 @@
     default=None,
     help="Task for calibration dataset selection (e.g., 'image-classification').",
 )
+@click.option(
+    "--model-id",
```
timenick: Nit: **CLI flag / JSON key asymmetry** — the CLI flag is `--model-id` but the JSON config key is `quant.model_name` (line 173 of this file maps one to the other). A user copying CLI args into a config file has to know to rename. Either rename the JSON key to `model_id` (and have `WinMLQuantizationConfig.from_dict()` accept both for back-compat), or document the mapping in the help text of `--model-id`.

🤖 Generated with Claude Code

src/winml/modelkit/quant/quantizer.py
```diff
@@ -175,6 +175,8 @@ def quantize_onnx(
         # resolution.  output_path.parent is guaranteed to exist (caller mkdir).
```
timenick, line 175: Stale comment: `# resolution.  output_path.parent is guaranteed to exist (caller mkdir).` is no longer accurate now that line 178 does the mkdir inside `quantize_onnx`. Update the comment (or drop the parenthetical) so a future reader doesn't add a redundant caller-side mkdir on the assumption that this function still requires it.

🤖 Generated with Claude Code

tests/unit/commands/test_compile_quantize_flags.py
```diff
@@ -149,3 +149,182 @@ def test_dml_ep_shows_gpu_device(self, tmp_path):
...
+            tmp_path,
+        )
+        assert cfg.weight_type == "int16"
+        # With --precision unset and JSON not setting activation_type, the
+        # resolver falls back to default uint8 for activation. The contract:
+        # JSON's silence about activation_type must not be misread as
+        # "user wants uint8" — it stays at the CLI-default sentinel which
+        # _resolve_quant_types then maps to uint8 (since precision is None).
+        # But weight_type comes from JSON unambiguously. This pins the
+        # weight_type value.
```
timenick, line 285: **Test does not assert the contract its comment describes.** The trailing comment explains what should happen to `activation_type` ("falls back to default uint8"), but the test only asserts `cfg.weight_type == "int16"`. The fall-through behavior is unpinned — if a regression flipped `activation_type` here, this test would still pass.

Add:
```python
assert cfg.activation_type == "uint8"
```
Same issue in `test_b5_partial_config_only_activation_type` below — should add `assert cfg.weight_type == "uint8"`.

🤖 Generated with Claude Code

tests/e2e/test_quantize_e2e.py
```diff
@@ -0,0 +1,825 @@
...
+#
+# Each fixture lazily exports the model via `winml export` to a persistent
+# cache under ~/.cache/winml/test_fixtures/ so that subsequent test runs
+# reuse the file (cold first run: ~30-90s per model; warm: ~0s).
+# Marked with `network` because the first run downloads from HuggingFace.
+# ---------------------------------------------------------------------------
+
+
+def _export_hf_to_onnx(hf_id: str, task: str, slug: str) -> Path:
+    cache = Path.home() / ".cache" / "winml" / "test_fixtures" / slug
```
timenick: **Persistent cache outside project tree** — `Path.home() / ".cache" / "winml" / "test_fixtures"` writes to user home. The project convention (see [`CLAUDE.md`](/CLAUDE.md)) is to use `temp/` in the project root for temp files. If the cross-session persistence is intentional (to avoid 30-90 s HF downloads on every run), please document why a user-home cache is required, or move the cache under `temp/` so the project owns the location and CI cleanup can find it.

🤖 Generated with Claude Code

tests/e2e/test_quantize_e2e.py
```diff
@@ -0,0 +1,825 @@
...
+        out = tmp_path / "d2.onnx"
+        r = _invoke(
+            runner,
+            [
+                "-m", str(onnx_imgcls), "-o", str(out),
+                "--task", "image-classification",
+                "--model-id", "microsoft/resnet-50",
+                "--samples", "4",
+            ],
+        )
```
timenick, line 484: **Per-task dataset tests don't verify the right dataset was actually selected.** `test_task_image_classification_dataset` (and the text/object-detection/segmentation siblings) only call `_assert_quantized_output`, which checks structural QDQ properties. If `ImageDataset` silently fell back to `RandomDataset` (e.g., due to a preprocessor load failure, a `model_id` resolution bug, or a mapping regression), all of these tests would still pass.

The ASR fallback test on line 530 already asserts `"falling back to RandomDataset" in r.output` — please add a complementary positive assertion here, e.g. grep stdout for a dataset-class marker or `patch` the dataset constructor to record which type was instantiated.

🤖 Generated with Claude Code

tests/e2e/test_quantize_e2e.py
```diff
@@ -0,0 +1,825 @@
...
+    @pytest.mark.network
+    def test_task_image_classification_dataset(
+        self, runner: CliRunner, onnx_imgcls: Path, tmp_path: Path
+    ):
+        out = tmp_path / "d2.onnx"
+        r = _invoke(
+            runner,
+            [
+                "-m", str(onnx_imgcls), "-o", str(out),
+                "--task", "image-classification",
```
timenick, line 480: Nit: **HF model IDs hardcoded** — `tests/e2e/conftest.py` already exposes `HUB_PAIRS` sourced from `src/winml/modelkit/data/hub_models.json`, and `microsoft/resnet-50` and `Intel/bert-base-uncased-mrpc` are already in that file. Consider parametrizing these per-task tests off the central hub list so the test corpus and the SDK's supported-model catalog stay in sync.

🤖 Generated with Claude Code

timenick (commented):
Review of #608. Comments inline.

🤖 Generated with Claude Code

src/winml/modelkit/commands/quantize.py
zhenchaoni: Thanks for reminder. Left the comment there. And I use timenick's suggestion of using model-id as the param name.

src/winml/modelkit/commands/quantize.py
zhenchaoni: This is correct, i will have a separate PR to fix this.

src/winml/modelkit/commands/quantize.py
zhenchaoni: I will use model_name, since it comes earlier and is already available in many places. Renaming model_name to model_id will be done in consistency work.

src/winml/modelkit/commands/quantize.py
zhenchaoni: I will use model_name for now. As the build config logic is using model_name. We can handle the renaming when we address consistency issue (#563) for model inputs.

src/winml/modelkit/quant/quantizer.py
zhenchaoni: Good catch. I keep the comment as this is intention. I made a change to let the caller to create the directory.

tests/unit/commands/test_compile_quantize_flags.py
zhenchaoni: applied your suggestion.

tests/e2e/test_quantize_e2e.py
zhenchaoni: Add an assertion to check the log which says the dataset used.

tests/e2e/test_quantize_e2e.py
zhenchaoni: "yolos-small and segformer-b0 are intentionally chosen for their small footprint in tests; they're not catalog entries. I will keep this part unchanged because this is a Nit.

src/winml/modelkit/commands/quantize.py
zhenchaoni: applied your suggestion on keeping load_build_config(). Other command have similar issues as well. We can do that in a separate PR in our quality control work.

src/winml/modelkit/commands/quantize.py
xieofxie: Then I think we should make this p0 as it is easy to fix. As we will ship publicly, we had better keep name consistent and fixed

zhenchaoni: done. marked that issue p0

xieofxie (approved):

timenick (approved):

PR #609, opened 2026-05-13
Fix empty node.name handling with deterministic stable-node keys and drift-safe indexing

fangyangci:
## Goal
This PR fixes failures and ambiguity when ONNX nodes have an empty node.name by enforcing one deterministic identifier path across pattern matching, analysis, and rewrite.

## Index Drift: Root Cause and Guarantee
Previous behavior could drift because unnamed-node identifiers were derived from mutable graph position and, in some paths, fell back to object-id-style keys. After graph mutation (delete/insert), lookups could point to the wrong node or become nondeterministic.

This PR guarantees:
- Internal node identity always uses a stable key policy: node.name if present, otherwise node_{index} from a preserved graph snapshot.
- Unknown unnamed external nodes are rejected explicitly (KeyError) instead of silently falling back to node_obj_{id(node)}.
- Rewrite-time key arrays stay synchronized with graph mutations, preventing key/index drift.

## Scope and Method
- Unify stable-key generation in shared utilities and consume it across analyze/pattern/rewrite paths.
- Extend match result data model with stable matched_node_keys and alignment validation.
- Build and propagate sidecar node-key maps keyed by object identity together with persistent graph snapshots for deterministic resolution.
- Remove all fallback schemes based on Python object id.
- Update conflict detection, pattern extraction, and runtime/doc checker flows to resolve by stable node key first.

## Test Coverage
Added/updated unit tests to cover:
- unnamed-node key generation and resolution consistency
- rejection of unknown unnamed external nodes
- rewrite key-drift regression scenarios
- matched_node_keys and node/key alignment invariants
- RuntimeCheckerQuery and DocConstraintChecker stable-key behavior

Executed targeted regression suites during this branch:
- 188 passed, 15 skipped
- 192 passed, 15 skipped
- 139 passed, 15 skipped
- 11 passed, 0 failed (post-cleanup key-focused suite)

No API/public behavior expansion beyond the intended deterministic-key and strict-resolution semantics.


tests/unit/analyze/core/test_pattern_deduplication.py
```diff
@@ -11,6 +11,7 @@
 import pytest
 from onnx import TensorProto, helper
 
+from tests.unit.test_helpers.stable_node_keys import stable_test_node_keys as _stable_test_node_keys
```
timenick: **Test import reaches into a submodule whose symbol is already re-exported.** [tests/unit/test_helpers/__init__.py](tests/unit/test_helpers/__init__.py) explicitly does `from .stable_node_keys import stable_test_node_keys` and lists it in `__all__`, so per CLAUDE.md test rule ("Never reach into internal submodules for symbols exported by `__init__.py`") the import should be:

```python
from tests.unit.test_helpers import stable_test_node_keys as _stable_test_node_keys
```

Same violation at `tests/unit/analyze/core/test_runtime_checker.py:22`, `tests/unit/analyze/models/test_patterns.py:20`, and `tests/unit/optim/pipes/test_pipe_rewrite.py:29` — please fix all four.

🤖 Generated with Claude Code

src/winml/modelkit/pattern/match.py
```diff
@@ -64,16 +64,24 @@ class SkeletonMatchResult:
...
     def matched_node_names(self) -> list[str]:
         """Get matched node names as strings.
 
         Returns:
-            List of node name strings. For unnamed nodes, returns formatted
-            name like "OpType_node" (e.g., "Conv_node").
+            List of stable internal node keys used by matcher internals.
         """
-        return [node.name if node.name else f"{node.op_type}_node" for node in self.matched_nodes]
+        return self.matched_node_keys
```
timenick: **`matched_node_names` is now a misleading alias for `matched_node_keys`.** The property's docstring openly says it returns "stable internal node keys" (e.g., `"node_0"` for unnamed nodes), but the name still implies these are ONNX `node.name` values — they aren't. Callers like [runtime_checker.py:421](src/winml/modelkit/analyze/core/runtime_checker.py#L421) and [information_engine.py:622](src/winml/modelkit/analyze/core/information_engine.py#L622) have already renamed the local variable to `node_key` while still reading from `.matched_node_names`, perpetuating the inconsistency.

Suggested fix: either
1. drop the `matched_node_names` property entirely and have call sites read `matched_node_keys` directly (single source of truth), or
2. keep `matched_node_names` returning real `node.name` strings (with empty for unnamed) and add a separate keys accessor.

The current state — a `_keys` field plus a `_names` property that returns keys — is the worst of both.

🤖 Generated with Claude Code

src/winml/modelkit/pattern/base.py
```diff
@@ -1695,13 +1696,19 @@ def _match_single_skeleton(
...
-                        if src >= 0 and edge_info.src_slot != src_slot:
-                            src_slot_matched = False
-                            break
+                    if dst_slot not in input_slots:
+                        continue
+
+                    edge_info = self.edge_info_by_name.get(input_edge, {}).get(node_name)
+                    if edge_info is None:
+                        src_slot_matched = False
+                        break
```
timenick: **Silent `.get().get()` fallback can mask edge-info population bugs.**

Before the PR this was `self.edge_info_by_name[input_edge][node.name]` — it would raise `KeyError` loudly if `edge_info_by_name` wasn't populated for a node/edge. After:

```python
edge_info = self.edge_info_by_name.get(input_edge, {}).get(node_name)
if edge_info is None:
    src_slot_matched = False
    break
```

If a future refactor stops populating `edge_info_by_name` for some edges (typo, off-by-one in init), the matcher will silently produce zero matches instead of failing visibly during development. Same pattern repeats at line ~1724.

At minimum: `logger.debug("missing edge_info: %s/%s", input_edge, node_name)` so the silent path is observable. Better: assert that the lookup must succeed given how `edge_info_by_name` is populated, and only fall through silently for inputs known not to have edge info (e.g., graph inputs / initializers).

🤖 Generated with Claude Code

timenick (commented):
Review of #609 — comments inline.

🤖 Generated with Claude Code

src/winml/modelkit/analyze/utils/node_key_utils.py
```diff
@@ -0,0 +1,58 @@
...
+    1) explicit id(node) sidecar mapping
+    2) named node fallback (node.name)
+    3) identity scan in graph snapshot for unnamed graph nodes
+    4) raise KeyError for unnamed external nodes
+    """
+    stable_key = node_key_by_node_id.get(id(node))
+    if stable_key is not None:
+        return stable_key
+
+    if node.name:
```
timenick, line 53: **Hidden caching side effect.** The identity-scan branch mutates the caller's `node_key_by_node_id` dict (`node_key_by_node_id[id(node)] = stable_key`). This caching is reasonable, but:

1. The docstring doesn't mention it — a reader who treats the dict as a read-only snapshot will be surprised.
2. The function isn't pure, so any caller that passes a shared/cached dict is implicitly opting into mutation.

Either document the side effect explicitly ("Note: caches resolved keys back into `node_key_by_node_id`") or move the caching to the call site that wants it.

🤖 Generated with Claude Code

tests/unit/test_helpers/__init__.py
```diff
@@ -0,0 +1,10 @@
+# -------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+# --------------------------------------------------------------------------
+"""Shared helpers for unit tests."""
+
+from .stable_node_keys import stable_test_node_keys
```
timenick, line 11: **`tests/unit/test_helpers/` is not in the CI path matrix.**

[.github/workflows/modelkit-ci.yml:23-38](.github/workflows/modelkit-ci.yml#L23-L38) enumerates the `tests/unit/<module>` subpaths CI actually runs. `test_helpers/` isn't listed — today the directory contains only helper modules so it doesn't matter, but the moment someone adds a `test_*.py` here it will be silently invisible to CI.

Two cleaner options:
1. Move the helper into a shared `conftest.py` at the appropriate scope (idiomatic pytest pattern, also makes the helper auto-discoverable via fixtures rather than relying on import paths).
2. Add `tests/unit/test_helpers/` to the CI matrix preemptively and add a guardrail like an `__init__.py` comment forbidding `test_*.py` files there.

🤖 Generated with Claude Code

src/winml/modelkit/analyze/models/onnx_model.py
```diff
@@ -151,3 +180,55 @@ def get_model(self) -> onnx.ModelProto:
...
+
+    def get_node_key(self, node: onnx.NodeProto) -> str:
+        """Get the stable sidecar key for a node."""
+        node_key_by_node_id = self._node_key_by_node_id
+        if node_key_by_node_id is None:
+            self._initialize_node_key_index()
+            node_key_by_node_id = self._node_key_by_node_id
+
+        if node_key_by_node_id is None:
+            raise RuntimeError("Node key index is unavailable.")
```
timenick: **Dead defensive check.** After `self._initialize_node_key_index()` runs, `_node_key_by_node_id` is unconditionally set to a dict (possibly empty) — never `None`. So the second `if node_key_by_node_id is None: raise RuntimeError(...)` can never trigger. Same applies to the `_graph_nodes is None` guard right below, and the parallel patterns in `get_node_by_key`, `get_node_by_name`, and `get_node_key_map`.

Either drop the redundant second guard, or document why init might "silently fail" to set the attribute (which would be a Pydantic edge case worth a real test, not a defensive `raise`).

🤖 Generated with Claude Code

tests/unit/analyze/core/model_validators/test_validators.py
```diff
@@ -59,6 +59,7 @@ def create_runtime_result_with_tags(
     skeleton_result = SkeletonMatchResult(
         pattern=pattern,
         matched_nodes=[node_proto],
+        matched_node_keys=[node_proto.name if node_proto.name else "node_0"],
```
timenick: **Inconsistent helper usage across the PR.** This site (and `test_node_checkers.py:58`, `test_output_aggregator.py:53`, `test_pattern_extractor.py:212-213`, `test_onnx_model.py`) inlines `node_proto.name if node_proto.name else "node_0"`, while [test_pattern_deduplication.py](tests/unit/analyze/core/test_pattern_deduplication.py), [test_runtime_checker.py](tests/unit/analyze/core/test_runtime_checker.py), [test_patterns.py](tests/unit/analyze/models/test_patterns.py), and [test_pipe_rewrite.py](tests/unit/optim/pipes/test_pipe_rewrite.py) all use the new `stable_test_node_keys` helper.

Since the whole PR's goal is to centralize the stable-key fallback rule, please use the helper everywhere. Otherwise a future change to `make_stable_node_key` (e.g. zero-padding to `node_{idx:04d}`) will silently break the inlined sites and pass the helper-based ones, which is exactly the drift the PR is trying to eliminate.

🤖 Generated with Claude Code

src/winml/modelkit/analyze/core/information_engine.py
```diff
@@ -613,14 +617,11 @@ def _query_doc_constraints(self, runtime_result: PatternRuntime, pattern_id: str
...
-            # Find the actual ONNX NodeProto from model
+            # Resolve ONNX NodeProto from stable sidecar key
             node_lookup_start = time.perf_counter()
-            model_proto = self._model.get_model()
-            node = None
-            for graph_node in model_proto.graph.node:
-                if graph_node.name == node_name:
-                    node = graph_node
-                    break
+            node = self._model.get_node_by_key(node_name)
```
timenick: Nit: the local variable is named `node_name`, but the value passed in is now a stable key (e.g. `"node_0"` for unnamed nodes), and it's used as `get_node_by_key(node_name)` first. Rename to `node_key` to match what `matched_node_names` actually returns post-PR — otherwise this reads like a real `node.name` lookup when it isn't.

🤖 Generated with Claude Code

src/winml/modelkit/analyze/models/onnx_model.py
```diff
@@ -124,8 +133,28 @@ def from_onnx_model(cls, model: onnx.ModelProto, model_path: str) -> ONNXModel:
...
+        graph_nodes = list(model.graph.node)
+        node_key_by_node_id = build_node_key_by_node_id(graph_nodes)
+        node_by_key: dict[str, onnx.NodeProto] = {}
+        node_by_name: dict[str, onnx.NodeProto] = {}
+
+        for index, node in enumerate(graph_nodes):
+            stable_key = make_stable_node_key(node, index)
+            node_by_key[stable_key] = node
+            if node.name and node.name not in node_by_name:
+                node_by_name[node.name] = node
```
timenick: Nit: `_node_by_name` silently drops nodes whose `.name` collides with a name already seen (`if node.name and node.name not in node_by_name`). ONNX permits this — the upstream optimizer may produce duplicate names. Today this is silent and the second-or-later node is unreachable via `get_node_by_name`.

At minimum, `logger.debug("duplicate node.name in graph: %s — only first node will be reachable via get_node_by_name", node.name)` so a future "why is get_node_by_name returning the wrong node?" debug session has something to grep for.

🤖 Generated with Claude Code

tests/unit/analyze/core/test_pattern_deduplication.py
fangyangci: Fixed. Updated all four test imports to use the package re-export: from tests.unit.test_helpers import stable_test_node_keys as _stable_test_node_keys.

src/winml/modelkit/pattern/match.py
fangyangci: Fixed. We now use matched_node_keys as the canonical skeleton key list, removed the misleading SkeletonMatchResult.matched_node_names alias, and updated internal call sites to consume matched_node_keys directly.

tests/unit/analyze/core/model_validators/test_validators.py
fangyangci: Fixed. Replaced inline key fallbacks with stable_test_node_keys helper in all listed test sites so the fallback policy is centralized and drift-safe.

src/winml/modelkit/pattern/base.py
fangyangci: Fixed. Added debug observability for missing edge_info lookups at both matcher sites (the early src-slot check and the mapping-expansion pass). Missing entries are still treated as unmatched, but now leave searchable diagnostics.

src/winml/modelkit/analyze/models/onnx_model.py
fangyangci: Fixed. Removed redundant post-initialization None guards by centralizing initialization in _ensure_node_key_index(), then using the initialized sidecar maps directly in get_node_key/get_node_by_key/get_node_by_name/get_node_key_map.

src/winml/modelkit/analyze/models/onnx_model.py
fangyangci: Fixed. Added debug logging when duplicate ONNX node.name values are encountered in _initialize_node_key_index(); behavior remains first-wins, but it is now explicitly observable.

src/winml/modelkit/analyze/utils/node_key_utils.py
fangyangci: Fixed. Documented the side effect in resolve_stable_node_key(): when identity-scan resolution succeeds, the key is cached back into node_key_by_node_id for subsequent lookups.

tests/unit/test_helpers/__init__.py
fangyangci: Fixed with option (2): added tests/unit/test_helpers to the CI matrix and added an explicit helper-only guardrail note in tests/unit/test_helpers/__init__.py (no test_*.py in that folder).

src/winml/modelkit/analyze/core/information_engine.py
fangyangci: Fixed. Renamed the local variable in _query_doc_constraints from node_name to node_key and updated related logging/timing fields to reflect key semantics.

timenick (approved):

DingmaomaoBJTU (approved):

PR #613, opened 2026-05-14
fix(perf): make `winml catalog` ~10x faster by deferring heavy data-pkg imports

### LINKED ISSUE #538: [winml hub] [P1] Default invocation takes 7 s for a static curated table
**Severity**: P1  
**Command**: `winml hub`

---
**Category**: Local Command Latency — informational queries should return in <500 ms warm.

**Repro**:
```
Measure-Command { uv run winml hub --task image-classification }
```

**Actual**: ~7.1 s warm, ~8.1 s on a freshly-warmed machine. Output is a static curated table.

**Expected**: <500 ms. The catalog should be a JSON file loaded lazily, not built by importing every model loader / EP probe at import time.

**Why it matters**: `winml hub` is the discovery entry point — users will run it dozens of times to browse models. Seven seconds per invocation feels broken.

vortex-captain:
Fixes #538.

## Root cause

`winml catalog` was taking **5–7 s** warm to render a static curated table.

The culprit was a side effect of how `_load_catalog` located its JSON file:

```python
pkg = importlib.resources.files("winml.modelkit.data")
data = (pkg / "hub_models.json").read_text(encoding="utf-8")
```

`importlib.resources.files("winml.modelkit.data")` executes `winml/modelkit/data/__init__.py`, which eagerly imports `image_classification_dataset`, which in turn eagerly imports `torchvision` and `datasets`. That single line transitively pulls in **torch (~2.2 s), torchvision (~4.3 s), datasets / pandas / pyarrow / huggingface_hub / aiohttp / requests (~1.2 s)** every time `winml catalog` runs.

`cProfile` confirmed >5 s of the wall time was spent in those imports, not in any catalog logic.

## Fix

Two small, local changes — no architecture-specific or hard-coded model logic:

1. **`commands/catalog.py::_load_catalog`** — resolve the JSON path via `__file__` instead of `importlib.resources.files`, so reading the catalog no longer triggers `data/__init__.py`.
2. **`data/image_classification_dataset.py`** — move the `torchvision` and `datasets.load_dataset` imports into the constructor body so they fire only when the dataset is actually instantiated.

## Measured impact

`winml catalog --task image-classification` end-to-end (5 warm runs):

| State | Wall time |
|---|---|
| Before | 5 000 – 6 000 ms |
| After this PR | **490 – 630 ms** |

Roughly **10× faster** at the CLI level. Remaining time is dominated by `import onnxruntime` via `utils/constants.py` (~210 ms) — out of scope for this PR; can be addressed in a follow-up if needed.

## Verification

- Existing pytest suite: 263 passed, 0 regressions (`pytest -k "constants or catalog or device"`).
- `winml catalog --task image-classification` output identical to pre-PR (5 image-classification models rendered correctly).

### DIFF
--- a/src/winml/modelkit/commands/catalog.py
+++ b/src/winml/modelkit/commands/catalog.py
@@ -19,7 +19,6 @@ Usage:
 
 from __future__ import annotations
 
-import importlib.resources
 import json
 import logging
 from pathlib import Path
@@ -54,12 +53,16 @@ _TYPE_PALETTE = ["cyan", "green", "yellow", "magenta", "blue", "red"]
 def _load_catalog() -> dict[str, Any]:
     """Load hub_models.json from package data.
 
+    Resolves the path via ``__file__`` rather than ``importlib.resources.files``
+    so that reading the catalog does not execute ``winml.modelkit.data``'s
+    package init — that init pulls in heavy optional dependencies (onnx, numpy,
+    torchvision via the dataset modules) that the catalog command never needs.
+
     Returns:
         Parsed catalog dict with ``version`` and ``models`` keys.
     """
-    pkg = importlib.resources.files("winml.modelkit.data")
-    data = (pkg / "hub_models.json").read_text(encoding="utf-8")
-    return json.loads(data)  # type: ignore[no-any-return]
+    data_path = Path(__file__).resolve().parent.parent / "data" / "hub_models.json"
+    return json.loads(data_path.read_text(encoding="utf-8"))  # type: ignore[no-any-return]
 
 
 def _filter_models(
--- a/src/winml/modelkit/data/image_classification_dataset.py
+++ b/src/winml/modelkit/data/image_classification_dataset.py
@@ -8,9 +8,6 @@ from __future__ import annotations
 
 from typing import TYPE_CHECKING
 
-from datasets import load_dataset
-from torchvision import transforms
-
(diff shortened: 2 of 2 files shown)


zhenchaoni (approved):

timenick (commented):
Nice perf fix. One suggestion to lock the win in.

The import-budget regression test at [`tests/cli/test_import_time.py:173`](tests/cli/test_import_time.py#L173) currently allows `torch`/`torchgen`/`torchvision` when importing `winml.modelkit.data`:

```python
("winml.modelkit.data", ("torch", "torchgen", "torchvision")),
```

That allowlist exception exists **only** because `image_classification_dataset.py` had a top-level `from torchvision import transforms`. After this PR moves those imports into the constructor, `dummy_dataset` and `random_dataset` are the only other modules in `data/` and neither pulls torch/torchvision — so importing `winml.modelkit.data` should no longer touch torchvision at all.

With the allowlist untouched, a future PR that re-adds a top-level `from torchvision import ...` anywhere under `data/` (refactor, new dataset, etc.) won't trip the test — the 5–7s regression would silently come back. Tightening this line to either:

```python
("winml.modelkit.data", ()),
```

or moving `winml.modelkit.data` up to the strict `test_module_no_heavy_deps` parametrize (no allowlist) is the natural way to convert the perf win into an enforced invariant.

🤖 Generated with Claude Code

PR #615, opened 2026-05-14
feat(analyze): disambiguate EPContext pattern_id by EP label, generalize display-name suffix stripping

xieofxie:
## Summary

Two related changes around how `winml analyze` reports per-EP results from `EPContext` nodes.

### 1. EPContextNodeChecker: suffix pattern_id with an EP label derived from `partition_name`

Multi-EP analysis runs the same EPContext pattern under multiple providers (QNN, DML, OpenVINO, ...), so the report needs the pattern_id rows to stay disambiguated. The checker now emits `pattern_id` suffixed with the EP label derived from the node's `partition_name`:

- `partition_name = None` → label is `"UNKNOWN_EP"` (attribute missing — surfaced explicitly instead of silently producing an empty label).
- Canonical ORT format `"<X>ExecutionProvider_..."` → label is `<X>` (e.g. `"QNNExecutionProvider_0"` → `"QNN"`). See [graph_partitioner.cc#L378](https://github.com/microsoft/onnxruntime/blob/7e1d818ba7923c1e44187284a6ca77bbe13aa1eb/onnxruntime/core/framework/graph_partitioner.cc#L378) for the source format.
- Anything else → `partition_name` used directly as the label.

Helper: `EPContextNodeChecker._ep_label(partition_name)`. The existing partition-name vs `ep_name` match check is unchanged.

### 2. `_display_name` in `commands/analyze.py`: strip any trailing ` (xxx)`

To keep the live table grouping by base op type, `_display_name` now strips any trailing ` (...)` annotation via a precompiled regex. This handles QDQ markers (` (QDQ)`), the new EP-prefix suffix from change 1, and any future ` (...)` decoration uniformly.

Consequences:
- `commands/analyze.py` no longer needs to call `.removesuffix(QDQ_SUFFIX)` after `_display_name`.
- `QDQ_SUFFIX` is no longer re-exported from `winml.modelkit.analyze` (it stays an implementation detail inside `core/runtime_checker_query.py` where it's actually used).

### Tests

- `test_node_checkers.py::test_pattern_id_suffixed_with_ep_label_from_partition_name` covers all three branches (`(QNN)`, `(UNKNOWN_EP)`, `(DML_0)`).
- `test_static_analyzer_cli.py::test_qdq_pattern_id_maps_to_base_op_for_table_key` now asserts `_display_name` standalone (no separate `removesuffix` step) and adds EP-suffix coverage.
- Full analyze suite (1378 tests) passes locally.

### DIFF
--- a/src/winml/modelkit/analyze/__init__.py
+++ b/src/winml/modelkit/analyze/__init__.py
@@ -21,7 +21,6 @@ from .core.onnx_loader import ONNXLoader
 from .core.output_aggregator import OutputAggregator
 from .core.pattern_extractor import PatternExtractor
 from .core.runtime_checker import RuntimeChecker
-from .core.runtime_checker_query import QDQ_SUFFIX
 from .models import (
     Action,
     ActionItem,
@@ -44,7 +43,6 @@ from .utils.rule_loader import RuleLoader
 
 
 __all__ = [
-    "QDQ_SUFFIX",
     "Action",
     "ActionItem",
     "ActionLevel",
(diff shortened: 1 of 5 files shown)


xieofxie:
result like
image

DingmaomaoBJTU (approved):

PR #619, opened 2026-05-14
Fix license issue for calculating cider metric

### LINKED ISSUE #610: pycocoevalcap ships GPL JARs as a runtime dep — should be optional
## Problem

`pycocoevalcap>=1.2` is declared as a hard runtime dependency in
[`pyproject.toml`](https://github.com/microsoft/WinML-ModelKit/blob/main/pyproject.toml#L45),
but the installed package bundles several GPL/LGPL-licensed Java
artifacts under `site-packages/pycocoevalcap/`:

| JAR | License |
|---|---|
| `tokenizer/stanford-corenlp-3.4.1.jar` | **GPL v3+** |
| `spice/spice-1.0.jar` | **GPL** |
| `meteor/meteor-1.5.jar` | **LGPL** |
| `spice/lib/Meteor-1.5.jar` | **LGPL** |

`pycocoevalcap` itself also has no `License` field in its PyPI METADATA
(reported as `UNKNOWN` by `pip-licenses`).

Since `winml-modelkit` is MIT-licensed and published to PyPI, every
`pip install winml-modelkit` drops GPL-licensed artifacts into the
user's environment. This is typically flagged by Microsoft OSPO as a
compliance blocker for MIT-licensed redistributables.

## Current usage

Only `pycocoevalcap.cider.cider.Cider` (pure-Python CIDEr-D) is used,
in [`eval/metrics/text_similarity.py`](https://github.com/microsoft/WinML-ModelKit/blob/main/src/winml/modelkit/eval/metrics/text_similarity.py#L88-L102).
The GPL JARs (SPICE / METEOR / Stanford CoreNLP) are **never invoked**
at runtime — they just sit on disk.

The import is already wrapped in `try/except ImportError` and degrades
gracefully to `None` when the package is missing — so the code already
treats it as optional.

## Suggested fix

Move `pycocoevalcap` from `[project.dependencies]` to an extras group,
e.g.:

```toml
optional-dependencies.eval-cider = [ "pycocoevalcap>=1.2" ]
```

Users who need the CIDEr metric opt in via
`pip install winml-modelkit[eval-cider]`. No code change required —
the existing `try/except` in `_cider()` handles the missing-dep path.

## Introduced by

- PR #408 (commit `0599836e`)

## References

- https://github.com/salaniz/pycocoevalcap (no LICENSE file at the
  repo root; bundles GPL jars)
- Stanford CoreNLP 3.4.1 licensing: GPL v3+
  (https://stanfordnlp.github.io/CoreNLP/)

zhenchaoni:
fix #610

CIDER metric is typically used to evaluate `image caption` models. We currently use `pycocoevalcap` package to calculate it. Although the package is well-known, it contains some GPL license JAR files in it. Even though we don't use those JAR files, we cannot ship with it.

The fix is to "hand implement" the CIDER metric calculation because there is no other popular python packages available for it. Considering the implementation logic is not complicated, we can take this approach. We have tested to ensure the handmaking calculation gives same output as `pycocoevalcap`

### DIFF
--- a/src/winml/modelkit/eval/metrics/cider.py
+++ b/src/winml/modelkit/eval/metrics/cider.py
@@ -1,0 +1,325 @@
+# -------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+# --------------------------------------------------------------------------
+"""In-house implementation of the CIDEr image-captioning metric.
+
+Reference:
+    Vedantam, Zitnick, Parikh. "CIDEr: Consensus-based Image Description
+    Evaluation." CVPR 2015. arXiv:1411.5726.
+
+Exposes a :class:`Cider` class whose :meth:`Cider.compute_score`
+accepts the standard ``(gts, res)`` dict-of-lists corpus format used
+throughout the image-captioning literature.
+"""
+
+from __future__ import annotations
+
+import math
+from collections import Counter
+from typing import TYPE_CHECKING
+
+import numpy as np
+
+
+if TYPE_CHECKING:
+    from collections.abc import Hashable, Sequence
+
+
+# Paper defaults. Tests pin these — do not change without a test update.
+_NGRAM_ORDER = 4
+_LENGTH_SIGMA = 6.0
+_SCORE_SCALE = 10.0
+
+
+# Internal type aliases for readability only.
+_Ngram = tuple[str, ...]
+_NgramCounts = Counter[_Ngram]
+_TfidfVector = dict[_Ngram, float]
+
+
+def _tokenize(sentence: str) -> list[str]:
+    """Whitespace tokenization.
+
+    Callers handle any required upstream normalization (case folding,
+    punctuation stripping). Keeping tokenization trivial here makes the
+    metric deterministic and free of external tokenizer dependencies.
+    """
+    return sentence.split()
+
+
+def _collect_ngrams(tokens: Sequence[str], order_max: int) -> _NgramCounts:
+    """Return n-gram counts for orders 1..``order_max`` over ``tokens``."""
+    counts: _NgramCounts = Counter()
+    n_tokens = len(tokens)
+    for order in range(1, order_max + 1):
+        last_start = n_tokens - order + 1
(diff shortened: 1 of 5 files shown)


xieofxie (approved):
I don't know if it is correct implemented or not, if you said so, then it is XD

src/winml/modelkit/eval/metrics/cider.py
```diff
@@ -0,0 +1,325 @@
...
+            dot_product += min(cand_weight, ref_weight) * ref_weight
+
+        denom = cand_norms[k] * ref_norms[k]
+        cosine_sim = (dot_product / denom) if denom > 0.0 else 0.0
+        similarities.append(cosine_sim * length_penalty)
+
+    return similarities
+
+
+class Cider:
```
timenick, line 175: The class is named `Cider` and `method()` returns `"CIDEr"`, but the implementation is actually **CIDEr-D** — `_pair_similarity` applies the Gaussian length penalty and the min-clipped cosine similarity, which are exactly the modifications introduced in §3.2 of Vedantam et al. (the "-D" variant), not vanilla CIDEr. The `_pair_similarity` docstring acknowledges this internally ("This is the '-D' variant's defense against length-based gaming"), but the public surface doesn't.

Consider either:
1. Rename the class to `CiderD` and have `method()` return `"CIDEr-D"`, or
2. Update the class-level docstring to state explicitly that this implements the CIDEr-D variant.

Option 2 is the smaller change. Either way, anyone reading the public API today would reasonably expect vanilla CIDEr and get -D scores instead, which can confuse cross-paper comparisons.

🤖 Generated with Claude Code

timenick:
> I don't know if it is correct implemented or not, if you said so, then it is XD

you can leverage copilot to check 😄

timenick (approved):

src/winml/modelkit/eval/metrics/cider.py
zhenchaoni: Fixed by option 2 (updating the doc string). I keep the variable name "cider" because `pycocoevalcap` also uses this name but calculates cider-d.

timenick (approved):

PR #622, opened 2026-05-14
fix(compile,analyze): reject unsupported EPContext targets and handle no-rule-data gracefully

DingmaomaoBJTU:
## 1. `compile` command — three bug fixes

**Bug 1: DML/CPU crashed instead of giving a clear error.**
`WinMLCompileConfig.for_provider()` returns `None` for any provider with
`enable_ep_context=False` (DML, CPU, CUDA, MIGraphX, NV TensorRT RTX).
The next line `config.validate = validate` raised
`AttributeError: 'NoneType' object has no attribute 'validate'`
instead of a user-readable error.

**Bug 2: Error message listed `dml` as a valid EPContext provider.**
The exception text read `(e.g. qnn, dml, openvino)`, implying DML produces
EPContext models. It does not — `for_dml()` has always had `enable_ep_context=False`.

**Bug 3: `--device` was not forwarded to `for_provider()`.**
`for_provider(provider)` was called without `device=`, so QNN and OpenVINO
compilation always produced empty `provider_options`. For QNN this meant the
EPContext file was named `..._qnn_ctx.onnx` instead of `..._npu_ctx.onnx` or
`..._gpu_ctx.onnx`, breaking output-path resolution downstream. For OpenVINO,
CPU and GPU builds shared the same cache key.

**Bonus cleanup:** `_EP_TO_DEVICE.get(provider, device)` on the Device display
line would print `Device: npu` for `--device gpu --ep qnn`. The fix prints the
user-supplied `--device` directly.

**Changes:** `commands/compile.py` — add `if config is None` guard → `ClickException`; pass `device=device` to `for_provider()`; simplify Device display; remove unused `_EP_TO_DEVICE` import.

**New tests** (`test_compile_command.py`): `test_gpu_device_raises_unsupported_error`, `test_ep_dml_raises_unsupported_error`, `test_cpu_device_raises_unsupported_error`, `test_device_label_reflects_device_flag`, `test_compile_device_propagates_to_provider_options`.

---

## 2. Static analyzer — graceful no-rule-data handling

**Bug 1: Pattern detection was silently dropped for no-rule-data EPs.**
A `has_rule_data_for_ep` gate at the top of the EP loop caused the analyzer to
`continue` before pattern extraction ran. Patterns detected in the model were
never surfaced in the Analysis Summary for EPs like DML or CPU.

**Fix:** Remove the pre-loop gate. `RuntimeChecker` is always instantiated;
pattern detection always runs. The rule-data check is moved into
`RuntimeChecker.op_support()` where it is actually needed, and pre-computed
once at `__init__` (no repeated filesystem probes).

**Bug 2: `infer_ihv_from_ep_name()` raised `ValueError` for Microsoft EPs.**
`CPUExecutionProvider` and `DmlExecutionProvider` were not in the IHV map, so
any code path that called this function for those EPs (e.g. `rule_loader`)
crashed with `ValueError: Unknown execution provider`.

**Fix:** Add `IHVType.MICROSOFT` and fall back to it for unrecognized EPs
instead of raising. Update `rule_loader` to use `prefix_map.get(ihv, ihv.lower())`
so unknown IHVs don't `KeyError` on the prefix lookup either.

**Bug 3: `exit(2)` misfired when rule data was merely absent.**
The early-exit logic in `analyze.py` did not distinguish between two different
situations: (a) the requested device is genuinely not supported by this EP, and
(b) the EP+device pair is valid but no rule data has been collected yet. Both
cases triggered `exit(2)`, making the command unusable on machines where data
collection hasn't been run.

**Fix:** Only `exit(2)` for case (a). For case (b), proceed and let
`RuntimeChecker` emit a clear diagnostic. Two distinct log messages tell the
user which situation they are in:
- *"No runtime check data found. Follow CONTRIBUTING.md…"* — no parquet files at all (setup needed)
- *"No runtime check data for {EP} on {device} — op analysis will return no_data."* — data exists but not for this EP+device

**Bug 4: No-data EPs polluted the progress UI with empty bars.**
OP CHECK progress bars were always created for every EP. For no-data EPs all
node counts came back 0, producing meaningless 0/0/0 rows in the build output
and `analyze` command display.

**Fix:** Skip the progress bar entirely for no-data EPs. In the Analysis
Summary, show a compact patterns-only row instead of the full op breakdown.

**Updated tests:** `test_analyzer.py`, `test_has_rule_data.py` — no `ValueError`
on unknown EPs, `RuntimeChecker` always instantiated, `trtrtx` alias added.


src/winml/modelkit/commands/build.py
```diff
@@ -814,7 +816,9 @@ def _on_ep_start(ep_name: str, operator_counts: dict) -> None:
                     f"[bold]Analyzing[/bold] [cyan]{total}[/cyan] nodes  "
                     f"[dim](iter {_current_iter[0]}/{_current_iter[1]})[/dim]"
                 )
```
timenick, line 823: **Regression: `--device auto` (the default) silently suppresses progress bars for every EP.**

`_on_ep_start` does:
```python
if has_rule_data_for_ep(ep_name, device or ""):
    _ep_bars[ep_name] = sl.ep_bar_add(ep_name, total=total)
```
The `device` here is the unresolved value captured from the outer closure. `winml build` defaults `--device` to `"auto"` (build.py:289-291), so this calls `has_rule_data_for_ep(ep_name, "auto")` → globs for `*_QNNExecutionProvider_AUTO_*.parquet` → never matches → returns False for every EP.

The analyzer resolves `"auto"` internally (analyzer.py: `if device.lower() == "auto": resolved, _ = resolve_device("auto")`), but the build callback doesn't see the resolved value. Net effect: a plain `winml build model.onnx` invocation loses all OP-CHECK progress bars.

Two ways to fix:
1. Resolve `device` to a concrete value before defining the callback (mirror what analyzer.py does).
2. Resolve inside `_on_ep_start` on first call and cache the result.

The analyze command isn't affected because its `--device` defaults to `"NPU"`, not `"auto"`.

🤖 Generated with Claude Code

src/winml/modelkit/analyze/analyzer.py
```diff
@@ -734,22 +733,6 @@ def analyze_from_proto(
...
-                        device_to_use,
-                    )
-                else:
-                    logger.warning(
-                        "No runtime check data for %s — skipping op analysis.",
-                        current_ep,
-                    )
-                continue
-
             logger.info("Checking runtime support for %s...", current_ep)
```
timenick, line 734: **Dead variable `skipped_ep_count`.** It was previously incremented inside the `has_rule_data_for_ep` skip block that this PR removed, but the initializer here and the `_log_timing(skipped_eps=skipped_ep_count)` call below are still in place. The telemetry field is now always `0`.

Either remove the variable entirely, or — if you want to keep emitting the count — derive it after the loop from the set of EPs whose `op_runtime_check_result` came back empty / no_data, since `_no_data_eps` lives in `commands/analyze.py` and isn't visible here.

🤖 Generated with Claude Code

src/winml/modelkit/commands/analyze.py
```diff
@@ -373,6 +374,22 @@ def _render_analysis_summary(
     for ep_support in results:
         ep_name = ep_support.ep_type
 
+        # For EPs with no rule data, skip op-level rows — only show patterns
+        if no_data_eps and ep_name in no_data_eps:
+            patterns = (ep_patterns or {}).get(ep_name, {})
+            if patterns:
+                console.print(f"   🔵 [bold bright_black]{ep_name}[/bold bright_black]:")
```
timenick: **No-data EPs disappear from the summary when no patterns are detected.**

```python
if no_data_eps and ep_name in no_data_eps:
    patterns = (ep_patterns or {}).get(ep_name, {})
    if patterns:
        console.print(f"   🔵 [bold bright_black]{ep_name}[/bold bright_black]:")
        console.print("      [dim]Op check skipped — no rule data[/dim]")
        ...
    continue
```

The `if patterns:` gate means that if a model has no detected patterns for a no-data EP, the EP is silently omitted from the Analysis Summary. The user only sees the `logger.info("…op analysis will return no_data")` line scrolled past in the console.

Consider rendering a one-line "EP not shown — no rule data, no patterns" row (or moving the EP header outside the `if patterns:` gate) so the user can still see which EPs were attempted.

🤖 Generated with Claude Code

src/winml/modelkit/commands/compile.py
```diff
@@ -188,7 +188,14 @@ def compile(
 
     # Resolve EP from device + ep flags
     provider = _resolve_compile_provider(device, ep)
-    config = WinMLCompileConfig.for_provider(provider)
+    config = WinMLCompileConfig.for_provider(provider, device=device)
```
timenick, line 191: **The new `if config is None` gate doesn't cover all "unsupported EPContext" cases.**

`for_provider` (configs.py:87-122) returns `None` only for **known** providers whose factory produces `enable_ep_context=False` (dml/cuda/nv_tensorrt_rtx/cpu/migraphx). For an **unknown** provider name it falls through to:
```python
return cls(ep_config=EPConfig(provider=provider, enable_ep_context=False))
```
i.e. it returns a non-None config with `enable_ep_context=False`. That bypasses this gate and proceeds into `compile_onnx` where nothing useful happens because the EPContext step is disabled.

Click's `--ep` Choice probably constrains the surface to a known set, so the user-facing impact is small, but the new error message ("Compile is only supported for providers that produce EPContext models") promises completeness it doesn't have. Two clean options:
1. Make `for_provider`'s generic fallback also return `None` when `enable_ep_context=False` (so the gate here covers everything).
2. Add a second gate here on `config.ep_config.enable_ep_context` and raise the same `ClickException`.

Not a blocker — this is a pre-existing seam, not introduced by the PR — but the new gate naturally invites the cleanup.

🤖 Generated with Claude Code

DingmaomaoBJTU (commented):
All four comments addressed — see latest commits.

src/winml/modelkit/commands/build.py
DingmaomaoBJTU: Fixed in 88821f5. Resolved `device or ""` by calling `resolve_device(device or "auto", ep=ep)` once before the callbacks are defined, storing the concrete result in `_resolved_device`. The callback now passes `_resolved_device` to `has_rule_data_for_ep` instead of the raw (potentially "auto") value.

src/winml/modelkit/analyze/analyzer.py
DingmaomaoBJTU: Fixed in 88821f5. Removed `skipped_ep_count = 0` and the `skipped_eps=skipped_ep_count` field from the `_log_timing` call — the count was always 0 after the pre-loop gate was removed.

src/winml/modelkit/commands/analyze.py
DingmaomaoBJTU: Fixed in 88821f5. Moved `console.print(ep_name header)` outside the `if patterns:` gate so the EP is always visible in the summary. The empty-patterns case now shows `"Op check skipped — no rule data, no patterns"`.

src/winml/modelkit/commands/compile.py
DingmaomaoBJTU: Fixed in 88821f5. Went with option 1: the generic fallback in `for_provider` now returns `None` instead of a config with `enable_ep_context=False`, so the `if config is None` gate in `compile.py` covers all unsupported providers uniformly. Updated `test_compiler_configs.py` accordingly.

xieofxie (approved):

PR #624, opened 2026-05-14
CLI: Batch fix of small UX/contract issues + -o/--output refactor

### LINKED ISSUE #514: [winml analyze] [P1] Lacks `-o` short alias; only `--output` works (drift)
**Severity**: P1  
**Command**: `winml analyze`

---
**Category**: Cross-Command Flag Consistency — same flag should have same form (long + short) across commands.

**Repro**:
```
uv run winml analyze -m temp\cli-audit\resnet.onnx --ep qnn -o temp\ux\out_analyze.json
```

**Actual**:
```
Usage: winml analyze [OPTIONS]
Try 'winml analyze --help' for help.

Error: No such option: -o
```
Exit 2.

**Expected**: Both `-o` and `--output` should work, the same way they do for `winml export`, `winml optimize`, `winml quantize`, `winml compile`, `winml hub`, `winml perf`, `winml config`, `winml eval`. Add the `-o` alias.

**Why it matters**: One subcommand using a different flag name forces users to memorize an exception. Most users will type `-o` from muscle memory and get an "unknown option" error.

### LINKED ISSUE #548: [winml optimize] [P1] Several `sys.exit()` paths in `optimize.py` instead of `click.ClickException`
**Severity**: P1  
**Command**: `winml optimize`

---
**Category**: Exit-Code Contract

**Repro**: Triggering an internal error (e.g. malformed `--preset-config` JSON) prints the error and `sys.exit(1)` immediately.

**Actual**: Exit 1 with no documented meaning vs. Click's exit 2 for usage errors. Inconsistent with the rest of the CLI (most paths use `click.ClickException` → exit 1).

**Expected**: Replace `sys.exit(N)` with `raise click.ClickException(...)` (exit 1) or `raise click.UsageError(...)` (exit 2), matching the contract proposed in another audit finding.

**Why it matters**: Same as another audit finding — CI scripts can't branch on exit codes.

vortex-captain:
## Summary

Batch fix of two small CLI papercuts surfaced by the recent UX audit, plus
a follow-on refactor that removes the now-duplicated ``--output`` option
declaration across every command.

### Fixed #514 — `winml analyze` lacks `-o` short alias
`winml analyze` was the only output-producing subcommand missing the `-o`
short alias for `--output`. Every other command (`export`, `optimize`,
`quantize`, `compile`, `perf`, `eval`, `config`, `catalog`, `run`) already
accepts both forms, so users hitting `-o` from muscle memory got
`Error: No such option: -o` and exit 2.

### Fixed #548 — `winml optimize` used `sys.exit(1)` for config validation
The capability-config validation path in `commands/optimize.py` called
`sys.exit(1)` directly, bypassing Click's exception contract. Replaced
with `raise click.UsageError(...)`:

* exit code 2 (matches Click's usage-error contract used by other commands)
* keeps the existing colored error list (now rendered via `click.style`)
* removes the unused `import sys`
* no log duplication — the validation errors are part of the `UsageError`
  message itself

### Refactor — shared ``-o/--output`` decorator

Added `cli_utils.output_option(help_text, required=False)` so the ten
`--output` declarations now collapse into one canonical form. The decorator:

* always uses `click.Path(path_type=Path)` — callbacks receive `pathlib.Path`
* takes a per-command `help_text` so each command keeps its specific
  guidance string
* handles `required=True` (export) by omitting `default=None`, since
  Click silently treats `required + default=None` as not-required

`config.py` and `run.py` previously declared `--output` without
`path_type=Path`, so their callbacks were redundantly wrapping the value
in `Path(output)`. Cleaned those up — callback type annotations widened
from `str | None` to `Path | None` and the manual wrap calls were removed.

Net diff is -40 lines across 11 source files; 13 file change in total.


src/winml/modelkit/commands/analyze.py
```diff
@@ -445,6 +445,7 @@ def _render_analysis_summary(
 @cli_utils.build_config_option
 @click.option(
```
xieofxie: should we use cli_utils?

xieofxie (approved):

src/winml/modelkit/commands/analyze.py
vortex-captain: fixed

PR #625, opened 2026-05-14
fix(perf,session): honour --ep cpu and surface actual EP in perf output

DingmaomaoBJTU:
## Summary

- **Bug fix**: `winml perf --ep cpu --device cpu` was silently picking OpenVINO over `CPUExecutionProvider` on systems where OV is registered. `_build_session_options` short-circuited `--ep cpu` and fell through to the `PREFER_CPU` policy, which let ORT prefer any CPU-capable EP it had loaded. Dropping the `self._ep != "cpu"` exception sends it through the same `add_provider_for_devices` path as every other `--ep` value, so the user's choice is honoured.
- **Visibility**: `winml perf` only printed the requested device string (e.g. `Device: cpu`), so users couldn't tell whether OV-on-CPU or CPU EP actually ran inference — even when `--ep` resolution silently disagreed with their input. New `WinMLSession.ep_name` exposes the primary provider after compile, and the perf header / console report / JSON now show `Device: cpu / OpenVINOExecutionProvider`.

## Before / after

```
# Before (--ep cpu silently ignored)
INFO: Using provider selection policy PREFER_CPU for device cpu
INFO: Session created with device cpu, providers: ['OpenVINOExecutionProvider', 'CPUExecutionProvider']
Device:      cpu

# After
INFO: Explicit EP: cpu (CPUExecutionProvider) device=cpu -> CPU
INFO: Session created with device cpu, providers: ['CPUExecutionProvider']
Device:      cpu / CPUExecutionProvider
```

The no-`--ep` path is unchanged (still `PREFER_CPU` policy), but the actual EP ORT lands on is now visible in the header — e.g. `Device: cpu / OpenVINOExecutionProvider` when no `--ep` is given on a system with OV registered.


src/winml/modelkit/session/session.py
```diff
@@ -489,7 +489,11 @@ def _build_session_options(self, device: str) -> ort.SessionOptions:
         # When device is also specified (non-"auto"), narrow by both EP name
         # and device type so e.g. `--ep qnn --device cpu` finds QNN-on-CPU
         # instead of the first QNN ep_device (which may report as NPU).
-        if self._ep and self._ep != "cpu":
+        # `--ep cpu` is honoured here too so the CPUExecutionProvider gets
+        # bound explicitly; otherwise PREFER_CPU policy lets ORT prefer
+        # OV-on-CPU (or any other registered CPU-capable EP) over the basic
+        # CPU EP, silently ignoring the user's --ep choice.
+        if self._ep:
```
timenick, line 500: Documentation around this block is now out of sync with the new behavior — please update both:

1. The `_build_session_options` docstring (lines 463–475) still says: *"When `self._ep` is set **(and not `"cpu"`)**, uses `add_provider_for_devices`... `"cpu"` falls through to policy-based selection so ORT handles CPU-only inference without any EP registration."* After this change, `"cpu"` also goes through `add_provider_for_devices`, so the docstring directly contradicts the new code.

2. The comment at lines 480–481 reads: *"Exception: when an explicit EP is set (e.g. `--ep openvino --device cpu`), fall through so the EP binding logic below can honour it."* `--ep cpu --device cpu` is now also part of this exception path; the example is misleading on its own and worth updating (or adding the `cpu` case alongside).

🤖 Generated with Claude Code

src/winml/modelkit/commands/perf.py
```diff
@@ -488,6 +491,7 @@ def _collect_results(self, stats: PerfStats) -> BenchmarkResult:
             # Actual values (resolved after build + compile)
             actual_device=self._model.device,
             actual_task=self._model.task or self.config.task or "auto-detected",
+            actual_ep=self._model._session.ep_name,
```
timenick: Minor consistency nit: `WinMLPreTrainedModel` exposes `device` as a public property delegating to `self._session.device` (see `models/winml/base.py:213`), and the call sites in this PR use that public property on line 300 (`device=self._model.device`). For symmetry, consider adding a `WinMLPreTrainedModel.ep_name` property that delegates to `self._session.ep_name`, so the two new call sites here (line 303 `ep_name=self._model._session.ep_name` and this line `actual_ep=self._model._session.ep_name`) can also use the public surface instead of reaching through `_session`.

Not blocking — `self._model._session` is already used elsewhere in this file (lines 294, 371, 391) — but adding the property keeps the model's public abstraction consistent.

🤖 Generated with Claude Code

timenick (commented):
Two notes beyond the inline comments:

**Missing regression test.** Per `CLAUDE.md`'s "Mandatory Test Verification" rule, a bug fix should ship with a regression test that pins the new behavior. The session change (removing the `self._ep != "cpu"` exception) and the new `WinMLSession.ep_name` property both have natural homes in `tests/unit/session/test_winml_session.py`:

- A test (in `TestWinMLSessionExplicitProviders` or similar) that constructs a `WinMLSession(ep="cpu", device="cpu")` against a discovery list including OV and asserts the primary provider is `CPUExecutionProvider` — this would have caught the original bug.
- A test for the new `ep_name` property covering both pre-compile (`None`) and post-compile states.

Without a test, the next refactor of `_build_session_options` could silently reintroduce the regression.

🤖 Generated with Claude Code

src/winml/modelkit/session/session.py
timenick, line 500: The behavior change here invalidates two nearby comments that no longer match the code:

1. The `_build_session_options` docstring (lines 465-468) still claims `"cpu"` falls through to policy-based selection so ORT handles CPU-only inference without any EP registration. After this PR, `--ep cpu` goes through `add_provider_for_devices` instead — please update the docstring to reflect the new contract.
2. The "Exception" example at line 480 only mentions `--ep openvino --device cpu`. Worth adding `--ep cpu` so readers see that explicit CPU now takes the same exception path.

🤖 Generated with Claude Code

src/winml/modelkit/session/session.py
```diff
@@ -684,6 +688,19 @@ def device(self) -> str:
         """Target device for this session."""
         return self._device
 
+    @property
+    def ep_name(self) -> str | None:
```
timenick: Per CLAUDE.md "Mandatory Test Verification", a bug fix should ship with a regression test. The new `ep_name` property and the `--ep cpu` honouring change in `_build_session_options` have no test coverage in this PR.

Suggested additions in `tests/unit/session/test_winml_session.py`:
- A test that `WinMLSession(ep="cpu", device="cpu")` produces a session whose `ep_name == "CPUExecutionProvider"` even on systems with OV registered (the original bug).
- A test that `ep_name` returns `None` before `compile()` and a string after.

These pin down the new contract so it can't regress silently.

🤖 Generated with Claude Code

src/winml/modelkit/commands/perf.py
```diff
@@ -298,6 +300,7 @@ def run(self) -> BenchmarkResult:
             self._model.io_config,
             task=self._model.task or self.config.task,
             device=self._model.device,
+            ep_name=self._model._session.ep_name,
```
timenick: Minor consistency: `WinMLPreTrainedModel` exposes `device` as a public property (used on the line above as `self._model.device`), but `ep_name` is accessed via the private `self._model._session.ep_name`. Adding an `ep_name` property on `WinMLPreTrainedModel` that delegates to `self._session.ep_name` would mirror the `device` pattern and keep callers off the underscore-prefixed attribute.

Non-blocking — `self._model._session` is already accessed elsewhere in this file, so the inconsistency predates this PR.

🤖 Generated with Claude Code

src/winml/modelkit/session/session.py
DingmaomaoBJTU: Fixed in 4da20b7 — rewrote the `_build_session_options` docstring to match the new contract, and added `--ep cpu --device cpu` to the `Exception:` example.

src/winml/modelkit/commands/perf.py
DingmaomaoBJTU: Added `WinMLPreTrainedModel.ep_name` delegation in 4da20b7; both call sites now use `self._model.ep_name` instead of reaching through `_session`.

src/winml/modelkit/session/session.py
DingmaomaoBJTU: Fixed in 4da20b7 — same change as the earlier thread on this hunk: docstring rewritten and `--ep cpu --device cpu` added to the `Exception:` example.

src/winml/modelkit/session/session.py
DingmaomaoBJTU: Added 3 regression tests in 4da20b7 (`tests/unit/session/test_winml_session.py`): `test_ep_name_is_none_before_compile`, `test_ep_name_after_compile`, and `test_explicit_ep_cpu_binds_cpu_execution_provider` — the last one pins the bug this PR fixes.

src/winml/modelkit/commands/perf.py
DingmaomaoBJTU: Added the `WinMLPreTrainedModel.ep_name` delegation in 4da20b7; both perf.py call sites switched to `self._model.ep_name`.

timenick (approved):

xieofxie (approved):

src/winml/modelkit/commands/perf.py
```diff
@@ -121,6 +121,7 @@ class BenchmarkResult:
     # Actual values used (after auto-detection)
     actual_device: str = ""
     actual_task: str = ""
+    actual_ep: str | None = None
```
xieofxie: EPName type

src/winml/modelkit/commands/perf.py
```diff
@@ -880,11 +886,13 @@ def _print_model_info(
     *,
     task: str | None = None,
     device: str = "auto",
+    ep_name: str | None = None,
```
xieofxie: EPName type

src/winml/modelkit/models/winml/base.py
```diff
@@ -214,6 +214,11 @@ def device(self) -> str:
         """Current device (delegates to session, resolved after compile)."""
         return self._session.device
 
+    @property
+    def ep_name(self) -> str | None:
```
xieofxie: EPName type

src/winml/modelkit/session/session.py
```diff
@@ -685,6 +692,19 @@ def device(self) -> str:
...
+        """Primary EP ORT actually bound, or None before compile.
+
+        Returns ``session.get_providers()[0]`` — the EP that owns node
+        partitioning. ``CPUExecutionProvider`` may still appear later
+        in the list as ORT's automatic fallback for unsupported ops.
+        """
+        if self._session is None:
+            return None
+        providers = self._session.get_providers()
+        return providers[0] if providers else None
```
xieofxie: you could cast

src/winml/modelkit/commands/perf.py
DingmaomaoBJTU: Fixed in c97e81e — `BenchmarkResult.actual_ep` now typed `EPName | None`.

src/winml/modelkit/commands/perf.py
DingmaomaoBJTU: Fixed in c97e81e — `_print_model_info`'s `ep_name` parameter now typed `EPName | None`.

src/winml/modelkit/models/winml/base.py
DingmaomaoBJTU: Fixed in c97e81e — `WinMLPreTrainedModel.ep_name` return type now `EPName | None`.

src/winml/modelkit/session/session.py
DingmaomaoBJTU: Done in c97e81e — return uses `cast("EPName", providers[0])` so the property's `EPName | None` annotation type-checks against ORT's runtime `str`.

xieofxie (approved):

PR #626, opened 2026-05-14
fix(eval): only apply w8a16 precision default on NPU, not CPU/GPU

DingmaomaoBJTU:
## Summary

- The eval runner was unconditionally passing `--precision w8a16` to `winml config` for all devices
- On GPU, this causes a QDQ-quantized model to be built; when ORT creates an InferenceSession with QNN GPU EP, the layout transformer inserts `com.ms.internal.nhwc` Conv nodes as requested by QNN GPU — but QNN GPU's `GetCapability` does not claim those nodes for QDQ-wrapped Conv patterns, leaving the graph invalid
- Float32 (non-quantized) models are unaffected; the same model runs fine on QNN NPU

**Fix:** introduce `_resolve_precision(device, explicit)` which returns `w8a16` only for `npu`, and `None` (no `--precision` flag passed) for `cpu`/`gpu`, letting `winml config` auto-detect the appropriate quantization for those devices. Per-model explicit precision in the eval list is always honoured.

### DIFF
--- a/scripts/e2e_eval/run_eval.py
+++ b/scripts/e2e_eval/run_eval.py
@@ -79,7 +79,22 @@ BASELINE_CACHE_PATH = Path(__file__).parent / "cache" / "baseline_cache.json"
 EVAL_DATASETS_CACHE = Path.home() / ".cache" / "winml" / "eval_datasets"
 TIMEOUT_SKIP_LIST_PATH = Path(__file__).parent / "cache" / "timeout_skip_list.json"
 _DEFAULT_SAMPLES = 1000
-_DEFAULT_PRECISION = "w8a16"
+_DEFAULT_PRECISION_NPU = "w8a16"
+
+
+def _resolve_precision(device: str, explicit: str | None) -> str | None:
+    """Return the precision to pass to winml config/perf.
+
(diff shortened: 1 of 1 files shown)


scripts/e2e_eval/run_eval.py
```diff
@@ -529,16 +544,17 @@ def run_model(
     """
     if not onnx_paths:
         # No pre-built paths: fall back to HF model ID (single model only)
+        precision = _resolve_precision(device, None)
```
timenick, line 548: Minor (non-blocking): the PR description says *"Per-model explicit precision in the eval list is always honoured"*, but this fallback path calls `_resolve_precision(device, None)` and silently drops `entry.precision`. The `_run_build` path forwards it correctly.

This is **pre-existing behavior** (the old code used `_DEFAULT_PRECISION` here too, not `entry.precision`), and in practice this branch may never execute since `main()` always calls `_run_build` first — so this is a latent inconsistency rather than a regression introduced by the PR. Worth a one-line fix for consistency though:

```python
precision = _resolve_precision(device, entry.precision)
```

🤖 Generated with Claude Code

timenick (approved):

xieofxie (approved):

PR #628, opened 2026-05-15
chore: remove CUDA EP support

### LINKED ISSUE #631: hide cuda ep
should we do it? Not supported in winml now

xieofxie:
Closes https://github.com/microsoft/winml-cli/issues/631

## Summary

CUDA is not part of our current target EP set, so disable the CUDA paths by deleting them.

## Source changes

| File | What |
|---|---|
| `utils/constants.py` | Drop `"CUDAExecutionProvider"` from `EPName` and `"cuda"` from `EPAlias`; drop matching entries in `EP_ALIASES` and `EP_NAME_TO_ALIAS` |
| `sysinfo/device.py` | Drop CUDA entry from `_EP_DEVICE_MAP` |
| `config/precision.py` | Drop CUDA entry from `_EP_TO_DEVICE` |
| `compiler/configs.py` | Drop CUDA factory dict entry and the `for_cuda` classmethod; update docstring |
| `export/htp/config_generator.py` | Drop `"cuda"` from `target_deployment` Literal, the `cuda` branch in `generate()`, and the `generate_for_cuda` helper |

## Test changes

- `compiler/test_compiler_configs.py` — comment out `test_for_cuda`; drop `"cuda"` from the `for_provider` parametrize and the direct-factory list; change the alternate-provider test from `"cuda"` to `"dml"`.
- `session/conftest.py` — drop `"cuda"` from `EP_NAME_MAP`; update the `ep` marker docstring.
- `e2e/test_session.py` — comment out the `cuda` `pytest.param` and drop it from `ids`.

## Out of scope (untouched on purpose)

- `_warnings.py` — filters `*CUDA*` warnings from diffusers, not CUDA-support code
- `analyze/onnx_opset/_impl/opset_com_microsoft1.py` — auto-generated ONNX schema text
- `utils\optimum_loader.py`
- `sys command`

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -22,7 +22,7 @@ jobs:
         uses: astral-sh/setup-uv@v4
 
       - name: Set up Python
-        run: uv python install 3.10
+        run: uv python install 3.11
 
       - name: Install dependencies
         run: uv sync --all-extras
--- a/.github/workflows/modelkit-ci.yml
+++ b/.github/workflows/modelkit-ci.yml
@@ -34,7 +34,7 @@ jobs:
           - group: remaining
             paths: >-
               tests/unit/core tests/unit/onnx tests/unit/cache
-              tests/unit/utils tests/unit/sysinfo tests/unit/inspect
+              tests/unit/utils tests/unit/test_helpers tests/unit/sysinfo tests/unit/inspect
               tests/unit/optracing tests/regression tests/cli
 
     name: test (${{ matrix.group }})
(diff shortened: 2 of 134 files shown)


timenick:
I don't think it's necessary to comment out the code; we can simply remove the CUDA-related parts directly. Since we have git history, we can always track these changes if needed.

xieofxie:
> I don't think it's necessary to comment out the code; we can simply remove the CUDA-related parts directly. Since we have git history, we can always track these changes if needed.

I am thinking we just remove it from ep click option

timenick (approved):

PR #632, opened 2026-05-15
Strong-type EP parameters across analyze/compiler/optracing

xieofxie:
## Summary

Replace `ep: str` with the Literal-based types defined in `utils/constants.py` (`EPName`, `EPAlias`) so EP parameters are checked statically rather than relying on runtime guards.

- **`EPName` (canonical, e.g. `"QNNExecutionProvider"`)** — post-normalization internal APIs: `RuntimeChecker`, `InformationEngine`, `_has_parquet_in_search_dir`, `OpTraceResult.ep` (`EPName | Literal[""]`).
- **`EPAlias` (shorthand, e.g. `"qnn"`)** — compiler pipeline that carries aliases via `CompileContext`: `GraphTransform.applies_to`, `get_transforms_for_ep`, `needs_format_conversion`, `CompileContext.execution_provider`.
- Drop the dead `if not ep or not ep.strip()` runtime guard in `RuntimeChecker.__init__` and the corresponding `test_requires_non_empty_ep` — the Literal annotation enforces the same constraint at type-check time, and the empty-string narrowing was widening `EPName` back to `str` for pyright.
- Minor: remove an unnecessary `ep.lower()` call in `session/session.py` so canonical EP strings flow through unchanged.


src/winml/modelkit/session/session.py
```diff
@@ -242,7 +242,7 @@ def __init__(
 
         # HF Pipeline may pass torch.device; coerce to string for downstream .lower() calls
         self._device = str(device) if not isinstance(device, str) else device
-        self._ep = ep.lower() if ep else None
+        self._ep: EPNameOrAlias | None = ep if ep else None
```
xieofxie: bug fix in 0.0.3

DingmaomaoBJTU (approved):

PR #638, opened 2026-05-15
fix compile device parameter

xieofxie:
- enable vitisai compilation
- remove hardcoded _DEVICE_TO_PROVIDER
- add resolve_eps


DingmaomaoBJTU (approved):

src/winml/modelkit/config/precision.py
```diff
@@ -300,7 +281,7 @@ def resolve_precision(
     compile_provider: EPName | None = (
         ep_canonical
         if (ep_canonical and ep_canonical != "CPUExecutionProvider")
-        else _DEVICE_TO_PROVIDER.get(resolved_device)
+        else None
```
timenick: **Regression: breaks `resolve_precision` for explicit devices.**

Changing the else-branch from `_DEVICE_TO_PROVIDER.get(resolved_device)` to `None` means `resolve_precision(device="npu")` now returns `compile_provider=None` instead of `"QNNExecutionProvider"`. Same for `device="gpu"` (was `DmlExecutionProvider`).

Downstream impact: `config/build.py:358` calls `WinMLCompileConfig.for_provider(policy.compile_provider, device=policy.device)`. With `None`, `for_provider` short-circuits at line 111-112 and returns `None`, so the entire `winml build` compile step is dropped for explicit-device-no-`--ep` runs.

Running `uv run pytest tests/unit/config/test_precision.py` on this branch yields **17 failures**, all of the form `assert None == 'QNNExecutionProvider'` / `'DmlExecutionProvider'`:
- `TestResolvePrecision::test_resolve_precision_matrix[npu-*]` (8 cases)
- `TestResolvePrecision::test_resolve_precision_matrix[gpu-*]` (6 cases)
- `TestEpOverride::test_ep_overrides_default_dml`
- `TestEpOverride::test_ep_none_uses_default_mapping`
- `TestMixedPrecisionAutoDevice::test_w8a16_auto_npu_full_policy`

If `_DEVICE_TO_PROVIDER` is being removed by design, `resolve_precision` needs an equivalent device→provider fallback (e.g. by calling `resolve_eps(resolved_device)[0]` here too), and the tests need to be updated rather than implicitly broken.

🤖 Generated with Claude Code

src/winml/modelkit/sysinfo/device.py
```diff
@@ -231,3 +232,9 @@ def resolve_device(
             sorted(available_eps),
         )
     return device, available_devices
+
+
+def resolve_eps(resolved_device: str) -> list[EPName]:
```
timenick, line 237: **New public symbol must be exported in `sysinfo/__init__.py`.**

Per CLAUDE.md import rules:

> Every symbol that external code needs must be exported in the package's `__init__.py`. When adding new public classes/functions, add them to both the import and `__all__`.

`resolve_eps` is public (no underscore prefix) but isn't added to [`src/winml/modelkit/sysinfo/__init__.py`](src/winml/modelkit/sysinfo/__init__.py), which currently re-exports `resolve_device` and `get_ep_device_map`. The PR itself reaches in via `from ..sysinfo.device import resolve_device, resolve_eps` rather than the package surface — that import should be `from ..sysinfo import resolve_device, resolve_eps` once the function is properly exported.

Also style nits on this signature line:
- Missing space before `->`: should be `def resolve_eps(resolved_device: str) -> list[EPName]:`. `uvx ruff format --check` flags it.
- Body can be a comprehension (ruff `PERF401`): `return [ep for ep in _DEVICE_EP_MAP.get(resolved_device, []) if ep in available_eps]`.
- `resolved_device` is not normalized — call `.lower()` (or document that callers must lowercase) so that `"NPU"` doesn't silently return `[]`.

🤖 Generated with Claude Code

tests/unit/commands/test_compile_quantize_flags.py
```diff
@@ -20,25 +20,41 @@
...
+    ``_resolve_compile_provider`` now consults ``resolve_eps`` (which
+    intersects ``_DEVICE_EP_MAP`` with the EPs ORT actually advertises on
+    the current host). Tests mock ``resolve_eps`` so they don't depend on
+    the runner's hardware.
+    """
 
-    def test_gpu_defaults_to_dml(self):
-        assert _resolve_compile_provider("gpu", None) == "DmlExecutionProvider"
+    def test_npu_picks_first_available_ep(self):
+        with patch(
```
timenick: **Test coverage shrinks instead of growing.**

The removed `test_auto_defaults_to_qnn` isn't replaced — there's now no test pinning what `--device auto` resolves to under the new pipeline. The new `resolve_eps()` function in `sysinfo/device.py` also has zero unit-test coverage.

Please add (under `tests/unit/sysinfo/test_device.py`, using the existing mock pattern in `TestGetAvailableDevices` that patches `_get_available_eps`):

- `resolve_eps("npu")` returns the expected priority list when multiple NPU EPs are available
- `resolve_eps("gpu")` returns DML when only DML is available (the common Windows case)
- `resolve_eps("NPU")` — case sensitivity
- `resolve_eps("unknown")` returns `[]`

And in `test_compile_quantize_flags.py`, replace the removed test with one that covers the new `resolve_device` → `_resolve_compile_provider` pipeline end-to-end (mock the available EPs so the test is deterministic regardless of host).

Without these, the failures the existing suite catches today (see other comments) won't have a regression net going forward.

🤖 Generated with Claude Code

src/winml/modelkit/commands/compile.py
```diff
@@ -204,9 +206,7 @@ def compile(
 
     # Show info
     console.print(f"[bold blue]Input:[/bold blue] {model}")
-    console.print(f"[bold blue]Device:[/bold blue] {device}")
-    if ep:
-        console.print(f"[bold blue]EP:[/bold blue] {ep}")
+    console.print(f"[bold blue]Device:[/bold blue] {resolved_device}")
```
timenick, line 209: **User-visible information loss: `--ep` no longer displayed.**

The old output showed both the user-supplied `--device` flag and, when present, the explicit `--ep` override:

```
Device: gpu
EP: migraphx
Provider: MIGraphXExecutionProvider
```

This PR removes the `EP:` line entirely. When a user explicitly overrides the EP with `--ep`, they should still see that confirmation in the summary — otherwise it's not obvious whether their override took effect (the `Provider:` line shows the canonical name, not the alias they typed).

Suggest keeping the `if ep:` block:

```python
console.print(f"[bold blue]Device:[/bold blue] {resolved_device}")
if ep:
    console.print(f"[bold blue]EP:[/bold blue] {ep}")
console.print(f"[bold blue]Provider:[/bold blue] {provider}")
```

🤖 Generated with Claude Code

src/winml/modelkit/commands/compile.py
```diff
@@ -25,8 +25,8 @@
 import click
 from rich.console import Console
```
timenick, line 27: **Lint failures: import grouping + ruff format/style.**

Three lint issues from `uvx ruff check` / `uvx ruff format --check` on this branch:

1. **Here (`compile.py:27`)** — `from ..sysinfo.device import resolve_device, resolve_eps` is placed adjacent to the third-party `from rich.console import Console` without a blank-line separator, mixing the two import groups. Ruff `I001` (import sorting) fails. The local import block starts at line 29 and that's where this line belongs.

2. **`sysinfo/device.py:237`** — `def resolve_eps(resolved_device: str)-> list[EPName]:` is missing the space before `->`. `uvx ruff format --check --diff` flags it.

3. **`sysinfo/device.py:240-243`** — Ruff `PERF401`: replace the for/append loop with a comprehension:
   ```python
   return [ep for ep in _DEVICE_EP_MAP.get(resolved_device, []) if ep in available_eps]
   ```

CLAUDE.md requires `uv run ruff check --fix` after Python edits — please run it and the formatter before merging.

🤖 Generated with Claude Code

timenick (commented):
Reviewed against project standards. The PR introduces multiple regressions confirmed by running the existing pytest suite on this branch: 17 failures in `tests/unit/config/test_precision.py` and 4 failures in `tests/unit/commands/test_compile_quantize_flags.py`. Detailed inline comments below.

🤖 Generated with Claude Code

src/winml/modelkit/commands/compile.py
xieofxie: add back

src/winml/modelkit/config/precision.py
xieofxie: good catch, use resolve_eps to replace _DEVICE_TO_PROVIDER

DingmaomaoBJTU:
### Code review

Found 2 issues:

1. `resolve_eps` bypasses `sysinfo/__init__.py` — both `compile.py` and `precision.py` import it directly from the internal submodule. [CLAUDE.md](https://github.com/microsoft/winml-cli/blob/925d11df5ff295b0e7d2166c0a58c297c7b883c0/CLAUDE.md) says: "Every symbol that external code needs must be exported in the package's `__init__.py`. When adding new public classes/functions, add them to both the import and `__all__`." and "Relative imports through `__init__.py`: `from ..onnx import ONNXDomain` (not `from winml.modelkit.onnx.domains import ONNXDomain`)". Fix: add `resolve_eps` to `sysinfo/__init__.py`'s import and `__all__`, then use `from ..sysinfo import resolve_eps` in both callers.

https://github.com/microsoft/winml-cli/blob/925d11df5ff295b0e7d2166c0a58c297c7b883c0/src/winml/modelkit/commands/compile.py#L28-L30

https://github.com/microsoft/winml-cli/blob/925d11df5ff295b0e7d2166c0a58c297c7b883c0/src/winml/modelkit/config/precision.py#L16-L18

2. `_EP_DEVICE_MAP` reordering puts `OpenVINOExecutionProvider` before `DmlExecutionProvider` in the GPU EP list. Since `resolve_eps` returns EPs in insertion order and `_resolve_compile_provider` picks `eps[0]`, on a Windows machine where both OpenVINO and DML are available (but no NVIDIA/AMD GPU EPs), `--device gpu` now defaults to OpenVINO instead of DML. DML is the native Windows GPU EP and should take priority over OpenVINO for the `gpu` device.

https://github.com/microsoft/winml-cli/blob/925d11df5ff295b0e7d2166c0a58c297c7b883c0/src/winml/modelkit/sysinfo/device.py#L53-L59

xieofxie:
> ### Code review
> Found 2 issues:
> 
> 1. `resolve_eps` bypasses `sysinfo/__init__.py` — both `compile.py` and `precision.py` import it directly from the internal submodule. [CLAUDE.md](https://github.com/microsoft/winml-cli/blob/925d11df5ff295b0e7d2166c0a58c297c7b883c0/CLAUDE.md) says: "Every symbol that external code needs must be exported in the package's `__init__.py`. When adding new public classes/functions, add them to both the import and `__all__`." and "Relative imports through `__init__.py`: `from ..onnx import ONNXDomain` (not `from winml.modelkit.onnx.domains import ONNXDomain`)". Fix: add `resolve_eps` to `sysinfo/__init__.py`'s import and `__all__`, then use `from ..sysinfo import resolve_eps` in both callers.
> 
> https://github.com/microsoft/winml-cli/blob/925d11df5ff295b0e7d2166c0a58c297c7b883c0/src/winml/modelkit/commands/compile.py#L28-L30
> 
> https://github.com/microsoft/winml-cli/blob/925d11df5ff295b0e7d2166c0a58c297c7b883c0/src/winml/modelkit/config/precision.py#L16-L18
> 
> 2. `_EP_DEVICE_MAP` reordering puts `OpenVINOExecutionProvider` before `DmlExecutionProvider` in the GPU EP list. Since `resolve_eps` returns EPs in insertion order and `_resolve_compile_provider` picks `eps[0]`, on a Windows machine where both OpenVINO and DML are available (but no NVIDIA/AMD GPU EPs), `--device gpu` now defaults to OpenVINO instead of DML. DML is the native Windows GPU EP and should take priority over OpenVINO for the `gpu` device.
> 
> https://github.com/microsoft/winml-cli/blob/925d11df5ff295b0e7d2166c0a58c297c7b883c0/src/winml/modelkit/sysinfo/device.py#L53-L59
> 
> 

the 2nd change is by design. We prefer ep by IHV instead of builtin

src/winml/modelkit/commands/compile.py
xieofxie: fixed

tests/unit/commands/test_compile_quantize_flags.py
xieofxie: added one with auto

src/winml/modelkit/sysinfo/device.py
xieofxie: fixed

src/winml/modelkit/sysinfo/device.py
```diff
@@ -50,10 +51,10 @@
     "VitisAIExecutionProvider": "npu",
     # Qualcomm (QNN supports both NPU and GPU via Adreno backend)
     "QNNExecutionProvider": "npu/gpu",
-    # Microsoft
-    "DmlExecutionProvider": "gpu",
     # Intel
     "OpenVINOExecutionProvider": "npu/gpu/cpu",
```
DingmaomaoBJTU, line 55: Placing `OpenVINOExecutionProvider` before `DmlExecutionProvider` changes `_DEVICE_EP_MAP["gpu"]` priority from `[..., DML, OpenVINO]` to `[..., OpenVINO, DML]`. Since `_resolve_compile_provider` now calls `resolve_eps(resolved_device)[0]`, on any machine with OpenVINO installed alongside DML, `--device gpu` without `--ep` will silently default to OpenVINO instead of DML. DML is the universal Windows GPU EP; OpenVINO is Intel-SDK-specific. Notably, the fixture comment in `conftest.py` explicitly warns that "OpenVINO would otherwise out-rank QNN/DML/CPU under the dynamic resolution in `resolve_precision`" and mocks `resolve_eps` to prevent it in tests — but production is unprotected. DML should remain above OpenVINO in this map.

src/winml/modelkit/config/precision.py
```diff
@@ -297,11 +279,12 @@ def resolve_precision(
...
     # For all other explicit EPs (canonical names), use ep as the provider.
-    compile_provider: EPName | None = (
-        ep_canonical
-        if (ep_canonical and ep_canonical != "CPUExecutionProvider")
-        else _DEVICE_TO_PROVIDER.get(resolved_device)
-    )
+    compile_provider: EPName | None =  ep_canonical
+    if not compile_provider:
+        eps = resolve_eps(resolved_device)
+        compile_provider = eps[0] if eps else None
```
DingmaomaoBJTU, line 285: On a machine with OpenVINO installed, `resolve_eps("cpu")` returns `["OpenVINOExecutionProvider", "CPUExecutionProvider"]` because OpenVINO maps to `"npu/gpu/cpu"` in `_EP_DEVICE_MAP`. `eps[0]` is therefore `"OpenVINOExecutionProvider"`, and the guard below only clears `compile_provider` when it equals `"CPUExecutionProvider"`, so OpenVINO silently becomes the `compile_provider` for `--device cpu` builds. Before this PR, `_DEVICE_TO_PROVIDER["cpu"] = None` ensured CPU always skipped EPContext compilation. Fix: short-circuit to `None` for `resolved_device == "cpu"` before calling `resolve_eps`.

src/winml/modelkit/commands/compile.py
```diff
@@ -25,8 +25,8 @@
...
 
-from ..config.precision import _DEVICE_TO_PROVIDER
 from ..onnx import is_compiled_onnx
+from ..sysinfo.device import resolve_device, resolve_eps
```
DingmaomaoBJTU: `resolve_eps` is imported directly from the internal submodule `..sysinfo.device` instead of through the package `__init__.py`. [CLAUDE.md](https://github.com/microsoft/winml-cli/blob/925d11df5ff295b0e7d2166c0a58c297c7b883c0/CLAUDE.md) requires: "Every symbol that external code needs must be exported in the package's `__init__.py`" with imports like `from ..sysinfo import resolve_device, resolve_eps`. Fix: add `resolve_eps` to `sysinfo/__init__.py`'s import and `__all__`, then change this to `from ..sysinfo import resolve_device, resolve_eps`.

src/winml/modelkit/config/precision.py
```diff
@@ -14,6 +14,7 @@
 import re
 from dataclasses import dataclass
 
+from ..sysinfo.device import resolve_eps
```
DingmaomaoBJTU: Same import rule violation as `compile.py:29` — `resolve_eps` is imported directly from the internal submodule rather than through `sysinfo/__init__.py`. Change to `from ..sysinfo import resolve_eps` after exporting it from `sysinfo/__init__.py`.

src/winml/modelkit/config/precision.py
xieofxie: this is expected. we 
- prefer ihv ep over built in
- we skip compiling for cpu ep, not cpu device

src/winml/modelkit/compiler/configs.py
```diff
@@ -174,7 +174,7 @@ def for_dml(cls) -> WinMLCompileConfig:
     def for_nv_tensorrt_rtx(cls) -> WinMLCompileConfig:
         """Factory for NvTensorRTRTX compilation."""
         return cls(
-            ep_config=EPConfig(provider="nv_tensorrt_rtx", enable_ep_context=False),
+            ep_config=EPConfig(provider="nv_tensorrt_rtx", enable_ep_context=True),
```
DingmaomaoBJTU: Do we need provider options here？

xieofxie: do we? I didn't test

src/winml/modelkit/session/session.py
```diff
@@ -303,9 +224,6 @@ def compile(self) -> None:
             logger.info("Using cached EPContext: %s", ctx_path)
 
         # Compile if needed (persist_jit=True and no cache)
-        # Native QNN SDK compiler writes progress to stdout/stderr;
```
xieofxie, line 306: will we support it in the future? now no one tests it seems like

PR #641, opened 2026-05-15
CLI: EP_SUPPORTED_DEVICES policy + trust_remote_code warning

### LINKED ISSUE #513: [winml analyze] [P1] `--ep cpu` silently changes target to NPU then exits 1
**Severity**: P1  
**Command**: `winml analyze`

---
**Category**: Conflicting Flag Combinations — when flags conflict, reject at entry; never silently override.

**Repro**:
```
uv run winml analyze -m temp\cli-audit\resnet.onnx --ep cpu
```

**Actual**:
```
Target: CPUExecutionProvider on NPU
WARNING: No runtime check data for CPUExecutionProvider on NPU — skipping op analysis.
No runtime check results for CPUExecutionProvider on NPU — no rule data available.
```
Exit 1. The user passed `--ep cpu` and did **not** pass `--device`, but the target line says `on NPU`. Then the rule lookup fails because the (CPU, NPU) pair is impossible, and the whole command exits 1.

**Expected**: When `--ep cpu` is given without `--device`, default `--device cpu`. When `--ep cpu --device npu` is given explicitly, reject at entry:
```
Error: --ep cpu cannot run on --device npu. CPUExecutionProvider runs on CPU.
```
Exit 2.

**Why it matters**: Silent override means the command appears to "do something" but produces no result. The user has no way to know their flag was ignored.

### LINKED ISSUE #516: [winml build] [P1] `--trust-remote-code` runs custom code with no warning
**Severity**: P1  
**Command**: `winml build`

---
**Category**: Security Flag Warning — flags that loosen security must emit a warning to stderr before doing the loosened operation.

**Repro**:
```
uv run winml build --trust-remote-code -m microsoft/resnet-50 -c temp\ux\build_cfg.json -o temp\ux\trc_out
```

**Actual**: stderr contains only the normal Setup banner and download progress; **no** warning about `--trust-remote-code` is emitted before the model is downloaded and its custom Python is executed. Filtering stderr for `WARNING|trust` returns zero matches.

**Expected**: Emit a warning to stderr **before** any download starts, not suppressed by `-q/--quiet`:
```
WARNING: --trust-remote-code allows running custom Python from microsoft/resnet-50.
         Proceed only if you trust the publisher.
```
Same warning should fire for `winml config --trust-remote-code` and `winml inspect --trust-remote-code`.

**Why it matters**: `--trust-remote-code` causes `transformers` to import and execute arbitrary Python from the model repo. Without a visible warning at the call site, this becomes an attack vector: a user copy-pasting a `winml` command from a blog post may not realize they just authorized arbitrary code execution.

### LINKED ISSUE #521: [winml compile] [P0] `--device cpu --ep qnn` silently overrides device to NPU and exits 0
**Severity**: P0  
**Command**: `winml compile`

---
**Category**: Conflicting Flag Combinations — when flags conflict, reject at entry; never silently override.

**Repro**:
```
uv run winml compile -m temp\cli-audit\resnet.onnx --device cpu --ep qnn -o temp\ux\out_cpu_qnn.onnx
```

**Actual** (also reproduces with `--device gpu --ep qnn`):
```
Input:    temp\cli-audit\resnet.onnx
Device:   npu          ← user passed cpu
EP:       qnn
Success! Model compiled
```
Exit 0. A 51 MB `out_cpu_qnn.onnx` is written and the user has no idea it was compiled for NPU.

**Expected**: At entry:
```python
if device == "cpu" and ep == "qnn":
    raise click.ClickException(
        "--device cpu and --ep qnn are incompatible. QNN runs on NPU only. "
        "Either pass --device npu, or use --ep CPUExecutionProvider for CPU."
    )
```
Exit 2.

**Why it matters**: Silent device override produces an artifact that does not match what the user asked for. If the user then ships that artifact thinking it runs on CPU, deployment fails in production.

vortex-captain:
## Summary

Batch fix in the ``reny/ep_device_pair`` line of work, addressing three
related CLI papercuts. Two commits:

1. ``EP_SUPPORTED_DEVICES`` policy + ``--ep``/``--device`` coordination
2. ``--trust-remote-code`` / ``trust_remote_code=True`` security warning

Fixed #513 — ``winml analyze --ep cpu`` silently falls back to ``"NPU"`` for ``--device``.

Fixed #521 — ``winml compile --device cpu --ep qnn`` silently overrides ``cpu`` and produces a QNN/NPU artifact.

Fixed #516 — ``--trust-remote-code`` runs custom Python from the model repo with no visible warning.

---

## Commit 1: ``EP_SUPPORTED_DEVICES`` policy table

Introduces a single canonical EP -> ordered supported devices table in
``utils/constants.py`` and threads it through ``analyze`` and ``compile``.

```python
EP_SUPPORTED_DEVICES: dict[EPName, tuple[str, ...]] = {
    "QNNExecutionProvider":           ("npu", "gpu"),
    "VitisAIExecutionProvider":       ("npu",),
    "OpenVINOExecutionProvider":      ("gpu", "npu", "cpu"),
    "DmlExecutionProvider":           ("gpu",),
    "MIGraphXExecutionProvider":      ("gpu",),
    "NvTensorRTRTXExecutionProvider": ("gpu",),
    "CUDAExecutionProvider":          ("gpu",),
    "CPUExecutionProvider":           ("cpu",),
}
```

- **First element = canonical default device** when only ``--ep`` is given.
- **Whole tuple = supported set** used to reject incompatible explicit pairs.

OpenVINO's default device kept as ``gpu`` (preserves the prior ``_EP_TO_DEVICE`` pick).

### Map consolidation

Two pre-existing maps are now derived from ``EP_SUPPORTED_DEVICES``:

| Previous location | New form |
|---|---|
| ``config/precision.py::_EP_TO_DEVICE`` | **removed**; all 3 call sites in ``resolve_precision`` now read ``EP_SUPPORTED_DEVICES`` directly |
| ``sysinfo/device.py::_EP_DEVICE_MAP`` | derived as ``"/"``-joined string for the legacy ``get_ep_device_map()`` accessor only; internal uses (``_DEVICE_EP_MAP`` builder, ``resolve_device``) now consume tuples directly |

``config/precision.py::_DEVICE_TO_PROVIDER`` is now total over its key set (``auto``/``cpu``/``gpu``/``npu`` all map to a real EP name). ``cpu`` maps to ``"CPUExecutionProvider"`` (was ``None``); ``resolve_precision`` keeps the "CPU produces no EPContext" semantic by nullifying ``compile_provider`` when it resolves to ``CPUExecutionProvider``.

### Behavioral fixes

**Analyze** (``commands/analyze.py``):
- ``device_option`` default changed from ``"NPU"`` to ``None``; the original ``has_rule_data_for_ep`` validation block is preserved untouched.
- After it, a small block fills ``device`` from the (ep, device) pair:
  - ``(None, None)`` -> ``"NPU"`` (legacy fallback; analyzer scans all EPs)
  - ``(ep, None)`` -> ``EP_SUPPORTED_DEVICES[ep][0].upper()``
  - ``(*, user value)`` -> respected as-is
- ``analyze/analyzer.py``'s "all EPs" list now includes ``NvTensorRTRTXExecutionProvider`` so it gets analyzed when ``--ep`` is omitted.

**Compile** (``commands/compile.py``):
- Callback infers ``device`` from ``--ep`` (via ``EP_SUPPORTED_DEVICES[ep][0]``) when ``--device`` was not CLI-provided.
- ``_resolve_compile_provider`` now:
  - rejects incompatible explicit pairs with ``click.UsageError`` (e.g. ``--device cpu --ep qnn``).
  - rejects devices outside ``{auto, cpu, gpu, npu}`` (the old QNN fallback for unknown devices is gone).
  - returns ``CPUExecutionProvider`` for ``--device cpu`` (instead of branching on a ``None`` from the map).

### E2E verification

| Invocation | Before | After |
|---|---|---|
| ``winml analyze -m x.onnx --ep cpu`` | ``Target: ... on NPU`` exit 1, no rule data | ``Target: CPUExecutionProvider on CPU`` |
| ``winml analyze -m x.onnx --ep cpu --device npu`` | rejected by rule-data block, exit 2 | unchanged (rule-data block preserved) |
| ``winml compile -m x.onnx --device cpu --ep qnn`` | ``Success! Model compiled`` exit 0 (wrong artifact) | ``Error: --ep qnn cannot run on --device cpu.`` exit 2 |
| ``winml compile -m x.onnx --ep dml`` | infers ``--device npu``, fails at config build | infers ``--device gpu``, clearer error |

---

## Commit 2: ``trust_remote_code`` security warning

Surfaces a bold-red stderr warning whenever ``trust_remote_code=True`` is honoured, covering both the CLI flag and the Python API kwarg.

```
WARNING: trust_remote_code is enabled - custom Python from 'microsoft/resnet-50'
         will be downloaded and executed. Proceed only if you trust the publisher.
```

### Wiring

- **CLI**: a click callback attached to ``--trust-remote-code`` fires the warning during option parsing — before any download begins. Covers ``winml build``, ``winml config``, ``winml eval`` (all consumers of ``trust_remote_code_option``).
- **API**: ``load_hf_model`` (``loader/hf.py``) and ``generate_hf_build_config`` (``config/build.py``) emit the warning before invoking ``transformers`` with ``trust_remote_code=True``.

Uses a shared stderr ``rich.Console`` so the warning renders in bold red and matches the rest of the CLI's output style. Going through ``Console`` instead of ``logging`` means ``-q/--quiet`` cannot hide it. Emitted on every call.

### E2E verification

```text
$ winml build --trust-remote-code -m fake/nonexistent
WARNING: trust_remote_code is enabled - custom Python will be downloaded ...

$ python -c "from winml.modelkit.loader.hf import load_hf_model; load_hf_model('microsoft/resnet-50', trust_remote_code=True)"
WARNING: trust_remote_code is enabled - custom Python from 'microsoft/resnet-50' will be downloaded ...
```

---

## Test coverage

- ``tests/unit/commands/test_ep_device_pair.py`` (new, 44 cases) — ``EP_SUPPORTED_DEVICES`` invariants + per-EP defaults; ``_resolve_compile_provider`` compat/incompat tables; analyze CLI default-device inference for 5 EPs.
- ``tests/unit/commands/test_trust_remote_code_warning.py`` (new, 10 cases) — helper-level dedup-free behaviour; CLI flag fires for ``build/config/eval``; API entry fires from ``load_hf_model``.
- Extended cases in ``tests/unit/commands/test_compile_quantize_flags.py`` (``test_incompatible_pair_rejected``, ``test_unknown_device_rejected``).
- Existing rule-data, sysinfo, and precision suites continue to cover the derived maps with only minor order tweaks (OpenVINO).

344+ tests pass in the impacted suites. Pre-existing ``TestConfigBadPath`` pollution failures (unrelated logging-stream race) are the only remaining non-green tests; they reproduce on plain ``main``.


xieofxie:
wait for the compile pr https://github.com/microsoft/winml-cli/pull/638

chinazhangchao (approved):

src/winml/modelkit/commands/analyze.py
```diff
@@ -459,7 +459,9 @@ def _render_analysis_summary(
     required=False, optional_message="If not specified, analyzes all supported EPs"
 )
 @cli_utils.device_option(
-    required=False, optional_message="If not specified, uses NPU as default", default="NPU"
+    required=False,
+    optional_message="If not specified, derived from --ep (default: NPU when --ep is omitted)",
+    default=None,
```
xieofxie, line 464: should we use None or a new value?

Also I suggest we typed device parameter first

vortex-captain: IMHO we should use `None` as default: if we use a specific device default, like "NPU", issue like https://github.com/microsoft/winml-cli/issues/513 would persist
Per discussion, let's keep device as str for now

src/winml/modelkit/commands/compile.py
```diff
@@ -265,6 +265,12 @@ def _resolve_compile_provider(resolved_device: str, ep: EPNameOrAlias | None) ->
         canonical = normalize_ep_name(ep)
         if canonical is None:
             raise click.UsageError(f"Unknown EP: {ep}")
+        supported = EP_SUPPORTED_DEVICES.get(canonical)
+        if supported is not None and resolved_device.lower() not in supported:
```
timenick: **nit:** `supported is not None` is unreachable. By this point `canonical` is a valid `EPName` (we returned at line 267 otherwise), and `EP_SUPPORTED_DEVICES.keys() == set(SUPPORTED_EPS)` — `test_keys_match_supported_eps` asserts this invariant. So `.get(canonical)` cannot return `None`. Suggested simplification:

```python
supported = EP_SUPPORTED_DEVICES[canonical]
if resolved_device.lower() not in supported:
    raise click.UsageError(
        f"--ep {ep} cannot run on --device {resolved_device}. "
        f"{canonical} supports: {', '.join(supported)}."
    )
```

Generated with Claude Code

timenick (approved):

src/winml/modelkit/commands/compile.py
```diff
@@ -265,6 +265,12 @@ def _resolve_compile_provider(resolved_device: str, ep: EPNameOrAlias | None) ->
         canonical = normalize_ep_name(ep)
         if canonical is None:
             raise click.UsageError(f"Unknown EP: {ep}")
+        supported = EP_SUPPORTED_DEVICES.get(canonical)
```
zhenchaoni: One missing check — **EP-not-available-on-host** — would fit naturally next to the new device-EP conflict block in `_resolve_compile_provider`:

```python
from ..session.ep_registry import WinMLEPRegistry

registry = WinMLEPRegistry.get_instance()
if not registry.is_ep_available(canonical):
    available = [e for e in EP_SUPPORTED_DEVICES if registry.is_ep_available(e)]
    raise click.UsageError(
        f"--ep {ep} ({canonical}) is not registered on this host. "
        f"Available EPs: {', '.join(available) if available else 'none'}."
    )
```

Without it, `winml compile --ep openvino` on a QNN-only host still proceeds — silently substitutes an EP or fails deep in ORT.

src/winml/modelkit/commands/compile.py
vortex-captain: fixed

src/winml/modelkit/commands/compile.py
vortex-captain: fixed

timenick (approved):

PR #645, opened 2026-05-15
Compile: Add e2e functional validation tests

### LINKED ISSUE #488: Identify functional scenarios and implement E2E tests for `winml compile`
### Description
Identify the functionality list for `winml compile` and implement E2E pytest cases covering it. Tests live in [tests/e2e](../tree/main/tests/e2e).

Coverage must include:

### Deliverables

### Acceptance Criteria
- `uv run pytest tests/e2e/ -k compile` is green.

zhenchaoni:
fix #488

# `winml compile` E2E coverage

EP-context EPs = `qnn`, `openvino` (produce an EPContext model).
Passthrough EPs = `cpu`, `dml` (CLI rejects with explicit error).
Note: vitisai and nv_tensorrt_rtx will be added later

Tests use `require_ep("<name>")` for EP gating: absent EPs auto-skip. QAIRT tests additionally skip if neither `QNN_SDK_ROOT` nor `QAIRT_SDK_ROOT` resolves to an existing directory.

## CLI surface (no EP runtime required)

- `winml compile --help` — every documented option present in stdout
- `winml compile` — fails with "missing option" / "model" error
- `winml compile --list` — exit 0, lists `ort`
- `winml compile --list --device npu` — lists both `ort` and `qairt`
- `winml compile -m FAKE_CTX.onnx` (already-compiled EPContext model) — fails with "already a compiled" / "cannot be re-compiled"

## Happy path — EP-context EPs

For each EP in `{qnn, openvino}`:

- `winml compile -m MODEL --ep <EP> -o OUT/out.onnx`
- Asserts: exit 0, `Success!` banner, valid EPContext artifact with sidecar `.bin`

## Unsupported EPs — explicit error

For each EP in `{cpu, cuda, dml, nv_tensorrt_rtx, vitisai, migraphx}`:

- `winml compile -m MODEL --ep <EP>`
- Asserts: exit ≠ 0, error contains `does not support EPContext compilation`, input ONNX byte-identical (sha256 unchanged)

## Output paths — EP-context EPs

For each EP in `{qnn, openvino}`:

- `winml compile -m MODEL --ep <EP> -o OUT/custom_name.onnx` — file at exact `-o` path
- `winml compile -m MODEL --ep <EP> --output-dir OUT` — exactly one EPContext `*.onnx` produced in `OUT/`

## `--embed` vs sidecar — EP-context EPs

For each EP in `{qnn, openvino}`:

- `winml compile -m MODEL --ep <EP> --embed -o OUT/embedded.onnx` — `embed_mode=1`, inlined bytes
- `winml compile -m MODEL --ep <EP> -o OUT/external.onnx` — `embed_mode=0`, sidecar `.bin` exists alongside

## Validation toggle — EP-context EPs

For each EP in `{qnn, openvino}`:

- `winml compile -m MODEL --ep <EP> --no-validate --verbose -o OUT/no_validate.onnx` — no validation log lines in stdout
- `winml compile -m MODEL --ep <EP>            --verbose -o OUT/validated.onnx`    — validation log lines in stdout

## QAIRT backend (qnn-only, requires QAIRT SDK)

EP: `qnn`. Skips when no QAIRT SDK directory is found.

- `winml compile -m MODEL --ep qnn --compiler qairt --qnn-sdk-root $SDK -o OUT/qairt.onnx --verbose` — banner shows `Compiler: qairt` + SDK root, sidecar EPContext artifact, intermediate `*_qnn_ctx_qnn.bin` and `*_cache_info.json` exist
- same plus `--embed` — `embed_mode=1`, ~28KB QNN payload inlined into the EPContext node

## `--device` → provider resolution (no `--ep`)

Pins `_resolve_compile_provider(ep=None)` for every click-allowed `--device`. Banner is observable even when compile is rejected, so the gpu/cpu rows run on any host.

- `winml compile -m MODEL --device npu -o OUT/npu.onnx` — banner `Device: npu` / `Provider: qnn`; EPContext artifact produced. EP: `qnn`.
- `winml compile -m MODEL --device gpu` — banner `Provider: dml` (gpu maps to dml via `_DEVICE_TO_PROVIDER`, **not** qnn); fails with `does not support EPContext compilation`. EP-agnostic.
- `winml compile -m MODEL --device cpu` — banner `Provider: cpu`; fails with `does not support EPContext compilation`. EP-agnostic.
- `winml compile -m MODEL --device auto -o OUT/auto.onnx` — pins the only path that hits the resolver's `else "qnn"` fall-through (auto is in click's Choice but absent from `_DEVICE_TO_PROVIDER`); banner `Provider: qnn`; EPContext artifact produced. EP: `qnn`.

## Already-compiled rejection — real round-trip

Complements the hand-crafted EPContext fixture with the realistic user mistake. Parametrized over `{qnn, openvino}`.

- `winml compile -m MODEL --ep <EP> -o OUT/first.onnx` then `winml compile -m OUT/first.onnx --ep <EP>` — second invocation fails with `already a compiled` / `cannot be re-compiled`.

## `--config` (`-c`) precedence

Config-file format: JSON. `compile` block keys: `execution_provider`, `compiler`, `embed_context`, `validate`, `verbose`.

- `winml compile -m MODEL -c FULL.json -o OUT/cfg_full.onnx` with config setting all 5 fields (provider=qnn, compiler=ort, embed_context=true, validate=false, verbose=true) — config values applied: `Provider: qnn`, embed=true, no validation logs. EP: `qnn`.
- `winml compile -m MODEL -c NO_EMBED.json --embed --validate --verbose -o OUT/cli_wins.onnx` with config `embed_context: false, validate: false` — CLI flags override config: embed=true, validation logs present. Guards against "config dataclass default wins over explicit CLI flag" regression. EP: `qnn`.
- `winml compile -m MODEL -c CPU.json` with config `execution_provider: cpu` — fails with `does not support EPContext compilation` (unsupported-EP gate fires for provider sourced from config). EP: `cpu`.

### DIFF
--- a/tests/e2e/require_ep.py
+++ b/tests/e2e/require_ep.py
@@ -1,0 +1,66 @@
+# -------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+# --------------------------------------------------------------------------
+"""Runtime EP-availability gate for e2e tests.
+
+Single supported usage — call :func:`require_ep` at the top of a test body
+with one EP name. Skips the test when the EP is not available on the host.
+
+The accepted name follows the same convention as the CLI ``--ep`` flag:
+short alias (``"qnn"``, ``"openvino"``, ``"ov"``, ``"dml"``, ...) or the
+full ORT provider name (``"QNNExecutionProvider"``). Resolution is done
+by :func:`winml.modelkit.utils.constants.normalize_ep_name`, so no local
+mapping lives here.
+
+Single test::
+
+    def test_openvino_specific():
+        require_ep("openvino")
+        ...
+
+Parametrized fan-out — follow the codebase convention of flat
+``@pytest.mark.parametrize`` and gate inside the body::
+
+    @pytest.mark.parametrize("ep", ["qnn", "openvino", "dml", "cpu"])
+    def test_compile_happy_path(ep):
+        require_ep(ep)
+        ...
+"""
+
+from __future__ import annotations
+
+import pytest
+
+
+def require_ep(ep: str) -> str:
+    """Skip the current test unless the requested EP is available.
+
+    Args:
+        ep: EP name. CLI alias (``"qnn"``) or full ORT provider name
+            (``"QNNExecutionProvider"``); both accepted.
+
+    Returns:
+        The full ORT provider name (e.g. ``"QNNExecutionProvider"``) that
+        satisfied the requirement. Handy when the caller wants to pass it
+        on to ``onnxruntime`` / ``WinMLSession``.
+
+    Raises:
+        pytest.skip.Exception: When ``ep`` is unknown or the corresponding
+            provider is not available on the host.
+    """
+    from winml.modelkit.session import WinMLEPRegistry
+    from winml.modelkit.utils.constants import normalize_ep_name
+
+    provider = normalize_ep_name(ep)
+    if provider is None:
+        pytest.skip(f"Unknown EP: {ep!r}")
+
+    # Singleton — first call probes; subsequent calls are free.
+    available = set(WinMLEPRegistry.get_instance().get_available_eps())
+    available.add("CPUExecutionProvider")  # always-on fallback
+
+    if provider not in available:
+        pytest.skip(f"EP not available on this host: {provider}")
+
+    return provider
--- a/tests/e2e/require_ep.py
+++ b/tests/e2e/require_ep.py
@@ -50,17 +50,21 @@ def require_ep(ep: str) -> str:
             provider is not available on the host.
     """
     from winml.modelkit.session import WinMLEPRegistry
-    from winml.modelkit.utils.constants import normalize_ep_name
+    from winml.modelkit.utils import normalize_ep_name
 
     provider = normalize_ep_name(ep)
     if provider is None:
         pytest.skip(f"Unknown EP: {ep!r}")
 
-    # Singleton — first call probes; subsequent calls are free.
-    available = set(WinMLEPRegistry.get_instance().get_available_eps())
-    available.add("CPUExecutionProvider")  # always-on fallback
+    # CPU is always available via ORT itself but is not enumerated by
+    # WinMLEPRegistry, which only tracks WinML-discoverable backend EPs
+    # (QNN, OpenVINO, DML, ...). Short-circuit before consulting the
+    # registry so `require_ep("cpu")` doesn't spuriously skip.
+    if provider == "CPUExecutionProvider":
+        return provider
 
-    if provider not in available:
+    # Singleton — first call probes; subsequent calls are free.
+    if not WinMLEPRegistry.get_instance().is_ep_available(provider):
         pytest.skip(f"EP not available on this host: {provider}")
 
     return provider
--- a/tests/e2e/test_compile_e2e.py
+++ b/tests/e2e/test_compile_e2e.py
@@ -36,7 +36,6 @@ import numpy as np
 import onnx
 import pytest
 from click.testing import CliRunner
-from onnx import TensorProto, helper
 
 from tests.e2e.require_ep import require_ep
 from winml.modelkit.commands.compile import compile as compile_cmd
@@ -147,22 +146,10 @@ def _find_qairt_sdk_root() -> Path | None:
 # Fixtures
 # ---------------------------------------------------------------------------
 
-
-@pytest.fixture
-def tiny_model(tmp_path: Path) -> Path:
-    """A minimal MatMul ONNX model usable by every EP."""
-    x = helper.make_tensor_value_info("input", TensorProto.FLOAT, [1, 4])
-    y = helper.make_tensor_value_info("output", TensorProto.FLOAT, [1, 4])
-    w_arr = np.eye(4, dtype=np.float32)
-    w = helper.make_tensor("weight", TensorProto.FLOAT, [4, 4], w_arr.flatten().tolist())
-    node = helper.make_node("MatMul", ["input", "weight"], ["output"], name="matmul")
-    graph = helper.make_graph([node], "tiny", [x], [y], [w])
-    model = helper.make_model(graph, opset_imports=[helper.make_opsetid("", 17)])
-    model.ir_version = 8
-    onnx.checker.check_model(model)
-    path = tmp_path / "tiny.onnx"
-    onnx.save(model, str(path))
-    return path
+# Note: the input ONNX model is provided by the conftest's `simple_matmul_onnx`
+# fixture (1×4 MatMul, opset 13). Compile tests don't depend on the specific
+# weight values or input names, so we reuse it instead of defining a local
+# `simple_matmul_onnx` fixture.
 
 
 def _write_json(path: Path, data: dict) -> Path:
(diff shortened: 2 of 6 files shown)


tests/e2e/test_compile_e2e.py
```diff
@@ -0,0 +1,714 @@
...
+    w_arr = np.eye(4, dtype=np.float32)
+    w = helper.make_tensor("weight", TensorProto.FLOAT, [4, 4], w_arr.flatten().tolist())
+    node = helper.make_node("MatMul", ["input", "weight"], ["output"], name="matmul")
+    graph = helper.make_graph([node], "tiny", [x], [y], [w])
+    model = helper.make_model(graph, opset_imports=[helper.make_opsetid("", 17)])
+    model.ir_version = 8
+    onnx.checker.check_model(model)
+    path = tmp_path / "tiny.onnx"
+    onnx.save(model, str(path))
+    return path
```
timenick: `tiny_model` duplicates the existing `simple_matmul_onnx` fixture in `tests/e2e/conftest.py` — same 1×4 MatMul, just different opset (17 vs 13) and weights (identity vs seeded random). The `tests/CLAUDE.md` convention says new fixtures should check existing `conftest.py` and either reuse or extend rather than duplicate.

Suggest either:
- Use `simple_matmul_onnx` directly (the EPs in scope don't care about weight values), or
- If you need the specific opset/weight shape, parametrize the existing fixture in conftest rather than introducing a parallel one.

The `_invoke()` helper above also overlaps with the conftest's `runner` fixture, though more thinly (it adds `obj={"debug": False}` binding).

🤖 Generated with Claude Code

tests/e2e/test_compile_e2e.py
```diff
@@ -0,0 +1,714 @@
...
+
+@pytest.mark.e2e
+class TestEpOverridesDevice:
+    # Coverage scope: only the ``--device auto --ep qnn`` combination is
+    # exercised here. Other device/ep combinations (e.g. ``--device gpu --ep
+    # qnn``, ``--device cpu --ep qnn``) currently exhibit a bug where the
+    # ``--device`` value propagates into the EP factory as ``device_type``
+    # and the resulting compile reports success while producing a non-
+    # EPContext artifact. That bug will be fixed in a separate PR; tests
+    # for those combinations will be added there.
```
timenick, line 649: The PR description advertises two combinations that are not actually tested here:

- `--device gpu --ep qnn`
- `--device npu --ep openvino`

This scope comment correctly explains they're deferred due to the `device_type` propagation bug, but the PR body presents them as covered. Please update the PR description to match what this class actually pins (only `--device auto --ep qnn`), so reviewers and future readers aren't misled into thinking those scenarios are guarded.

🤖 Generated with Claude Code

tests/e2e/require_ep.py
```diff
@@ -0,0 +1,66 @@
...
+    from winml.modelkit.session import WinMLEPRegistry
+    from winml.modelkit.utils.constants import normalize_ep_name
+
+    provider = normalize_ep_name(ep)
+    if provider is None:
+        pytest.skip(f"Unknown EP: {ep!r}")
+
+    # Singleton — first call probes; subsequent calls are free.
+    available = set(WinMLEPRegistry.get_instance().get_available_eps())
+    available.add("CPUExecutionProvider")  # always-on fallback
```
timenick: Two small things on this block:

1. `WinMLEPRegistry` already exposes `is_ep_available(ep_name)` — it's a more direct fit than building a set from `get_available_eps()` keys.
2. The manual `available.add("CPUExecutionProvider")` suggests the registry doesn't always report CPU as available. If that's actually true, that's a registry bug worth fixing upstream rather than papering over in every caller. If the registry does report CPU, this line is dead code.

Worth a quick check before merging.

🤖 Generated with Claude Code

tests/e2e/require_ep.py
```diff
@@ -0,0 +1,66 @@
...
+        The full ORT provider name (e.g. ``"QNNExecutionProvider"``) that
+        satisfied the requirement. Handy when the caller wants to pass it
+        on to ``onnxruntime`` / ``WinMLSession``.
+
+    Raises:
+        pytest.skip.Exception: When ``ep`` is unknown or the corresponding
+            provider is not available on the host.
+    """
+    from winml.modelkit.session import WinMLEPRegistry
+    from winml.modelkit.utils.constants import normalize_ep_name
```
timenick: `normalize_ep_name` is not exported in `winml/modelkit/utils/__init__.py`, so this reaches into the `utils.constants` submodule. The source code does the same (`from ..utils.constants import normalize_ep_name`), so this appears intentional — but per the test-code import rule in `CLAUDE.md`, public test imports should come through the package's `__init__.py`.

If the team considers `normalize_ep_name` part of the public API (it's used in `analyze/`, `commands/`, etc.), please add it to `winml/modelkit/utils/__init__.py` + `__all__`. Otherwise this is acceptable.

🤖 Generated with Claude Code

tests/e2e/test_compile_e2e.py
```diff
@@ -0,0 +1,714 @@
...
+        assert result.exit_code == 0, result.output
+        assert_epcontext_artifact(out, tiny_model, embed=False)
+
+
+# ===========================================================================
+# Validation toggle (--validate / --no-validate)
+# ===========================================================================
+
+
+_VALIDATE_TOKENS = ("validating compiled model", "validation")
```
timenick: The second token `"validation"` is overly lenient — it'll match unrelated output like `"skipping validation"` or any error containing the word, defeating the assertion in `test_no_validate_skips_validation`. The actual emitter is a single line at `src/winml/modelkit/compiler/stages/compile.py:153` (`"Validating compiled model..."`).

Suggest dropping the second token and asserting on `"validating compiled model"` only.

🤖 Generated with Claude Code

timenick (commented):
Review of E2E compile coverage. The test scope is solid; a few items below.

**Blocking — existing unit test will fail after merge**

`tests/unit/session/test_qairt_session.py:260` still asserts the old filename:

```python
assert session._ctx_path == model_dir / f"{model_stem}_qnn_ctx.onnx"
```

This PR renames `_ctx_path` in `qairt_session.py:63` from `{stem}_qnn_ctx.onnx` to `{stem}_ctx.onnx`, so this assertion needs to be updated in lockstep. Please include the fix in this PR; otherwise the unit suite breaks the moment this lands. (That file is not part of this diff so I couldn't leave an inline comment there.)

Remaining items are inline.

🤖 Generated with Claude Code

timenick:
@DingmaomaoBJTU could you please comment on the EP support list? 

We might need to add new tests once this PR has been merged. https://github.com/microsoft/winml-cli/pull/638

tests/e2e/test_compile_e2e.py
zhenchaoni: fixed, using the simple_matmul_onnx

tests/e2e/test_compile_e2e.py
zhenchaoni: Updated the PR description. These 2 are existing bugs I found during test. I need to use a different PR to fix it and add e2e test.

tests/e2e/require_ep.py
zhenchaoni: fixed according to your suggestion

tests/e2e/require_ep.py
zhenchaoni: fixed

tests/e2e/test_compile_e2e.py
zhenchaoni: fixed

DingmaomaoBJTU:
> @DingmaomaoBJTU could you please comment on the EP support list?
> 
> We might need to add new tests once this PR has been merged. #638

CUDA is not supported, and MIGraphX is not supported for now.
Other than CPU and DML, all execution providers support compile.
Be cautious: the compile result can differ depending on the EP/device. For example, OV-CPU and OV-GPU may produce different results.

zhenchaoni:
> > @DingmaomaoBJTU could you please comment on the EP support list?
> > We might need to add new tests once this PR has been merged. #638
> 
> CUDA is not supported, and MIGraphX is not supported for now. Other than CPU and DML, all execution providers support compile. Be cautious: the compile result can differ depending on the EP/device. For example, OV-CPU and OV-GPU may produce different results.

Updated: The current e2e test will cover qnn, ov, cpu and dml. Vitisai and tensor rt will be supported later when they are good.

DingmaomaoBJTU (approved):

PR #647, opened 2026-05-15
fix(_warnings): downgrade huggingface_hub symlinks UserWarning to INFO

DingmaomaoBJTU:
## Summary

- Adds `_HFSymlinksInfoFilter` on the `py.warnings` logger in `_warnings.py`
- When `logging.captureWarnings(True)` is active (build/eval pipeline), the filter mutates the record `levelno`/`levelname` from WARNING → INFO for huggingface_hub symlinks messages
- 4 unit tests in `tests/unit/test_warnings_configure.py` cover: filter installation, level downgrade, and two negative cases (non-symlinks message, symlinks from unrelated module)

## Context

On Windows without Developer Mode, `huggingface_hub` emits a `UserWarning` every time it populates the cache because symlinks aren't supported and it falls back to file copies. The warning is cosmetic — caching still works — but it was surfacing at `WARNING` level in the Rich CLI output, which is misleading and noisy.

The fix follows the existing pattern in `_warnings.py` (logging filters on named loggers) rather than `filterwarnings("ignore")`, so the message is still visible at `INFO` level with `-v` instead of being silently dropped.


timenick:
Could you provide an example that illustrates the difference?

DingmaomaoBJTU:
Added a before/after example directly in the docstring:

```
Before (WARNING level, always visible):
    [09:12:34] WARNING  `huggingface_hub` cache-system uses symlinks by
                        default to efficiently store duplicated files but
                        your machine does not support them in
                        C:\...\huggingface_hub\file_download.py

After (INFO level, only visible with -v / --verbose):
    [09:12:34] INFO     `huggingface_hub` cache-system uses symlinks by
                        default to efficiently store duplicated files but
                        your machine does not support them in
                        C:\...\huggingface_hub\file_download.py
```

src/winml/modelkit/_warnings.py
```diff
@@ -94,6 +94,27 @@ def filter(self, record: logging.LogRecord) -> bool:
...
+    # py.warnings logger filters (for warnings routed via logging.captureWarnings)
+    # =========================================================================
+
+    class _HFSymlinksInfoFilter(logging.Filter):
+        """Downgrade the huggingface_hub symlinks UserWarning from WARNING to INFO.
+
+        On Windows without Developer Mode, huggingface_hub warns that symlinks
+        are unsupported and the cache will use copies instead. This is cosmetic —
+        the cache still works, just without deduplication. WARNING is misleading
+        here; INFO is the appropriate level.
```
timenick, line 107: **This filter never matches in production.** When warnings are routed through `logging.captureWarnings(True)`, `logging._showwarning` calls `getLogger("py.warnings").warning("%s", formatted)`. The LogRecord's `pathname` is then resolved by `findCaller`, which skips `logging`-module frames and lands on **`warnings.py`** — never on the original `huggingface_hub` caller.

Verified empirically in this repo's venv:

```
pathname basename: warnings.py
message contains huggingface_hub: True
message contains symlinks: True
```

So `"huggingface_hub" in record.pathname` is always `False` here, and the filter is a no-op. The original path *is* present, but inside `record.getMessage()` — `formatwarning` prepends `"<filename>:<lineno>: UserWarning: ..."` to the body.

Suggested fix — match against the formatted message instead:

```python
def filter(self, record: logging.LogRecord) -> bool:
    msg = record.getMessage()
    if "huggingface_hub" in msg and "symlinks" in msg:
        record.levelno = logging.INFO
        record.levelname = "INFO"
    return True
```

Generated with Claude Code

tests/unit/test_warnings_configure.py
```diff
@@ -0,0 +1,92 @@
...
+        record = _make_record(
+            message="Some other huggingface_hub warning",
+            pathname=r"C:\some\path\huggingface_hub\file_download.py",
+        )
+
+        result = f.filter(record)
+
+        assert result is True
+        assert record.levelno == logging.WARNING
+
```
timenick: **These tests give false confidence — they validate the broken predicate against fabricated input.**

All three behavioural tests construct LogRecords with `pathname=r"C:\some\path\huggingface_hub\file_download.py"`, but real records from `logging.captureWarnings(True)` always have `pathname == ".../warnings.py"` (see comment on `_warnings.py:107`). The synthetic value masks the production bug — the filter passes the unit tests but is a no-op when actually invoked.

Suggested approach: drive a real warning through the capture path instead of constructing records by hand. Something like:

```python
def test_downgrade_to_info(caplog) -> None:
    import logging, warnings
    import winml.modelkit._warnings  # noqa: F401

    logging.captureWarnings(True)
    try:
        with caplog.at_level(logging.INFO, logger="py.warnings"):
            warnings.warn_explicit(
                message="`huggingface_hub` cache-system uses symlinks ...",
                category=UserWarning,
                filename=r"C:/fake/huggingface_hub/file_download.py",
                lineno=1,
            )
    finally:
        logging.captureWarnings(False)

    rec = next(r for r in caplog.records if "symlinks" in r.getMessage())
    assert rec.levelno == logging.INFO
```

That shape would have caught the `pathname` issue in CI.

Generated with Claude Code

src/winml/modelkit/_warnings.py
DingmaomaoBJTU: Fixed — the predicate now matches against `getMessage()` instead of `record.pathname`:

```python
def filter(self, record: logging.LogRecord) -> bool:
    msg = record.getMessage()
    if "symlinks" in msg and "huggingface_hub" in msg:
        record.levelno = logging.INFO
        record.levelname = "INFO"
    return True
```

The docstring was also updated to explain why: `warnings.formatwarning()` embeds the source filename in the message body, so `huggingface_hub` appears there.

tests/unit/test_warnings_configure.py
DingmaomaoBJTU: Fixed — tests now drive real warnings through the `captureWarnings` path via `warnings.warn_explicit` instead of constructing `LogRecord`s by hand:

```python
def _emit_warning(message: str, filename: str) -> None:
    logging.captureWarnings(True)
    try:
        warnings.warn_explicit(
            message=message,
            category=UserWarning,
            filename=filename,
            lineno=1,
            registry={},  # fresh registry — always route to showwarning
        )
    finally:
        logging.captureWarnings(False)
```

All three behavioural tests now use this helper with `caplog` to assert on the actual mutated record. This approach would have caught the `pathname` issue in CI.

timenick (approved):

PR #653, opened 2026-05-18
fix: HWMonitor uses session ep to align with session; _find_ep_device uses fixed order instead of ort.get_ep_devices

xieofxie:
Before the fix
- `winml perf -m microsoft/resnet-50 --device gpu --iterations 1000 --monitor` session and monitor could use different ones (session openvino, monitor dml)
-  `winml perf -m model.onnx --device gpu --iterations 1000 --monitor` could use random ones depend on registering order like dml instead of openvino

### DIFF
--- a/src/winml/modelkit/commands/perf.py
+++ b/src/winml/modelkit/commands/perf.py
@@ -409,11 +409,10 @@ class PerfBenchmark:
         from ..utils.constants import normalize_ep_name
 
         monitor_device = self._model.device or self.config.device or "auto"
-        ep_name = normalize_ep_name(self.config.ep) if self.config.ep else None
         hw_monitor = HWMonitor(
             poll_interval_ms=_HW_POLL_INTERVAL_MS,
             device=monitor_device,
-            ep_name=ep_name,
+            ep_name=session.ep_name,
         )
 
         # EP-specific proof-of-execution monitor.
@@ -663,7 +662,7 @@ def _perf_modules(
                         hw_ctx = HWMonitor(
                             poll_interval_ms=_HW_POLL_INTERVAL_MS,
                             device=resolved_device,
-                            ep_name=normalize_ep_name(ep) if ep else None,
+                            ep_name=session.ep_name,
                         )
 
                 if hw_ctx:
@@ -1025,7 +1024,7 @@ def _run_onnx_benchmark(
             hw_ctx = HWMonitor(
                 poll_interval_ms=_HW_POLL_INTERVAL_MS,
                 device=session.device or device,
-                ep_name=normalize_ep_name(config.ep) if config.ep else None,
+                ep_name=session.ep_name,
             )
         else:
             Console(stderr=True).print(
--- a/src/winml/modelkit/commands/perf.py
+++ b/src/winml/modelkit/commands/perf.py
@@ -627,7 +627,7 @@ def _perf_modules(
             cfg.quant = None
             cfg.compile = None
 
-        with tempfile.TemporaryDirectory() as tmpdir:
+        with tempfile.TemporaryDirectory(ignore_cleanup_errors=True) as tmpdir:
             try:
                 build_result = build_hf_model(
                     config=cfg,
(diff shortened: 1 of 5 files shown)


xieofxie:
this should be covered by fangyang's update or refactory in the end

src/winml/modelkit/session/session.py
```diff
@@ -563,19 +589,6 @@ def _prepare_inputs(
 
         return ort_inputs
 
-    def _detect_best_device(self) -> str:
```
xieofxie, line 566: A no-op and misleading, so removed

DingmaomaoBJTU:
### Code review

Found 1 issue:

1. `session.ep_name` is always `None` in `_perf_modules` because `session.compile()` has not been called when `HWMonitor` is constructed — `ep_name=None` is silently passed, defeating the PR's stated goal of aligning the monitor with the session's resolved EP.

The `WinMLSession` is created at line 643 but compilation is deferred until the first `session.run()` call (line 668 inside `session.perf()`). The `ep_name` property documents `None` before compile. The other two call sites (`_run_benchmark_monitored` line 413 and `_run_onnx_benchmark` line 1023) are correct because those sessions are compiled before `HWMonitor` is instantiated.

https://github.com/microsoft/winml-cli/blob/b14375438949ea1d87c17ebc5b0dd7bcd3c928dc/src/winml/modelkit/commands/perf.py#L643-L663

src/winml/modelkit/commands/perf.py
```diff
@@ -657,13 +654,12 @@ def _perf_modules(
...
                 if monitor:
                     from ..session.monitor.hw_monitor import HWMonitor
-                    from ..utils.constants import normalize_ep_name
 
                     if HWMonitor.is_available():
                         hw_ctx = HWMonitor(
                             poll_interval_ms=_HW_POLL_INTERVAL_MS,
                             device=resolved_device,
-                            ep_name=normalize_ep_name(ep) if ep else None,
+                            ep_name=session.ep_name,
```
timenick, line 665: **Question:** Is it expected that `session.ep_name` resolves correctly on this codepath?

In `_perf_modules`, `session = WinMLSession(...)` is created at line 643 but I don't see an explicit `session.compile()` before this line. `session.io_config` at line 648 only loads ONNX metadata without compiling, and the auto-compile inside `session.run()` doesn't fire until line 668. Since the `ep_name` property returns `None` while `self._session is None`, it looks like HWMonitor would receive `ep_name=None` here and fall back to device-only resolution — which seems to defeat the alignment this PR introduces for the other two call sites:
- `_run_benchmark_monitored` (line 413) — explicit compile at line 297
- `_run_onnx_benchmark` (line 1023) — explicit compile at line 1005

Did I miss a place where compile is triggered before this line? If not, would adding `session.compile()` before line 662 (matching the other two paths) be appropriate?

🤖 Generated with Claude Code

src/winml/modelkit/session/session.py
```diff
@@ -489,6 +487,7 @@ def _find_ep_device(device: str, ep_name: EPName | None = None) -> Any:
         Returns:
             The matching OrtEpDevice, or None if not found.
         """
+        from ..sysinfo.device import get_device_ep_map, get_ep_device_map
```
timenick: **Import style:** This reaches into the submodule, but both `get_device_ep_map` and `get_ep_device_map` are now exported via `..sysinfo/__init__.py` in this PR. Per CLAUDE.md import rules, source code should go through `__init__.py`:

```python
from ..sysinfo import get_device_ep_map, get_ep_device_map
```

Note: line 281 in this file already uses the same submodule-reaching style (pre-existing, not introduced by this PR), but the new import here can follow the package-level convention.

🤖 Generated with Claude Code

xieofxie: updated

src/winml/modelkit/commands/perf.py
xieofxie: fixed. Also call compile

src/winml/modelkit/commands/perf.py
```diff
@@ -630,7 +627,7 @@ def _perf_modules(
             cfg.quant = None
             cfg.compile = None
 
-        with tempfile.TemporaryDirectory() as tmpdir:
+        with tempfile.TemporaryDirectory(ignore_cleanup_errors=True) as tmpdir:
```
xieofxie, line 630: Failed to remove `compiled_OpenVINOExecutionProvider.bin`, so add this

Issue #596, opened 2026-05-12
perf: HF model ID uses AOT build pipeline, ONNX file uses raw JIT — inconsistent results and divergent code paths


xieofxie:
## Summary

`winml perf` currently dispatches by argument shape and runs two completely different benchmark pipelines for what users perceive as the same operation:

- `winml perf -m hf/model` → `PerfBenchmark.run()` (`commands/perf.py:280`), which calls `WinMLAutoModel.from_pretrained(...)`. This runs the **full AOT pipeline**: export → optimize → quantize → compile, then benchmarks the compiled artifact.
- `winml perf -m path/to/model.onnx` → `_run_onnx_benchmark(...)` (`commands/perf.py:952`), which calls `WinMLSession(onnx_path=...)` directly. This **bypasses the build pipeline entirely** — it's effectively a raw ORT JIT load with whatever EP the device resolves to.

The branch is at `commands/perf.py:1297` (`is_onnx = model_path.suffix.lower() == \".onnx\"`).

## Why this matters

1. **Apples-to-oranges numbers.** If a user runs `winml perf -m microsoft/resnet-50` and then `winml perf -m <the_output.onnx>`, the two latency numbers don't agree, even though the file was just produced by the first command. The first goes through quantize+compile (so the benchmarked artifact may not even be `<the_output.onnx>` — it's the compiled context model). The second loads the raw file as-is.

2. **Flag semantics silently change.** Several CLI flags are HF-path only:
   - `--no-quantize`, `--rebuild`, `--ignore-cache`, `--precision`: silently no-op on the ONNX path.
   - `--shape-config`: ONNX path prints a warning and discards it (`commands/perf.py:1301`).
   The user has no way to ask \"benchmark this ONNX after quantize+compile\".

3. **Duplicated benchmark logic.** The perf loop, hardware-monitor setup, model-info printing, and stats collection exist in **both** `PerfBenchmark.run()` and `_run_onnx_benchmark()` (with helpers `_run_simple_loop` / `_run_monitored_loop`). Any future change to the timing/stats code has to be made in two places.

4. **Dead code.** `PerfBenchmark._load_model` at `commands/perf.py:322-353` has a branch for `is_onnx and model_path.exists()` calling `WinMLAutoModel.from_onnx(...)`. This branch is unreachable from the CLI because the outer dispatcher at `commands/perf.py:1297` already routes `.onnx` files away.

## Suggested direction

[...]

PR #659, opened 2026-05-19
fix(perf): unify HF and ONNX paths through PerfBenchmark

xieofxie:
## Summary

Part 1 of #596.

- `winml perf -m hf/model` and `winml perf -m model.onnx` previously ran two completely different pipelines: HF went through the full AOT build (export → optimize → quantize → compile) via `PerfBenchmark`, while `.onnx` files bypassed the pipeline entirely and ran a raw ORT JIT load through `_run_onnx_benchmark`. Same user-facing command, non-comparable numbers, and several flags (`--no-quantize`, `--rebuild`, `--ignore-cache`, `--precision`) silently no-oped on the ONNX path.
- Both inputs now flow through `PerfBenchmark`, which dispatches to `WinMLAutoModel.from_pretrained` or `.from_onnx`. The `is_onnx` branch in `_load_model` (previously dead code) is now the live entry point, so an `.onnx` file runs optimize → [quantize] → [compile] like the HF flow minus export.
- Delete `_run_onnx_benchmark` and the duplicate hardware-monitor / stats-collection logic it carried. The CLI keeps `is_onnx` only for the file-exists check, the `--shape-config` warning (shapes are baked into pre-exported ONNX), and feeding `--op-tracing` the raw input path.
- Refresh docstrings on the `perf` command, `PerfBenchmark._load_model`, and the loop helpers to drop stale references; update the CLI test to assert ONNX inputs route through `PerfBenchmark.run`.
- add skip_build option


xieofxie:
could wait for perf e2e

xieofxie:
> could wait for perf e2e

Done and tested in qnn

src/winml/modelkit/commands/perf.py
```diff
@@ -317,7 +317,13 @@ def run(self) -> BenchmarkResult:
...
-        """Load model via WinMLAutoModel (handles both HF and ONNX)."""
+        """Load model via WinMLAutoModel.
+
+        Both HF model IDs and pre-exported .onnx files flow through this
+        single path so latency numbers stay comparable: HF runs export →
+        optimize → [quantize] → [compile], and ONNX runs the same pipeline
+        minus export.
+        """
         from ..config import WinMLBuildConfig
         from ..models import WinMLAutoModel
```
DingmaomaoBJTU, line 334: **suggestion (robustness):** `_load_model` still gates the ONNX branch on `model_path.exists()` (line 326). If someone uses `PerfBenchmark` programmatically (not through the CLI) with a non-existent `.onnx` path, `is_onnx` silently becomes `False` and it falls through to `from_pretrained`, producing the confusing "not a valid JSON file" error from AutoConfig.

The CLI now guards this (the new early `FileNotFoundError` is great), but `_load_model` itself is still fragile. Consider splitting the check:

```python
is_onnx = model_path.suffix.lower() == ".onnx"
if is_onnx and not model_path.exists():
    raise FileNotFoundError(f"ONNX file not found: {model_path}")
```

This way the intent ("this *is* an ONNX path, but the file is missing") is surfaced clearly regardless of entry point.

src/winml/modelkit/commands/perf.py
```diff
@@ -837,8 +843,7 @@ def display_console_report(result: BenchmarkResult, console: Console) -> None:
             )
         else:
             console.print(
-                f"  CPU: {cpu.get('mean_pct', 0):.1f}% avg  |  "
-                f"Mem: {ram.get('used_mb', 0):.0f} MB"
+                f"  CPU: {cpu.get('mean_pct', 0):.1f}% avg  |  Mem: {ram.get('used_mb', 0):.0f} MB"
```
DingmaomaoBJTU, line 874: **nit:** This cosmetic change (collapsing a two-line implicit string concat into one line) is unrelated to the ONNX unification. Not a problem per se, but it adds noise to the diff and makes `git blame` less useful for the surrounding block. Consider reverting or splitting into a separate commit if you do more formatting cleanup.

tests/unit/commands/test_perf_cli.py
```diff
@@ -279,14 +279,22 @@ def test_no_quantize_false_passes_no_override(self) -> None:
...
+        from `winml perf -m hf/id` and `winml perf -m <built.onnx>` are
+        comparable (issue #596).
+        """
         onnx_file = tmp_path / "model.onnx"
         onnx_file.write_bytes(b"fake onnx")
 
         with (
-            patch(
-                "winml.modelkit.commands.perf._run_onnx_benchmark",
+            patch.object(
```
DingmaomaoBJTU, line 296: Good switch to `patch.object` — more resilient to internal refactors than string-patching.

One thing: since the mock replaces `PerfBenchmark.run` entirely, this test only verifies the CLI *calls* `PerfBenchmark.run`, but not that ONNX-specific CLI-level behavior (e.g., `shape_config` being set to `None` on the config, the "Benchmarking ONNX" console line) actually fires before the benchmark runs.

Consider adding a side-effect or assertion on the `config` object passed into `PerfBenchmark()` to verify ONNX-specific setup happened. For example, you could capture the config and assert `config.shape_config is None` when `--shape-config` is provided with an ONNX input. (Low priority — existing `test_cli_onnx_not_found_error` covers the existence check path.)

DingmaomaoBJTU (commented):
Clean, well-motivated change. Unifying both paths through `PerfBenchmark` removes ~100 lines of duplicated benchmark logic and makes latency numbers directly comparable — solid improvement.

A few inline comments below, mostly nits and one suggestion for robustness.

src/winml/modelkit/commands/perf.py
xieofxie: no difference. Caused by code formatter

src/winml/modelkit/commands/perf.py
xieofxie: cool

tests/unit/commands/test_perf_cli.py
xieofxie: added

xieofxie:
--ep cpu needs fix..

already fixed

xieofxie:
WinMLAutoModel.from_onnx has at least two issues:
- when `perf --device gpu` without ep, it will analyze on all eps but will only perf on one ep
- for same model path, if run first with --device cpu, it will cache a cpu model (openvino for example), when running again with --device gpu, it will still load the cpu cache but could not run (dml for example)

xieofxie:
> WinMLAutoModel.from_onnx has at least two issues:
> 
> * when `perf --device gpu` without ep, it will analyze on all eps but will only perf on one ep
> * for same model path, if run first with --device cpu, it will cache a cpu model (openvino for example), when running again with --device gpu, it will still load the cpu cache but could not run (dml for example)

Tracked https://github.com/microsoft/winml-cli/issues/827

src/winml/modelkit/inference/engine.py
```diff
@@ -334,7 +339,7 @@ def load(
                 else:
```
DingmaomaoBJTU, line 354: `skip_build` is silently dropped when the model path is a directory and `_load_from_build_dir` succeeds. `_load_from_build_dir` calls `WinMLAutoModel.from_onnx(... skip_build=True)` hardcoded (unchanged by this PR), so a user who runs `winml run --no-skip-build /path/to/build-dir/` will get no build step even though they explicitly asked for one. The docstring says 'ONNX-only; ignored for HF model IDs and build output directories' — but a build directory *is* an ONNX source. Either (a) thread `skip_build` into `_load_from_build_dir`, or (b) update the docstring to say 'ignored for HF model IDs and pre-built build directories (use an explicit .onnx path instead)'.

tests/unit/commands/test_config_value_priority.py
```diff
@@ -279,26 +279,33 @@ def _run_perf(
...
 
     captured: dict[str, Any] = {}
 
-    def fake_run(model_path, *, config=None, **_kw):
+    def fake_benchmark(config):
         captured["config"] = config
-        return MagicMock()
+        instance = MagicMock()
+        instance.run.return_value = MagicMock()
+        return instance
```
DingmaomaoBJTU, line 300: `fake_benchmark` configures `instance.run.return_value` but never asserts that `run()` was actually called. If `perf` constructs `PerfBenchmark` but skips `.run()` (e.g., due to an early return), all priority tests would still pass. Consider adding `assert instance.run.call_count == 1` inside the captured check, or asserting it after the invoke.

src/winml/modelkit/eval/config.py
```diff
@@ -118,6 +118,7 @@ class WinMLEvaluationConfig:
     dataset: DatasetConfig = field(default_factory=DatasetConfig)
     output_path: Path | None = field(default=None, metadata={"cli_name": "output"})
     mode: EvalMode = "onnx"
+    skip_build: bool = True
```
DingmaomaoBJTU, line 122: `to_dict()` only serializes `skip_build` when it's `False` (line 142: `if not self.skip_build`). This means a config exported with `skip_build=True` won't include the field — a consumer re-reading the JSON gets the same default back, so round-tripping is consistent. But if `skip_build` is ever changed from True to False in a future refactor (e.g., to force build by default for eval), a previously exported config without the field would silently revert to the new default rather than preserving the old 'True'. Consider always serializing `skip_build` for explicit auditability, similar to how `device` is always written.

src/winml/modelkit/inference/engine.py
xieofxie: Updated the load() docstring to call out that skip_build is ignored for build directories and document the workaround (point at the explicit .onnx file inside the build dir). Went with the docstring clarification rather than threading skip_build into _load_from_build_dir since build dirs hold already-built artifacts — re-building is uncommon enough that the explicit-path workaround beats adding a new code path.

tests/unit/commands/test_config_value_priority.py
xieofxie: Done — added assert captured["instance"].run.call_count == 1 after the CLI invoke, with a comment explaining what it guards against (an early-return path that constructs PerfBenchmark but skips .run() would otherwise let the priority assertions pass against a config that never reached the benchmark).

src/winml/modelkit/eval/config.py
xieofxie: Good catch — done. Dropped the if not self.skip_build: guard so to_dict() always emits skip_build, matching how device is unconditionally serialized. A future flip of the default now safely round-trips through exported configs.

src/winml/modelkit/inference/engine.py
```diff
@@ -307,13 +308,20 @@ def load(
...
             allow_unsupported_nodes: If True, warn instead of raising when the
                 analyzer reports unsupported nodes during an HF build. Note: has
-                no effect when loading from a pre-built ONNX file or a cached
-                build directory (no build/analyze step runs in those paths).
+                no effect on raw .onnx files or pre-built build directories
+                (no build/analyze step runs in those paths).
         """
         self._model_path = str(model_path)
-        self._device = device
         self._ep = ep
```
DingmaomaoBJTU, line 325: `self._device = device` was removed from `load()` in this PR (the diff shows it as a deletion at line 380 area). However `self._device` is still read in two places: (1) `reload()` passes it back to `load()` as `device=self._device`; (2) the `device` property returns it. After this change, calling `engine.load('model', device='npu')` leaves `self._device` at its `__init__` default of `'auto'`, so `engine.device` returns `'auto'` instead of `'npu'`, and `engine.reload()` will call `load()` with `device='auto'` — re-loading on the wrong device. The line must be restored.

tests/unit/commands/test_config_value_priority.py
```diff
@@ -279,26 +279,34 @@ def _run_perf(
...
     captured: dict[str, Any] = {}
 
-    def fake_run(model_path, *, config=None, **_kw):
+    def fake_benchmark(config):
         captured["config"] = config
-        return MagicMock()
+        instance = MagicMock()
+        instance.run.return_value = MagicMock()
+        captured["instance"] = instance
+        return instance
```
DingmaomaoBJTU, line 300: `skip_build` is a new `BenchmarkConfig` field wired through `collect_cli_overrides`, but it is not represented in `PERF_CASES`. `TestPerfPriority` only exercises `ep`; if the CLI→config-file priority for `skip_build` broke (e.g. a name mismatch or a default-shadowing issue), no test would catch it. Consider adding a `FieldCase` for `skip_build` / `--no-skip-build` and exposing `cfg.skip_build` in `_run_perf`'s return dict.

src/winml/modelkit/inference/engine.py
xieofxie: good catch! not sure why it is removed..

tests/unit/commands/test_config_value_priority.py
xieofxie: collect_cli_overrides is not used in perf, but add test cases

tests/unit/commands/test_config_value_priority.py
```diff
@@ -877,20 +894,14 @@ def test_t2_beats_t3(self, case: FieldCase, tmp_path: Path) -> None:
...
-        self, case: FieldCase, tmp_path: Path
-    ) -> None:
+    def test_t3_not_shadowed_by_empty_section(self, case: FieldCase, tmp_path: Path) -> None:
         _check_t3_not_shadowed_by_empty_section(_run_eval, case, tmp_path)
 
     @pytest.mark.parametrize("case", EVAL_CASES, ids=lambda c: c.field)
-    def test_t3_not_shadowed_by_absent_section(
-        self, case: FieldCase, tmp_path: Path
-    ) -> None:
+    def test_t3_not_shadowed_by_absent_section(self, case: FieldCase, tmp_path: Path) -> None:
```
DingmaomaoBJTU, line 902: `TestEvalPriority` is missing `skip_build` coverage. Unlike `perf`'s `skip_build` (no JSON config source, tested via dedicated boolean-flag tests), `eval`'s `skip_build` has a full Tier-2 path: `{"eval": {"skip_build": false}}` flows through `merge_config(cfg, eval_data)`. The full CLI > config-file > default contract should be verified.

Two things needed:

1. Add `"skip_build": cfg.skip_build` to the `_run_eval` return dict (line 544–550).

2. Add tests to `TestEvalPriority`:
```python
def test_skip_build_default_is_true(self, tmp_path):
    assert _run_eval([], None, tmp_path)["skip_build"] is True

def test_no_skip_build_flag_sets_false(self, tmp_path):
    assert _run_eval(["--no-skip-build"], None, tmp_path)["skip_build"] is False

def test_json_skip_build_false_applies(self, tmp_path):
    # Tier-2: config file {"eval": {"skip_build": false}} must take effect
    eff = _run_eval([], {"eval": {"skip_build": False}}, tmp_path)
    assert eff["skip_build"] is False

def test_cli_beats_json_skip_build(self, tmp_path):
    # CLI True must win over JSON False
    eff = _run_eval(["--skip-build"], {"eval": {"skip_build": False}}, tmp_path)
    assert eff["skip_build"] is True
```

DingmaomaoBJTU (approved):
Overall the unification looks solid — both HF and ONNX paths now flow through PerfBenchmark, skip_build is properly wired across perf/eval/run/engine, and the priority tests cover the key contracts. One minor gap still open: TestEvalPriority is missing skip_build coverage (see inline comment), but that shouldn't block the merge.

DingmaomaoBJTU (approved):

PR #661, opened 2026-05-19
EP: Implement compile validation and e2e

zhenchaoni:
Our supported device for each ep is:
- qnn: npu and gpu
- openvino: npu, gpu and cpu
- vitisai: npu
- nv_tensorrt_rtx: gpu

This PR
1. Add "good input" e2e test cases, it permutates the above support matrix to generate --device and --ep combinations for e2e test.
2. Add a validation logic. If the user provides --device and --ep which doesn't match the above matrix, or the user provides ep that is not applicable to his machine, a validation error will be raised.

### DIFF
--- a/tests/e2e/test_compile_e2e.py
+++ b/tests/e2e/test_compile_e2e.py
@@ -642,10 +642,6 @@ def test_reject_recompile_of_real_compiled_output(
 # Device/EP good inputs
 # ===========================================================================
 
-
-# Per-EP set of valid concrete ``--device`` values, keyed by EP alias.
-# Derived from the canonical ``EP_SUPPORTED_DEVICES`` policy table (PR #641)
-# so this suite tracks the production source of truth instead of duplicating it.
 _EPCONTEXT_CAPABLE_EPS = ("qnn", "openvino", "vitisai", "nv_tensorrt_rtx")
 EP_DEVICE_SUPPORT: dict[str, tuple[str, ...]] = {
     alias: EP_SUPPORTED_DEVICES[normalize_ep_name(alias)]
@@ -697,10 +693,22 @@ def test_good_input_compiles_and_runs(
     assert "Success! Model compiled" in result.output, result.output
     assert_epcontext_artifact(out, simple_matmul_onnx, embed=False)
     assert_banner_matches_artifact(result, out)
+    # When ``--ep`` is explicit, bind that EP. When it's omitted, the resolver
+    # picks an EP based on host registry order, which is not deterministic
+    # across hosts — read the actual EP back from the artifact's
+    # ``EPContext.source`` so the inference binding matches what was compiled.
+    if ep is not None:
+        runtime_ep = ep
+    else:
+        source = _epcontext_attrs(out)["source"]
+        assert isinstance(source, bytes) and source, (
+            f"Artifact at {out} has no EPContext.source attribute"
+        )
+        runtime_ep = source.decode("utf-8")
     assert_by_run_inference(
         out,
         device=device if device is not None else "auto",
-        ep=require_ep_name,
+        ep=runtime_ep,
         sample_input=sample_input,
     )
 
@@ -756,15 +764,8 @@ def test_bad_input_device_ep_conflict(
     )
 
 
-# EPs that are valid choices for ``--ep`` (registered on at least some hosts)
-# but whose provider has no EPContext compile path. ``cpu`` is omitted:
-# ``CPUExecutionProvider`` is not enumerated by ``WinMLEPRegistry`` so
-# ``require_ep("cpu")`` cannot meaningfully gate this path.
-_UNSUPPORTED_EPCONTEXT_EPS = ("dml", "cuda", "migraphx")
-
-
 @pytest.mark.e2e
-@pytest.mark.parametrize("ep", _UNSUPPORTED_EPCONTEXT_EPS)
+@pytest.mark.parametrize("ep", ("dml", "cuda", "migraphx"))
 def test_bad_input_unsupported_ep(ep: str, simple_matmul_onnx: Path) -> None:
     """``--ep X`` is rejected when X does not produce EPContext.
 
@@ -779,11 +780,12 @@ def test_bad_input_unsupported_ep(ep: str, simple_matmul_onnx: Path) -> None:
     )
 
 
-# Pair each EP with a device it supports so the policy check (which would
-# otherwise fire) is bypassed and the host-state rejection surfaces.
+# Pair each EP with a device it supports. With no ``--device``, ``sysinfo``
+# falls back to CPU when the requested EP isn't registered, which then
+# trips the policy check ("cannot run on --device cpu") instead of the
+# host-state rejection this test targets.
 _EP_NOT_REGISTERED_PARAMS = [
-    ("qnn", "npu"),
-    ("openvino", "gpu"),
+    (ep, EP_DEVICE_SUPPORT[ep][0]) for ep in _EPCONTEXT_CAPABLE_EPS
 ]
 
 
@@ -810,12 +812,20 @@ def test_bad_input_ep_not_registered(
 
 @pytest.mark.e2e
 def test_bad_input_no_ep_covers_device(simple_matmul_onnx: Path) -> None:
-    """``--device cpu`` (no ``--ep``) on a host whose only registered EP does
-    not support cpu is rejected by the ``sysinfo`` device-resolution preflight.
+    """``--device cpu`` (no ``--ep``) on a host with no EPContext-capable EP
+    covering cpu is rejected.
+
+    ``CPUExecutionProvider`` is always implicitly available, so
+    ``sysinfo``'s preflight resolves to it rather than raising. The
+    rejection then comes from the capability check in
+    ``commands/compile.py``: CPU cannot produce EPContext models.
     """
     require_not_ep("openvino")
     src_hash = _sha256(simple_matmul_onnx)
     result = _invoke("-m", str(simple_matmul_onnx), "--device", "cpu")
     _assert_rejected(
-        result, "no compatible EP is available", src_hash, simple_matmul_onnx
+        result,
+        "does not support EPContext compilation",
+        src_hash,
+        simple_matmul_onnx,
     )
--- a/tests/e2e/test_compile_e2e.py
+++ b/tests/e2e/test_compile_e2e.py
@@ -39,9 +39,11 @@ import onnx
 import pytest
 from click.testing import CliRunner
 
-from tests.e2e.require_ep import require_ep
+from tests.e2e.require_ep import require_ep, require_not_ep
 from winml.modelkit.commands.compile import compile as compile_cmd
 from winml.modelkit.onnx import is_compiled_onnx
+from winml.modelkit.utils import normalize_ep_name
+from winml.modelkit.utils.constants import EP_SUPPORTED_DEVICES
 
 
 if TYPE_CHECKING:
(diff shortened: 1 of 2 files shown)


DingmaomaoBJTU:
### Code review

Found 1 issue:

1. `_BAD_INPUT_PARAMS` row 3 uses the wrong EP guard — `require_ep("openvino")` only guarantees OpenVINO is present on the host, but the test asserts that QNN is *absent* (expected error: `"is not available on this host"`). On a machine with both OpenVINO and QNN installed, the test will run (OpenVINO guard passes), the compile with `--ep qnn` will *succeed* (QNN is available), and the assertion `exit_code != 0` will fail.

https://github.com/microsoft/winml-cli/blob/ed3191733f8780af128d74bf1984c5028ef1f855/tests/e2e/test_compile_e2e.py#L706-L711

DingmaomaoBJTU (approved):

tests/e2e/test_compile_e2e.py
```diff
@@ -706,3 +634,142 @@ def test_reject_recompile_of_real_compiled_output(
...
+@pytest.mark.e2e
+@pytest.mark.parametrize("device,ep", _BAD_INPUT_CONFLICT_PARAMS)
+def test_bad_input_device_ep_conflict(
+    device: str, ep: str, simple_matmul_onnx: Path
+) -> None:
+    """``--device X --ep Y`` is rejected when Y does not support X."""
+    require_ep(ep)
+    src_hash = _sha256(simple_matmul_onnx)
+    result = _invoke("-m", str(simple_matmul_onnx), "--device", device, "--ep", ep)
+    _assert_rejected(result, "not supported by EP", src_hash, simple_matmul_onnx)
```
timenick: **Blocker: this test references an error string that doesn't exist in `src/`.**

The PR description states *"Add a validation logic. If the user provides --device and --ep which doesn't match the above matrix ... a validation error will be raised."* — but `git diff origin/main...HEAD` shows the PR touches only `tests/`. No source change adds a `"not supported by EP"` rejection:

```
$ grep -r "not supported by EP" src/ tests/
tests/e2e/test_compile_e2e.py:741:    _assert_rejected(result, "not supported by EP", ...)
```

On current `main`, `_resolve_compile_provider("cpu", "qnn")` happily returns `"QNNExecutionProvider"` and the compile proceeds — no rejection at all. So this parametrized test will fail unconditionally.

Same issue applies to `test_bad_input_ep_not_registered` (`"is not available on this host"` — also nowhere in src).

If this PR is intended to land on top of #641 (EP_SUPPORTED_DEVICES), note that #641's actual rejection message is `"--ep X cannot run on --device Y. Z supports: ..."` — still not `"not supported by EP"`. Either:
1. Push the corresponding validation source change in this PR, **or**
2. Update the assertions to match the existing/upcoming source-side message and call out the cross-PR dependency.

🤖 Generated with Claude Code

tests/e2e/test_compile_e2e.py
```diff
@@ -706,3 +634,142 @@ def test_reject_recompile_of_real_compiled_output(
...
+def _expand_good_inputs(
+    support: dict[str, tuple[str, ...]],
+) -> list[tuple[str | None, str | None, str]]:
+    """Cartesian-expand ``EP_DEVICE_SUPPORT`` into ``(device, ep, require_ep)`` rows."""
+    rows: list[tuple[str | None, str | None, str]] = []
+    for require_ep_name, devices in support.items():
+        rows.extend(
+            (device, ep, require_ep_name)
+            for device in (*devices, "auto", None)
+            for ep in (require_ep_name, None)
```
timenick, line 661: **Some `(device, ep)` rows are inconsistent with the runtime assertion on non-QNN EPs.**

This expansion produces rows like `(device=None, ep=None, require_ep_name="openvino")` and `(device="auto", ep=None, require_ep_name="vitisai")`. Trace the OpenVINO/None-None row:

- `--device` click default is `"npu"` ([commands/compile.py:62](../blob/cce8f22b94fb735b86926ac4b3e1d950083ce3e1/src/winml/modelkit/commands/compile.py#L62))
- `_resolve_compile_provider("npu", None)` → `_DEVICE_TO_PROVIDER["npu"]` → `"QNNExecutionProvider"`
- Compile uses **QNN**, not OpenVINO

On an OpenVINO host that doesn't also have QNN, the compile step fails outright (require_ep skips on OV-less hosts but not on QNN-less hosts). Even when both are present, the compiled artifact is a **QNN** EPContext, but `assert_by_run_inference(..., ep=require_ep_name="openvino")` then asks ORT to load it via OpenVINO → mismatch.

The `(auto, None)` rows for openvino/vitisai/nv_tensorrt_rtx have the same problem (`auto` falls through to QNN in `_resolve_compile_provider`'s `else` branch).

Suggested: skip `(None, None)` and `(auto, None)` rows for any non-QNN `require_ep_name`, e.g.:

```python
for device in (*devices, "auto", None) if require_ep_name == "qnn" else devices
```

🤖 Generated with Claude Code

tests/e2e/test_compile_e2e.py
```diff
@@ -706,3 +634,142 @@ def test_reject_recompile_of_real_compiled_output(
...
+# ===========================================================================
+
+
+# Per-EP set of valid concrete ``--device`` values, keyed by EP alias.
+EP_DEVICE_SUPPORT: dict[str, tuple[str, ...]] = {
+    "qnn": ("npu", "gpu"),
+    "openvino": ("npu", "gpu", "cpu"),
+    "vitisai": ("npu",),
+    "nv_tensorrt_rtx": ("gpu",),
+}
```
timenick, line 649: **This duplicates the canonical `sysinfo.device._EP_DEVICE_MAP`.**

The public accessor `winml.modelkit.sysinfo.device.get_ep_device_map()` already encodes EP→device support, and `tests/unit/utils/test_ep_constants.py::test_matches_ep_device_map_keys` enforces parity with `SUPPORTED_EPS`. Hand-maintaining the same matrix here will drift the moment QNN gains CPU support or another EPContext-capable EP is added — and there's no test to catch the drift.

Suggested: derive from the public map, filtered to the EPContext-capable subset. For example:

```python
from winml.modelkit.sysinfo import get_ep_device_map
from winml.modelkit.utils.constants import normalize_ep_name

_EPCONTEXT_CAPABLE = ("qnn", "openvino", "vitisai", "nv_tensorrt_rtx")
EP_DEVICE_SUPPORT = {
    alias: tuple(get_ep_device_map()[normalize_ep_name(alias)].split("/"))
    for alias in _EPCONTEXT_CAPABLE
}
```

(The "EPContext-capable" list itself can also be derived from `for_provider` returning non-None, but keeping it explicit is fine.)

🤖 Generated with Claude Code

tests/e2e/test_compile_e2e.py
```diff
@@ -706,3 +634,142 @@ def test_reject_recompile_of_real_compiled_output(
...
+    src_hash = _sha256(simple_matmul_onnx)
+    result = _invoke("-m", str(simple_matmul_onnx), "--device", device, "--ep", ep)
+    _assert_rejected(result, "not supported by EP", src_hash, simple_matmul_onnx)
+
+
+@pytest.mark.e2e
+@pytest.mark.parametrize("ep", ["dml", "cpu", "cuda", "migraphx"])
+def test_bad_input_unsupported_ep(ep: str, simple_matmul_onnx: Path) -> None:
+    """``--ep X`` is rejected when X does not produce EPContext."""
+    require_ep("openvino")
```
timenick: **`require_ep("openvino")` is dead code here.**

The rejection path under test is the static EPContext-capability check at [commands/compile.py:194](../blob/cce8f22b94fb735b86926ac4b3e1d950083ce3e1/src/winml/modelkit/commands/compile.py#L194):

```python
config = WinMLCompileConfig.for_provider(provider, device=device)
if config is None:  # for_provider returns None for non-EPContext EPs
    raise click.ClickException("Provider '{...}' does not support EPContext compilation. ...")
```

`for_provider("DmlExecutionProvider", ...)` returns None deterministically, regardless of which EPs are registered on the host. So this test will produce the same result on every host. Guarding it with `require_ep("openvino")` only causes silent skips on OV-less hosts, shrinking coverage for no benefit.

Drop the `require_ep` call. (Or, if a sentinel-available EP is needed for an unrelated setup reason, use `"cpu"` — always available.)

🤖 Generated with Claude Code

tests/e2e/test_compile_e2e.py
```diff
@@ -706,3 +634,142 @@ def test_reject_recompile_of_real_compiled_output(
...
+    src_hash = _sha256(simple_matmul_onnx)
+    result = _invoke("-m", str(simple_matmul_onnx), "--ep", ep)
+    _assert_rejected(result, "is not available on this host", src_hash, simple_matmul_onnx)
+
+
+@pytest.mark.e2e
+def test_bad_input_no_ep_covers_device(simple_matmul_onnx: Path) -> None:
+    """``--device cpu`` (no ``--ep``) on a QNN-only host is rejected."""
+    require_ep("qnn")
+    require_not_ep("openvino")
```
timenick: **Same dead-`require` issue as `test_bad_input_unsupported_ep`.**

`--device cpu` + no `--ep` traces deterministically:
- `_resolve_compile_provider("cpu", None)` → `"CPUExecutionProvider"` (the `_DEVICE_TO_PROVIDER["cpu"] is None` branch)
- `for_provider("CPUExecutionProvider", ...)` returns None because `cls.for_cpu().ep_config.enable_ep_context` is False
- → `"does not support EPContext compilation"`

This is purely a static capability check. Host EP availability doesn't enter the picture, so `require_ep("qnn")` and `require_not_ep("openvino")` only constrain the test to a narrow CI environment without changing the assertion.

The docstring says *"on a QNN-only host is rejected"* — that framing implies the test wants to verify a **dynamic** check (no registered EP supports `cpu` for EPContext, so reject). But the source-side code doesn't implement that check; it just falls through to CPU's static `enable_ep_context=False`. So the two requires are effectively decorative.

Either:
1. Drop both requires — the test still verifies the existing static rejection on any host, **or**
2. Wait for the dynamic host-aware validation to land in src and tighten the assertion to that specific message.

🤖 Generated with Claude Code

tests/e2e/test_compile_e2e.py
zhenchaoni: ﻿Fix by rebasing on top of #641 (EP_SUPPORTED_DEVICES policy) and #660 (sysinfo preflight) and updating both assertions to `"no compatible EP is available"`, which is the actual ValueError sysinfo's device-resolution preflight now raises for these inputs. Verified passing on both an OpenVINO-only host and a QNN-only host (13 passed, 0 failed each).

tests/e2e/test_compile_e2e.py
zhenchaoni: Fix by reading the actual EP back from the compiled artifact's `EPContext.source` attribute when `--ep` is omitted, so the inference binding always matches what was compiled — regardless of which EP the resolver picked based on host registry order. Combined with #641's policy resolver and the existing `assert_banner_matches_artifact` check, the resolver/banner/artifact/runtime chain is fully consistent. Verified on OV-only and QNN-only hosts.

tests/e2e/test_compile_e2e.py
zhenchaoni: Fix by deriving `EP_DEVICE_SUPPORT` from the canonical `EP_SUPPORTED_DEVICES` dict added in #641 (single source of truth that also drives `sysinfo.device`):

```python
EP_DEVICE_SUPPORT: dict[str, tuple[str, ...]] = {
    alias: EP_SUPPORTED_DEVICES[normalize_ep_name(alias)]
    for alias in _EPCONTEXT_CAPABLE_EPS
}
```

The only hand-maintained list now is `_EPCONTEXT_CAPABLE_EPS` itself — kept explicit so adding an EPContext-capable EP requires a deliberate test opt-in.

tests/e2e/test_compile_e2e.py
zhenchaoni: The analysis pre-dates #641. After #641, the EP-registration check in `_resolve_compile_provider` fires *before* `for_provider`'s capability check, so on a host without the requested EP the rejection comes from registration (`"is not registered on this host"`) rather than from capability (`"does not support EPContext compilation"`). `require_ep(ep)` is now load-bearing — it ensures the test reaches and asserts the capability-check message it actually targets. (On hosts where the EP isn't registered, that path is instead exercised by `test_bad_input_ep_not_registered`.)

tests/e2e/test_compile_e2e.py
zhenchaoni: The analysis pre-dates #660. After #660, `sysinfo`'s preflight resolves `--device cpu` against the actual registry: on a host where OpenVINO is registered, it picks OpenVINO (which supports cpu) and the compile *succeeds* — the test would fail. `require_not_ep("openvino")` is now load-bearing: it scopes the test to hosts where no EPContext-capable EP covers cpu, which is exactly the condition under which the static `CPUExecutionProvider` capability rejection (`"does not support EPContext compilation"`) actually fires. Docstring updated to reflect this. Verified: OV host skips correctly; QNN-only host runs and the assertion passes.

vortex-captain (approved):

PR #665, opened 2026-05-19
test(e2e): add perf CLI tests for ONNX and per-module

xieofxie:
## Summary
- Refactor `tests/e2e/test_perf_e2e.py` into a shared `_PerfBenchmarkSuite` base class so the seven cross-cutting perf tests (CPU, verbose, GPU+monitor, NPU+monitor, auto device, QNN, QNN-on-GPU) only live in one place.
- Add `TestPerfONNXDirect` (existing ONNX-fixture coverage)
- Add `TestPerfModule` with two tests covering `--module` mode on `microsoft/resnet-50`: a successful `ResNetStage` per-module benchmark on CPU, and an invalid-module case that asserts the CLI fails and prints the list of available module class names.

### DIFF
--- a/tests/e2e/test_perf_e2e.py
+++ b/tests/e2e/test_perf_e2e.py
@@ -189,7 +189,7 @@ class TestPerfONNXDirect:
                 "--device",
                 "gpu",
                 "--iterations",
-                "3",
+                "100",
                 "--warmup",
                 "1",
                 "-o",
@@ -207,6 +207,9 @@ class TestPerfONNXDirect:
         assert data["benchmark_info"]["device"] == "gpu"
         assert data["latency_ms"]["mean"] > 0
         assert "hw_monitor" in data, "hw_monitor section missing with --monitor"
+        assert data["hw_monitor"]["device_kind"] == "gpu"
+        assert data["hw_monitor"]["adapter_luid"] is not None
+        assert data["hw_monitor"]["gpu"]["mean_pct"] > 0
 
     def test_onnx_benchmark_npu_monitor(self, tmp_path: Path, onnx_model_path: Path):
         """ONNX direct benchmark on NPU with --monitor.
@@ -227,7 +230,7 @@ class TestPerfONNXDirect:
                 "--device",
                 "npu",
                 "--iterations",
-                "3",
+                "100",
                 "--warmup",
                 "1",
                 "-o",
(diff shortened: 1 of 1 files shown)


xieofxie:
PerfHuggingFace will be added once analyze issue is resolved - only check one ep - or it will be too slow

tests/e2e/test_perf_e2e.py
```diff
@@ -97,26 +145,314 @@ def test_onnx_benchmark_cpu(self, tmp_path: Path, onnx_model_path: Path):
...
+            obj={},
+            catch_exceptions=False,
+        )
+        assert result.exit_code == 0, f"perf failed (exit {result.exit_code}):\n{result.output}"
+
+        assert output_file.exists()
+        data = json.loads(output_file.read_text())
+        assert data["benchmark_info"]["device"] == "auto"
+        # At leat a non-cpu should exist and picked up
+        assert data["benchmark_info"]["ep"] != "CPUExecutionProvider"
```
timenick, line 289: The docstring says "CPU is the universal fallback" and this test isn't gated by any GPU/NPU skip, but the assertion forbids `CPUExecutionProvider`. On a CPU-only host (e.g. CI without an accelerator) the test will fail even though `auto` correctly resolved to CPU. Either drop this assert (the `latency_ms.mean > 0` check below is sufficient to prove the run worked) or add a hardware-gate skip if the real intent is "auto must resolve to a non-CPU EP when one is present".

Also: `# At leat` → "At least".

🤖 Generated with Claude Code

tests/e2e/test_perf_e2e.py
```diff
@@ -97,26 +145,314 @@ def test_onnx_benchmark_cpu(self, tmp_path: Path, onnx_model_path: Path):
...
+# ===========================================================================
+# Concrete suites
+# ===========================================================================
+
+
+class TestPerfONNXDirect(_PerfBenchmarkSuite):
+    """Benchmark a pre-exported ONNX file directly via WinMLSession."""
+
+    @pytest.fixture
+    def model_arg(self, onnx_model_path: Path | None = None) -> str:
```
timenick: `Path | None = None` default and the `RuntimeError` branch are unreachable. Pytest fixtures are resolved by name, not by parameter defaults; if `onnx_model_path` were missing, pytest itself would raise a clear `fixture not found` error before this body runs. Reduce to:

```python
@pytest.fixture
def model_arg(self, onnx_model_path: Path) -> str:
    return str(onnx_model_path)
```

🤖 Generated with Claude Code

tests/e2e/test_perf_e2e.py
xieofxie: It will not run on CI and DML also always exist and should be preferred to CPU

tests/e2e/test_perf_e2e.py
xieofxie: CC should be clever。 Revert copilot's comment

DingmaomaoBJTU (approved):

PR #667, opened 2026-05-19
Build: auto-select EP via resolve_device, forward --ep from run_eval (#663)

### LINKED ISSUE #663: Black nodes from prior EP analysis block `winml build --ep <other-ep>`
## Repro

```
python.exe -m winml.modelkit.cli build \
  -c eval_results\2026-05-15\models\BAAI__bge-base-en-v1.5__sentence-similarity\build_config.json \
  -m BAAI/bge-base-en-v1.5 \
  --use-cache
```

Intent: build for **TrtRtx**. Actual: build is blocked by *persisting black nodes* left over from an analysis pass targeting **OpenVINO**.

## Root cause

`winml build` has no `--ep` passed by the caller, so it falls into the auto-select branch in `src/winml/modelkit/commands/build.py:380-394`:

```python
candidate_eps: list[EPName] = [
    \"QNNExecutionProvider\",
    \"OpenVINOExecutionProvider\",
    \"VitisAIExecutionProvider\",
]
for candidate_ep in candidate_eps:
    if registry.is_ep_available(candidate_ep):
        ep = candidate_ep
        break
```

That list is **device-agnostic**: on a TrtRtx GPU box where OpenVINO is also installed, OV wins the lookup. Analyze then runs OV rules, marks ops as black, and the artifacts persist in the cache. A subsequent build (or even the same build trying to compile for TrtRtx) sees those black nodes and refuses to proceed.

The eval driver in \`scripts/e2e_eval/run_eval.py\` makes this easier to hit, because the loop already knows the target EP (\`args.ep\`) but does not forward it to the build subprocess — see \`run_eval.py:436-444\`:

```python
build_args = [
    *WINML_CLI,
    \"build\",
    \"-c\", str(sub_cfg),
    \"-m\", entry.hf_id,
    \"--use-cache\",
]
```

So whatever EP eval was configured for, build silently chooses a different one.

## Proposed fix

Two changes, both small:

**1. \`scripts/e2e_eval/run_eval.py\`** — forward \`--ep\` to the build command, the same way it is already forwarded to perf/eval at lines 408/562/592/730:

```python
build_args = [*WINML_CLI, \"build\", \"-c\", str(sub_cfg), \"-m\", entry.hf_id, \"--use-cache\"]
if ep:
    build_args += [\"--ep\", ep]
```

**2. \`src/winml/modelkit/commands/build.py\`** — when \`--ep\` is unspecified, resolve the **device** first, then ask \`resolve_eps(device)\` for the device-compatible EPs and pick the first. Replaces the hardcoded \`[QNN, OV, VitisAI]\` walk that ignores the host device:

```python
if ep is None:
    from ..sysinfo import resolve_device, resolve_eps
    resolved_device, _ = resolve_device(device=device or \"auto\")
    compatible_eps = resolve_eps(resolved_device)
    if compatible_eps:
        ep = compatible_eps[0]
        logger.info(\"EP unspecified for build, auto-selecting %s for device %s\", ep, resolved_device)
```

(1) closes the gap from the eval driver side; (2) makes plain \`winml build\` correct even when called directly. Both helpers (\`resolve_device\`, \`resolve_eps\`) already exist in \`src/winml/modelkit/sysinfo/device.py\`.

vortex-captain:
## Summary

Fixes #663.

`winml build` used to auto-select an EP from a hardcoded list `[QNN, OV, VitisAI]` and pick the first one installed, regardless of the host device. On a TrtRtx GPU box where OpenVINO is also installed, OV would win — the analyze pass then marked ops black for OV and the residual state blocked a later TrtRtx-targeted build.

### Changes

**`src/winml/modelkit/commands/build.py`**
- When `--ep` is omitted, resolve the target device via `resolve_device(device)` and pick the first EP from `resolve_eps(resolved_device)`. `resolve_device` either returns a device with at least one available EP (auto-mode walks the priority list, falls back to CPU which is always valid) or raises `ValueError` for an explicit device with no compatible EP — surfaced as `click.UsageError`. So `resolve_eps(resolved_device)[0]` is safe whenever `resolve_device` returns.
- Switched the `--device` option to `cli_utils.device_option(required=False, default=\"auto\", include_auto=True, optional_message=\"Default: auto-detect.\")` to match `perf`/`run`/`serve`. Guarantees `device` is never `None` in the auto-select block.

**`scripts/e2e_eval/run_eval.py`**
- Forward `--ep` to the build subprocess when set, matching how perf/eval already do (`run_eval.py:408/562/592/730`).

**`tests/unit/commands/test_build.py`**
- New `TestBuildEpAutoSelection` (6 tests) covering: auto-select per resolved device (npu→QNN, gpu→Dml, cpu→CPU), explicit `--ep` bypassing auto-selection, `resolve_device` `ValueError` → `UsageError`, and `resolve_eps` priority order.
- Existing device assertions updated from `\"NPU\"` to `\"npu\"` to match `click.Choice(case_sensitive=False)` normalization.

### Verification

- `python -m pytest tests/unit/commands/test_build.py` → 78 passed.
- E2E: `python scripts/e2e_eval/run_eval.py --hf-model BAAI/bge-base-en-v1.5 --ep cpu --eval-type perf` → exit 0, `[PASS]`, `--ep cpu` reached the build + perf subprocesses.

### DIFF
--- a/scripts/e2e_eval/run_eval.py
+++ b/scripts/e2e_eval/run_eval.py
@@ -442,6 +442,8 @@ def _run_build(
             entry.hf_id,
             "--use-cache",
         ]
+        if ep:
+            build_args += ["--ep", ep]
 
         build_proc = _run_subprocess(build_args, timeout)
         last_proc = build_proc
--- a/src/winml/modelkit/commands/build.py
+++ b/src/winml/modelkit/commands/build.py
@@ -284,10 +284,11 @@ def _build_modules(
     required=False,
     optional_message="Falls back to compile config EP if not set.",
 )
-@click.option(
-    "--device",
+@cli_utils.device_option(
+    required=False,
     default="auto",
-    help="Target device ('auto', 'npu', 'gpu', 'cpu'). Default: auto-detect.",
+    include_auto=True,
+    optional_message="Default: auto-detect.",
 )
 @click.option(
     "--no-analyze",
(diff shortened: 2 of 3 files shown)


xieofxie (approved):

DingmaomaoBJTU:
### Code review

Found 1 issue:

1. `_resolve_eps(resolved_device)[0]` raises `IndexError` when no EPs are installed for the resolved device — the comment at line 386 claims this is always safe because `resolve_device` falls back to CPU, but `resolve_eps` filters by *available* EPs (those found via DLL scanning). On a machine where no WinML/ORT EP DLLs are present, `resolve_eps("cpu")` returns `[]` and `[0]` crashes unconditionally. The old code had an explicit fallback (`logger.warning(...)` + continuing without EP hints) that has been removed without a replacement guard.

https://github.com/microsoft/winml-cli/blob/b632086de74c8206f90b1c8219d90be9a9789e01/src/winml/modelkit/commands/build.py#L386-L401

PR #669, opened 2026-05-19
test(cli): add CLI surface tests for `winml catalog`

DingmaomaoBJTU:
## Summary

`winml catalog` has had no test coverage since it shipped. This PR adds `tests/cli/test_catalog_cli.py` — 48 tests across 8 classes that exercise the full CLI surface and filtering contract of the catalog command through `CliRunner`.

The structure follows the pattern established by #672 (`winml --help` CLI tests).

## Why `tests/cli/`, not `tests/e2e/`

`winml catalog` is pure data filtering against `hub_models.json` — no model downloads, no hardware, no network. Every test completes in under 10 s on any host. Placing them in `tests/e2e/` behind `pytest.mark.e2e` would hide them from the default CI run for no reason.

### CLI surface

- `winml catalog --help` — every documented option present in stdout
- `winml catalog` (no filter flags) — exits 0, returns the full non-empty catalog
- `winml catalog --ep <bogus>` — Click rejects with `Invalid value for '--ep'`, exit 2
- `winml catalog --device TPU` — Click rejects with `Invalid value for '--device'`, exit 2
- `winml catalog -t <type> -k <task>` — short flags `-t` / `-k` accepted

### Filter by `--model-type` / `-t`

- Known type returns only entries with that `model_type`
- Filter is **case-insensitive**
- Known type produces a strict subset of the unfiltered catalog
- Unknown type writes `[]` to `--output`, exits 0
- Two distinct types produce **disjoint** model-ID sets

### Filter by `--task` / `-k`

- Known task returns only matching entries
- Case-insensitive
- Strict subset of unfiltered catalog
- Unknown task writes `[]`, exits 0
- Two distinct tasks are disjoint

### Filter by `--ep`

- `--ep qnn` / `--ep vitisai` — returned models all carry that EP in `supported_eps`
- `vitisai` returns a strict subset (not universally supported)
- `--ep ov` and `--ep openvino` resolve to the **same model set**
- `--ep vitisai` and `--ep VitisAIExecutionProvider` resolve to the **same model set**

### Filter by `--device`

- `--device NPU`, `--device CPU`, `--device GPU` — returned models all support that device
- Case-insensitive (`NPU` == `npu`)
- `--device FPGA` rejected by Click (exit 2)

### Combined filters

- `--model-type <t> --task <k>` — intersection returns only matching entries
- Intersection is ≤ either filter alone
- `--ep <ep> --model-type <t>` — returned models satisfy both constraints
- `--ep qnn --device NPU` — independent filters applied; models have QNN and support NPU
- Disjoint `(model_type, task)` pair → `[]`
- `--ep vitisai --task <k>` — subset of `--ep vitisai` alone

### `--output` / `-o` file

- Writes a valid JSON **list** (not a dict)
- Every entry has required fields: `model_id`, `task`, `model_type`, `supported_eps`
- Combined with a filter → only matching entries written
- Unknown filter → writes `[]`
- **Creates parent directories** that don't exist yet
- Short flag `-o` accepted

### EP + device combination

- Only `--ep` given → every returned model contains that EP
- Only `--device` given → every returned model supports that device
- Both given → result is ⊆ each filter alone (intersection)
- Neither given → full catalog returned

## Design notes

- **All data assertions go through `--output` JSON**, not Rich console stdout. The module-level `Console` in `catalog.py` is initialised before `CliRunner` patches `sys.stdout`, so `result.output` is not a reliable carrier of Rich table content.
- **Test values derived from `hub_models.json` at import time** — no hardcoded architecture or task names (CLAUDE.md Cardinal Rule #1). `@pytest.mark.skipif` guards on catalog shape are evaluated at collection time, so skipped tests are visible without running.
- **No EP runtime required** — catalog is pure data filtering; all 48 tests run on any host.

## Running the tests

```
uv run pytest tests/cli/test_catalog_cli.py -v
```


tests/e2e/test_catalog_e2e.py
```diff
@@ -0,0 +1,519 @@
...
+    assert isinstance(data, list), f"--output JSON must be a list, got {type(data)}"
+    return data
+
+
+# ---------------------------------------------------------------------------
+# CLI surface
+# ---------------------------------------------------------------------------
+
+
+@pytest.mark.e2e
```
zhenchaoni: Redundant marker: `pytestmark = [pytest.mark.e2e]` at line 57 already applies `e2e` to every test in this file. The per-class `@pytest.mark.e2e` decorators on lines 88, 158, 201, 241, 291, 326, 422, 481 are duplicates. Suggest removing all of them.

tests/e2e/test_catalog_e2e.py
```diff
@@ -0,0 +1,519 @@
...
+
+    def test_invalid_device_choice_exits_two(self) -> None:
+        result = _invoke("--device", "TPU")
+        assert result.exit_code == 2
+        assert "Invalid value for '--device'" in result.output
+
+    def test_short_flags_accepted(self, tmp_path: Path) -> None:
+        """-t and -k short aliases are accepted by the parser."""
+        models = _invoke_json(tmp_path / "out.json", "-t", "bert", "-k", "text-classification")
+        assert result_is_subset_of_full_catalog(
```
zhenchaoni: Vacuous-truth risk: `result_is_subset_of_full_catalog([], ...)` returns `True`, so this test would also pass if `-t bert -k text-classification` returned zero models. Add `assert len(models) > 0` before the subset check (or assert on the exact types/tasks present).

zhenchaoni (commented):
Solid coverage — the data-filter-via-`--output`-JSON approach is the right call given the module-level `Console`. Ran the suite locally against current `main`: all 43 tests pass in ~2s. A few inline notes; nothing blocking.

tests/e2e/test_catalog_e2e.py
```diff
@@ -0,0 +1,519 @@
...
+            models, model_type="bert", task="text-classification"
+        )
+
+
+# ---------------------------------------------------------------------------
+# Helpers for content assertions
+# ---------------------------------------------------------------------------
+
+
+def result_is_subset_of_full_catalog(
```
zhenchaoni: Naming inconsistency: `result_is_subset_of_full_catalog` is the only module-level helper without a leading underscore — `_invoke`, `_invoke_json`, `_all_model_ids`, `_all_model_types`, `_all_tasks`, `_ep_keys`, `_supports_device` are all private. Suggest `_result_is_subset_of_full_catalog`, or inline the check into `test_short_flags_accepted` (its only caller).

tests/e2e/test_catalog_e2e.py
```diff
@@ -0,0 +1,519 @@
...
+        assert 0 < len(bert_models) < len(all_models)
+
+    def test_unknown_model_type_returns_empty_list(self, tmp_path: Path) -> None:
+        out = tmp_path / "empty.json"
+        result = _invoke("--model-type", "nonexistent_arch_xyz", "--output", str(out))
+        assert result.exit_code == 0, result.output
+        assert out.is_file()
+        assert json.loads(out.read_text()) == []
+
+    def test_different_model_types_return_disjoint_sets(self, tmp_path: Path) -> None:
```
zhenchaoni: Hardcoded architecture names (`bert`/`vit`/`resnet`) and task names (`text-classification`/`image-classification`) appear in many tests across the file. If any are dropped from `hub_models.json` later, the suite silently regresses with a misleading message. Suggest picking two distinct types and two distinct tasks at module load from `hub_models.json` (the same source-of-truth pattern `tests/e2e/conftest.py::_unique_pairs` uses) and parametrising. Keeps the contract "filtering is correct" intact while letting the catalog evolve.

tests/e2e/test_catalog_e2e.py
```diff
@@ -0,0 +1,519 @@
...
+            tmp_path / "combined.json",
+            "--model-type",
+            "bert",
+            "--task",
+            "text-classification",
+        )
+        assert len(combined) <= len(bert_all)
+        assert len(combined) <= len(textcls_all)
+
+    def test_ep_and_model_type_intersection(self, tmp_path: Path) -> None:
```
zhenchaoni: Missing existence check: if no bert model happens to be vitisai-capable, `models == []` and the `for` body never executes, so the test trivially passes. Add `assert len(models) > 0` after the call. (If an empty result is genuinely acceptable, rename the test to reflect that.)

tests/e2e/test_catalog_e2e.py
```diff
@@ -0,0 +1,519 @@
...
+        assert isinstance(data, list)
+
+
+# ---------------------------------------------------------------------------
+# EP column XOR logic
+# ---------------------------------------------------------------------------
+
+
+@pytest.mark.e2e
+class TestCatalogEpColumnXor:
```
zhenchaoni: Class name `TestCatalogEpColumnXor` and its docstring imply this verifies the 5th-column rendering (`Devices` vs `EPs`), but the assertions only check the underlying filtered model set — the column header is never validated. As the module docstring acknowledges, Rich output isn't captured by `CliRunner` here. Two options:

1. Rename to `TestCatalogEpAndDeviceCombination` (or similar) and update the docstring so the scope isn't overstated.
2. Follow-up: monkey-patch the module-level `console` in `commands/catalog.py` to `Console(file=io.StringIO(), force_terminal=True)` via a fixture and assert the column headers (`"Devices"` / `"EPs"`) directly.

tests/e2e/test_catalog_e2e.py
```diff
@@ -0,0 +1,519 @@
...
+        assert result.exit_code == 0
+        assert json.loads(out.read_text()) == []
+
+    def test_ep_and_task_intersection_subset_of_ep_alone(self, tmp_path: Path) -> None:
+        vitisai_all = _invoke_json(tmp_path / "vitisai.json", "--ep", "vitisai")
+        vitisai_imgcls = _invoke_json(
+            tmp_path / "vitisai_imgcls.json",
+            "--ep",
+            "vitisai",
+            "--task",
```
zhenchaoni: Optional coverage gap: the `_load_catalog` failure → `ClickException` branch in `commands/catalog.py` is never exercised. One small test with `monkeypatch.setattr(...)` to make `_load_catalog` raise would close it. Pure helpers `_fmt_size` / `_type_color` are unit-test territory rather than e2e — worth a quick follow-up file.

tests/e2e/test_catalog_e2e.py
DingmaomaoBJTU: Fixed: removed all per-class `@pytest.mark.e2e` decorators from `TestCatalogCliSurface`, `TestCatalogFilterModelType`, `TestCatalogFilterTask`, `TestCatalogFilterEp`, `TestCatalogFilterDevice`, `TestCatalogCombinedFilters`, `TestCatalogOutputFile`, and `TestCatalogEpAndDeviceCombination`. The module-level `pytestmark` already applies.

tests/e2e/test_catalog_e2e.py
DingmaomaoBJTU: Fixed: added `assert len(models) > 0, "Expected at least one bert/text-classification model"` before the subset check in `test_short_flags_accepted`.

tests/e2e/test_catalog_e2e.py
DingmaomaoBJTU: Fixed: renamed to `_result_is_subset_of_full_catalog` to match the leading-underscore convention of all other module-level helpers.

tests/e2e/test_catalog_e2e.py
DingmaomaoBJTU: Fixed: added `assert len(bert) > 0` and `assert len(vit) > 0` in `test_different_model_types_return_disjoint_sets`, and similar existence checks in `test_different_tasks_return_disjoint_sets`. An empty filter result can no longer trivially satisfy `isdisjoint()`.

tests/e2e/test_catalog_e2e.py
DingmaomaoBJTU: Fixed: added `assert len(models) > 0, "Expected at least one bert model with vitisai support"` after the `_invoke_json` call. Verified against `hub_models.json` — bert models with vitisai support exist.

tests/e2e/test_catalog_e2e.py
DingmaomaoBJTU: Fixed: renamed to `TestCatalogEpAndDeviceCombination` and updated the docstring to accurately state that the class tests filtered model sets (not Rich column headers, which `CliRunner` cannot capture when `Console` is module-level).

tests/e2e/test_catalog_e2e.py
DingmaomaoBJTU: Noted — tracked as a follow-up. The `_load_catalog` failure path and pure helper unit tests (`_fmt_size`, `_type_color`) are out of scope for this e2e PR.

tests/e2e/test_catalog_e2e.py
```diff
@@ -0,0 +1,517 @@
...
+        assert len(vit) > 0, "Expected at least one vit model — add 'vit' to hub_models.json"
+        assert _all_model_ids(bert).isdisjoint(_all_model_ids(vit)), (
+            "bert and vit filters must not overlap"
+        )
+
+
+# ---------------------------------------------------------------------------
+# Filter by task
+# ---------------------------------------------------------------------------
+
```
timenick: The whole suite hard-codes specific architectures (`bert`, `vit`, `resnet`) and tasks (`text-classification`, `image-classification`) as required catalog contents — e.g. `test_different_model_types_return_disjoint_sets` on line 188 will fail outright if `vit` is ever removed from `hub_models.json`. The PR even acknowledges this with hints like *"add ''bert'' to hub_models.json"* in the failure messages.

That clashes with `CLAUDE.md` → Cardinal Rule #1:

> Never hardcode model architecture names, node/operator names, input/output tensor names, layer naming patterns, or any model-specific logic. All solutions must be universal and architecture-agnostic.

Please pick types/tasks dynamically from the catalog at fixture scope so the tests assert *filter behaviour* against whatever data is actually present:

```python
from collections import Counter

@pytest.fixture(scope="module")
def catalog_data(tmp_path_factory) -> list[dict]:
    return _invoke_json(tmp_path_factory.mktemp("catalog") / "all.json")

@pytest.fixture(scope="module")
def two_model_types(catalog_data) -> tuple[str, str]:
    counts = Counter(m["model_type"] for m in catalog_data)
    distinct = [t for t, _ in counts.most_common() if t]
    if len(distinct) < 2:
        pytest.skip("catalog needs ≥2 model_types for this test")
    return distinct[0], distinct[1]
```

Same pattern for tasks and EPs. The suite then locks down *invariants* (subsets, disjointness, case-insensitivity) instead of pinning the current contents of `hub_models.json`, and stays green as the catalog evolves.

tests/e2e/test_catalog_e2e.py
```diff
@@ -0,0 +1,517 @@
...
+def _invoke(*args: str) -> Result:
+    """Run ``winml catalog <args>`` through a fresh CliRunner."""
+    return CliRunner().invoke(catalog, list(args), obj={})
+
+
+def _invoke_json(out_path: Path, *args: str) -> list[dict]:
+    """Run ``winml catalog --output <out_path> <args>`` and return parsed JSON.
+
+    Asserts exit 0 and a valid JSON list before returning.
```
timenick: `tests/e2e/conftest.py` already exposes a `runner` fixture (lines 133-137). Per `CLAUDE.md` → Testing:

> New fixtures or mocks should check if an existing `conftest.py` already provides equivalent functionality that can be extended, rather than duplicating in a local scope.

Two options that are cleaner than the current free function:

1. Take the existing fixture and build on it:

   ```python
   @pytest.fixture
   def invoke_catalog(runner):
       def _invoke(*args: str) -> Result:
           return runner.invoke(catalog, list(args), obj={})
       return _invoke
   ```

2. Or promote `_invoke` / `_invoke_json` to `tests/e2e/conftest.py` so future catalog (and similar Click-command) E2E suites can reuse them, with the existing `runner` fixture as the underlying invoker.

Either way avoids constructing a fresh `CliRunner()` per call and keeps the E2E layer''s test-infra in one place.

tests/e2e/test_catalog_e2e.py
```diff
@@ -0,0 +1,517 @@
...
+        for opt in ("--model-type", "--task", "--ep", "--device", "--output"):
+            assert opt in result.output, f"--help missing option {opt!r}"
+
+    def test_no_args_exits_zero_and_returns_all_models(self, tmp_path: Path) -> None:
+        """``winml catalog`` with no filters returns the full catalog."""
+        models = _invoke_json(tmp_path / "all.json")
+        assert len(models) > 0, "Catalog must not be empty"
+
+    def test_invalid_ep_choice_exits_two(self) -> None:
+        result = _invoke("--ep", "totally_unknown_ep_xyz")
```
timenick: Minor naming nit: this test is `test_no_args_exits_zero_and_returns_all_models`, but it actually invokes catalog *with* `--output <path>` (via `_invoke_json`). A reader scanning the test list will read "no args" and expect a no-`--output` invocation.

Rename to e.g. `test_no_filter_args_returns_all_models` so the name matches what the test exercises. (If you also want a true no-args case — bare `winml catalog` without `--output` — that would be a separate test asserting `result.exit_code == 0` and non-empty stdout.)

tests/e2e/test_catalog_e2e.py
DingmaomaoBJTU: Addressed by reworking the entire test file: the suite is now in `tests/cli/test_catalog_cli.py` (no `e2e` marker, no hardcoded names). All test values — model types, tasks, EP/type pairs, and the disjoint type/task pair — are derived from `hub_models.json` at module import time via module-level constants (`_MODEL_TYPES`, `_TASKS`, `_TYPE_TASK_PAIR`, `_EP_MODEL_TYPE_PAIR`, `_DISJOINT_TYPE_TASK`). The suite asserts filter invariants against whatever data is actually present, so it stays green as the catalog evolves.

tests/e2e/test_catalog_e2e.py
DingmaomaoBJTU: Addressed. The test file was moved to `tests/cli/test_catalog_cli.py` following the PR #672 pattern. `_run` creates a fresh `CliRunner()` per call (same style as `test_inspect_cli.py`) — no module-level singleton. The `runner` fixture from `conftest.py` is not used here because the catalog-specific helper is local to this file and not shared across suites.

tests/e2e/test_catalog_e2e.py
DingmaomaoBJTU: Fixed. Renamed to `test_no_filter_args_exits_zero` in `tests/cli/test_catalog_cli.py`.

tests/cli/test_catalog_cli.py
```diff
@@ -0,0 +1,509 @@
...
+_HUB_JSON = (
+    Path(__file__).resolve().parents[2] / "src" / "winml" / "modelkit" / "data" / "hub_models.json"
+)
+_CATALOG: list[dict] = json.loads(_HUB_JSON.read_text(encoding="utf-8"))["models"]
+
+_MODEL_TYPES: list[str] = [
+    t
+    for t, _ in Counter(m["model_type"] for m in _CATALOG if m.get("model_type")).most_common()
+    if t
+]
```
timenick: Two related issues with reading `hub_models.json` directly at module-import time:

**1. Collection-time failure mode.** This runs during pytest collection, so a missing file, a moved file, or a top-level schema change (e.g. `{"version": ..., "models": [...]}` → `{"models": [...]}` reversed) doesn''t fail this test — it fails *collection of the entire `tests/cli/` directory*, including `test_main.py`, `test_inspect_cli.py`, and `test_import_time.py`. CI shows `error during collection` with a `FileNotFoundError` / `KeyError: ''models''` traceback rather than a meaningful test failure. Note that with the v2 move out of `tests/e2e/`, this whole file now runs on every PR via the `remaining` matrix group (`.github/workflows/modelkit-ci.yml:38`), so the blast radius is larger than v1.

**2. Circular dependency on the command''s own data layout.** `winml catalog`''s whole job is to be the user-facing view of this JSON. By reading the same file the command reads, the test couples to (a) the on-disk path `src/winml/modelkit/data/hub_models.json`, and (b) the implicit `{"models": [...]}` schema. If `catalog` ever changes its data source (packaged resource, env-var override, remote), this test would silently fall out of sync — it would keep passing while no longer testing the user-visible CLI surface.

The cleaner inversion is to invoke `catalog` once at module/session scope and derive `_MODEL_TYPES` / `_TASKS` / `_TYPE_TASK_PAIR` from *that* output:

```python
@pytest.fixture(scope="module")
def catalog_models(tmp_path_factory) -> list[dict]:
    out = tmp_path_factory.mktemp("cat") / "all.json"
    result = CliRunner().invoke(catalog, ["--output", str(out)], obj={})
    if result.exit_code != 0:
        pytest.skip(f"catalog command unavailable: {result.output}")
    return json.loads(out.read_text())

@pytest.fixture(scope="module")
def model_types(catalog_models) -> list[str]:
    return [
        t for t, _ in Counter(m["model_type"] for m in catalog_models).most_common() if t
    ]
```

The `@pytest.mark.skipif(not _MODEL_TYPES, ...)` guards then move inside individual tests (or into the fixture itself):

```python
def test_known_type_returns_only_matching_entries(self, model_types, tmp_path):
    if not model_types:
        pytest.skip("catalog has no model_type entries")
    ...
```

Upshot:
- Tests assert behaviour against whatever the *command* says is in the catalog, not whatever we can scrape out of a file path.
- File missing / schema changed → graceful skip, no collection crash.
- No `parents[2] / "src" / "winml" / "modelkit" / "data" / ...` hardcoded path leaking the source tree layout into a test file.

tests/cli/test_catalog_cli.py
```diff
@@ -0,0 +1,509 @@
...
+        assert "--device" in _run("--help").output
+
+    def test_help_documents_output_flag(self) -> None:
+        assert "--output" in _run("--help").output
+
+    def test_no_filter_args_exits_zero(self, tmp_path: Path) -> None:
+        """``winml catalog`` with no filters returns the full catalog."""
+        models = _run_json(tmp_path / "all.json")
+        assert len(models) > 0, "Catalog must not be empty"
+
```
timenick, line 176: These five `test_help_documents_*_flag` methods (lines 138-151) all do the same thing — invoke `_run("--help")` and assert one flag string is present in the output. Two observations:

1. **Each call re-renders the whole help page.** Catalog''s `--help` path runs `_load_catalog()`, which reads `hub_models.json`, so we''re re-doing that 5x for what is one observation about a static string.
2. **Adding a new top-level flag means copy-pasting another method.** That''s easy to forget — the test suite drifts out of sync with the option list.

Collapse to one parametrized test, ideally backed by a class-scoped fixture so `--help` is only invoked once for the whole group:

```python
@pytest.fixture(scope="class")
def help_output(self) -> str:
    return _run("--help").output

@pytest.mark.parametrize(
    "flag",
    ["--model-type", "--task", "--ep", "--device", "--output"],
)
def test_help_documents_flag(self, help_output: str, flag: str) -> None:
    assert flag in help_output
```

Same coverage, one invocation, and adding a new top-level option only requires extending the `parametrize` list.

tests/cli/test_catalog_cli.py
DingmaomaoBJTU: Fixed — removed all module-level `hub_models.json` reads. All derived constants (`_MODEL_TYPES`, `_TASKS`, `_TYPE_TASK_PAIR`, `_EP_MODEL_TYPE_PAIR`, `_DISJOINT_TYPE_TASK`) are replaced by module-scoped pytest fixtures (`catalog_models`, `model_types`, `tasks`, `type_task_pair`, `ep_model_type_pair`, `disjoint_type_task`) that invoke the `catalog` command itself. If the command fails, affected tests skip gracefully — no collection crash. `@pytest.mark.skipif` decorators became inline `pytest.skip()` guards inside each test body.

tests/cli/test_catalog_cli.py
DingmaomaoBJTU: Fixed — collapsed the 5 `test_help_documents_*_flag` methods into a single `@pytest.mark.parametrize` test backed by a `@pytest.fixture(scope="class")` `help_output` fixture (one `--help` invocation per class).

timenick (approved):

tests/cli/test_catalog_cli.py
```diff
@@ -0,0 +1,556 @@
...
+@pytest.fixture(scope="module")
+def tasks(catalog_models: list[dict]) -> list[str]:
+    """Most-common task values present in the catalog."""
+    return [
+        t for t, _ in Counter(m["task"] for m in catalog_models if m.get("task")).most_common() if t
+    ]
+
+
+@pytest.fixture(scope="module")
+def type_task_pair(catalog_models: list[dict]) -> tuple[str, str]:
```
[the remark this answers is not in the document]
DingmaomaoBJTU: Fixed — replaced bare `pytest.skip()` at the end of `type_task_pair` and `ep_model_type_pair` with `return pytest.skip(...)`. `pytest.skip()` is typed `NoReturn` so the `return` is unreachable at runtime, but every exit path is now an explicit return and CodeQL's `py/mixed-returns` rule is satisfied.

tests/cli/test_catalog_cli.py
```diff
@@ -0,0 +1,556 @@
...
+    """First (model_type, task) pair that exists in the catalog."""
+    for m in catalog_models:
+        if m.get("model_type") and m.get("task"):
+            return m["model_type"], m["task"]
+    pytest.skip("catalog has no model with both model_type and task")
+
+
+@pytest.fixture(scope="module")
+def ep_model_type_pair(catalog_models: list[dict]) -> tuple[str, str]:
```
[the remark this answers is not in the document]
DingmaomaoBJTU: Same fix as above — `return pytest.skip(...)` in `ep_model_type_pair`.

timenick (approved):

PR #672, opened 2026-05-19
test(cli): add CLI surface tests for `winml` and `winml --help`

### LINKED ISSUE #481: Identify functional scenarios and implement E2E tests for `winml`
### Description
Identify the functionality list for `winml` and implement E2E pytest cases covering it. Tests live in [tests/e2e](../tree/main/tests/e2e).

Coverage must include:

### Deliverables

### Acceptance Criteria
- `uv run pytest tests/e2e/ -k top_level` is green.

DingmaomaoBJTU:
Closes #481

## Summary

Adds `tests/cli/test_help_cli.py` — **48 tests** pinning the output contract of `winml` (no args) and `winml --help` / `winml -h`. Supersedes and consolidates the overlapping `TestBanner` and `TestDisabledCommands` classes previously in `test_main.py`.

```
uv run pytest tests/cli/test_help_cli.py -v
```

### `TestWinmlNoArgs` — `winml`
- Exit 0
- `Usage:`, `Options`, `Commands` sections present
- Banner ("Windows ML") on `stderr`; current package version in banner
- Description ("WinML CLI") present

### `TestWinmlHelp` — `winml --help` / `-h`
- Exit 0 for both `--help` and `-h`
- `--help` and `-h` produce **identical** output
- `winml` and `winml --help` produce **identical** output
- Banner and description present
- Subcommand `--help` does **not** emit the top-level banner

### `TestCommandList` — Commands section
- Every enabled command listed — parametrized over the full set discovered from `_COMMANDS_DIR`
- Disabled commands (`run`, `serve`) **not** listed — checked via `splitlines()` + `startswith` (no false-positive substring matches)
- Disabled commands **rejected on invocation** — exit non-zero with `"disabled"` in output
- Every listed command has **non-empty AST-extracted help text** (catches missing docstrings or broken `_parse_click_help`)

### `TestOptionsSection` — Options section
- All documented options present: `--version`, `--verbose`/`-v`, `--quiet`/`-q`, `--help`/`-h`
- `winml --version` exits 0 and emits the current `__version__`

## Design notes

- Imports `_COMMANDS_DIR` and `_DISABLED_COMMANDS` directly from `cli.py` — CLAUDE.md Import Rules explicitly permits private-symbol imports in tests for implementation details. The public aliases added in a previous iteration have been reverted.
- `TestBanner` and `TestDisabledCommands` removed from `test_main.py`; the one unique test (`test_subcommand_help_has_no_banner`) is folded into `TestWinmlHelp`. Single source of truth, no double CI cost.


tests/e2e/test_help_e2e.py
```diff
@@ -0,0 +1,191 @@
...
+from __future__ import annotations
+
+import pytest
+from click.testing import CliRunner
+
+from winml.modelkit import __version__
+from winml.modelkit.cli import _COMMANDS_DIR, _DISABLED_COMMANDS, main
+
+
+pytestmark = [pytest.mark.e2e]
```
xieofxie: **Wrong test directory.** `tests/CLAUDE.md` says: *"Place cross-cutting CLI-surface tests (startup, import budget, arg parsing, command discovery, **version/help output**) under `tests/cli/`"* ([source](https://github.com/microsoft/winml-cli/blob/98c9013d334d5370facbeef39985089cf47f0139/tests/CLAUDE.md#L7-L13)). This file pins help output, command discovery, and option parsing — exactly the categories the rule names, so it belongs in `tests/cli/`, not `tests/e2e/`.

Side effect: the `@pytest.mark.e2e` marker plus the default `-m "not e2e"` CI filter means these 47 tests will not run in regular CI.

tests/cli/test_help_cli.py
```diff
@@ -0,0 +1,191 @@
...
+        out = _invoke("--help").output
+        commands_start = out.find("Commands")
+        assert commands_start != -1, "Commands section not found"
+        commands_block = out[commands_start:]
+        for line in commands_block.splitlines():
+            stripped = line.strip()
+            if stripped.startswith((cmd + " ", cmd + "\t")):
+                tail = stripped[len(cmd) :].strip()
+                assert tail, f"'{cmd}' has no help text in winml --help"
+                break
```
xieofxie, line 187: **Silent pass when no command row is found.** The `for` loop only asserts inside the `if stripped.startswith((cmd + " ", cmd + "\t"))` branch. If Click renders a row without the expected separator — e.g., the docstring is empty so the row is just `"cmdname"` after `strip()` — the loop falls through without asserting and the parametrized test passes. That's the exact "missing help text" case this test is supposed to catch.

Fix: add an `else:` clause on the `for` (or `pytest.fail`) so an unmatched row is a hard failure.

tests/e2e/test_help_e2e.py
```diff
@@ -0,0 +1,191 @@
...
+
+    def test_commands_section_present(self) -> None:
+        assert "Commands" in _invoke().output
+
+    def test_options_section_present(self) -> None:
+        assert "Options" in _invoke().output
+
+    def test_banner_present(self) -> None:
+        """The 'Windows ML' footer line of the gradient banner must appear."""
+        assert "Windows ML" in _invoke().output
```
xieofxie: **Banner asserted on the mixed `.output` stream, not `.stderr`.** The module docstring states the contract is *"the gradient banner on stderr and the Click help text ... on stdout"* ([L7-L9](https://github.com/microsoft/winml-cli/blob/98c9013d334d5370facbeef39985089cf47f0139/tests/e2e/test_help_e2e.py#L7-L9)), but `CliRunner` defaults to `mix_stderr=True`, so `.output` conflates both streams. This assertion passes whether the banner lands on stdout or stderr — a regression that moved the banner to stdout would slip through, and the stdout/stderr split documented at the top of the file isn't actually pinned.

Fix: use `CliRunner(mix_stderr=False)` and assert banner content on `result.stderr` (same pattern as `tests/cli/test_main.py::TestBanner`). Same applies to `test_banner_shows_current_version` (L98) and `TestWinmlHelp::test_banner_present` (L129-L130).

xieofxie (commented):
Code review — 3 issues, posted inline so each can be resolved individually.

tests/e2e/test_help_e2e.py
```diff
@@ -0,0 +1,191 @@
...
+    uv run pytest tests/e2e/ -m e2e -k top_level
+"""
+
+from __future__ import annotations
+
+import pytest
+from click.testing import CliRunner
+
+from winml.modelkit import __version__
+from winml.modelkit.cli import _COMMANDS_DIR, _DISABLED_COMMANDS, main
```
zhenchaoni, line 41: This pulls two underscore-prefixed symbols (`_COMMANDS_DIR`, `_DISABLED_COMMANDS`) from `cli.py` just to mirror its filesystem discovery. Per CLAUDE.md, e2e tests should import the public API and assert behaviour, not snapshot internals. Also: this makes the test slightly tautological — if the disabled set changes in `cli.py`, both sides move in lock-step and a regression isn't caught.

Two options:
1. Promote `COMMANDS_DIR` / `DISABLED_COMMANDS` to public names in `cli.py` (drop the leading underscore). The test is now their second consumer, unlikely to be the last.
2. Compute the expected sets independently here (own `Path(__file__).parents[2] / 'src' / ... / 'commands'` walk + explicit `DISABLED = {'run', 'serve'}`). Diverging values would then surface a real bug.

tests/e2e/test_help_e2e.py
```diff
@@ -0,0 +1,191 @@
...
+# Expected command sets — derived from production code, stay in sync
+# ---------------------------------------------------------------------------
+
+ENABLED_COMMANDS: list[str] = sorted(
+    p.stem
+    for p in _COMMANDS_DIR.glob("*.py")
+    if not p.name.startswith("_") and p.stem not in _DISABLED_COMMANDS
+)
+
+DISABLED_COMMANDS: list[str] = sorted(_DISABLED_COMMANDS)
```
zhenchaoni: Coverage gap: the suite verifies `run` / `serve` are *absent* from `--help`, but not that `winml run` / `winml serve` are actually rejected by `LazyGroup.get_command`'s `ctx.fail(...)` path in `cli.py:178-183`. That's the most likely silent-regression spot: if someone removes a name from `_DISABLED_COMMANDS`, the help auto-updates but a failure to reject the invocation is much harder to spot. Suggest adding:

```python
@pytest.mark.parametrize('cmd', DISABLED_COMMANDS)
def test_disabled_command_invocation_fails(self, cmd: str) -> None:
    result = _invoke(cmd)
    assert result.exit_code != 0
    assert 'disabled' in result.output.lower()
```

tests/e2e/test_help_e2e.py
```diff
@@ -0,0 +1,191 @@
...
+
+
+# ---------------------------------------------------------------------------
+# Helper
+# ---------------------------------------------------------------------------
+
+
+def _invoke(*args: str):
```
zhenchaoni: Minor: missing return type annotation. Other e2e helpers in this repo (e.g. `tests/e2e/test_compile_e2e.py::_invoke`) annotate `-> Result`.

tests/cli/test_help_cli.py
```diff
@@ -0,0 +1,191 @@
...
+    @pytest.mark.parametrize("cmd", ENABLED_COMMANDS)
+    def test_enabled_command_listed(self, cmd: str) -> None:
+        """Every enabled command must appear by name in the Commands section."""
+        assert cmd in _invoke("--help").output
+
+    @pytest.mark.parametrize("cmd", DISABLED_COMMANDS)
+    def test_disabled_command_not_listed(self, cmd: str) -> None:
+        """Disabled commands must not appear in the Commands section."""
+        out = _invoke("--help").output
+        commands_block = out.split("Commands")[-1] if "Commands" in out else ""
```
zhenchaoni, line 155: Brittleness: `commands_block.split()` tokenises on all whitespace, so the assertion only protects against the bare command name appearing in isolation. If a future top-level option's help text contains the word `run` or `serve` (e.g. `"...used by serve mode"`), nothing here changes — but the test stays green even if the disabled command leaks back into the Commands section as a substring of another name. More robust: iterate `commands_block.splitlines()` and check no line, after stripping leading whitespace, starts with `cmd + ' '` (the format `format_commands` produces). Same pattern as `test_enabled_command_has_help_text` already uses on line 158.

tests/cli/test_help_cli.py
```diff
@@ -0,0 +1,191 @@
...
+
+@pytest.mark.e2e
+class TestOptionsSection:
+    """Every documented top-level option appears in ``winml --help``."""
+
+    @pytest.mark.parametrize(
+        "opt",
+        ["--version", "--verbose", "-v", "--quiet", "-q", "--help", "-h"],
+    )
+    def test_option_present(self, opt: str) -> None:
```
zhenchaoni, line 207: Coverage gap: this asserts the *string* `--version` appears in `--help` output, but never invokes `winml --version` to confirm it actually exits 0 and emits the version. Small addition with high value:

```python
def test_version_flag_executes(self) -> None:
    r = _invoke('--version')
    assert r.exit_code == 0
    assert __version__ in r.output
```

tests/cli/test_help_cli.py
```diff
@@ -0,0 +1,191 @@
...
+        """Each command row must show non-empty AST-extracted help text.
+
+        ``LazyGroup.format_commands`` parses each module's docstring via
+        AST without importing it.  A blank column means the docstring is
+        missing or the parse failed.
+        """
+        out = _invoke("--help").output
+        commands_start = out.find("Commands")
+        assert commands_start != -1, "Commands section not found"
+        commands_block = out[commands_start:]
```
zhenchaoni, line 181: Minor: the loop only inspects lines that pass `stripped.startswith((cmd + ' ', cmd + '\t'))`. If `format_commands` ever switches to indented two-column rendering with multi-char separators (e.g. `'export  '` followed by a justified gap), this still works because of the leading `.strip()`. Just worth a one-line note: this depends on Click's `HelpFormatter.write_dl` putting `<cmd><spaces><help>` on a single line — fine today, fragile if Click changes its dl rendering.

zhenchaoni (commented):
Clean and focused — pinning `winml` / `winml --help` is a useful contract test. Ran locally against current `main`: all 47 tests pass. Two coverage gaps on adjacent code paths and a couple of robustness notes inline.

tests/e2e/test_help_e2e.py
DingmaomaoBJTU: Fixed — moved to `tests/cli/test_help_cli.py`. The file no longer exists in `tests/e2e/`.

tests/cli/test_help_cli.py
DingmaomaoBJTU: Fixed — replaced the bare loop with a `for/else` pattern. The `else` clause calls `pytest.fail()` when the command row is never found, so a missing docstring now causes an explicit failure instead of a silent pass.

tests/e2e/test_help_e2e.py
DingmaomaoBJTU: Fixed — assertions now check `result.stderr` instead of `result.output`. The banner is written to `stderr` via `Console(stderr=True)`, so stdout was always empty for it.

tests/e2e/test_help_e2e.py
DingmaomaoBJTU: Fixed — promoted `_COMMANDS_DIR` → `COMMANDS_DIR` and `_DISABLED_COMMANDS` → `DISABLED_COMMANDS` in `cli.py`, with underscore aliases kept for backward compatibility. Tests now import from the public names.

tests/e2e/test_help_e2e.py
DingmaomaoBJTU: Fixed — added `test_disabled_command_invocation_fails` as a parametrized test covering both `run` and `serve`. It verifies `exit_code != 0` and that `"disabled"` appears in the output.

tests/e2e/test_help_e2e.py
DingmaomaoBJTU: Fixed — added `-> Result` return type annotation to `_invoke`.

tests/cli/test_help_cli.py
DingmaomaoBJTU: Fixed — the check now uses `splitlines()` + `strip().startswith((cmd + " ", cmd + "\t"))` instead of `split()`. This handles multi-word command names and indented rows without false-positive word matches.

tests/cli/test_help_cli.py
DingmaomaoBJTU: Fixed — added `test_version_flag_executes` which asserts `exit_code == 0` and that the version string from `__version__` appears in the output.

tests/cli/test_help_cli.py
DingmaomaoBJTU: Acknowledged — thanks for flagging. The mixed-stream behavior is a Click version quirk; `result.stderr` still works correctly here because CliRunner merges streams by default and the banner text is distinct enough to assert on.

src/winml/modelkit/cli.py
```diff
@@ -34,7 +34,8 @@
 
 logger = logging.getLogger(__name__)
 
-_COMMANDS_DIR = Path(__file__).parent / "commands"
+COMMANDS_DIR = Path(__file__).parent / "commands"
+_COMMANDS_DIR = COMMANDS_DIR  # backward-compat alias
```
timenick: These two renames (`_COMMANDS_DIR` → `COMMANDS_DIR` here and `_DISABLED_COMMANDS` → `DISABLED_COMMANDS` below) plus the `# backward-compat alias` lines exist only so the new test file can import them as public symbols. That permanently widens the public API surface and creates two names for the same constant.

`CLAUDE.md` → Import Rules → Test code explicitly carves out this case:

> **Exception**: private `_`-prefixed functions may use internal imports for testing implementation details

These constants are implementation details of `LazyGroup` — please revert both `cli.py` hunks and change the test import to:

```python
from winml.modelkit.cli import _COMMANDS_DIR, _DISABLED_COMMANDS, main
```

(and update the two references inside the test module).

tests/cli/test_help_cli.py
```diff
@@ -0,0 +1,204 @@
+# -------------------------------------------------------------------------
```
timenick, line 1: This file substantially overlaps with the existing `tests/cli/test_main.py`:

| New test | Already covered in `test_main.py` |
|---|---|
| `TestWinmlNoArgs.test_exits_zero` / `test_banner_present` / `test_description_present` | `TestBanner.test_no_args_exits_zero` / `test_no_args_prints_banner_to_stderr`, `TestCLIBasics.test_help` |
| `TestWinmlHelp.test_long_flag_exits_zero` / `test_banner_present` | `TestBanner.test_help_exits_zero` / `test_help_prints_banner_to_stderr` |
| `TestCommandList.test_disabled_command_not_listed` / `test_disabled_command_invocation_fails` (parametrized over `run`, `serve`) | `TestDisabledCommands.test_run_not_in_help` / `test_serve_not_in_help` / `test_run_invocation_rejected` / `test_serve_invocation_rejected` |
| `TestOptionsSection.test_version_flag_executes` | `TestCLIBasics.test_version` |

Consider folding the genuinely new coverage (parametrization over the full enabled command set, `-h`/`--help` parity, no-args/`--help` output equality, AST-extracted help text non-empty, full options-present check) into `test_main.py` (or replace the duplicated `TestBanner`/`TestDisabledCommands` classes there) so we have a single source of truth for CLI-surface contracts and don't double the CI cost on every PR.

tests/e2e/test_help_e2e.py
DingmaomaoBJTU: Addressed — moved to `tests/cli/test_help_cli.py` from the start (the original commit landed it there). The file no longer exists in `tests/e2e/`.

tests/cli/test_help_cli.py
DingmaomaoBJTU: Fixed — added a `for/else` clause: the `else` branch calls `pytest.fail()` when no matching command row is found in the Commands section, turning the silent pass into an explicit failure.

tests/e2e/test_help_e2e.py
DingmaomaoBJTU: Fixed — banner assertions now check `result.stderr`. The module docstring contract (banner → stderr, Click help → stdout) is correctly pinned.

tests/e2e/test_help_e2e.py
DingmaomaoBJTU: Reconsidered and reverted: `COMMANDS_DIR`/`DISABLED_COMMANDS` are back to private `_COMMANDS_DIR`/`_DISABLED_COMMANDS` in `cli.py`. `test_help_cli.py` imports the private names directly, which CLAUDE.md Import Rules explicitly permits for testing implementation details.

tests/e2e/test_help_e2e.py
DingmaomaoBJTU: Added — `test_disabled_command_invocation_fails` is a parametrized test over `_DISABLED_LIST` (currently `["run", "serve"]`). It asserts `exit_code != 0` and `"disabled" in result.output.lower()`.

tests/e2e/test_help_e2e.py
DingmaomaoBJTU: Fixed — `_invoke` is annotated `-> Result` and `Result` is imported from `click.testing` at the top of the file.

tests/cli/test_help_cli.py
DingmaomaoBJTU: Fixed — the disabled-command check now uses `splitlines()` + `strip().startswith((cmd + " ", cmd + "\t"))` instead of `.split()`, so it matches only command rows (not substring occurrences in option help text).

tests/cli/test_help_cli.py
DingmaomaoBJTU: Added — `test_version_flag_executes` invokes `winml --version`, asserts `exit_code == 0`, and checks that `__version__` from the package appears in the output.

tests/cli/test_help_cli.py
DingmaomaoBJTU: Acknowledged — the dependency on Click's `write_dl` single-line rendering is documented in the test's docstring.

src/winml/modelkit/cli.py
DingmaomaoBJTU: Reverted: both public aliases removed from `cli.py`. `test_help_cli.py` now imports `_COMMANDS_DIR` and `_DISABLED_COMMANDS` directly — the CLAUDE.md Import Rules exception for private symbols in test code applies here.

tests/cli/test_help_cli.py
DingmaomaoBJTU: Consolidated: `TestBanner` (4 overlapping tests) and `TestDisabledCommands` (all 4 tests) removed from `test_main.py`. The one unique test from `TestBanner` (`test_subcommand_help_has_no_banner`) is moved into `TestWinmlHelp` in `test_help_cli.py`. Single source of truth for all `winml`/`winml --help` surface contracts; no double CI cost. 71 tests pass across `tests/cli/`.

tests/cli/test_help_cli.py
```diff
@@ -0,0 +1,210 @@
...
+    p.stem
+    for p in _COMMANDS_DIR.glob("*.py")
+    if not p.name.startswith("_") and p.stem not in _DISABLED_COMMANDS
+)
+
+_DISABLED_LIST: list[str] = sorted(_DISABLED_COMMANDS)
+
+
+# ---------------------------------------------------------------------------
+# Helper
```
timenick, line 62: `ENABLED_COMMANDS` is computed at module-import time by globbing `_COMMANDS_DIR`. If `commands/` is ever renamed, moved, or accidentally emptied during a refactor, this evaluates to `[]` — and `pytest` will silently generate **zero** parametrized cases for `test_enabled_command_listed` and `test_enabled_command_has_help_text`. The suite stays green while a critical regression (no enabled commands at all!) goes uncaught.

Add a sanity guard so a zero-command catalog fails loudly instead:

```python
ENABLED_COMMANDS: list[str] = sorted(
    p.stem
    for p in _COMMANDS_DIR.glob("*.py")
    if not p.name.startswith("_") and p.stem not in _DISABLED_COMMANDS
)
assert ENABLED_COMMANDS, (
    f"No enabled commands discovered under {_COMMANDS_DIR}; "
    "refusing to silently turn parametrized tests into zero cases."
)
```

Or if you''d rather not fail at collection: add a `TestCommandList.test_at_least_one_enabled_command` method that asserts `len(ENABLED_COMMANDS) > 0`. Either way the regression becomes visible.

DingmaomaoBJTU: Fixed — added `assert ENABLED_COMMANDS, ...` immediately after the list comprehension. If `commands/` is ever empty or missing, collection fails loudly with a clear message instead of silently generating zero parametrized cases.

timenick (approved):

PR #678, opened 2026-05-20
fix(models): forward explicit --ep to analyzer in HF build path

xieofxie:
## Summary
- `WinMLAutoModel.from_pretrained` was deriving the analyzer EP solely from `config.compile.ep_config.provider`. On compile-less paths (e.g. `--device cpu`, where `config.compile` is `None`), an explicit `--ep cpu` was silently dropped and `build_hf_model` / `analyze_onnx` were called with `ep=None`, causing the analyzer to fall back to all-EP aggregation and emit `analyze_onnx called with ep=None — results will aggregate all EPs.`
- Fix: prefer the user-supplied `ep` (via `kwargs.get("ep")`) and fall back to the compile-derived value only when the user did not pass one.
- Added `tests/unit/models/auto/test_from_pretrained_ep.py` with four monkeypatch-based unit tests covering: the bug's exact scenario, the compile-derived fallback, user-supplied wins over compile-derived, and the both-absent legacy `None` case.

### DIFF
--- a/src/winml/modelkit/models/auto.py
+++ b/src/winml/modelkit/models/auto.py
@@ -410,8 +410,17 @@ class WinMLAutoModel:
         # =====================================================================
         from ..build import build_hf_model
 
-        # Pass resolved EP so the static analyzer targets only this EP
-        resolved_ep = config.compile.ep_config.provider if config.compile is not None else None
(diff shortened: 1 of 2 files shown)


xieofxie:
when testing `perf --ep cpu --device cpu`

timenick (approved):

PR #679, opened 2026-05-20
fix(inspect): distinguish local-path-not-found from network errors

### LINKED ISSUE #542: [winml inspect] [P0] Bogus HF id and missing local file both reported as "Network error"
**Severity**: P0  
**Command**: `winml inspect`

---
**Category**: Error Messages — three distinct conditions need three distinct messages.

**Repro**:
```
uv run winml inspect -m totally-bogus/does-not-exist
uv run winml inspect -m .\does-not-exist.onnx
```

**Actual**:
```
Error: Network error: totally-bogus/does-not-exist is not a local folder and is not a valid model identifier listed on 'https://huggingface.co/models'
If this is a private repository, make sure to pass a token …

Error: Network error: Can't load the configuration of '.\does-not-exist.onnx' …
If you were trying to load it from 'https://huggingface.co/models' …
```

A 404 reported as "Network error". A missing local file reported as "Network error". The "private repository" hint fires unconditionally.

**Expected**: At entry, classify the input:
- Looks like a path or ends in `.onnx` → check existence first; report `Error: Local file '.\does-not-exist.onnx' does not exist.`
- Matches HF id pattern (`org/name`) → catch `RepositoryNotFoundError` and report `Error: Model 'totally-bogus/does-not-exist' not found on Hugging Face Hub.`
- Real network failure → `Error: Network error fetching '...': <DNS / TLS / timeout>.`

**Why it matters**: "Network error" sends users down the wrong debugging path (proxy, VPN, firewall) when the real issue is a typo. P0 because it's the most common failure mode for new users.

DingmaomaoBJTU:
## Summary

Fixes microsoft/winml-cli#542. Two distinct failure modes were both surfacing as `"Network error"`:

- A bogus HF ID (`totally-bogus/does-not-exist`) → `RepositoryNotFoundError` is an `OSError` subclass, so the broad `except OSError → NetworkError` handler caught it
- A missing local file (`.\does-not-exist.onnx`) → `AutoConfig.from_pretrained` raised a plain `OSError` with a confusing HF Hub message including "Network error"

**Fix 1 — local path pre-check in `inspect()`** ([inspect.py](src/winml/modelkit/commands/inspect.py))

Before touching HF Hub, classify the input. If it looks like a local path (absolute, backslash, dot-prefix, or has a file extension — none of which appear in normal `org/model` HF IDs) and does not exist, fail early with:
```
Error: Local path '.\does-not-exist.onnx' does not exist.
```

**Fix 2 — explicit `RepositoryNotFoundError` catch in `_inspect_model_v2()`** ([inspect.py](src/winml/modelkit/commands/inspect.py))

Catch `RepositoryNotFoundError` before the broad `OSError` handler so HF 404s map to `ModelNotFoundError` and surface as:
```
Error: Model not found: Model 'totally-bogus/does-not-exist' not found on Hugging Face Hub.
```

**Tests** ([test_inspect_cli.py](tests/unit/commands/test_inspect_cli.py))
- `test_missing_local_file_shows_path_error` — absolute tmp path without `.onnx`
- `test_missing_local_relative_path_shows_path_error` — `./does-not-exist.onnx`
- `test_bogus_hf_id_shows_model_not_found` — patches `AutoConfig.from_pretrained` to raise `RepositoryNotFoundError`

### DIFF
--- a/src/winml/modelkit/commands/inspect.py
+++ b/src/winml/modelkit/commands/inspect.py
@@ -138,14 +138,25 @@ def inspect(
             "Use --list-tasks to see available tasks."
         )
 
-    # Handle ONNX file input
+    # Classify the input before hitting HF Hub: local paths must exist.
+    # HF IDs (org/model) have no file extension and no path separators other
+    # than a single forward slash, so the heuristics below have no false-positives
+    # for normal HF model identifiers.
     from pathlib import Path
 
-    if model_id and model_id.endswith(".onnx") and Path(model_id).is_file():
-        raise click.ClickException(
-            "ONNX file inspection is not yet supported. "
-            "Use 'winml config -m model.onnx' for ONNX build config."
+    if model_id:
+        _p = Path(model_id)
+        _is_local = (
+            _p.is_absolute() or "\\" in model_id or model_id.startswith(".") or bool(_p.suffix)
         )
+        if _is_local:
+            if _p.suffix == ".onnx" and _p.is_file():
+                raise click.ClickException(
+                    "ONNX file inspection is not yet supported. "
+                    "Use 'winml config -m model.onnx' for ONNX build config."
+                )
+            if not _p.exists():
+                raise click.ClickException(f"Local path '{model_id}' does not exist.")
 
     from ..inspect import InspectError, ModelNotFoundError, NetworkError
     from ..inspect.formatter import output_json, output_table
@@ -246,6 +257,8 @@ def _inspect_model_v2(
     # =========================================================================
     # STEP 2: Shared loader resolution (same call as config command)
     # =========================================================================
+    from huggingface_hub.utils import RepositoryNotFoundError
+
     try:
         loader_config, hf_config, _resolved_class = resolve_loader_config(
             model_id,
@@ -253,6 +266,8 @@ def _inspect_model_v2(
             model_type=model_type_override,
             model_class=model_class_override,
         )
+    except RepositoryNotFoundError as e:
+        raise ModelNotFoundError(f"Model '{model_id}' not found on Hugging Face Hub.") from e
     except ValueError as e:
         err_str = str(e).lower()
         if "not found" in err_str or "404" in err_str:
--- a/src/winml/modelkit/commands/inspect.py
+++ b/src/winml/modelkit/commands/inspect.py
@@ -31,6 +31,14 @@ logger = logging.getLogger(__name__)
 console = Console()
 
 
+def _looks_like_local_path(model_id: str) -> bool:
+    """Return True when model_id is explicitly a local path."""
+    from pathlib import Path
+
+    _p = Path(model_id).expanduser()
+    return _p.exists() or _p.is_absolute() or "\\" in model_id or model_id.startswith((".", "~"))
+
+
 @click.command("inspect")
 @click.option(
     "-m",
@@ -138,17 +146,13 @@ def inspect(
             "Use --list-tasks to see available tasks."
         )
 
-    # Classify the input before hitting HF Hub: local paths must exist.
-    # HF IDs (org/model) have no file extension and no path separators other
-    # than a single forward slash, so the heuristics below have no false-positives
-    # for normal HF model identifiers.
-    from pathlib import Path
-
+    # Classify the input before hitting HF Hub: explicitly-local paths must exist.
+    # Keep this conservative to avoid misclassifying valid HF IDs as local paths.
     if model_id:
-        _p = Path(model_id)
-        _is_local = (
-            _p.is_absolute() or "\\" in model_id or model_id.startswith(".") or bool(_p.suffix)
-        )
+        from pathlib import Path
+
+        _p = Path(model_id).expanduser()
+        _is_local = _looks_like_local_path(model_id)
         if _is_local:
             if _p.suffix == ".onnx" and _p.is_file():
                 raise click.ClickException(
--- a/src/winml/modelkit/commands/inspect.py
+++ b/src/winml/modelkit/commands/inspect.py
@@ -285,14 +285,21 @@ def _inspect_model_v2(
             model_class=model_class_override,
         )
     except RepositoryNotFoundError as e:
-        raise ModelNotFoundError(f"Model '{model_id}' not found on Hugging Face Hub.") from e
+        # Direct HF Hub 404 — keep full message (includes private-repo hint).
+        raise ModelNotFoundError(str(e)) from e
     except ValueError as e:
         err_str = str(e).lower()
         if "not found" in err_str or "404" in err_str:
             raise ModelNotFoundError(str(e)) from e
         raise InspectError(str(e)) from e
     except OSError as e:
-        raise NetworkError(str(e)) from e
+        # transformers wraps RepositoryNotFoundError as a plain OSError with a
+        # recognizable message.  Detect that pattern so users see "Model not found"
+        # (with the original hint text) rather than the misleading "Network error".
+        err_msg = str(e)
+        if "is not a valid model identifier" in err_msg or "is not a local folder" in err_msg:
+            raise ModelNotFoundError(err_msg) from e
+        raise NetworkError(err_msg) from e
 
     if parent_hf_config is None:
         parent_hf_config = hf_config
--- a/src/winml/modelkit/commands/inspect.py
+++ b/src/winml/modelkit/commands/inspect.py
@@ -30,13 +30,28 @@ from rich.console import Console
 logger = logging.getLogger(__name__)
 console = Console()
 
+# File extensions that unambiguously indicate a local file path.
+# HF model IDs routinely contain dots in version numbers (Phi-3.5, Qwen2.5, …)
+# so matching on any suffix would cause false-positives; restrict to known extensions.
+_LOCAL_FILE_EXTS = frozenset({".onnx", ".pt", ".pth", ".safetensors", ".bin"})
+
 
 def _looks_like_local_path(model_id: str) -> bool:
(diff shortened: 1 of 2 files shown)


src/winml/modelkit/commands/inspect.py
```diff
@@ -138,14 +138,25 @@ def inspect(
...
+    # for normal HF model identifiers.
     from pathlib import Path
 
-    if model_id and model_id.endswith(".onnx") and Path(model_id).is_file():
-        raise click.ClickException(
-            "ONNX file inspection is not yet supported. "
-            "Use 'winml config -m model.onnx' for ONNX build config."
+    if model_id:
+        _p = Path(model_id)
+        _is_local = (
```
xieofxie: @copilot is the right way to test if input to huggingface is local folder or hf id?

timenick: **Blocking regression.** `bool(_p.suffix)` here misclassifies many real HF model IDs as local paths, because `Path.suffix` returns *everything after the first dot* in the basename. Any HF ID with a version number lands in the broken bucket:

| `model_id` | `Path(model_id).suffix` | classified as local? |
|---|---|---|
| `microsoft/Phi-3.5-mini-instruct` | `''.5-mini-instruct''` | ❌ **yes** — regression |
| `Qwen/Qwen2.5-7B-Instruct` | `''.5-7B-Instruct''` | ❌ **yes** — regression |
| `meta-llama/Llama-3.1-8B` | `''.1-8B''` | ❌ **yes** — regression |
| `google/flan-t5-base-v1.1` | `''.1''` | ❌ **yes** — regression |
| `microsoft/resnet-50` | `''''` | ✅ ok |
| `bert-base-uncased` | `''''` | ✅ ok |

All four regressing cases will now exit with `Local path ''X'' does not exist` *before any Hub lookup is attempted* — exactly the user-facing failure mode from #542, just relocated to a different (and very common) class of model IDs.

### Suggested fix

Restrict the file-extension check to a known set, and tighten the leading-dot heuristic to `./` / `../` so single-dot prefixes can''t false-match:

```python
_KNOWN_LOCAL_EXTS = {".onnx", ".pt", ".pth", ".safetensors", ".bin"}

_p = Path(model_id)
_is_local = (
    _p.is_absolute()
    or "\\" in model_id
    or model_id.startswith(("./", "../"))
    or _p.suffix.lower() in _KNOWN_LOCAL_EXTS
)
```

That keeps the original target cases (`.\does-not-exist.onnx`, `C:\…`, `/abs/path`) classified as local while letting HF IDs with version dots flow to the Hub path.

tests/unit/commands/test_inspect_cli.py
```diff
@@ -263,3 +263,39 @@ def test_inspect_error(self, runner: CliRunner) -> None:
             result = runner.invoke(inspect, ["-m", "test"], obj={})
             assert result.exit_code != 0
             assert "Inspection error" in result.output
+
+    def test_missing_local_file_shows_path_error(
+        self, runner: CliRunner, tmp_path: pytest.TempPathFactory
```
timenick: Type annotation mismatch: `tmp_path: pytest.TempPathFactory` should be `tmp_path: Path`. The `pytest.TempPathFactory` type is for the **`tmp_path_factory`** fixture (a separate fixture that returns a factory). `tmp_path` itself is just `pathlib.Path`. Replace with:

```python
def test_missing_local_file_shows_path_error(
    self, runner: CliRunner, tmp_path: Path
) -> None:
```

(and add `from pathlib import Path` at the top of the file if it isn''t already imported).

tests/unit/commands/test_inspect_cli.py
```diff
@@ -263,3 +263,39 @@ def test_inspect_error(self, runner: CliRunner) -> None:
...
+    ) -> None:
+        """Absolute path to a missing file shows a clear local error, not 'Network error'."""
+        from winml.modelkit.commands.inspect import inspect
+
+        missing = str(tmp_path / "missing.onnx")
```
timenick, line 276: Small inconsistency with the PR description: the body says *"absolute tmp path without `.onnx`"*, but the test uses `tmp_path / "missing.onnx"`. The test still passes because `_p.is_file()` returns False on a missing path so it falls through to the existence check — but the description doesn''t match what the test exercises. Either drop the `.onnx` suffix here to match the description, or update the PR body to say *"absolute tmp path with a non-existent `.onnx` file"*.

tests/unit/commands/test_inspect_cli.py
```diff
@@ -263,3 +263,39 @@ def test_inspect_error(self, runner: CliRunner) -> None:
...
+        from winml.modelkit.commands.inspect import inspect
+
+        with patch(
+            "transformers.AutoConfig.from_pretrained",
+            side_effect=RepositoryNotFoundError("totally-bogus/does-not-exist"),
+        ):
+            result = runner.invoke(inspect, ["-m", "totally-bogus/does-not-exist"], obj={})
+            assert result.exit_code != 0
+            assert "Model not found" in result.output
+            assert "Network error" not in result.output
```
timenick, line 328: Worth adding one more positive-path test that would have caught the dotted-HF-ID regression flagged on `inspect.py:149`:

```python
def test_dotted_hf_id_attempts_hub_lookup(self, runner: CliRunner) -> None:
    """HF IDs with version dots (e.g. Phi-3.5) must reach the Hub path, not be classified as local."""
    from winml.modelkit.commands.inspect import inspect

    with patch(
        "transformers.AutoConfig.from_pretrained",
        side_effect=RuntimeError("intentional — proves we reached the Hub path"),
    ):
        result = runner.invoke(inspect, ["-m", "microsoft/Phi-3.5-mini-instruct"], obj={})
    # The exact error doesn''t matter — what matters is that we got past the local-path branch.
    assert "does not exist" not in result.output, (
        "dotted HF ID was misclassified as a local path"
    )
```

This pins the contract that dotted IDs must flow to the Hub branch, so any future heuristic tweak that re-breaks them fails CI immediately.

timenick (changes requested):
Found a blocking regression in the local-path heuristic — `bool(_p.suffix)` will misclassify popular HF model IDs with dotted version numbers (Phi-3.5, Qwen2.5, Llama-3.1, …) as local paths and fail before any Hub lookup. See inline. Three smaller nits also flagged.

src/winml/modelkit/commands/inspect.py
DingmaomaoBJTU: Good question.  already handles both paths and HF IDs internally (it checks  first), but the problem is that on failure it always raises  regardless of root cause — a 404 and a missing local file are indistinguishable at that layer. The pre-check in  runs *before* we call into HF so we can give a precise error. The heuristic uses only unambiguous indicators: absolute path, path separators, `./`/`../`/`~/` prefixes, and a known-extension allowlist (`.onnx`, `.pt`, `.pth`, `.safetensors`, `.bin`). Version-dotted IDs like `Phi-3.5` or `Qwen2.5` match none of these and flow through to the Hub correctly.

DingmaomaoBJTU: Fixed — replaced `bool(_p.suffix)` with a `_LOCAL_FILE_EXTS` allowlist (`{.onnx, .pt, .pth, .safetensors, .bin}`) and moved it into the `_looks_like_local_path` helper at module level. Also tightened the dot-prefix check to `startswith(("./", "../", "~/"))` so bare version-dotted names aren't caught. `microsoft/Phi-3.5-mini-instruct`, `Qwen/Qwen2.5-7B-Instruct`, `meta-llama/Llama-3.1-8B`, and `google/flan-t5-base-v1.1` all flow through to the Hub path. Covered by the new `test_dotted_hf_id_reaches_hub_path` test.

tests/unit/commands/test_inspect_cli.py
DingmaomaoBJTU: Fixed — changed to `tmp_path: Path` and moved the import into a `TYPE_CHECKING` block (also required by ruff TC003).

tests/unit/commands/test_inspect_cli.py
DingmaomaoBJTU: Kept the `.onnx` suffix — the test intentionally exercises the extension-based detection path in `_looks_like_local_path`. Updated the docstring to say "absolute tmp path with a non-existent .onnx file" to match.

tests/unit/commands/test_inspect_cli.py
DingmaomaoBJTU: Added `test_dotted_hf_id_reaches_hub_path` — patches `AutoConfig.from_pretrained` to raise `RuntimeError` and asserts the output does not contain "does not exist" for `microsoft/Phi-3.5-mini-instruct`. This pins the contract that dotted HF IDs always reach the Hub branch.

timenick (approved):
LGTM. All four prior review comments cleanly addressed:

- `bool(_p.suffix)` regression → replaced with `_LOCAL_FILE_EXTS` allowlist via the new `_looks_like_local_path()` helper (now correctly handles Phi-3.5 / Qwen2.5 / Llama-3.1 / flan-t5-v1.1)
- Test type annotation fixed via `TYPE_CHECKING` import of `Path`
- Test docstring now matches the actual `missing.onnx` test code
- `test_dotted_hf_id_reaches_hub_path` added as the suggested regression-pin

Bonus fix in `e0f4941`: caught that `transformers.AutoConfig.from_pretrained` re-wraps `RepositoryNotFoundError` as plain `OSError`, so the original `except RepositoryNotFoundError` handler would never have fired on the common path. The added message-pattern match plus `test_bogus_hf_id_shows_model_not_found` (which uses the real transformers error text and preserves the private-repo hint) closes this gap properly.

PR #680, opened 2026-05-20
Fix quantization P0 bugs

### LINKED ISSUE #433: Bug: Quantize stage in build shows train output
## Description

When running the `build` command, the quantize stage displays train output instead of the expected quantize output.

Image

## Steps to Reproduce
Clear dataset cache.
Run `modelkit build` with a model that includes a quantize stage and observe the console output during the quantize step.

## Expected Behavior

The quantize stage should display quantize-relevant output.

## Actual Behavior

Train output is shown during the quantize stage.

## Source

Reported via user feedback from Panther Lake testing.

### LINKED ISSUE #555: [winml quantize] [P0] `--precision banana` silently falls back to defaults, prints "Success!", exits 1
**Severity**: P0  
**Command**: `winml quantize`

---
**Category**: Input Validation — reject unknown values at parse time; do not contradict the exit code.

**Repro**:
```
uv run winml quantize -m temp\cli-audit\resnet.onnx --precision banana -o temp\ux\bad.onnx
```

**Actual**:
```
Precision: banana          ← lying — actually used uint8/uint8
Weight type: uint8
Activation type: uint8
Success! Model quantized
QDQ nodes inserted: 256
```
Followed by two near-duplicate ORT WARNING lines (~7 s apart). Exit code **1** despite the `Success!` line in stdout.

**Expected**: Declare `--precision` as `click.Choice([auto, fp32, fp16, int8, int16, w4a16, w8a8, w8a16])`. Fail at parse time:
```
Error: Invalid value for --precision: 'banana' is not one of [auto, fp32, fp16, int8, int16, w4a16, w8a8, w8a16].
```
Exit 2.

**Why it matters**: Three things wrong at once: (a) the flag value is silently ignored; (b) stdout says "Success!"; (c) exit code is 1. Any one of those is a P1; together they form a P0.

zhenchaoni:
## PR Title

`fix(quantize/eval): validate --precision, stream default image dataset, support pixel_mask padding`

### Summary

Three independent fixes bundled together — two are user-reported bugs, one is a side-fix found while validating object detection.

### Changes

**1. Reject invalid `--precision` values (fixes [#555](https://github.com/microsoft/winml-cli/issues/555))**

Previously, `winml quantize --precision banana` (or any unrecognized string) silently fell through to `uint8/uint8` defaults and reported "Success!", masking user typos like `w4a16` or `fp64`. `_resolve_quant_types` now raises `click.BadParameter` for any precision that is not `None`, `"auto"` (case-insensitive), or a value accepted by `is_quantized_precision`. The error message lists the accepted forms.

**2. Stream the default image calibration dataset (fixes [#433](https://github.com/microsoft/winml-cli/issues/433))**

The default calibration dataset (`timm/mini-imagenet`) was downloading the entire training split (~5 GB) just to pull 10 calibration samples. `ImageDataset._get_default_dataset` now sets `streaming=True` by default, and the shared `_load_and_sample` helper takes the streaming path when `streaming` is enabled **and** `max_samples` is set (streaming without a sample cap would be strictly worse than a bulk download, so it falls back automatically). `ObjectDetectionDataset` reuses the same helper, eliminating duplicated load/sample logic.

Verified: with the HF cache cleared, running `winml quantize -m resnet-50 --task image-classification --model-name microsoft/resnet-50 --samples 10` leaves `datasets--timm--mini-imagenet` at 0 MB on disk.

**3. Support `pixel_mask` input for DETR-family object detection**

Some object detection exports (DETR, etc.) declare a `pixel_mask` input alongside `pixel_values`, which the HF image processor only populates meaningfully when padding to a fixed canvas. `WinMLObjectDetectionEvaluator.prepare_pipeline` now detects this case and switches the processor to aspect-preserving resize (`shortest_edge`/`longest_edge`) + `pad_size` + `do_pad=True`. Models without `pixel_mask` keep the previous exact-resize, no-padding behavior.

### Tests

- `TestQuantizePrecisionValidation` — parametrized over `banana`, `w4a16`, `int4`, `fp64`; asserts `quantize_onnx` is never invoked.
- `TestPreparePipeline::test_pixel_mask_path_enables_padding` and `test_no_pixel_mask_path_disables_padding` — cover both branches with a sentinel to prove the wrong branch did not run. Uses `monkeypatch` for the parent-class `prepare_pipeline` stub so the override is reverted on teardown.

All 84 affected unit tests pass.

### Manual verification

| Issue | Verification |
|---|---|
| #555 | `winml quantize -p banana` → `BadParameter: 'banana' is not a supported quantization precision. Expected: auto \| int8 \| int16 \| w{x}a{y} ...` |
| #433 | Cleared HF cache → ran quantize → `datasets--timm--mini-imagenet` = 0.0 MB on disk; only 10 examples flowed through `Map`/`Aligning the labels` |
| pixel_mask | Covered by new unit tests |

### DIFF
--- a/src/winml/modelkit/commands/quantize.py
+++ b/src/winml/modelkit/commands/quantize.py
@@ -192,8 +192,6 @@ def quantize(
     # Show info
     console.print(f"[bold blue]Input:[/bold blue] {model}")
     console.print(f"[bold blue]Output:[/bold blue] {output}")
-    if precision:
-        console.print(f"[bold blue]Precision:[/bold blue] {precision}")
     console.print(f"[bold blue]Weight type:[/bold blue] {resolved_weight}")
     console.print(f"[bold blue]Activation type:[/bold blue] {resolved_activation}")
     console.print(f"[bold blue]Samples:[/bold blue] {samples}")
@@ -260,8 +258,14 @@ def _resolve_quant_types(
 
     if precision and is_quantized_precision(precision):
         default_w, default_a = resolve_quant_types(precision)
-    else:
+    elif precision is None or precision.lower() == "auto":
         default_w, default_a = "uint8", "uint8"
+    else:
+        raise click.BadParameter(
+            f"'{precision}' is not a supported quantization precision. "
+            "See `winml quantize --help` for details.",
+            param_hint="'-p' / '--precision'",
+        )
 
     # Explicit flags override precision defaults
     resolved_w = weight_type if weight_type else default_w
--- a/src/winml/modelkit/commands/quantize.py
+++ b/src/winml/modelkit/commands/quantize.py
@@ -48,7 +48,8 @@ console = Console()
     "-p",
     type=str,
     default=None,
-    help="Quantization precision: int8, int16, or w{x}a{y} (e.g., w8a16). "
+    help="Quantization precision. Accepted: auto, int8, int16, or w{x}a{y} "
+    "where x,y in {8,16} (e.g., w8a8, w8a16, w16a16). "
     "Overridden by explicit --weight-type/--activation-type.",
 )
 @click.option(
@@ -192,6 +193,7 @@ def quantize(
     # Show info
     console.print(f"[bold blue]Input:[/bold blue] {model}")
     console.print(f"[bold blue]Output:[/bold blue] {output}")
+    console.print(f"[bold blue]Precision:[/bold blue] {precision or 'auto'}")
     console.print(f"[bold blue]Weight type:[/bold blue] {resolved_weight}")
     console.print(f"[bold blue]Activation type:[/bold blue] {resolved_activation}")
     console.print(f"[bold blue]Samples:[/bold blue] {samples}")
@@ -263,7 +265,8 @@ def _resolve_quant_types(
     else:
         raise click.BadParameter(
             f"'{precision}' is not a supported quantization precision. "
-            "See `winml quantize --help` for details.",
+            "Accepted: auto, int8, int16, or w{x}a{y} with x,y in {8,16} "
+            "(e.g., w8a8, w8a16, w16a16).",
             param_hint="'-p' / '--precision'",
         )
 
--- a/src/winml/modelkit/datasets/image.py
+++ b/src/winml/modelkit/datasets/image.py
@@ -50,6 +50,52 @@ class ImageDataset(BaseTaskDataset):
         if self._dataset_name is None:
             self._dataset_name = "timm/mini-imagenet"
             self._data_split = "train"
+            self._config.setdefault("streaming", True)
+
+    def _load_and_sample(self) -> Any:
+        """Load the configured dataset and apply sample/shuffle.
+
+        Shared by ImageDataset and ObjectDetectionDataset. Column detection
+        is *not* done here — callers run their own detection on the returned
+        dataset because the column schema differs by task.
+
+        Returns:
+            A materialized arrow Dataset of up to ``self._max_samples`` rows.
+        """
+        # Streaming only helps when capped by max_samples; otherwise we'd
+        # iterate the full remote stream into memory, which is worse than a
+        # bulk download.
+        streaming = self._config.get("streaming", False) and self._max_samples is not None
+        logger.info(f"Loading dataset: {self._dataset_name} with split: {self._data_split}")
+        try:
+            dataset = load_dataset(self._dataset_name, split=self._data_split, streaming=streaming)
+        except Exception as e:
+            logger.error(f"Failed to load dataset {self._dataset_name}: {e}")
+            raise
+
+        shuffle = self._config.get("shuffle", False)
+        seed = self._config.get("seed", 42)
+
+        if streaming:
+            # Streaming datasets aren't indexable: shuffle reservoir-samples
+            # within a buffer; take() pulls only the slice we need.
+            if shuffle:
+                dataset = dataset.shuffle(seed=seed, buffer_size=1000)
+            dataset = dataset.take(self._max_samples)
+            from datasets import Dataset as ArrowDataset
+            dataset = ArrowDataset.from_list(list(dataset), features=dataset.features)
+        elif self._max_samples is not None:
+            max_samples = min(self._max_samples, len(dataset))
+            indices = (
+                Random(seed).sample(range(len(dataset)), max_samples)
+                if shuffle
+                else list(range(max_samples))
+            )
+            dataset = dataset.select(indices)
+        elif shuffle:
+            dataset = dataset.shuffle(seed=seed)
+
+        return dataset
 
     def _initialize(self) -> None:
         """Initialize the image classification dataset.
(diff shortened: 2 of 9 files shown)


src/winml/modelkit/commands/quantize.py
```diff
@@ -260,8 +258,14 @@ def _resolve_quant_types(
 
     if precision and is_quantized_precision(precision):
         default_w, default_a = resolve_quant_types(precision)
-    else:
+    elif precision is None or precision.lower() == "auto":
```
timenick, line 263: **Error message redirects to `--help`, but `--help` is incomplete.**

The error says *"See `winml quantize --help` for details."* but the existing help string (line 51, unchanged in this PR) is just:

```
Quantization precision: int8, int16, or w{x}a{y} (e.g., w8a16).
Overridden by explicit --weight-type/--activation-type.
```

No mention of `auto` (which the new validation explicitly accepts), and no enumeration of which `w{x}a{y}` combinations qualify (`w4a16` is rejected per the new test, but a user only finds out by trying). The PR description promises *"The error message lists the accepted forms"* — please deliver one of:

1. **Inline the accepted values** in the error message:
   ```python
   raise click.BadParameter(
       f"''{precision}'' is not a supported quantization precision. "
       f"Expected: auto | int8 | int16 | {{w8a8, w8a16, ...}}",
       param_hint="''-p'' / ''--precision''",
   )
   ```
2. **Or expand the help text** on line 51 to include `auto` and a representative list of supported `w{x}a{y}` widths.

Otherwise users follow the breadcrumb and dead-end.

src/winml/modelkit/commands/quantize.py
```diff
@@ -192,8 +192,6 @@ def quantize(
     # Show info
```
timenick, line 193: This removes the `Precision:` console echo that previously printed `Precision: auto` (or whatever the user passed) right alongside `Input`/`Output`/`Weight type`/`Activation type`.

The removal isn''t mentioned in the PR description (which lists three fixes — precision validation, streaming, pixel_mask) and is a minor UX regression: a user running `winml quantize --precision auto` now has to mentally back-translate `Weight type: uint8 / Activation type: uint8` into "yes, that''s what `auto` resolved to".

Either restore the echo or call out the removal explicitly in the PR description and commit message.

timenick (commented):
Five findings — none blocking, but worth addressing before merge. The biggest one is that the new error message redirects users to `--help`, but `--help` doesn''t yet mention `auto` as an accepted value, so users hitting the BadParameter still won''t know how to fix it.

src/winml/modelkit/datasets/image.py
```diff
@@ -50,36 +50,41 @@ def _get_default_dataset(self) -> None:
         if self._dataset_name is None:
             self._dataset_name = "timm/mini-imagenet"
             self._data_split = "train"
+            self._config.setdefault("streaming", True)
```
timenick, line 53: The PR''s headline win (5 GB → 0 MB on disk for the default dataset) is only covered by *manual* verification in the PR description. A small regression test would be cheap insurance:

```python
class TestDefaultDatasetStreaming:
    def test_default_dataset_enables_streaming(self):
        ds = ImageDataset(model_name="x", max_samples=10)
        ds._get_default_dataset()
        assert ds._config.get("streaming") is True

    def test_custom_dataset_does_not_force_streaming(self):
        ds = ImageDataset(model_name="x", dataset_name="cifar10", max_samples=10)
        ds._get_default_dataset()  # no-op when dataset_name set
        assert ds._config.get("streaming") in (None, False)

    def test_streaming_without_max_samples_degrades_to_bulk(self, monkeypatch):
        ds = ImageDataset(model_name="x", dataset_name="cifar10")
        ds._config["streaming"] = True
        called = {}
        def fake_load(name, split, streaming):
            called["streaming"] = streaming
            return _FakeDataset()  # whatever your test fixture is
        monkeypatch.setattr("winml.modelkit.datasets.image.load_dataset", fake_load)
        ds._load_and_sample()
        assert called["streaming"] is False  # silent fallback
```

The last test in particular pins the documented "streaming + no `max_samples` → bulk download" fallback, which is currently undocumented except for the inline code comment.

tests/unit/eval/test_object_detection_evaluator.py
```diff
@@ -195,3 +195,72 @@ def test_user_mapping_overrides_model(self):
...
+
+def _patch_super_prepare_pipeline(monkeypatch, fake_pipe) -> None:
+    """Patch the *parent* prepare_pipeline so super() in the subclass returns fake_pipe.
+
+    Uses monkeypatch so the override is reverted on teardown — directly assigning
+    to a parent class attribute leaks across tests in the same session.
+    """
+    parent = WinMLObjectDetectionEvaluator.__mro__[1]
+    monkeypatch.setattr(parent, "prepare_pipeline", lambda self: fake_pipe)
+
```
timenick, line 228: `__mro__[1]` is fragile. If a mixin is ever inserted between `WinMLObjectDetectionEvaluator` and `WinMLEvaluator` (e.g., a logging mixin, a telemetry mixin), this patches the mixin instead of `WinMLEvaluator` and the test silently passes for the wrong reason — `super().prepare_pipeline()` would no longer route through the mock.

Be explicit instead:

```python
from winml.modelkit.eval import WinMLEvaluator  # or the actual parent module path

def _patch_super_prepare_pipeline(monkeypatch, fake_pipe) -> None:
    monkeypatch.setattr(WinMLEvaluator, "prepare_pipeline", lambda self: fake_pipe)
```

If the parent class ever moves, the import fails loudly at collection time rather than producing a silently-wrong test.

tests/unit/eval/test_object_detection_evaluator.py
```diff
@@ -195,3 +195,72 @@ def test_user_mapping_overrides_model(self):
...
+        pipe = ev.prepare_pipeline()
+        proc = pipe.image_processor
+        assert proc.size == {"shortest_edge": 800, "longest_edge": 800}
+        assert proc.pad_size == {"height": 800, "width": 800}
+        assert proc.do_pad is True
+
+    def test_no_pixel_mask_path_disables_padding(self, monkeypatch) -> None:
+        """No pixel_mask input: exact resize, no padding (legacy behavior)."""
+        ev = make_evaluator()
+        ev.model.io_config = {
```
timenick, line 260: Both `pixel_mask` tests use `h == w == 800`, which makes `min(h, w)` and `max(h, w)` collapse to the same value. The new code path:

```python
pipe.image_processor.size = {
    "shortest_edge": min(h, w),
    "longest_edge": max(h, w),
}
pipe.image_processor.pad_size = {"height": h, "width": w}
```

is really only exercised when h != w — the canonical DETR setup is 800×1333. Suggest parametrizing:

```python
@pytest.mark.parametrize(
    ("h", "w", "shortest", "longest"),
    [
        (800, 800, 800, 800),
        (800, 1333, 800, 1333),
        (1333, 800, 800, 1333),
    ],
)
def test_pixel_mask_path_enables_padding(self, monkeypatch, h, w, shortest, longest) -> None:
    ...
    assert proc.size == {"shortest_edge": shortest, "longest_edge": longest}
    assert proc.pad_size == {"height": h, "width": w}
```

That catches an accidental `min`/`max` swap or `height`/`width` swap in `pad_size`, both of which the current test would pass through.

src/winml/modelkit/datasets/image.py
```diff
@@ -50,36 +50,41 @@ def _get_default_dataset(self) -> None:
...
-        # 5. Efficient sampling and processing pipeline
         shuffle = self._config.get("shuffle", False)
         seed = self._config.get("seed", 42)
 
-        if self._max_samples is not None:
+        if streaming:
+            # Streaming datasets aren't indexable: shuffle reservoir-samples
+            # within a buffer; take() pulls only the slice we need.
+            if shuffle:
+                dataset = dataset.shuffle(seed=seed, buffer_size=1000)
```
vortex-captain, line 84: nit: would `buffer_size=min(1000, self._max_samples)` be slightly better?

vortex-captain (approved):

src/winml/modelkit/commands/quantize.py
zhenchaoni: Done both. Expanded the --precision help text to enumerate auto | int8 | int16 | w{x}a{y} and inlined the accepted forms in the BadParameter message so users don't need to follow the breadcrumb.

src/winml/modelkit/commands/quantize.py
zhenchaoni: Restored. The Precision: line is back in the banner alongside Input / Output / Weight type / Activation type.

src/winml/modelkit/datasets/image.py
zhenchaoni: Added [test_image_streaming.py](vscode-file://vscode-app/c:/Users/zhenni/AppData/Local/Programs/Microsoft%20VS%20Code/f6cfa2ea24/resources/app/out/vs/code/electron-browser/workbench/workbench.html) covering: default dataset enables streaming=True, custom dataset_name does not force streaming, streaming=True + no max_samples falls back to bulk, and the buffer_size=1000 reservoir contract. 6 cases total, all passing.

tests/unit/eval/test_object_detection_evaluator.py
zhenchaoni: Fixed. _patch_super_prepare_pipeline() now imports WinMLEvaluator explicitly and patches that class directly — any future parent-class move surfaces at import time.

tests/unit/eval/test_object_detection_evaluator.py
zhenchaoni: Parametrized test_pixel_mask_path_enables_padding over (800,800), (800,1333), (1333,800). Now catches accidental min/max or height/width swaps.

src/winml/modelkit/datasets/image.py
zhenchaoni: Kept at 1000 intentionally. The buffer is the shuffle reservoir window for class diversity on class-ordered streams like timm/mini-imagenet; lowering it to max_samples would degrade class coverage when max_samples is small (e.g. 100 samples → only 100 consecutive same-class items). Added a clarifying comment.

timenick (approved):

PR #681, opened 2026-05-20
feat(export): normalize exported ONNX in-place via optimize_onnx

### LINKED ISSUE #677: winml analyze in v0.0.3 reports all ops as Unknown (regression from v0.0.2 - see screencaps)
## Summary

`winml analyze` against an exported HuggingFace model in **v0.0.3** classifies almost every operator as **Unknown** and emits a `derive_properties: 'NoneType' object is not iterable` warning at the top of the report. The same model + command on **v0.0.2** correctly classifies ops as Supported / Partial / Unsupported.

## Repro

Same model, same command, two versions of the CLI installed side-by-side in separate venvs.

```powershell
# 1. Inspect + export the model
winml inspect -m cardiffnlp/twitter-roberta-base-sentiment-latest
winml export  -m cardiffnlp/twitter-roberta-base-sentiment-latest -o sentiment/model.onnx

# 2. Analyze for OpenVINO NPU
winml analyze -m sentiment/model.onnx --ep openvino --device NPU
```

## Expected (v0.0.2)

Operators are bucketed into Supported / Partial / Unsupported with the colored bars and an Analysis Summary listing concrete unsupported ops.

`TOTAL  251/322/111/2`  (S/P/U/Unknown)

_Screenshot: v0.0.2 analyze output_ 

## Actual (v0.0.3)

Top of the report shows:

> Incomplete model information for derive_properties: 'NoneType' object is not iterable

Every op larger than a Constant/Cast/Sqrt is classified as **Unknown** (435 of them). The Analysis Summary's Unsupported and Partial buckets are empty; everything lands in `? Unknown:`.

`TOTAL  251/0/0/435`  (S/P/U/Unknown)

_Screenshot: v0.0.3 analyze output_ 

## Environment

- WinML CLI: **0.0.3** (regression) vs **0.0.2** (working)
- Wheel names confirm package rename: `winml_modelkit-0.0.2-py3-none-any.whl` → `winml_cli-0.0.3-py3-none-any.whl`
- OS: Windows 11 Enterprise 24H2 (build 26200), x64
- Python: 3.10.20 (uv-managed venv)
- Hardware: Intel Core Ultra 7 268V (Lunar Lake), Intel AI Boost NPU
- EP under test: OpenVINOExecutionProvider, NPU
- Model: `cardiffnlp/twitter-roberta-base-sentiment-latest` (RoBERTa, text-classification)

Image
Image

vortex-captain:
Fixes #677.

## Summary
- Adds a post-export normalization step inside `export_pytorch()`, the single function used by every export path (the `winml export` CLI, `build_hf_model`, the `winml build` command, and direct Python API callers), so normalization runs automatically for all of them.
- New helper `_normalize_exported_model(output_path) -> bool` in `src/winml/modelkit/export/pytorch.py`:
  - logs `Normalizing model`
  - runs `optimize_onnx()` into a fresh `tempfile.mkdtemp()`
  - on success, replaces the original `.onnx` (and `.data` sidecar, if any) via `copy_onnx_model()` — confirmed to overwrite both files correctly by new tests in `tests/unit/onnx/test_external_data.py`
  - on failure, logs a warning and leaves the original export untouched
  - cleans up the temp directory in a `finally` clause either way
  - records the outcome in `stats["model_normalization_succeeded"]`

## Tests
- `tests/unit/export/test_pytorch_export.py` — success path populates `value_info`; mocked-failure path leaves the model un-shape-inferenced and reports `model_normalization_succeeded=False`.
- `tests/unit/onnx/test_external_data.py` — two new overwrite tests covering `copy_onnx_model` against a pre-existing dst with and without external data sidecars.

### DIFF
--- a/src/winml/modelkit/export/pytorch.py
+++ b/src/winml/modelkit/export/pytorch.py
@@ -138,7 +138,10 @@ def _normalize_exported_model(output_path: Path) -> bool:
     from ..optim import optimize_onnx
 
     logger.info("Normalizing model")
-    tmp_dir = Path(tempfile.mkdtemp())
+    # Place the temp dir next to the output so copy_onnx_model stays on the
+    # same volume — avoids a cross-volume data transfer for multi-GB models
+    # and keeps the system drive's %TEMP% free of large sidecars.
+    tmp_dir = Path(tempfile.mkdtemp(dir=output_path.parent))
     tmp_path = tmp_dir / output_path.name
 
     try:
--- a/src/winml/modelkit/export/pytorch.py
+++ b/src/winml/modelkit/export/pytorch.py
@@ -102,12 +102,12 @@ def export_pytorch(
             **kwargs,
         )
 
-    if normalize:
-        stats["model_normalization_status"] = (
-            "succeeded" if _normalize_exported_model(output_path) else "failed"
-        )
-    else:
-        stats["model_normalization_status"] = "not_run"
+        if normalize:
+            stats["model_normalization_status"] = (
+                "succeeded" if _normalize_exported_model(output_path) else "failed"
+            )
+        else:
+            stats["model_normalization_status"] = "not_run"
 
     return stats
 
--- a/src/winml/modelkit/export/pytorch.py
+++ b/src/winml/modelkit/export/pytorch.py
@@ -47,6 +47,7 @@ def export_pytorch(
     task: str | None = None,
     verbose: bool = False,
     enable_reporting: bool = False,
+    normalize: bool = True,
     **kwargs: Any,
 ) -> dict[str, Any]:
     """Export a PyTorch nn.Module to ONNX.
@@ -63,9 +64,15 @@ def export_pytorch(
         task: Task for auto-input generation fallback.
         verbose: Enable verbose logging.
         enable_reporting: Generate export report file.
+        normalize: If True (default), run optimize_onnx on the exported model
+            to apply graph-level optimizations and shape inference. Set False
+            to keep the raw torch.onnx.export output (useful when debugging
+            the exporter or running custom downstream optimization).
 
     Returns:
-        Export statistics dict from HTPExporter.
+        Export statistics dict from HTPExporter, with an extra
+        `model_normalization_status` entry: one of `"not_run"` (when
+        `normalize=False`), `"succeeded"`, or `"failed"`.
     """
     from .htp.exporter import HTPExporter
 
@@ -95,7 +102,12 @@ def export_pytorch(
             **kwargs,
         )
 
-    stats["model_normalization_succeeded"] = _normalize_exported_model(output_path)
+    if normalize:
+        stats["model_normalization_status"] = (
+            "succeeded" if _normalize_exported_model(output_path) else "failed"
+        )
+    else:
+        stats["model_normalization_status"] = "not_run"
 
     return stats
 
@@ -105,13 +117,19 @@ def _normalize_exported_model(output_path: Path) -> bool:
 
     Writes the normalized model into a temporary directory, then replaces
     the original export (and its `.data` sidecar, if any) via
-    copy_onnx_model — which overwrites both the `.onnx` file and its
-    external-data sidecar by design. On any failure during optimization
-    or copy, logs a warning and leaves the original export untouched.
-    The temp directory is removed in either case.
+    copy_onnx_model. The temp directory is removed either way.
+
+    Failure modes are not symmetric:
+    - An optimize_onnx failure leaves the original export untouched: the
+      temp directory is the only write target, and it is cleaned up.
+    - A copy_onnx_model failure may leave the original `.onnx` and/or
+      `.data` sidecar partially overwritten: copy_onnx_model writes
+      directly to the destination (no temp-and-rename), so a process
+      kill or full disk mid-copy can corrupt the destination.
 
     Returns:
-        True if normalization succeeded, False otherwise.
+        True if normalization succeeded, False otherwise. On False, the
+        traceback is included in the warning log to aid debugging.
     """
     import shutil
     import tempfile
@@ -126,8 +144,11 @@ def _normalize_exported_model(output_path: Path) -> bool:
     try:
         optimize_onnx(model=output_path, output=tmp_path)
         copy_onnx_model(tmp_path, output_path)
-    except Exception as e:
-        logger.warning("Normalization failed; keeping un-normalized export: %s", e)
+    except Exception:
+        logger.warning(
+            "Normalization failed; keeping un-normalized export",
+            exc_info=True,
+        )
         return False
     finally:
         shutil.rmtree(tmp_dir, ignore_errors=True)
--- a/src/winml/modelkit/export/pytorch.py
+++ b/src/winml/modelkit/export/pytorch.py
@@ -86,7 +86,7 @@ def export_pytorch(
     )
     with warnings.catch_warnings():
         warnings.filterwarnings("ignore")
-        return exporter.export(
+        stats = exporter.export(
             model=model,
             output_path=str(output_path),
             export_config=export_config,
@@ -94,3 +94,41 @@ def export_pytorch(
             task=task,
             **kwargs,
         )
+
+    stats["model_normalization_succeeded"] = _normalize_exported_model(output_path)
+
+    return stats
+
+
+def _normalize_exported_model(output_path: Path) -> bool:
+    """Normalize the exported ONNX in-place via optimize_onnx.
+
+    Writes the normalized model into a temporary directory, then replaces
+    the original export (and its `.data` sidecar, if any) via
+    copy_onnx_model — which overwrites both the `.onnx` file and its
+    external-data sidecar by design. On any failure during optimization
+    or copy, logs a warning and leaves the original export untouched.
+    The temp directory is removed in either case.
+
+    Returns:
+        True if normalization succeeded, False otherwise.
+    """
+    import shutil
+    import tempfile
+
+    from ..onnx import copy_onnx_model
+    from ..optim import optimize_onnx
+
+    logger.info("Normalizing model")
+    tmp_dir = Path(tempfile.mkdtemp())
+    tmp_path = tmp_dir / output_path.name
+
+    try:
+        optimize_onnx(model=output_path, output=tmp_path)
+        copy_onnx_model(tmp_path, output_path)
+    except Exception as e:
+        logger.warning("Normalization failed; keeping un-normalized export: %s", e)
+        return False
+    finally:
+        shutil.rmtree(tmp_dir, ignore_errors=True)
+    return True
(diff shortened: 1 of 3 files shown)


DingmaomaoBJTU (approved):

src/winml/modelkit/export/pytorch.py
```diff
@@ -86,11 +86,49 @@ def export_pytorch(
...
+    copy_onnx_model — which overwrites both the `.onnx` file and its
+    external-data sidecar by design. On any failure during optimization
+    or copy, logs a warning and leaves the original export untouched.
+    The temp directory is removed in either case.
+
+    Returns:
+        True if normalization succeeded, False otherwise.
+    """
+    import shutil
+    import tempfile
```
timenick, line 135: **Perf — temp dir defaults to the system drive.**

`tempfile.mkdtemp()` with no `dir=` argument creates the temp directory under `%TEMP%` (Windows) / `/tmp` (Linux). For multi-GB models (the primary WinML target), this forces `copy_onnx_model` into a cross-volume copy — `shutil.copy2` cannot use the rename fast-path across mount points, so we end up paying the cost of a full data transfer twice (export → temp, temp → output) plus the disk-space tax in `%TEMP%`.

Simple fix:

```python
tmp_dir = Path(tempfile.mkdtemp(dir=output_path.parent))
```

Same-volume placement means the eventual `copy_onnx_model` is filesystem-rename-fast for the `.data` sidecar, and the temp footprint disappears from the system drive.

src/winml/modelkit/export/pytorch.py
```diff
@@ -86,11 +86,49 @@ def export_pytorch(
...
             model=model,
             output_path=str(output_path),
             export_config=export_config,
             model_name_or_path=model_name_or_path,
             task=task,
             **kwargs,
         )
+
+    stats["model_normalization_succeeded"] = _normalize_exported_model(output_path)
+
```
timenick, line 111: **API — no opt-out for the new default behavior.**

`export_pytorch()` is the single public entry point used by every export path (the `winml export` CLI, `build_hf_model`, the `winml build` command, and direct Python API callers per the PR description). Making normalization unconditional means callers who specifically need the raw `torch.onnx.export` output can no longer get it:
- Regression tests that assert torch-exporter semantics
- Custom downstream pipelines that run their own optimization
- Anyone debugging the exporter itself (now sees a post-shape-inference model and can''t tell what came from torch vs. what came from `optimize_onnx`)

Suggest a keyword argument:

```python
def export_pytorch(
    model, output_path, export_config, *,
    normalize: bool = True,
    **kwargs,
) -> dict:
    ...
    if normalize:
        stats["model_normalization_succeeded"] = _normalize_exported_model(output_path)
    return stats
```

Default-on preserves the new behavior for every existing caller; opt-out is one keyword away when needed.

src/winml/modelkit/export/pytorch.py
```diff
@@ -86,11 +86,49 @@ def export_pytorch(
...
+    Returns:
+        True if normalization succeeded, False otherwise.
+    """
+    import shutil
+    import tempfile
+
+    from ..onnx import copy_onnx_model
+    from ..optim import optimize_onnx
+
+    logger.info("Normalizing model")
```
timenick, line 140: **DX — broad `except Exception` swallows programming errors silently.**

This catches `AttributeError`, `ImportError`, `TypeError`, etc. the same as legitimate `optimize_onnx`/`copy_onnx_model` failures. If a future refactor breaks this helper (renamed function in `..optim`, signature change in `copy_onnx_model`, etc.), every export silently logs a warning at WARN level and continues — CI tests like `test_normalization_succeeds_and_shape_inferences` would flip to failing the `value_info` assertion, but the failure mode wouldn''t obviously point at the broken helper.

At minimum, include the traceback:

```python
except Exception:
    logger.warning(
        "Normalization failed; keeping un-normalized export", exc_info=True
    )
    return False
```

Better, narrow the catch to specific failures we expect (`OSError` from copy, `RuntimeError`/`ValueError` from optimize) and let the rest propagate during development.

src/winml/modelkit/export/pytorch.py
```diff
@@ -86,11 +86,49 @@ def export_pytorch(
...
+
+
+def _normalize_exported_model(output_path: Path) -> bool:
+    """Normalize the exported ONNX in-place via optimize_onnx.
+
+    Writes the normalized model into a temporary directory, then replaces
+    the original export (and its `.data` sidecar, if any) via
+    copy_onnx_model — which overwrites both the `.onnx` file and its
+    external-data sidecar by design. On any failure during optimization
+    or copy, logs a warning and leaves the original export untouched.
```
timenick: **Correctness — the docstring''s atomicity claim isn''t backed by `copy_onnx_model`.**

The docstring promises: *"On any failure during optimization or copy, logs a warning and leaves the original export untouched."*

True for `optimize_onnx` failures (the temp file is thrown away without touching the original). **Not true** for `copy_onnx_model` failures. Tracing through `src/winml/modelkit/onnx/external_data.py`:
- `_copy_single_external` calls `shutil.copy2(src_data, dst_data)` directly into the destination — a partial write leaves `dst.onnx.data` half-written.
- It then calls `onnx.save_model(model, str(dst))` — also writes directly, no temp-and-rename.

If the disk fills, permissions are revoked, or the process is killed mid-copy, the user is left with a corrupted original `.onnx` and/or `.data` sidecar.

If the docstring claim is intended to hold, the helper should do a temp-and-rename in this function as well:

```python
final_tmp = output_path.with_suffix(output_path.suffix + ".tmp")
final_tmp_data = output_path.with_suffix(output_path.suffix + ".data.tmp")
try:
    copy_onnx_model(tmp_path, final_tmp)
    final_tmp.replace(output_path)
    if final_tmp_data.exists():
        final_tmp_data.replace(output_path.with_suffix(output_path.suffix + ".data"))
except Exception:
    final_tmp.unlink(missing_ok=True)
    final_tmp_data.unlink(missing_ok=True)
    raise
```

Alternatively, soften the docstring claim to acknowledge that `copy_onnx_model` failures may corrupt the destination.

tests/unit/export/test_pytorch_export.py
```diff
@@ -268,6 +284,36 @@ def forward(self, ids):
...
+        )
+        with patch(
+            "winml.modelkit.export.pytorch._normalize_exported_model",
+            return_value=False,
+        ):
+            result = export_pytorch(model, tmp_path / "model.onnx", config)
+
+        assert result["model_normalization_succeeded"] is False
+
+        onnx_model = onnx.load(str(tmp_path / "model.onnx"))
```
timenick, line 331: This test patches `_normalize_exported_model` to return False, so it only verifies the trivial wiring ("when the helper returns False, the stats field is False and `value_info` isn''t populated"). It doesn''t actually cover the failure paths *inside* the helper:

- Does `_normalize_exported_model` correctly return False when `optimize_onnx` raises?
- Does it correctly return False when `copy_onnx_model` raises?
- Is the temp dir cleaned up in both failure modes?
- Is the original `.onnx` left readable after a failure?

A more rigorous test that closes the most-important gap:

```python
def test_normalization_optimize_failure_returns_false(self, tmp_path) -> None:
    """optimize_onnx raising must propagate to the success flag without corrupting the export."""
    model = TwoLayerNet()
    config = WinMLExportConfig(
        input_tensors=[InputTensorSpec(name="x", dtype="float32", shape=(1, 10))],
    )
    with patch(
        "winml.modelkit.export.pytorch.optimize_onnx",
        side_effect=RuntimeError("simulated optimize failure"),
    ):
        result = export_pytorch(model, tmp_path / "model.onnx", config)

    assert result["model_normalization_succeeded"] is False
    # Original export must still be a loadable, structurally-valid ONNX model.
    onnx_model = onnx.load(str(tmp_path / "model.onnx"))
    onnx.checker.check_model(onnx_model)
```

tests/unit/export/test_pytorch_export.py
```diff
@@ -268,6 +284,36 @@ def forward(self, ids):
             # This is expected — the test verifies the flow works up to export
             pass
 
+    def test_normalization_succeeds_and_shape_inferences(self, tmp_path) -> None:
```
timenick, line 287: **Test coverage gap — no integration test for the external-data path.**

The two new `copy_onnx_model` overwrite tests cover `copy_onnx_model` *directly*. The two new `export_pytorch` tests use `TwoLayerNet` — small enough that no external-data sidecar is created. So the full chain that matters in production:

```
export_pytorch  →  _normalize_exported_model  →  optimize_onnx (writes tmp_path + tmp_path.data)
                                              →  copy_onnx_model (overwrites output_path + output_path.data)
```

is only covered for the no-external-data case. With multi-GB WinML models — the primary motivation for this PR — every real-world invocation goes through the external-data branch, which is exactly the branch with no integration coverage.

Suggest one integration test that forces external data (large initializer or `save_as_external_data=True` shim) and asserts both `.onnx` and `.onnx.data` exist and load correctly after normalization.

timenick (commented):
Six findings — four substantive (perf, API contract, error handling, atomicity claim) and two test gaps. None are strict blockers, but #1 and #4 in particular are worth addressing before this lands as the default for every export path.

tests/unit/export/test_pytorch_export.py
vortex-captain: added UTs for `copy_onnx_model` would suffice

DingmaomaoBJTU:
### Code review

Found 2 issues:

1. `_normalize_exported_model` is called outside the `warnings.catch_warnings()` block, leaking optimizer/shape-inference warnings to callers. A previous PR (#460) explicitly centralized all warning suppression into this block to prevent stderr noise; the new call undoes that.

https://github.com/microsoft/winml-cli/blob/39e73f7cedecff0c89f914db559032144c8a8725/src/winml/modelkit/export/pytorch.py#L96-L100

2. Normalization is unconditional — there is no parameter (e.g. `normalize: bool = True`) to skip it. `export_pytorch` is the single shared entry point for all export paths, so callers that need the raw torch-exporter output (debugging, downstream pipelines, tests asserting exporter semantics) have no escape hatch.

https://github.com/microsoft/winml-cli/blob/39e73f7cedecff0c89f914db559032144c8a8725/src/winml/modelkit/export/pytorch.py#L87-L100

src/winml/modelkit/export/pytorch.py
vortex-captain: fixed

src/winml/modelkit/export/pytorch.py
vortex-captain: fixed docstring

src/winml/modelkit/export/pytorch.py
vortex-captain: fixed

tests/unit/export/test_pytorch_export.py
vortex-captain: fixed

src/winml/modelkit/export/pytorch.py
vortex-captain: fixed

DingmaomaoBJTU (approved):

PR #689, opened 2026-05-21
Fix onnxruntime search dll path.

chinazhangchao:
Windows ships C:\Windows\System32\onnxruntime.dll (older API version) as part of the system WindowsML component.
When WinML EP plugin DLLs are loaded (via EpCatalog), they import "onnxruntime.dll" by base name and the loader binds them to the system copy, producing  "The requested API version [N] is not available" errors.
Loading the wheel-bundled DLL by full path first makes later base-name imports resolve to the already-loaded module.

### DIFF
--- a/src/winml/modelkit/__init__.py
+++ b/src/winml/modelkit/__init__.py
@@ -42,6 +42,37 @@ for _stream in (sys.stdout, sys.stderr):
 
 logging.getLogger(__name__).addHandler(logging.NullHandler())
 
+
+def _preload_bundled_onnxruntime_dll() -> None:
+    # Windows ships C:\Windows\System32\onnxruntime.dll (older API version)
+    # as part of the system WindowsML component. When WinML EP plugin DLLs
+    # are loaded (via EpCatalog), they import "onnxruntime.dll" by base name
+    # and the loader binds them to the system copy, producing
+    # "The requested API version [N] is not available" errors.
+    # Loading the wheel-bundled DLL by full path first makes later base-name
+    # imports resolve to the already-loaded module.
+    if sys.platform != "win32":
+        return
+    try:
+        import ctypes
+        import os
+        from importlib.util import find_spec
+        from pathlib import Path
+
+        spec = find_spec("onnxruntime")
+        if spec is None or spec.origin is None:
+            return
+        dll = Path(spec.origin).parent / "capi" / "onnxruntime.dll"
+        if not dll.is_file():
+            return
+        os.add_dll_directory(str(dll.parent))
+        ctypes.WinDLL(str(dll))
+    except Exception as _e:  # pragma: no cover - best-effort preload
+        print(f"Warning: failed to preload bundled onnxruntime.dll: {_e}", file=sys.stderr)
+
+
+_preload_bundled_onnxruntime_dll()
+
 from . import _warnings  # Configure warning filters before importing subpackages
 
 
--- a/tests/unit/test_modelkit_init.py
+++ b/tests/unit/test_modelkit_init.py
@@ -1,0 +1,67 @@
+# -------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+# --------------------------------------------------------------------------
+"""Regression tests for ``_preload_bundled_onnxruntime_dll``.
+
+The preload is the load-order fix for the System32 vs. wheel-bundled
+``onnxruntime.dll`` collision on Windows. It must run before any subpackage
+import that transitively touches onnxruntime, otherwise the System32 copy
+wins and produces "requested API version [N] is not available" errors on
+end-user machines. These tests pin the contract without needing a real
+Windows host or a real wheel.
+"""
+
+from __future__ import annotations
+
+import sys
+import types
+from unittest import mock
+
+
+def _reimport_modelkit():
+    # Drop the cached package so its __init__.py (and the preload call at
+    # module scope) executes again under the active mocks.
+    for name in [
+        m for m in sys.modules if m == "winml.modelkit" or m.startswith("winml.modelkit.")
+    ]:
+        sys.modules.pop(name, None)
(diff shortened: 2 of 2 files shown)


xieofxie (approved):
will it slow much? do we need to find the exact place before ort is loaded?

chinazhangchao:
> will it slow much? do we need to find the exact place before ort is loaded?

Didn't feel slow. Almost the same.

timenick (commented):
(cleanup: probe comment removed)

timenick (commented):
(cleanup: probe comment removed)

timenick (commented):
(cleanup: probe comment removed)

DingmaomaoBJTU (approved):

src/winml/modelkit/__init__.py
```diff
@@ -42,6 +42,37 @@
 
 logging.getLogger(__name__).addHandler(logging.NullHandler())
 
+
+def _preload_bundled_onnxruntime_dll() -> None:
```
timenick, line 46: **Missing regression coverage for this preload.**

This function is the load-order fix for the System32 vs wheel-bundled `onnxruntime.dll` collision, but there is no test pinning its behavior. The mechanism is easy to silently regress — for example, if someone later moves `_preload_bundled_onnxruntime_dll()` below `from . import _warnings` (or any other subpackage import that transitively touches `onnxruntime`), the System32 DLL will win again and produce the very `"requested API version [N] is not available"` errors this PR fixes, but only on end-user Windows machines.

A small mock-based unit test would lock in the contract without needing real Windows or a real wheel:

```python
# tests/unit/test_modelkit_init.py
import sys
import types
from unittest import mock


def _reimport_modelkit():
    sys.modules.pop("winml.modelkit", None)
    import importlib
    return importlib.import_module("winml.modelkit")


def test_preload_is_noop_on_non_windows(monkeypatch):
    monkeypatch.setattr(sys, "platform", "linux")
    with mock.patch("ctypes.WinDLL", create=True) as windll:
        _reimport_modelkit()
        windll.assert_not_called()


def test_preload_skips_when_bundled_dll_missing(monkeypatch, tmp_path):
    monkeypatch.setattr(sys, "platform", "win32")
    fake_spec = types.SimpleNamespace(origin=str(tmp_path / "onnxruntime" / "__init__.py"))
    with mock.patch("importlib.util.find_spec", return_value=fake_spec), \
         mock.patch("ctypes.WinDLL", create=True) as windll, \
         mock.patch("os.add_dll_directory", create=True) as add_dir:
        _reimport_modelkit()
        windll.assert_not_called()
        add_dir.assert_not_called()


def test_preload_loads_dll_on_windows_when_present(monkeypatch, tmp_path):
    monkeypatch.setattr(sys, "platform", "win32")
    capi = tmp_path / "onnxruntime" / "capi"
    capi.mkdir(parents=True)
    (capi / "onnxruntime.dll").write_bytes(b"")
    fake_spec = types.SimpleNamespace(origin=str(tmp_path / "onnxruntime" / "__init__.py"))
    with mock.patch("importlib.util.find_spec", return_value=fake_spec), \
         mock.patch("ctypes.WinDLL", create=True) as windll, \
         mock.patch("os.add_dll_directory", create=True) as add_dir:
        _reimport_modelkit()
        add_dir.assert_called_once_with(str(capi))
        windll.assert_called_once_with(str(capi / "onnxruntime.dll"))
```

The three branches cover the platform short-circuit, the silent skip when the DLL is missing (so a future bundling regression is observable), and the happy path (so the load-order property is enforced).

timenick (commented):
Review: missing regression coverage for the DLL preload mechanism.

src/winml/modelkit/__init__.py
chinazhangchao: Added unit tests.

timenick (approved):

tezheng:
Do we have proof that "C:\Windows\System32\onnxruntime.dll" is indeed loaded? how can I repro if yest?

tezheng:
At least we should try this first, and see which DLLs get loaded

```python
import os, ctypes, onnxruntime as ort
from ctypes import wintypes

print("ort version:", ort.__version__)
print("ort package:", ort.__file__)
capi = os.path.join(os.path.dirname(ort.__file__), "capi")
for name in ("onnxruntime.dll", "onnxruntime_pybind11_state.pyd", "onnxruntime_providers_shared.dll", "DirectML.dll"):
    p = os.path.join(capi, name)
    print(f"  {'OK ' if os.path.isfile(p) else 'NO '} {p}")

k32 = ctypes.WinDLL("kernel32")
k32.GetCurrentProcess.restype = wintypes.HANDLE

psapi = ctypes.WinDLL("psapi")
psapi.EnumProcessModules.argtypes = [wintypes.HANDLE, ctypes.POINTER(ctypes.c_void_p), wintypes.DWORD, ctypes.POINTER(wintypes.DWORD)]
psapi.EnumProcessModules.restype = wintypes.BOOL
psapi.GetModuleFileNameExW.argtypes = [wintypes.HANDLE, ctypes.c_void_p, wintypes.LPWSTR, wintypes.DWORD]

mods = (ctypes.c_void_p * 4096)(); need = wintypes.DWORD()
psapi.EnumProcessModules(k32.GetCurrentProcess(), mods, ctypes.sizeof(mods), ctypes.byref(need))
print("--- onnxruntime* modules currently mapped in process ---")
for i in range(need.value // ctypes.sizeof(ctypes.c_void_p)):
    buf = ctypes.create_unicode_buffer(32768)
    psapi.GetModuleFileNameExW(k32.GetCurrentProcess(), mods[i], buf, len(buf))
    if "onnxruntime" in os.path.basename(buf.value).lower():
        print(" ", buf.value)
```

xieofxie:
> Do we have proof that "C:\Windows\System32\onnxruntime.dll" is indeed loaded? how can I repro if yest?

yes, I have tested this both by process explorer and also log show it
image

PR #693, opened 2026-05-21
fix(telemetry): emit deviceid in CS 4.0 r:<uuid> form so OneCollector accepts events

### LINKED ISSUE #691: Telemetry device_id format violates CS 4.0 ext.device.localId schema; all events rejected
### Summary

Every telemetry batch is rejected by OneCollector because `ext.device.localId` doesn't match the CS 4.0 schema. `winml sys -v` shows:

```
DEBUG  telemetry backend returned 400: error='No events passed validation.'
       body={"acc":0,"rej":2,"efi":{"InvalidDeviceLocalId":[0,1]}}
```

`acc:0` across all batches means **no telemetry has ever reached the backend** since this code shipped. Failure is silent at normal log level.

### Root cause

[`deviceid.py:65-66`](src/winml/modelkit/telemetry/deviceid/deviceid.py#L65-L66) generates the id as a SHA256 hex digest:

```python
def _hash_uuid(value: uuid.UUID) -> str:
    return hashlib.sha256(str(value).encode("utf-8")).hexdigest()
```

→ 64-char lowercase hex, no prefix, no dashes.

The CS 4.0 `ext.device.localId` field requires `<scope>:<canonical-uuid>` (e.g. `r:550e8400-e29b-41d4-a716-446655440000`). The current value satisfies neither requirement, so the backend rejects every event with `InvalidDeviceLocalId`.

The SHA256 step is also pointless on its own merits: the input is already `uuid.uuid4()` (cryptographically random), so hashing doesn't add privacy — it just reshapes the value into something the schema rejects.

### Fix

[`deviceid.py`](src/winml/modelkit/telemetry/deviceid/deviceid.py):

```python
new_id = f"r:{uuid.uuid4()}"   # 'r:' = random scope per CS 4.0
```

Drop `_hash_uuid`. Existing persisted 64-hex values must be regenerated — there's no continuity to preserve since the backend has rejected 100% of events using them.

timenick:
## Summary

Closes #691.

`device_id` was emitted as a 64-char SHA256 hex digest. OneCollector requires `ext.device.localId` to be `<scope>:<canonical-uuid>` and rejected every batch with `InvalidDeviceLocalId` — `acc:0` across all sends since this code shipped, silent at normal log level.

- Generate `f"r:{uuid.uuid4()}"` (random-scope per CS 4.0) in [`src/winml/modelkit/telemetry/deviceid/deviceid.py`](src/winml/modelkit/telemetry/deviceid/deviceid.py). Drop `_hash_uuid` — hashing a uuid4 added no privacy and broke the schema.
- Stored 64-hex values from releases ≤ 0.0.4 fail `_LOCAL_ID_RE` on read and self-heal on the next call. No migration script, no user action required.
- Unit tests updated; added `test_legacy_sha256_hex_value_is_regenerated` to lock in the migration path.

## Verification

End-to-end against the **real** OneCollector backend (in-process monkeypatched iKey, no source knobs):

```
[0] POST 1 envelope(s) → HTTP 200  Collector-Error=None
    body: {"acc":1}
```

`acc:1, rej:0` — backend accepts the event. Pre-fix produced `{"acc":0,"rej":N,"efi":{"InvalidDeviceLocalId":[...]}}`.

`uv run pytest tests/unit/telemetry/` → 185 passed.

### DIFF
--- a/src/winml/modelkit/telemetry/deviceid/deviceid.py
+++ b/src/winml/modelkit/telemetry/deviceid/deviceid.py
@@ -3,17 +3,18 @@
 # Licensed under the MIT License.
 # --------------------------------------------------------------------------
 
-"""Stable-per-device ID used as the telemetry device_id.
+r"""Stable-per-device ID used as the telemetry device_id.
 
-Derived from a random UUID4, hashed with SHA256, and persisted so the same
-machine reports the same id across sessions. Users can reset by removing the
-stored value (registry on Windows, state file elsewhere).
+A random UUID4 in CS 4.0 ``ext.device.localId`` form (``r:<canonical-uuid>``,
+where ``r`` is the random scope). Persisted so the same machine reports the
+same id across sessions. Users can reset by removing the stored value
+(``HKCU\SOFTWARE\Microsoft\DeveloperTools\.modelkit``).
 """
 
 from __future__ import annotations
 
-import hashlib
 import logging
+import re
 import uuid
 from enum import Enum
 
@@ -22,6 +23,11 @@ from . import _store
 
 _LOGGER = logging.getLogger(__name__)
 _STORAGE_KEY = "deviceid"
+# OneCollector rejects any localId that isn't <scope>:<canonical-uuid>; the
+# 'r' (random) scope is what we generate. Stored values that don't match
+# (e.g. SHA256 hex digests written by releases <= 0.0.4) are treated as
+# absent and regenerated — see issue #691.
+_LOCAL_ID_RE = re.compile(r"^r:[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$")
 
 
 class IdStatus(str, Enum):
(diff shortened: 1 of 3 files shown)


src/winml/modelkit/telemetry/deviceid/deviceid.py
```diff
@@ -22,6 +23,11 @@
 
 _LOGGER = logging.getLogger(__name__)
 _STORAGE_KEY = "deviceid"
+# OneCollector rejects any localId that isn't <scope>:<canonical-uuid>; the
+# 'r' (random) scope is what we generate. Stored values that don't match
+# (e.g. SHA256 hex digests written by releases <= 0.0.4) are treated as
+# absent and regenerated — see issue #691.
+_LOCAL_ID_RE = re.compile(r"^r:[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$")
```
xieofxie: adding this check to a non release code is strange.. how about just change key to device_id instead?

timenick: Good point — the regex is permanent runtime defense for a one-time migration shape. Switching to a key rename:

- `_STORAGE_KEY = "device_id"` (matches the OTel resource attribute name), drop `_LOCAL_ID_RE` and the validation branch.
- Old code's `"deviceid"` value goes unread; the legacy SHA256-hex lingers as an orphan REG_SZ in HKCU. Harmless and naturally cleaned up by uninstall / manual reset.
- Migration test becomes `test_legacy_deviceid_key_is_ignored` — writes to old key, asserts new key returns NEW.

Will push the change. Let me know if you'd rather also `delete_key("deviceid")` on first NEW write to avoid the orphan — adds 3 lines, marginal.

Generated with Claude Code

xieofxie (approved):

DingmaomaoBJTU (approved):

PR #694, opened 2026-05-21
Improve winml eval UX: validate inputs upfront, friendlier errors, structured --schema

### LINKED ISSUE #216: winml eval fails for text-classification models: label alignment error with auto-selected dataset
## Description

`winml eval` fails for text-classification models when the auto-selected dataset's label set does not match the model's label set. The evaluator raises a `RuntimeError` during label alignment.

## Repro

```bash
winml eval -m ProsusAI/finbert --device npu --samples 100
```

## Error

```
RuntimeError: Label alignment failed for dataset 'nyu-mll/glue': 'equivalent'
Error: Evaluation failed: Label alignment failed for dataset 'nyu-mll/glue': 'equivalent'
```

## Root Cause

`winml eval` auto-selects `nyu-mll/glue` for text-classification tasks. However, `ProsusAI/finbert` uses labels `positive`, `negative`, `neutral`, which do not align with GLUE's label schema. The label alignment logic in `base_evaluator.py` raises an error rather than gracefully handling the mismatch.

## Expected

Either auto-select a compatible dataset for the specific model, or surface a clear error message suggesting the user specify a dataset manually via `--dataset`.

## Workaround

None currently. Image-classification models are unaffected.

## Environment

- OS: Windows 11
- winml-modelkit: 0.0.1.dev1
- Model: ProsusAI/finbert (text-classification, bert)

### LINKED ISSUE #528: [winml eval] [P0] Bogus or mismatched `--dataset` leaks raw multi-frame `datasets`/internal traceback
**Severity**: P0  
**Command**: `winml eval`

---
**Category**: Error Messages — never leak third-party traceback frames at default verbosity.

**Repro 1 — dataset that doesn't exist on the Hub**:
```
uv run winml eval -m microsoft/resnet-50 --dataset does-not-exist --samples 3
```

**Actual** (default verbosity, last 10 lines):
```
  File "...\datasets\load.py", line 1393, in dataset_module_factory
    raise DatasetNotFoundError(...) from e
datasets.exceptions.DatasetNotFoundError: Dataset 'does-not-exist' doesn't exist on the Hub or cannot be accessed.
Error: Evaluation failed: Dataset 'does-not-exist' doesn't exist on the Hub or cannot be accessed.
```
Four frames of `datasets/load.py` internals are printed before the friendly error.

**Repro 2 — valid but task-incompatible dataset**:
```
uv run winml eval -m microsoft/resnet-50 --dataset glue --dataset-name mrpc --samples 3
```

**Actual**: ~18 frames of traceback through `datasets/arrow_dataset.py`, `winml/modelkit/eval/base_evaluator.py`, ending in:
```
KeyError: 'equivalent'
...
RuntimeError: Label alignment failed for dataset 'glue': 'equivalent'
Error: Evaluation failed: Label alignment failed for dataset 'glue': 'equivalent'
```
No guidance to the user about *why* `glue/mrpc` is incompatible with an image-classification model, or how to find a compatible dataset.

**Expected**:
1. Wrap dataset loading in a guard that converts `DatasetNotFoundError` into a clean Click error with no traceback at default verbosity:
   ```
   Error: Dataset 'does-not-exist' was not found on the Hugging Face Hub.
          Check the spelling, or run 'winml eval --help' for the auto-discovered defaults.
          Use -v to see the underlying loader exception.
   ```
2. Validate dataset/task compatibility *before* iterating samples. If the model task is `image-classification` and the dataset has no image column or has incompatible labels, fail with:
   ```
   Error: Dataset 'glue/mrpc' is not compatible with task 'image-classification'.
          Expected an image column and integer labels; found columns: sentence1, sentence2, label (string).
          Run 'winml eval --schema --task image-classification' for compatible dataset shape.
   ```
3. Only show the raw traceback when `-v` is passed.

**Why it matters**: Eval is a high-traffic command and `--dataset` is the flag users are most likely to typo or misuse. Today the user sees pages of traceback through unfamiliar libraries (`datasets`, `arrow`) and has no clue whether the bug is in `winml`, `datasets`, or their own input. This is the single biggest UX cliff in eval.

### LINKED ISSUE #529: [winml eval] [P1] `eval` accepts a pre-built ONNX whose I/O signature does not match `--task`; failure surfaces late as an unrelated pipeline/dataset error
**Severity**: P1  
**Command**: `winml eval`

---
**Category**: Validate Inputs Up-Front + Actionable Errors — `eval` should cross-check the ONNX graph's input names against the requested task's expected inputs at load time.

**Status**: Behavior derived from static code path inspection (`commands/eval.py`, `eval/evaluate.py`). Runtime reproduction was blocked by an in-progress merge conflict in `src/winml/modelkit/commands/eval.py` on the audit branch; the reproduction string below is the path the code will follow once the conflict is resolved.

**Repro**:
```
# 1. Export a real image-classification ONNX:
uv run winml export -m microsoft/resnet-50 -o temp\resnet50.onnx
# 2. Ask eval to run it as a text-generation task:
uv run winml eval -m temp\resnet50.onnx --model-id microsoft/resnet-50 --task text-generation --samples 5
```

**Actual** (from code-path analysis):
- `_resolve_task(cfg)` returns `"text-generation"` because `cfg.task` is non-`None`. **No validation** that the task is even a known evaluator key, and **no validation** that the task is compatible with the ONNX graph.
- `WinMLAutoModel.from_onnx("temp\\resnet50.onnx", task="text-generation")` succeeds — `from_onnx` does not inspect the graph's `input` / `output` names against the task's expected I/O contract.
- `_EVALUATOR_REGISTRY.get("text-generation")` returns the text-generation evaluator, which constructs an HF `pipeline(task="text-generation", model=<resnet-onnx>)`.
- Failure surfaces several seconds later, deep in dataset preparation or first inference, as an `IndexError`/`KeyError`/missing-`input_ids` error. The user has no signal that the *task* flag is the cause; the symptom looks like a dataset or model-load bug.

**Expected**: At ONNX load time, when the user has explicitly passed `--task <T>`, compare the ONNX graph's input names against the task's documented input contract (e.g. `image-classification` → `pixel_values`; `text-generation` → `input_ids` + `attention_mask`). On mismatch, exit 2 with one line:
```
Error: ONNX file temp\resnet50.onnx exposes inputs [pixel_values] which do not
match task 'text-generation' (expects [input_ids, attention_mask]).
Did the model export with --task image-classification?
```
Secondary fix: validate `--task` against the evaluator registry at command entry; reject unknown task strings up front instead of letting `_EVALUATOR_REGISTRY.get(...)` fail late.

**Why it matters**: A user who passes the wrong `--task` (typo, wrong recall, or copy-paste from another command) gets neither a useful error nor a hint that `--task` is the wrong knob. Worst case, an evaluator that does not strictly check input shape could *complete* against random model output and report bogus accuracy numbers. Violates R4.1 (validate inputs at entry), R4.2 (reject unsupported combinations), R3.1 (actionable errors).

### LINKED ISSUE #532: [winml eval] [P2] Confusing dual `-m/--model` (`multiple=True`) + `--model-id`; no mutual-exclusion check; no role-key list in `--help`
**Severity**: P2  
**Command**: `winml eval`

---
**Category**: Canonical Flag Names — one concept, one flag.

**Repro**:
```
uv run winml eval --help | Select-String '\-\-model|\-\-model-id'
```

**Actual**: `--help` shows two model-input flags:
- `-m, --model TEXT  HuggingFace model id or path (can be passed multiple times with role keys e.g. -m generator=foo -m discriminator=bar)`
- `--model-id TEXT   Hugging Face model id` (single).

The role-key syntax (`generator=foo`) is mentioned but the list of valid role keys is **not** in `--help`. Passing both `-m` and `--model-id` is not flagged as conflicting.

**Expected**: One canonical input flag (`-m/--model`, `multiple=True`). Document the role keys in `--help` (`Valid roles: generator, discriminator, embedder, …`). Drop `--model-id` (or alias it to `-m`).

**Why it matters**: Three ways to spell "the model to evaluate" and zero documentation of the role keys forces users to read source.

zhenchaoni:
## Improve `winml eval` UX: validate inputs upfront, friendlier errors, structured `--schema`

Addresses [#216](https://github.com/microsoft/winml-cli/issues/216), [#528](https://github.com/microsoft/winml-cli/issues/528), [#529](https://github.com/microsoft/winml-cli/issues/529), [#532](https://github.com/microsoft/winml-cli/issues/532).

### What changes

**Task validation (#529)**
- `_resolve_task()` now validates the resolved task against `_EVALUATOR_REGISTRY` for both explicit `--task` and HF-inferred paths. Unsupported tasks fail immediately with `Task 'xyz' is not supported. Supported tasks: ...`, before model load and before the misleading "no default dataset" branch.
- `evaluate()` wraps `_load_model()` and `compute()` in actionable error messages pointing the user at `winml eval --schema --task <task>`.
- CLI no longer dumps a traceback at default verbosity; `-v` re-enables `logger.exception`.

**Dataset column validation (#216, #528)**
- New `utils/eval_utils.py` centralizes per-task schemas (`TaskSchema`, `SchemaItem`, `TASK_SCHEMAS`) plus `validate_dataset_columns()` and `DatasetValidationError`.
- `WinMLEvaluator.load_dataset()` validates required columns up-front; missing columns raise `DatasetValidationError` (re-raised by `evaluate()` as `Dataset '...' is not compatible with task '...': missing required column(s) ...`).
- Label-alignment failures and empty-after-filter cases now raise `DatasetValidationError` instead of `RuntimeError` / `ValueError`, so users get the same friendly wrap-up (e.g. the finbert + GLUE case in #216).
- Per-evaluator hard-coded column-default constants removed; defaults come from the shared schema table.

**`--schema` redesign (#532)**
- Fast path served from the lightweight `TASK_SCHEMAS` table — no longer imports the heavy `winml.modelkit.eval` package.
- New output format documents required dataset columns, optional config params, `--column` overrides, and (for split-encoder tasks) the `-m role=path` forms.
- `--schema --task <bad>` and `--schema` without `--task` print a clear usage hint listing supported tasks.

**Model flag cleanup (#532)**
- `-m / --model` help text simplified to one sentence; role-key syntax now documented in `--schema` output rather than the option help.
- `_resolve_model_path()` rejects `-m <hf_id> --model-id <other_id>` with `Cannot pass both ...`. The existing rule that `-m <onnx>` requires `--model-id` is unchanged.

**Misc**
- `evaluate.print_config()` lost the `used_default` flag; the dataset line now prints `ds.path` plainly.
- `commands/eval.py` example block reduced to two canonical invocations (HF id, ONNX path).

### Files
28 files changed, +601/-480. New: eval_utils.py. Bulk per-evaluator edits remove dead column constants and route through the shared schema.

### Verification
- `ruff check` clean
- `pytest tests/unit/` — 4953 passed
- `pytest test_eval_e2e.py -m e2e` — 45 passed (one e2e adjusted: `test_image_to_text_fp16` now passes `--column label_column=caption` because `lmms-lab/flickr30k` uses `caption`, which the new validator correctly flags)

### Not in this PR
- Static ONNX-input-vs-task contract check (the "expected `[input_ids, attention_mask]`" wording from #529). Existing wrapped `_load_model` / `compute()` errors already point users at `--schema`, so this is a follow-up enhancement rather than a fix.
- Dropping/aliasing `--model-id` (#532). Done conservatively: kept the flag, added the conflict check; flag-name consolidation is a separate UX decision.

### DIFF
--- a/src/winml/modelkit/commands/eval.py
+++ b/src/winml/modelkit/commands/eval.py
@@ -15,6 +15,7 @@ from typing import TYPE_CHECKING
 import click
 
 from ..utils import cli as cli_utils
+from ..utils.eval_utils import TASK_SCHEMAS, TaskSchema
 
 
 if TYPE_CHECKING:
@@ -31,13 +32,7 @@ logger = logging.getLogger(__name__)
     type=str,
     multiple=True,
     default=(),
-    help=(
-        "Model to evaluate. Accepts three forms: "
-        "(1) HuggingFace model ID, e.g. `-m <hf_model_id>`. "
-        "(2) ONNX file path, e.g. `-m model.onnx` (requires --model-id). "
-        "(3) Composite / split-encoder model as repeated role=path pairs, "
-        "e.g. `-m image-encoder=vision.onnx -m text-encoder=text.onnx`."
-    ),
+    help="Model to evaluate. Accepts a HuggingFace model ID or an ONNX file path.",
 )
 @click.option(
     "--model-id",
@@ -166,25 +161,38 @@ def eval(
     show_schema: bool,
     config_file: Path | None,
 ) -> None:
-    r"""Evaluate model accuracy on a dataset.
+    r"""Evaluate a model for a task.
 
-    If --dataset is not provided, a default dataset is used based on the task.
-
-    \b
     Examples:
-        # Use default dataset (auto-detected from task)
         winml eval -m microsoft/resnet-50
-        winml eval -m model.onnx --model-id dslim/bert-base-NER
 
-        # Specify dataset explicitly
-        winml eval -m microsoft/resnet-50 --dataset imagenet-1k
-        winml eval -m model.onnx --model-id microsoft/resnet-50 --dataset imagenet-1k
+        winml eval -m model.onnx --model-id microsoft/resnet-50
 
-        # Multi-config dataset with column overrides
-        winml eval -m model.onnx --model-id Intel/bert-base-uncased-mrpc \\
-            --dataset glue --dataset-name mrpc \\
-            --column input_column=sentence1
+    Run `winml eval --schema --task <task>` to see the dataset columns
+    and options expected by each task.
     """
+    # ── 0. --schema fast path: served from a local lightweight schema table
+    #       so this branch does not import the heavy winml.modelkit.eval package.
+    if show_schema:
+        task_arg: str | None = ctx.params.get("task")
+        if task_arg is None:
+            task_list = "\n  ".join(sorted(TASK_SCHEMAS))
+            click.echo(
+                "--schema requires --task <task>.\n\n"
+                f"Supported tasks:\n  {task_list}\n\n"
+                "Example: winml eval --schema --task image-classification"
+            )
+            return
+        schema = TASK_SCHEMAS.get(task_arg)
+        if schema is None:
+            supported = ", ".join(sorted(TASK_SCHEMAS))
+            raise click.UsageError(
+                f"Task '{task_arg}' is not supported by `winml eval`. "
+                f"Supported tasks: {supported}."
+            )
+        _print_schema(task_arg, schema)
+        return
+
     if verbose or (ctx.obj and ctx.obj.get("debug")):
         logging.getLogger("winml.modelkit").setLevel(logging.DEBUG)
 
@@ -192,20 +200,6 @@ def eval(
 
     # ── 1. Build config: defaults ← config file ← CLI ──
     cfg = _build_eval_config(ctx, config_file, column, label_mapping)
-
-    if show_schema:
-        from ..eval.evaluate import get_evaluator_class
-
-        if cfg.task is None:
-            raise click.UsageError(
-                "--schema requires --task. Example: winml eval --schema --task object-detection"
-            )
-        try:
-            cls = get_evaluator_class(cfg.task)
-        except ValueError as e:
-            raise click.UsageError(str(e)) from e
-        _print_schema(cfg.task, cls.schema_info())
-        return
 
     # ── 2. Resolve in place ──
     _resolve_model(cfg, model, model_id)
@@ -227,7 +221,8 @@ def eval(
         result = evaluate(cfg)
         _write_and_display(result, cfg.output_path)
     except Exception as e:
-        logger.exception("Evaluation failed")
+        if verbose:
+            logger.exception("Evaluation failed")
         raise click.ClickException(f"Evaluation failed: {e}") from e
 
 
@@ -475,6 +470,11 @@ def _resolve_model_path(
                 "for preprocessor and config resolution."
             )
         return value, model_id
+    if model_id is not None and model_id != value:
+        raise click.UsageError(
+            "Cannot pass both `-m <hf_id>` and `--model-id`. "
+            "Use `--model-id` only together with an ONNX file path in `-m`."
+        )
     return None, model_id or value
 
 
@@ -539,24 +539,41 @@ def display_eval_report(result: object, console: object) -> None:
     console.print()
 
 
-def _print_schema(task: str, schema: list, indent: int = 0) -> None:
-    """Format and print structured schema info."""
-    prefix = "  " * indent
-    if indent == 0:
-        click.echo(f"Dataset schema ({task}):\n")
-        click.echo(f"{prefix}{'Column':<20} {'Type':<25} {'Override (--column)'}")
-        click.echo(f"{prefix}{'-' * 20} {'-' * 25} {'-' * 25}")
+def _print_schema(task: str, schema: TaskSchema) -> None:
+    """Render the human-readable input schema for *task*."""
+    width = 50
+    title = f"Input schema for {task} models"
+    click.echo(title)
+    click.echo("=" * width)
+    click.echo()
 
-    for col in schema:
-        marker = "*" if col.required else " "
-        override_str = f"--column {col.override}={col.name}" if col.override else ""
-        click.echo(f"{prefix}{marker} {col.name:<18} {col.type:<25} {override_str}")
-        if col.description:
-            click.echo(f"{prefix}  {' ' * 18} {col.description}")
-        for child in col.children:
-            co = f"--column {child.override}={child.name}" if child.override else ""
-            click.echo(f"{prefix}    .{child.name:<16} {child.type:<25} {co}")
-            if child.description:
-                click.echo(f"{prefix}    {' ' * 18} {child.description}")
+    click.echo("--column option schema")
+    click.echo()
+    click.echo("Evaluating needs a dataset with the following columns:")
+    for item in schema.columns:
+        click.echo(f"  {item.name}")
+        click.echo(f"      {item.description} (default: {item.default})")
 
-    click.echo(f"\n{prefix}* = required")
+    if schema.params:
+        click.echo()
+        click.echo("Additional configuration parameters:")
+        for p in schema.params:
+            click.echo(f"  {p.name}")
+            click.echo(f"      {p.description} (default: {p.default})")
+
+    overrides = [c for c in (*schema.columns, *schema.params) if c.remap_hint]
+    if overrides:
+        click.echo()
+        click.echo("Override any default with --column:")
+        for c in overrides:
+            click.echo(f"  --column {c.name}={c.remap_hint}")
+
+    if schema.roles:
+        click.echo()
+        click.echo("-" * width)
+        click.echo("-m option schema")
+        click.echo()
+        click.echo("Use one of the following model input forms:")
+        click.echo("  1. use huggingface id: -m <hf-id>")
+        model_args = " ".join(f"-m {r}=<{r}.onnx>" for r in schema.roles)
+        click.echo(f"  2. use onnx file: {model_args} --model-id <hf-id>")
--- a/src/winml/modelkit/commands/eval.py
+++ b/src/winml/modelkit/commands/eval.py
@@ -32,7 +32,10 @@ logger = logging.getLogger(__name__)
     type=str,
     multiple=True,
     default=(),
-    help="Model to evaluate. Accepts a HuggingFace model ID or an ONNX file path.",
+    help=(
+        "Model to evaluate. Accepts a HuggingFace model ID, an ONNX file path "
+        "(requires --model-id), or split-encoder role=path pairs (see --schema)."
+    ),
 )
 @click.option(
     "--model-id",
@@ -174,7 +177,7 @@ def eval(
     # ── 0. --schema fast path: served from a local lightweight schema table
     #       so this branch does not import the heavy winml.modelkit.eval package.
     if show_schema:
-        task_arg: str | None = ctx.params.get("task")
+        task_arg = task
         if task_arg is None:
             task_list = "\n  ".join(sorted(TASK_SCHEMAS))
             click.echo(
--- a/src/winml/modelkit/eval/evaluate.py
+++ b/src/winml/modelkit/eval/evaluate.py
@@ -12,6 +12,8 @@ from copy import deepcopy
 from dataclasses import dataclass, field, replace
 from typing import TYPE_CHECKING, Any
 
+from rich.console import Console
+
 from .base_evaluator import WinMLEvaluator
 from .config import WinMLEvaluationConfig
 from .feature_extraction_evaluator import WinMLFeatureExtractionEvaluator
@@ -31,6 +33,7 @@ if TYPE_CHECKING:
     from ..models.winml.base import WinMLPreTrainedModel
 
 logger = logging.getLogger(__name__)
+console = Console()
 
 _EVALUATOR_REGISTRY: dict[str, type[WinMLEvaluator]] = {
     "image-classification": WinMLEvaluator,
@@ -197,19 +200,24 @@ def _load_model(config: WinMLEvaluationConfig) -> WinMLPreTrainedModel:
 
 
 def _resolve_task(config: WinMLEvaluationConfig) -> str:
-    """Resolve task from config or model's HF config."""
+    """Resolve task from config or model's HF config, and validate it is supported."""
     if config.task is not None:
-        return config.task
+        task = config.task
+    else:
+        if config.model_id is None:
+            raise ValueError("Cannot infer task without model_id. Provide --task.")
 
-    if config.model_id is None:
-        raise ValueError("Cannot infer task without model_id. Provide --task.")
+        from transformers import AutoConfig
 
-    from transformers import AutoConfig
+        from ..loader.task import _detect_task_from_config
 
-    from ..loader.task import _detect_task_from_config
+        hf_config = AutoConfig.from_pretrained(config.model_id)
+        task = _detect_task_from_config(hf_config)
 
-    hf_config = AutoConfig.from_pretrained(config.model_id)
-    return _detect_task_from_config(hf_config)
+    if task not in _EVALUATOR_REGISTRY:
+        supported = ", ".join(sorted(_EVALUATOR_REGISTRY))
+        raise ValueError(f"Task '{task}' is not supported. Supported tasks: {supported}.")
+    return task
 
 
 def evaluate(config: WinMLEvaluationConfig) -> EvalResult:
@@ -220,7 +228,13 @@ def evaluate(config: WinMLEvaluationConfig) -> EvalResult:
     config and any module-level defaults remain untouched.
     """
     config = replace(config, task=_resolve_task(config), dataset=deepcopy(config.dataset))
-    model = _load_model(config)
+    try:
+        model = _load_model(config)
+    except Exception as e:
+        raise ValueError(
+            f"Failed to load model '{config.model_id}'. "
+            f"Run 'winml eval --schema --task {config.task}' to see expected model inputs.",
+        ) from e
 
     if config.dataset.path is None:
         default = _DEFAULT_DATASETS.get(config.task)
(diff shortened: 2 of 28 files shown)


src/winml/modelkit/eval/evaluate.py
```diff
@@ -230,19 +244,57 @@ def evaluate(config: WinMLEvaluationConfig) -> EvalResult:
...
+    try:
+        task_evaluator = cls(config, model)
+        metrics = task_evaluator.compute()
+    except DatasetValidationError as e:
+        raise ValueError(
+            f"Dataset '{config.dataset.path}' is not compatible with task "
+            f"'{config.task}': {e}. Use --dataset to specify a different dataset, "
+            f"or run 'winml eval --schema --task {config.task}' to see the expected schema.",
+        ) from e
+    except Exception as e:
```
timenick: **Broad `except Exception` masks programmer bugs with a misleading schema hint.**

This fallback catches every non-`DatasetValidationError` exception thrown by `cls(config, model)` or `compute()` and rewrites it as `"Failed to compute metrics ... Run 'winml eval --schema --task X' to see the expected schema."`

The problem: this also catches programmer errors inside the evaluator code path (AttributeError, TypeError, KeyError on internal dicts, `transformers` API mismatches, OOMs, etc.) and points the user at the **dataset** schema — which is the wrong direction when the bug isn't dataset-shaped. From the user's POV, every internal failure looks like "your dataset is wrong" even when their dataset is fine.

Suggested narrowing:

```python
except (KeyError, ValueError) as e:
    # Most known dataset/label-shape problems surface as KeyError/ValueError
    # from datasets/HF. Wrap with the schema hint.
    raise ValueError(
        f"Failed to compute metrics for task '{config.task}' on dataset "
        f"'{config.dataset.path}'. "
        f"Run 'winml eval --schema --task {config.task}' to see the expected schema.",
    ) from e
```

Anything else (programmer errors, transformers internal failures) should propagate so it gets the existing `-v` traceback path in `commands/eval.py` rather than being relabelled as a schema issue.

src/winml/modelkit/utils/eval_utils.py
```diff
@@ -0,0 +1,308 @@
...
+        ),
+        SchemaItem(
+            "label_column", "integer class label",
+            default="label", remap_hint="<your_label_column>",
+        ),
+    ),
+    roles=("image-encoder", "text-encoder"),
+)
+
+TASK_SCHEMAS: dict[str, TaskSchema] = {
```
timenick, line 245: **`TASK_SCHEMAS` and `_EVALUATOR_REGISTRY` are now two separate sources of truth that must stay in sync by hand.**

Currently the two dicts have matching keys, but nothing enforces it. Drift modes are silent and asymmetric:

- **Schema-only**: task in `TASK_SCHEMAS` but not `_EVALUATOR_REGISTRY` → `--schema --task X` prints fine, but `winml eval --task X` fails at `get_evaluator_class` with a different message.
- **Registry-only**: task in `_EVALUATOR_REGISTRY` but not `TASK_SCHEMAS` → `validate_dataset_columns()` silently no-ops (no `schema`), so the entire upfront-validation goal of this PR is defeated for that task, *and* `--schema --task X` reports "not supported by `winml eval`" even though the evaluator exists.

Two ways to make this safe:

1. **Single source of truth**: attach `TaskSchema` to each evaluator class (e.g. `WinMLEvaluator.SCHEMA: ClassVar[TaskSchema | None]`), and derive `TASK_SCHEMAS` from `_EVALUATOR_REGISTRY`. The `--schema` fast path then loads only the lightweight `TaskSchema` instances (which still don't drag in transformers/torch if they live in `utils/eval_utils.py`).

2. **Lightweight invariant test** (cheaper to add now):

```python
def test_task_schemas_match_evaluator_registry():
    from winml.modelkit.eval.evaluate import _EVALUATOR_REGISTRY
    from winml.modelkit.utils.eval_utils import TASK_SCHEMAS
    assert set(_EVALUATOR_REGISTRY) == set(TASK_SCHEMAS), (
        f"Drift detected. Only in registry: {set(_EVALUATOR_REGISTRY) - set(TASK_SCHEMAS)}; "
        f"only in schemas: {set(TASK_SCHEMAS) - set(_EVALUATOR_REGISTRY)}"
    )
```

Without this, the next task added has a ~50% chance of becoming a silent-no-op or a UX gap.

src/winml/modelkit/eval/evaluate.py
```diff
@@ -220,7 +228,13 @@ def evaluate(config: WinMLEvaluationConfig) -> EvalResult:
     config and any module-level defaults remain untouched.
     """
     config = replace(config, task=_resolve_task(config), dataset=deepcopy(config.dataset))
-    model = _load_model(config)
+    try:
+        model = _load_model(config)
+    except Exception as e:
+        raise ValueError(
+            f"Failed to load model '{config.model_id}'. "
+            f"Run 'winml eval --schema --task {config.task}' to see expected model inputs.",
```
timenick: **"to see expected model inputs" is misleading wording.**

This fires when `_load_model()` raises. The hint sends the user to `winml eval --schema --task X`, but `--schema` output documents **dataset columns, config params, and (for split-encoder tasks) `-m role=path` forms**. It does not document ONNX model input tensor names/shapes.

For split-encoder tasks (`image-to-text`, `zero-shot-image-classification`), the schema is genuinely useful here because users likely passed `-m foo.onnx` instead of the required role form. For single-encoder tasks it just sends them on a wild goose chase.

Two options:

1. Soften the wording: `"Run 'winml eval --schema --task {task}' to see expected inputs and dataset schema."`
2. Only suggest `--schema` for tasks where `TASK_SCHEMAS[task].roles is not None`; for the rest, point at the model docs or omit the hint and let the wrapped exception speak for itself.

src/winml/modelkit/commands/eval.py
```diff
@@ -166,25 +161,38 @@ def eval(
...
-        winml eval -m model.onnx --model-id Intel/bert-base-uncased-mrpc \\
-            --dataset glue --dataset-name mrpc \\
-            --column input_column=sentence1
+    Run `winml eval --schema --task <task>` to see the dataset columns
+    and options expected by each task.
     """
+    # ── 0. --schema fast path: served from a local lightweight schema table
+    #       so this branch does not import the heavy winml.modelkit.eval package.
+    if show_schema:
+        task_arg: str | None = ctx.params.get("task")
```
timenick: **Non-idiomatic: reads `--task` via `ctx.params.get("task")` when it's already a function parameter.**

The `eval()` function signature already binds `task: str | None` (line 152). Going through `ctx.params.get("task")` instead of using `task` directly is slightly fragile: if the option name (or the function parameter name) ever changes, `ctx.params.get("task")` will silently start returning `None` and the `--schema` no-arg branch will accidentally fire for every call.

Suggested:

```python
if show_schema:
    if task is None:
        ...
    schema = TASK_SCHEMAS.get(task)
    ...
```

(Same applies to `task_arg` on the next line.)

src/winml/modelkit/eval/evaluate.py
```diff
@@ -31,6 +33,7 @@
     from ..models.winml.base import WinMLPreTrainedModel
 
 logger = logging.getLogger(__name__)
+console = Console()
```
timenick: **Module-level `Console()` is an import-time side effect.**

A module-level `console = Console()`:

1. Snapshots terminal width / color support at **import time** rather than at print time — wrong if the env changes (e.g. tests that set `COLUMNS`, redirection that happens after import).
2. Makes `print_config()` harder to test: there's no clean way to redirect or capture output without monkey-patching the module global.
3. Adds a (small) cost to every importer of `winml.modelkit.eval.evaluate`, even when `print_config` is never called.

Prefer one of:

```python
# Option A: lazy in print_config
def print_config(config: WinMLEvaluationConfig) -> None:
    console = Console()
    ...

# Option B: explicit parameter (testable)
def print_config(config, console: Console | None = None) -> None:
    console = console or Console()
    ...
```

src/winml/modelkit/eval/evaluate.py
```diff
@@ -230,19 +244,57 @@ def evaluate(config: WinMLEvaluationConfig) -> EvalResult:
...
-            "(or equivalent fields in the eval config) are ignored — the keys "
-            "below take precedence:\n%s",
+            "--dataset not specified; attempting default dataset '%s' for task '%s'. "
+            "Any --split / --column / --streaming / --dataset-name options are ignored.",
+            config.dataset.path,
             config.task,
-            _json.dumps(default, indent=2),
         )
 
+    print_config(config)
```
timenick, line 245: **`print_config(config)` runs only after `_load_model()` succeeds — the worst case for UX.**

The friendly load-error wrap above ("Failed to load model '...'.") fires before `print_config` is reached. So the *exact* user who would benefit most from seeing what config they were running — the one whose model didn't load — never sees it.

Move `print_config(config)` to **before** the `_load_model()` try block (right after `_resolve_task`, where `task` is finalized). That way the user always sees the effective config first, then any error.

src/winml/modelkit/commands/eval.py
```diff
@@ -31,13 +32,7 @@
...
     multiple=True,
     default=(),
-    help=(
-        "Model to evaluate. Accepts three forms: "
-        "(1) HuggingFace model ID, e.g. `-m <hf_model_id>`. "
-        "(2) ONNX file path, e.g. `-m model.onnx` (requires --model-id). "
-        "(3) Composite / split-encoder model as repeated role=path pairs, "
-        "e.g. `-m image-encoder=vision.onnx -m text-encoder=text.onnx`."
-    ),
+    help="Model to evaluate. Accepts a HuggingFace model ID or an ONNX file path.",
```
timenick: **`-m / --model` help text drops the split-encoder form, but that form is still real and now strict.**

The simplification reads: `"Model to evaluate. Accepts a HuggingFace model ID or an ONNX file path."`

This loses two pieces of information that the runtime still enforces:

1. The split-encoder `-m role=path` form (still supported, only documented in `--schema` output now — but only for tasks that have `schema.roles`, so users won't necessarily think to look there).
2. The rule that `-m model.onnx` **requires** `--model-id` (existing, unchanged).

The new `_resolve_model_path` also adds a hard rule (`Cannot pass both -m <hf_id> and --model-id`) with no hint of it in the help.

Consider keeping the help concise but mentioning the ONNX-needs-`--model-id` constraint at minimum:

```python
help=(
    "Model to evaluate. Accepts a HuggingFace model ID, an ONNX file path "
    "(requires --model-id), or split-encoder `role=path` pairs (see --schema)."
),
```

timenick (commented):
Reviewed the eval-UX refactor end-to-end. Direction looks great. Seven findings, ordered roughly by impact (substantive correctness/design first, then minor/style).

src/winml/modelkit/eval/evaluate.py
zhenchaoni: Good catch — narrowed the catch as suggested. The orchestrator now has a dedicated `except DatasetValidationError` branch that emits the schema-mismatch hint, plus `except (KeyError, ValueError)` for the remaining dataset/label-shape failures from HF that don't carry our custom exception. Everything else (`AttributeError`, `TypeError`, transformers internals, OOM, etc.) now propagates and gets the existing `-v` traceback path in `commands/eval.py`.

While testing this I also pushed dataset-load errors deeper: `WinMLEvaluator.prepare_data` now wraps `load_dataset` / `load_from_disk` and re-raises as `DatasetValidationError("Failed to load dataset '<path>' (name=…, split='…'): …")`. So things like "wrong split for this dataset" or "dataset not found" now hit the dedicated `DatasetValidationError` branch with the actionable hint, instead of being relabelled as a generic compute failure.

src/winml/modelkit/utils/eval_utils.py
zhenchaoni: Went with option 2 for now (lightweight invariant test). Added `TestGetEvaluatorClass::test_evaluator_registry_matches_schema_tasks` in `tests/unit/eval/test_eval.py` that asserts `set(_EVALUATOR_REGISTRY) == set(TASK_SCHEMAS)` and prints both drift directions in the failure message.

Option 1 (single source of truth via `WinMLEvaluator.SCHEMA`) is the cleaner long-term answer; happy to follow up in a separate PR since it touches every evaluator class.

src/winml/modelkit/eval/evaluate.py
zhenchaoni: Reworded. The `_load_model` failure hint no longer claims `--schema` documents model inputs. It now reads:

> "Check `--model`, `--model-id`, `--task`, device, and EP settings. For composite models, run `winml eval --schema --task <task>` to see supported `role=path` model options."

So users on split-encoder tasks still get pointed at the genuinely useful `--schema` output, and single-encoder users aren't sent on a wild goose chase.

src/winml/modelkit/commands/eval.py
zhenchaoni: Switched to using the bound `task` parameter directly: `task_arg = task` inside the `show_schema` branch, then `schema = TASK_SCHEMAS.get(task_arg)`. The `ctx.params.get(...)` indirection is gone.

src/winml/modelkit/eval/evaluate.py
zhenchaoni: Removed the module-level `console = Console()`. `print_config` now constructs its own `Console()` lazily (`output_console = Console()` inside the function), so terminal width / color is sampled at print time and importers of `winml.modelkit.eval.evaluate` no longer pay for it.

src/winml/modelkit/eval/evaluate.py
zhenchaoni: Moved `print_config(config)` to run right after `_resolve_task` (and the default-dataset resolution), **before** the `_load_model` try block. The user now always sees the effective config first, then any error — including the "Failed to load model …" wrap.

src/winml/modelkit/commands/eval.py
zhenchaoni: Help text restored to mention both constraints:

```python
help=(
    "Model to evaluate. Accepts a HuggingFace model ID, an ONNX file path "
    "(requires --model-id), or split-encoder role=path pairs (see --schema)."
),
```

Added `TestEvalHelp::test_model_help_mentions_onnx_model_id_and_role_path` in `tests/unit/commands/test_eval.py` to lock the two key phrases (`"requires --model-id"` and `"role=path"`) into the `--help` output so the simplification doesn't silently regress again.

timenick (approved):

timenick (approved):

PR #695, opened 2026-05-21
Revert "Fix onnxruntime search dll path."

chinazhangchao:
Reverts microsoft/winml-cli#689


timenick (approved):

DingmaomaoBJTU (commented):
**Overall:** This cleanly reverts PR #689 (the _preload_bundled_onnxruntime_dll workaround). The change itself is mechanically sound — removes the function, its call site, and the associated unit tests.

**Key concern:** The original PR #689 was introduced to fix a real issue: Windows System32 ships an older onnxruntime.dll that causes version mismatch errors when WinML EP plugins load the DLL by base name. If this revert lands, what alternative mechanism now prevents the problem? A few possibilities:
1. The ORT wheel fixed its own DLL search order
2. The WinML EP catalog loading was updated to use full paths
3. The system32 DLL was updated

Without clarity on which of these (or another) fix supersedes #689, this revert could regress the user experience on Windows machines with the System32 copy.

**Suggestion:** Add a note in the PR description explaining why the preload is no longer needed, or link to the upstream fix that makes it safe to remove.

xieofxie:
Just close as we kept it

PR #698, opened 2026-05-21
test(e2e): expand winml perf coverage across EPs and devices

### LINKED ISSUE #502: Identify functional scenarios and implement E2E tests for `winml perf`
### Description
Identify the functionality list for `winml perf` and implement E2E pytest cases covering it. Tests live in [tests/e2e](../tree/main/tests/e2e).

Coverage must include:

### Deliverables

### Acceptance Criteria
- `uv run pytest tests/e2e/ -k perf` is green.

### LINKED ISSUE #688: EP discovery should be resilient to per-provider failures
### Summary

[`WinMLEPRegistry._load_ep_catalog`](src/winml/modelkit/session/ep_registry.py#L74-L86) calls `provider.ensure_ready()` in a loop without per-provider error handling. The outer `try/except` in [`_discover_eps`](src/winml/modelkit/session/ep_registry.py#L61-L72) catches at the loop level, so a single failing provider sinks the whole discovery: `_winml_available = False`, `_ep_paths = {}`, downstream sees no EPs at all.

### Repro on a system with legacy + current EP framework packages installed side-by-side

`winml sys` (and any code path going through `WinMLEPRegistry`) logs:

```
WARNING  WinML EP discovery failed: [WinError -2140733418] Package dependency criteria could not be resolved
```

`EpCatalog.find_all_providers()` returns 3 entries; the third is a duplicate `NvTensorRTRTXExecutionProvider` backed by a legacy framework package that the current `windowsml` runtime can't mount. The first two (`OpenVINOExecutionProvider`, `NvTensorRTRTXExecutionProvider`) are usable but get discarded together with the bad one.

### Root cause

`windowsml` runtime issue (tracked separately with the WinML team) — `EpCatalog` emits duplicate-named entries for EPs whose legacy framework package is still registered, even though the current runtime can't mount them.

### Workaround in ModelKit

Per-provider try/except in [src/winml/modelkit/session/ep_registry.py:81-86](src/winml/modelkit/session/ep_registry.py#L81-L86):

```python
for provider in providers:
    try:
        provider.ensure_ready()
    except OSError as e:
        logger.debug("Skip EP %s: 0x%08X %s",
                     getattr(provider, "name", "<?>"),
                     e.winerror & 0xFFFFFFFF, e.strerror)
        continue
    lib = provider.library_path
    if not lib:
        continue
    self._ep_paths[cast("EPName", provider.name)] = lib
```

`logger.debug` rather than `warning` — duplicate/legacy entries are an expected transient state during the publisher migration. Keep the outer `warning` for catalog-creation/enumeration failures.

Remove once upstream stops emitting un-mountable entries.

xieofxie:
## Summary
- Expands `tests/e2e/test_perf_e2e.py` with parametrized `--ep` tests across `qnn`, `vitisai`, `openvino`, `dml`, `nv_tensorrt_rtx`, `migraphx`, combined with `--device cpu/gpu/npu`. Adds a CPU `--monitor` test and a new `TestPerfHuggingFace` class that runs the same EP/device matrix against `microsoft/resnet-50` loaded via the perf command.
- Refactors duplicated CLI-invocation and result-assertion code into `_build_perf_args`, `_assert_hw_monitor_section`, and `_assert_monitor_result` helpers so each scenario is a few lines instead of a hand-rolled argv list.
- Hardens `WinMLEPRegistry`: when a provider's `ensure_ready()` raises, log a warning and skip it instead of aborting registry initialization, so one broken EP no longer prevents the rest from being discovered.
- `tests/e2e/require_ep.py`: short-circuit `DmlExecutionProvider` alongside `CPUExecutionProvider` since DML ships with ORT and is not enumerated by `WinMLEPRegistry`, so `require_ep("dml")` no longer spuriously skips.
- `WinMLSession`: drop the `WINMLSESSION_VERBOSE` env-var gating; "Compiling for device" / "Already compiled" now log unconditionally at INFO.
- `test_benchmark_auto`: accept that `--device auto` resolves to a concrete `gpu`/`npu` in the output, since `resolve_device` no longer leaves the literal `"auto"` in `benchmark_info`.

Closes #502
Closes https://github.com/microsoft/winml-cli/issues/688

### DIFF
--- a/src/winml/modelkit/session/ep_registry.py
+++ b/src/winml/modelkit/session/ep_registry.py
@@ -79,7 +79,11 @@ class WinMLEPRegistry:
         providers = self._catalog.find_all_providers()
 
         for provider in providers:
-            provider.ensure_ready()
+            try:
+                provider.ensure_ready()
+            except Exception as e:
+                logger.warning("Failed to ensure EP %s is ready: %s", provider.name, e)
+                continue
             if provider.library_path == "":
                 continue
             self._ep_paths[cast("EPName", provider.name)] = provider.library_path
--- a/src/winml/modelkit/session/session.py
+++ b/src/winml/modelkit/session/session.py
@@ -195,12 +195,12 @@ class WinMLSession:
         """
         # If already compiled, ignore (idempotent)
         if self._session is not None:
-            logger.info("Already compiled for %s", self._device)
+            logger.debug("Already compiled for %s", self._device)
             return
 
         target_device = self._device
 
-        logger.info("Compiling for device: %s", target_device)
+        logger.debug("Compiling for device: %s", target_device)
 
         # Determine model path (original or EPContext)
         ctx_path = self._onnx_path.parent / f"{self._onnx_path.stem}_{target_device}_ctx.onnx"
--- a/src/winml/modelkit/session/session.py
+++ b/src/winml/modelkit/session/session.py
@@ -7,7 +7,6 @@
 from __future__ import annotations
 
 import logging
-import os
 from contextlib import contextmanager
 from enum import Enum
 from pathlib import Path
@@ -196,14 +195,12 @@ class WinMLSession:
         """
         # If already compiled, ignore (idempotent)
         if self._session is not None:
-            if self._is_verbose():
-                logger.info("Already compiled for %s", self._device)
+            logger.info("Already compiled for %s", self._device)
             return
 
         target_device = self._device
 
-        if self._is_verbose():
-            logger.info("Compiling for device: %s", target_device)
+        logger.info("Compiling for device: %s", target_device)
 
         # Determine model path (original or EPContext)
         ctx_path = self._onnx_path.parent / f"{self._onnx_path.stem}_{target_device}_ctx.onnx"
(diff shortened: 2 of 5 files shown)


src/winml/modelkit/session/ep_registry.py
```diff
@@ -79,7 +79,11 @@ def _load_ep_catalog(self) -> None:
         providers = self._catalog.find_all_providers()
 
         for provider in providers:
-            provider.ensure_ready()
+            try:
```
xieofxie, line 82: since we try-catch register, we should also try-catch ensure_ready to not block user

src/winml/modelkit/session/session.py
```diff
@@ -367,10 +364,6 @@ def __del__(self) -> None:
         except Exception:
             pass  # Suppress errors during interpreter shutdown
 
-    def _is_verbose(self) -> bool:
-        """Check if verbose logging is enabled via environment variable."""
-        return os.getenv("WINMLSESSION_VERBOSE", "").lower() in ("1", "true", "yes")
```
xieofxie, line 372: not meaningful

timenick (approved):

PR #699, opened 2026-05-21
Implement suggestion for command typo

### LINKED ISSUE #508: [winml] [P1] `winml exprt` (typo) returns "No such command" with no suggestion
**Severity**: P1  
**Command**: `winml`

---
**Category**: Did-you-mean — typos on subcommand names should suggest the closest valid name.

**Repro**:
```
uv run winml exprt
```

**Actual**:
```
Usage: winml [OPTIONS] COMMAND [ARGS]...
Try 'winml --help' for help.

Error: No such command 'exprt'.
```
Exit 2.

**Expected**: Suggest the closest match, the way `git`, `gh`, `cargo`, and `kubectl` do.
```
Error: No such command 'exprt'. Did you mean 'export'?
```

**Why it matters**: First-time users will mistype subcommand names. A bare "no such command" forces them back to `winml --help`; a one-line suggestion keeps them moving. Implementation is a one-method override of `LazyGroup.resolve_command` using `difflib.get_close_matches`.

zhenchaoni:
Fix #508 

### Before

```
$ winml exprt
Usage: winml [OPTIONS] COMMAND [ARGS]...
Try 'winml --help' for help.

Error: No such command 'exprt'.
```

The bare error gives no recourse — users had to re-read `winml --help` to find the intended name. This forced a round-trip for what is almost always a one-character typo.

### After

```
$ winml exprt
Usage: winml [OPTIONS] COMMAND [ARGS]...
Try 'winml --help' for help.

Error: No such command 'exprt'. Did you mean 'export'?
```

### Fix

Click 8.4 ships a built-in typo suggester: `Group.resolve_command` raises `NoSuchCommand(cmd_name, possibilities=self.commands, ctx=ctx)`, and `NoSuchCommand` runs `difflib.get_close_matches` over the keys of `possibilities` to append `Did you mean 'X'?` to the error message.

The catch: `LazyGroup` discovers subcommands by scanning `src/winml/modelkit/commands/*.py` at runtime (via `list_commands`) and never eagerly populates `self.commands`. So Click had nothing to compare against.

This PR is a 3-line adapter on top of Click's built-in path:

```python
def resolve_command(self, ctx: click.Context, args: list[str]):
    """Seed ``self.commands`` so Click can emit a did-you-mean hint on typos."""
    # Click's NoSuchCommand exception uses self.commands to find suggestions.
    for name in self.list_commands(ctx):
        self.commands.setdefault(name, None)  # type: ignore[arg-type]
    return super().resolve_command(ctx, args)

### DIFF
--- a/src/winml/modelkit/cli.py
+++ b/src/winml/modelkit/cli.py
@@ -204,38 +204,12 @@ class LazyGroup(ActionGroup):
                 discovered = attr
         return discovered
 
-    def resolve_command(
-        self,
-        ctx: click.Context,
-        args: list[str],
-    ) -> tuple[str | None, click.Command | None, list[str]]:
-        """Resolve subcommand; append a did-you-mean hint on unknown names.
-
-        Click's default ``No such command 'X'.`` is unhelpful on typos. Mirror
-        the UX of ``git``/``gh``/``cargo``/``kubectl`` by computing the closest
-        known command name with :func:`difflib.get_close_matches` and rewriting
-        the error message in-place. When no close match exists, the original
-        error is re-raised untouched (issue #508).
-        """
-        try:
-            return super().resolve_command(ctx, args)
-        except click.exceptions.UsageError as exc:
-            bad = args[0] if args else ""
-            if (
-                bad
-                and exc.message
-                and f"No such command '{bad}'" in exc.message
-            ):
-                known_commands = self.list_commands(ctx)
-                if bad not in known_commands:
-                    import difflib
-                    matches = difflib.get_close_matches(bad, known_commands, n=1)
-                    if matches:
-                        exc.message = (
-                            f"No such command '{bad}'. "
-                            f"Did you mean '{matches[0]}'?"
-                        )
-            raise
+    def resolve_command(self, ctx: click.Context, args: list[str]):
+        """Seed ``self.commands`` so Click can emit a did-you-mean hint on typos."""
+        # Click's NoSuchCommand exception uses self.commands to find suggestions.
+        for name in self.list_commands(ctx):
+            self.commands.setdefault(name, None)  # type: ignore[arg-type]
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: src/winml/modelkit/cli.py [hunk 1]
```diff
@@ -204,6 +204,39 @@ def get_command(self, ctx: click.Context, cmd_name: str) -> click.Command | None
...
+        except click.exceptions.UsageError as exc:
+            bad = args[0] if args else ""
+            if (
+                bad
+                and exc.message
+                and f"No such command '{bad}'" in exc.message
+            ):
+                known_commands = self.list_commands(ctx)
+                if bad not in known_commands:
+                    import difflib
```
xieofxie: does click has built-in function for this?

zhenchaoni: I hope so but no, it doesn't have a built-in way.

xieofxie (approved):

src/winml/modelkit/cli.py
zhenchaoni: I actually find a later `click` package supports this. I updated the code.

src/winml/modelkit/cli.py
```diff
@@ -204,6 +204,13 @@ def get_command(self, ctx: click.Context, cmd_name: str) -> click.Command | None
                 discovered = attr
         return discovered
 
+    def resolve_command(self, ctx: click.Context, args: list[str]):
+        """Seed ``self.commands`` so Click can emit a did-you-mean hint on typos."""
+        # Click's NoSuchCommand exception uses self.commands to find suggestions.
+        for name in self.list_commands(ctx):
+            self.commands.setdefault(name, None)  # type: ignore[arg-type]
```
xieofxie, line 211: since it will only resolve once, so I suppose this is OK?

xieofxie (approved):

PR #702, opened 2026-05-22
Clean up winml analyze: EP/device selection, op-check skipped UI, merged optim config

### LINKED ISSUE #697: [winml analyze] Output need improvement in UX
Image

Image

Above pics results are incorrect

WARNING: Multiple EP/device targets selected; writing optimization config for the first target: QNNExecutionProvider (NPU)

Test machine does not have qnn device, it should not write config for qnn

vortex-captain:
Fixes #697.

## Summary

- **EP/device selection** — replace the large mode-matrix branching with a straightforward `eps × devices` construction filtered against `EP_SUPPORTED_DEVICES`. When `device=auto and ep=auto`, restrict to locally-available pairs; when only one of them is auto, keep all pairs but warn about ones not supported in the local environment instead of silently dropping them.
- **Op-check skipped table** — render a title-only "Skipped - no rule data" table for EP/device pairs with no rule data, so the per-pair output stays consistent.
- **Analysis summary** — fix a key-shape mismatch where `ep_instance_counts` was being rewrapped with `"EP@DEVICE"` string keys while the renderer looked them up by `(ep, device)` tuple, which always missed and printed `0/0/0`. Pass the tuple-keyed dict directly.
- **Counts format** — always show four numbers `S/P/U/Unk` (Supported / Partial / Unsupported / Unknown) in both the OP CHECK table column and the ANALYSIS SUMMARY line. Update column header and legend.
- **Optimization config** — when multiple analysis pairs are selected, merge their `get_optimization_config(ep=…)` outputs into a single config (union of keys). On conflicting values for the same key, log a warning naming each pair's value; the merged config keeps the first pair's value.
- Drop the now-unused `pair_hints` plumbing and redundant local-vs-rule hint lines.

### DIFF
--- a/src/winml/modelkit/commands/analyze.py
+++ b/src/winml/modelkit/commands/analyze.py
@@ -183,10 +183,6 @@ def _build_analysis_table(
 
     if op_check_skipped:
         title += "  Skipped - no rule data"
-    elif complete:
-        title += "  [bold green]✅ Complete[/bold green]"
-
-    if op_check_skipped:
         table = Table(
             title=title,
             show_header=False,
@@ -196,8 +192,12 @@ def _build_analysis_table(
             expand=False,
             width=80,
         )
+        # add_column is required even though no rows are added — without it the
+        # empty table doesn't render the centered title.
         table.add_column("")
         return table
+    if complete:
+        title += "  [bold green]✅ Complete[/bold green]"
 
     # Build display order: all_ops sorted by count, or just data if no all_ops
     if all_ops:
@@ -770,6 +770,7 @@ def analyze(
             for candidate_device in devices
             if candidate_device.lower() in EP_SUPPORTED_DEVICES[candidate_ep]
         ]
+        execution_pairs = _sort_ep_device_pairs(execution_pairs)
 
         local_pairs = set(_get_local_ep_device_pairs())
 
@@ -779,18 +780,16 @@ def analyze(
             unsupported_pairs = [pair for pair in execution_pairs if pair not in local_pairs]
             if unsupported_pairs:
                 logger.warning(
-                    "device is set to auto (inferred from local availability) but ep is "
-                    "not; some (ep, device) pairs to "
-                    "analyze are not supported in local environment: %s",
+                    "--device auto resolves from local availability, but --ep is pinned;"
+                    " the following pairs are not available on this machine: %s",
                     ", ".join(_ep_name_device_display_name(e, d) for e, d in unsupported_pairs),
                 )
         elif ep == "auto":
             unsupported_pairs = [pair for pair in execution_pairs if pair not in local_pairs]
             if unsupported_pairs:
                 logger.warning(
-                    "ep is set to auto (inferred from local availability) but device is "
-                    "not; some (ep, device) pairs to "
-                    "analyze are not supported in local environment: %s",
+                    "--ep auto resolves from local availability, but --device is pinned;"
+                    " the following pairs are not available on this machine: %s",
                     ", ".join(_ep_name_device_display_name(e, d) for e, d in unsupported_pairs),
                 )
 
@@ -1171,9 +1170,9 @@ def analyze(
 
                 per_pair_values: dict[str, list[tuple[tuple[str, str], object]]] = {}
                 for (target_ep, target_device), run_result in zip(
-                    execution_pairs, analysis_results, strict=False
+                    execution_pairs, analysis_results, strict=True
                 ):
-                    pair_config = run_result.get_optimization_config(ep=target_ep)
+                    pair_config = run_result.get_optimization_config(ep=target_ep).to_dict()
                     for key, value in pair_config.items():
                         per_pair_values.setdefault(key, []).append(
                             ((target_ep, target_device), value)
--- a/src/winml/modelkit/commands/analyze.py
+++ b/src/winml/modelkit/commands/analyze.py
@@ -473,13 +473,13 @@ def _render_analysis_summary(
 
         # List ops by non-white support level (skip when op check was skipped \u2014
         # the classification would be all-unknown with no useful detail).
+        _issue_sections = [
+            (SupportLevel.UNSUPPORTED, "red", "\u26d4 Unsupported"),
+            (SupportLevel.PARTIAL, "yellow", "\u26a0\ufe0f  Partial"),
+            (SupportLevel.UNKNOWN, "bright_black", "\u2753 Unknown"),
+        ]
+        classification = ep_support.classification
         if not op_check_skipped:
-            classification = ep_support.classification
-            _issue_sections = [
-                (SupportLevel.UNSUPPORTED, "red", "\u26d4 Unsupported"),
-                (SupportLevel.PARTIAL, "yellow", "\u26a0\ufe0f  Partial"),
-                (SupportLevel.UNKNOWN, "bright_black", "\u2753 Unknown"),
-            ]
             for level, color, heading in _issue_sections:
                 ops = classification.get(level, [])
                 if ops:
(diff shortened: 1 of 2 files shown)


timenick:
Does this PR address the issue #697 ?

DingmaomaoBJTU (commented):
Overall this is a solid cleanup — the EP×device product approach is much easier to reason about than the old mode-matrix branching, and the  key-shape fix is a genuine bug. A few things to check before merging:

src/winml/modelkit/commands/analyze.py
```diff
@@ -175,7 +175,30 @@ def _build_analysis_table(
...
+        table = Table(
+            title=title,
+            show_header=False,
+            header_style="bold",
+            box=None,
+            padding=(0, 1),
+            expand=False,
+            width=80,
+        )
+        table.add_column("")
```
DingmaomaoBJTU, line 197: is a no-op here:  and no rows are ever added to this table, so the column has no effect. Can be removed.

src/winml/modelkit/commands/analyze.py
```diff
@@ -175,7 +175,30 @@ def _build_analysis_table(
...
+    title = "📊 OP CHECK"
+    if ep_device_pair_display_name:
+        title += f" — [bold cyan]{ep_device_pair_display_name}[/bold cyan]"
+
+    if op_check_skipped:
+        title += "  Skipped - no rule data"
+    elif complete:
+        title += "  [bold green]✅ Complete[/bold green]"
+
+    if op_check_skipped:
```
DingmaomaoBJTU: Two consecutive `if op_check_skipped:` checks (lines 184 and 189) is a minor readability bump. The title suffix and the early-return can be merged into one branch:

```python
if op_check_skipped:
    title += "  Skipped - no rule data"
    table = Table(title=title, show_header=False, box=None, padding=(0, 1), expand=False, width=80)
    return table
if complete:
    title += "  [bold green]✅ Complete[/bold green]"
```

src/winml/modelkit/commands/analyze.py
```diff
@@ -717,196 +718,61 @@ def analyze(
...
+            execution_pairs = [pair for pair in execution_pairs if pair in local_pairs]
+        elif device == "auto":
+            unsupported_pairs = [pair for pair in execution_pairs if pair not in local_pairs]
+            if unsupported_pairs:
+                logger.warning(
+                    "device is set to auto (inferred from local availability) but ep is "
+                    "not; some (ep, device) pairs to "
+                    "analyze are not supported in local environment: %s",
+                    ", ".join(_ep_name_device_display_name(e, d) for e, d in unsupported_pairs),
                 )
```
DingmaomaoBJTU, line 786: Behavioral change worth calling out: when `--device auto` is set (devices inferred from local availability) but `--ep` is pinned to a specific value, the code now **warns** about non-local pairs but still runs analysis on them. The old code would `sys.exit(2)` in this case. Is silently proceeding on pairs the local machine can't execute the intended behavior? If so, what does the analysis output for those pairs look like — will they produce meaningful results?

Also, the warning message text reads a bit awkwardly due to the mid-phrase line split:
```
"device is set to auto (inferred from local availability) but ep is "
"not; some (ep, device) pairs to "
"analyze are not supported in local environment: %s"
```
These three strings are concatenated at compile time into one log message. Consider phrasing it as a single coherent sentence, e.g.:
```python
"--device auto resolves from local availability, but --ep is pinned; "
"the following pairs are not available on this machine: %s"
```

src/winml/modelkit/commands/analyze.py
```diff
@@ -1272,19 +1136,39 @@ def on_node_result(pattern_runtime):
             import json
 
             try:
-                config_ep = execution_pairs[0][0]
-                if len(execution_pairs) > 1:
+                # Merge optimization configs from all execution pairs; warn on conflicts.
+
+                per_pair_values: dict[str, list[tuple[tuple[str, str], object]]] = {}
+                for (target_ep, target_device), run_result in zip(
+                    execution_pairs, analysis_results, strict=False
```
DingmaomaoBJTU: `strict=False` silently truncates if `execution_pairs` and `analysis_results` have different lengths (e.g., if an analysis run fails partway through). These two lists should always be in 1:1 correspondence, so `strict=True` would catch any mismatch as a bug rather than silently omitting pairs from the merged config.

src/winml/modelkit/commands/analyze.py
```diff
@@ -1272,19 +1136,39 @@ def on_node_result(pattern_runtime):
...
+                        key,
+                        detail,
+                        merged[key],
                     )
-                config = analysis_results[0].get_optimization_config(ep=config_ep)
+
+                merged = dict(sorted(merged.items()))
                 optim_config.parent.mkdir(parents=True, exist_ok=True)
-                optim_config.write_text(json.dumps(config.to_dict(), indent=2), encoding="utf-8")
+                optim_config.write_text(json.dumps(merged, indent=2), encoding="utf-8")
```
DingmaomaoBJTU, line 1201: The old code called `config.to_dict()` before passing to `json.dumps`. The new code writes `merged` directly, where values come from `pair_config.items()`. If `get_optimization_config` returns an object with a `.to_dict()` method (rather than a plain `dict`), the values in `merged` may not be JSON-serializable and this will raise a `TypeError` at runtime. Please confirm whether `pair_config` is already a plain dict, or add a serialization step (e.g., call `.to_dict()` on values that need it).

src/winml/modelkit/commands/analyze.py
```diff
@@ -717,196 +718,61 @@ def analyze(
...
+        else:
+            # ep is a specific EP or alias
+            eps = [normalize_ep_name(ep)]
+
+        execution_pairs = [
+            (candidate_ep, candidate_device)
+            for candidate_ep in eps
+            for candidate_device in devices
+            if candidate_device.lower() in EP_SUPPORTED_DEVICES[candidate_ep]
+        ]
```
DingmaomaoBJTU, line 772: The old code had a `_sort_ep_device_pairs` helper that was called in many places to produce a stable ordering. The new code sorts `devices` (line 729) but the final `execution_pairs` list inherits the ordering of `eps` (which comes from `_get_available_eps()` / `SUPPORTED_EPS` with no further sort). Is the ordering of `execution_pairs` stable and deterministic across runs? If `_sort_ep_device_pairs` is now dead code, it should be deleted.

src/winml/modelkit/commands/analyze.py
```diff
@@ -1197,25 +1066,20 @@ def on_node_result(pattern_runtime):
...
-                        for (ep_name, device), counts in ep_instance_counts.items()
-                    }
                     _render_analysis_summary(
                         console,
                         result.output.results,
-                        summary_ep_instance_counts,
+                        ep_instance_counts,
                         ep_patterns=ep_patterns,
                         ep=target_ep,
                         device=target_device,
```
DingmaomaoBJTU, line 1102: Nice catch. The old `summary_ep_instance_counts` was rebuilding the dict with `"EP@DEVICE"` string keys but `_render_analysis_summary` looked them up by `(ep, device)` tuple, so the lookup always missed and printed `0/0/0` for every pair. Passing `ep_instance_counts` directly fixes that.

vortex-captain:
> Does this PR address the issue #697 ?

Yes

src/winml/modelkit/commands/analyze.py
vortex-captain: Actually, without `table.add_column("")`, the empty table (which we need to display the title centered) will not show

timenick (approved):

src/winml/modelkit/commands/analyze.py
vortex-captain: when user sets `device` to `"auto"` but not `ep`, the 2 args are somehow contradictory in the first place, and I think it would be sensible to run locally unsupported (device, ep) pairs while warning the user. In addition, the analysis in such case is still meaningful, since the analysis does not require running nodes locally (unless --run-unknown-op is set, but in this case, the flag is resolved to `False` for unsupported pairs in `_resolve_run_unknown_op`)

src/winml/modelkit/commands/analyze.py
vortex-captain: fixed

src/winml/modelkit/commands/analyze.py
vortex-captain: warning string fixed

src/winml/modelkit/commands/analyze.py
vortex-captain: fixed by adding `.to_dict()` after `run_result.get_optimization_config(ep=target_ep)`

src/winml/modelkit/commands/analyze.py
vortex-captain: now `execution_pairs` is sorted by `_sort_ep_device_pairs`

src/winml/modelkit/commands/analyze.py
vortex-captain: fixed

timenick (approved):

PR #705, opened 2026-05-22
fix(sys): report host arch via IsWow64Process2 on Windows ARM64

### LINKED ISSUE #704: winml sys reports Machine: AMD64 on Windows ARM64 hosts
## Bug
`winml sys` reports `Machine: AMD64` on Windows ARM64 hosts (Snapdragon X with Prism emulation).

## Cause
The project venv uses x64 Python because `torch` has no Windows-ARM64 wheel. `platform.machine()` returns the **process** architecture, not the host. `PROCESSOR_ARCHITEW6432` is empty on Snapdragon X — Prism emulation doesn't set it — so reading that env var doesn't help either.

## Fix
Use the Win32 `IsWow64Process2` API to read the host machine type, falling back to `platform.machine()` if the call fails.

## Repro
On a Snapdragon X dev box using the project's default x64 venv:
- Before: `Machine: AMD64`
- After:  `Machine: ARM64`

timenick:
## Summary
- Fixes #704 — `winml sys` showed `Machine: AMD64` on Windows ARM64 hosts.
- `platform.machine()` returns the process arch, so an x64 Python on ARM64 (under Prism emulation) misreports the host. `PROCESSOR_ARCHITEW6432` is empty on Snapdragon X, so it's not a reliable fallback.
- Added `_get_windows_native_machine()` using the Win32 `IsWow64Process2` API to read the native machine type; falls back to `platform.machine()` if the call fails.
- Added unit coverage for the emulation case and the fallback path; updated existing Windows tests to mock the new helper so they stay deterministic across host architectures.

## Verification
- `winml sys` on a Snapdragon X box now shows `Machine: ARM64` (was `AMD64`).
- `uv run pytest tests/unit/sysinfo/test_sysinfo.py` → 14 passed.
- `ruff check` clean.

### DIFF
--- a/src/winml/modelkit/commands/sys.py
+++ b/src/winml/modelkit/commands/sys.py
@@ -23,6 +23,8 @@ Usage:
 
 from __future__ import annotations
 
+import ctypes
+import ctypes.wintypes
 import json
 import logging
 import platform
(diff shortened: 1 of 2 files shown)


xieofxie (approved):
if you tested, just approved

PR #708, opened 2026-05-22
feat: use resolve device in session creation

### LINKED ISSUE #656: perf: mixed resolved device and input device (auto) are displayed
Maybe resolve the AOT JIT issue first

```
winml perf -m "~\.cache\winml\artifacts\microsoft_resnet-50\imgcls_b112acc9e68a9479_optimized.onnx" --ep ov --iterations 1000 --monitor -v                                                       
Benchmarking ONNX: ~\.cache\winml\artifacts\microsoft_resnet-50\imgcls_b112acc9e68a9479_optimized.onnx
[2026-05-18T15:24:56] DEBUG: Found EP: OpenVINOExecutionProvider at C:\Program Files\WindowsApps\MicrosoftCorporationII.WinML.Intel.OpenVINO.EP.1.8_1.8.79.0_x64__8wekyb3d8bbwe\ExecutionProvider\onnxruntime_providers_openvino_plugin.dll
[2026-05-18T15:24:56] DEBUG: WinML EP discovery successful: ['OpenVINOExecutionProvider']
The requested API version [24] is not available, only API versions [1, 23] are supported in this build. Current ORT Version is: 1.23.5
The requested API version [24] is not available, only API versions [1, 23] are supported in this build. Current ORT Version is: 1.23.5
[2026-05-18T15:24:56] DEBUG: Registered EP: OpenVINOExecutionProvider -> C:\Program Files\WindowsApps\MicrosoftCorporationII.WinML.Intel.OpenVINO.EP.1.8_1.8.79.0_x64__8wekyb3d8bbwe\ExecutionProvider\onnxruntime_providers_openvino_plugin.dll
[2026-05-18T15:24:56] INFO: WinML EPs registered: ['OpenVINOExecutionProvider']
[2026-05-18T15:24:56] INFO: WinMLSession initialized: ~\.cache\winml\artifacts\microsoft_resnet-50\imgcls_b112acc9e68a9479_optimized.onnx
[2026-05-18T15:24:56] DEBUG: Loading ONNX model from ~\.cache\winml\artifacts\microsoft_resnet-50\imgcls_b112acc9e68a9479_optimized.onnx (weights=False)
[2026-05-18T15:24:56] DEBUG: Loaded ONNX model: 122 nodes, 108 initializers
[2026-05-18T15:24:56] DEBUG: Loaded value_ranges from ~\.cache\winml\artifacts\microsoft_resnet-50\imgcls_0a9c736dd54f206e_winml_build_config.json: {'pixel_values': [0, 1]}
[2026-05-18T15:24:56] INFO: Generated 'pixel_values': shape=[1, 3, 224, 224], dtype=float32
[2026-05-18T15:24:56] INFO: Explicit EP: ov (OpenVINOExecutionProvider) device=auto -> NPU
[2026-05-18T15:24:57] INFO: Session created with device auto, providers: ['OpenVINOExecutionProvider', 'CPUExecutionProvider']

Device:      auto / OpenVINOExecutionProvider
Inputs:      pixel_values         [1, 3, 224, 224]       float32
Outputs:     logits               [1, 1000]

[2026-05-18T15:24:57] DEBUG: Resolved NPU LUID via ORT (ep=OpenVINOExecutionProvider, ep_name=OpenVINOExecutionProvider): 0x00000000_0x00010193
[2026-05-18T15:24:57] INFO: Adapter query unavailable for NPU LUID 0x00000000_0x00010193 (Engine type 'Compute' not found on adapter 0x00000000_0x00010193. Available: ['Neural']); monitoring CPU/RAM only
[2026-05-18T15:24:57] INFO: No NPU/GPU found via PDH; monitoring CPU/RAM only
[2026-05-18T15:24:57] DEBUG: PdhPoller started (interval=200.0ms)
╭───────────────────────────────────────────────────────────────────────────────────────────────── HW Monitor - imgcls_b112acc9e68a9479_optimized.onnx ─────────────────────────────────────────────────────────────────────────────────────────────────╮
│   Utilization (██ CPU %)                                                                                                                                                                                                                              │
│    ┌───────────────────────────────────────────────────────────────────────────┐                                                                                                                                                                      │
│ 100┤                                                                           │                                                                                                                                                                      │
│    │                                                                           │                                                                                                                                                                      │
│  80┤                                                                           │                                                                                                                                                                      │
│    │                                                                           │                                                                                                                                                                      │
│  60┤                                                                           │                                                                                                                                                                      │
│    │         ⣀⢴ ⡾⡀⡀                                                            │                                                                                                                                                                      │
│  40┤        ⡜⠈⢸ ⡇⠙⡇                                                            │                                                                                                                                                                      │
│    │       ⢰⠁ ⢸⢠⠃ ⢇ ⢠⡆                                                         │                                                                                                                                                                      │
│  20┤      ⣴⢸  ⠘⠇  ⢸ ⡸⠘⠄                                                        │                                                                                                                                                                      │
│    │⡠⠔⠒⠣⢴⡠⠃⠻      ⠘⢴⠁                                                          │                                                                                                                                                                      │
│   0┤⡇    ⠁                                                                     │                                                                                                                                                                      │
│    └┬──────────────────┬─────────────────┬──────────────────┬─────────────────┬┘                                                                                                                                                                      │
│    0.0                2.5               5.0                7.5             10.0                                                                                                                                                                       │
│ Usage %                              Time (s)                                                                                                                                                                                                         │
│   [====================] 100%   |  Iter: 1000/1000  |  Device: auto                                                                                                                                                                                   │
│   CPU: 24.1%  |  Sys Mem: 42554 MB                                                                                                                                                                                                                    │
│   Latency: 2.41 ms        |  Throughput: ~415 smp/s                                                                                                                                                                                                   │
╰───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
[2026-05-18T15:25:03] DEBUG: PdhPoller stopped: 0 util samples, 0 local mem, 0 shared mem, 25 cpu samples

Device:      auto / OpenVINOExecutionProvider
Task:        n/a (direct ONNX)

Latency (ms)
┏━━━━━━┳━━━━━━┳━━━━━━┳━━━━━━┳━━━━━━━┳━━━━━━┳━━━━━━━┳━━━━━━┓
┃  Avg ┃  P50 ┃  P90 ┃  P95 ┃   P99 ┃  Min ┃   Max ┃  Std ┃
┡━━━━━━╇━━━━━━╇━━━━━━╇━━━━━━╇━━━━━━━╇━━━━━━╇━━━━━━━╇━━━━━━┩
│ 2.76 │ 2.46 │ 2.88 │ 3.14 │ 10.64 │ 1.62 │ 23.82 │ 1.58 │
└──────┴──────┴──────┴──────┴───────┴──────┴───────┴──────┘

Throughput: 362.24 samples/sec

Hardware (during benchmark)
  CPU: 24.1% avg
  Sys Mem: 42554 MB
```

### LINKED ISSUE #700: bug: if ep is not set and device is set (not auto) and that device doesn't support, we should exit(1)
for example in ModelKit\src\winml\modelkit\session\session.py

xieofxie:
Use resolve_device in session creation to align device ep resolution with other commands.

Currently we use a manual mapping in resolve_device / resolve_eps/

We could migrate to PREFER_XPU resolution with a temp model + real session later, but it is no harm to unify how we do it.

Closes https://github.com/microsoft/winml-cli/issues/700
Closes https://github.com/microsoft/winml-cli/issues/656

### DIFF
--- a/src/winml/modelkit/session/session.py
+++ b/src/winml/modelkit/session/session.py
@@ -384,10 +384,10 @@ class WinMLSession:
                     device,
                     resolved_device,
                 )
-                return opts
+                return opts, resolved_device, resolved_ep
 
         raise DeviceNotAvailableError(
-            message=f"No available device found for policy '{device}' and EP '{self._ep}'"
+            message=f"No available provider found for device '{device}' and EP '{self._ep}'"
         )
 
     def _validate_inputs(self, inputs: dict[str, Any]) -> None:
--- a/src/winml/modelkit/session/session.py
+++ b/src/winml/modelkit/session/session.py
@@ -162,7 +162,7 @@ class WinMLSession:
 
         # HF Pipeline may pass torch.device; coerce to string for downstream .lower() calls
         self._device = str(device) if not isinstance(device, str) else device
-        self._ep = ep if ep else None
+        self._ep = ep
         self._persist_jit = ep_config.enable_ep_context if ep_config else False
         self._embed_context = ep_config.embed_context if ep_config else False
         self._provider_options = ep_config.provider_options if ep_config else {}
@@ -199,6 +199,7 @@ class WinMLSession:
             return
 
         target_device = self._device
+        resolved_device = target_device
 
         logger.debug("Compiling for device: %s", target_device)
 
@@ -223,7 +224,7 @@ class WinMLSession:
                 logger.info("Model already compiled (EPContext), skipping ModelCompiler")
             else:
                 try:
-                    sess_options = self._build_session_options(target_device)
+                    sess_options, resolved_device, _ = self._build_session_options(target_device)
                     model_compiler = ort.ModelCompiler(
                         sess_options,
                         str(self._onnx_path),
@@ -245,7 +246,7 @@ class WinMLSession:
             # EP is either configured via add_provider_for_devices (WinML EP
             # registry, e.g. QNN) or left to ORT's device policy (fallback).
             # Never pass providers= — WinML-registered EPs don't support it.
-            sess_options = self._build_session_options(target_device)
+            sess_options, resolved_device, _ = self._build_session_options(target_device)
             session = ort.InferenceSession(str(model_path), sess_options=sess_options)
 
         except Exception as ep_err:
@@ -275,12 +276,8 @@ class WinMLSession:
 
         # Resolve device label from the primary provider ORT actually selected
         if self._device == "auto" and actual_providers:
-            from ..sysinfo.device import get_ep_device_map
-
-            ep_map = get_ep_device_map()
-            resolved = ep_map.get(actual_providers[0])
-            if resolved and "/" not in resolved:
-                self._device = resolved
+            self._device = resolved_device
+            logger.info("Auto-resolved device: %s", self._device)
 
     def run(
         self,
@@ -364,165 +361,34 @@ class WinMLSession:
         except Exception:
             pass  # Suppress errors during interpreter shutdown
 
-    def _build_session_options(self, device: str) -> ort.SessionOptions:
+    def _build_session_options(self, device: str) -> tuple[ort.SessionOptions, str, EPName]:
         """Build ORT SessionOptions from instance session_options and device.
 
-        When ``self._ep`` is set, uses ``add_provider_for_devices`` to
-        explicitly bind that EP — including ``"cpu"``, so the
-        CPUExecutionProvider isn't silently displaced by another CPU-capable
-        EP (e.g. OpenVINO) under PREFER_CPU policy.
-        When ``self._ep`` is not set, the path forks on device: ``"cpu"``
-        falls through to PREFER_CPU policy (skipping EP discovery so non-CPU
-        EPs aren't probed), while other devices query ``get_ep_devices()``
-        to discover an available EP. Policy-based selection is the
-        last-resort fallback.
-
-        Note: Returns a **fresh** SessionOptions when using explicit EP to
-        avoid "already registered" errors from repeated calls.
+        Returns a **fresh** SessionOptions.
         """
-        # CPU never needs EP binding — skip device discovery entirely so that
-        # non-CPU EPs (e.g. OpenVINO) are not probed via get_ep_devices(),
-        # which would trigger their native shared-library load and emit
-        # version-mismatch warnings even when the model runs on CPU.
-        # Exception: when an explicit EP is set (e.g. --ep openvino --device cpu,
-        # or --ep cpu --device cpu), fall through so the EP binding logic
-        # below can honour it.
-        if device.lower() == "cpu" and not self._ep:
-            opts = self._session_options
-            opts.set_provider_selection_policy(DEVICE_POLICY_MAP["cpu"])
-            logger.info("Using PREFER_CPU policy for device cpu")
-            return opts
+        from ..sysinfo.device import resolve_device, resolve_eps
+        from ..utils.constants import DEVICE_TO_DEVICE_TYPE, normalize_ep_name
 
-        # Explicit EP targeting: create fresh opts to avoid double-registration.
-        # When device is also specified (non-"auto"), narrow by both EP name
-        # and device type so e.g. `--ep qnn --device cpu` finds QNN-on-CPU
-        # instead of the first QNN ep_device (which may report as NPU).
-        # `--ep cpu` is honoured here too so the CPUExecutionProvider gets
-        # bound explicitly; otherwise PREFER_CPU policy lets ORT prefer
-        # OV-on-CPU (or any other registered CPU-capable EP) over the basic
-        # CPU EP, silently ignoring the user's --ep choice.
-        if self._ep:
-            from ..utils.constants import normalize_ep_name
+        resolved_device, _ = resolve_device(device, ep=self._ep)
+        resolved_ep = normalize_ep_name(self._ep) if self._ep else resolve_eps(resolved_device)[0]
+        device_type = DEVICE_TO_DEVICE_TYPE.get(resolved_device.upper())
 
-            target_name = normalize_ep_name(self._ep)
-            if target_name:
-                matched = self._find_ep_device(ep_name=target_name, device=device)
-                if matched:
-                    from ..utils.constants import DEVICE_TYPE_TO_DEVICE
-
-                    opts = ort.SessionOptions()
-                    opts.add_provider_for_devices([matched], self._provider_options)
-                    resolved = DEVICE_TYPE_TO_DEVICE.get(
-                        matched.device.type, str(matched.device.type)
-                    )
-                    logger.info(
-                        "Explicit EP: %s (%s) device=%s -> %s",
-                        self._ep,
-                        target_name,
-                        device,
-                        resolved,
-                    )
-                    return opts
-                logger.warning(
-                    "EP '%s' (%s) not found for device '%s'",
+        opts = ort.SessionOptions()
+        for ep_dev in ort.get_ep_devices():
+            if ep_dev.ep_name == resolved_ep and ep_dev.device.type == device_type:
+                opts.add_provider_for_devices([ep_dev], self._provider_options)
+                logger.info(
+                    "Create session with EP: %s (%s) device=%s -> %s",
+                    resolved_ep,
                     self._ep,
-                    target_name,
                     device,
+                    resolved_device,
                 )
-
-        # No explicit EP — discover available EP for this device type
-        if not self._ep and device.lower() != "cpu":
-            matched = self._find_ep_device(device=device)
-            if matched:
-                opts = ort.SessionOptions()
-                opts.add_provider_for_devices([matched], self._provider_options)
-                logger.info("Discovered EP for %s: %s", device, matched.ep_name)
                 return opts
 
-        # Policy-based selection (last resort)
-        opts = self._session_options
-        policy = DEVICE_POLICY_MAP.get(
-            device.lower(), ort.OrtExecutionProviderDevicePolicy.PREFER_NPU
+        raise DeviceNotAvailableError(
+            message=f"No available device found for policy '{device}' and EP '{self._ep}'"
         )
-        opts.set_provider_selection_policy(policy)
-        logger.info("Using provider selection policy %s for device %s", policy, device)
-
-        return opts
-
-    @staticmethod
-    def _find_ep_device(device: str, ep_name: EPName | None = None) -> Any:
-        """Find the first OrtEpDevice matching the given filters.
-
-        Behavior:
-            - ``ep_name`` set, ``device == "auto"`` → aggregate ep_devices
-              matching ``ep_name`` and pick the first one whose device type
-              appears earliest in that EP's preferred device list
-              (``get_ep_device_map()``).
-            - ``ep_name`` unset, ``device`` is a concrete type → aggregate
-              ep_devices matching that device type and pick the first one
-              whose EP name appears earliest in that device's EP priority
-              list (``get_device_ep_map()``).
-            - Both set (and ``device != "auto"``) → ep_device must satisfy
-              both filters (or None).
-            - ``ep_name`` unset and ``device == "auto"`` → ``None`` (no
-              effective filter — refuse to pick an arbitrary ep_device).
-
-        Note: When the ORT EP registry order disagrees with the priority
-        maps, the priority maps win — selection is deterministic and
-        independent of registry order.
-
-        Args:
-            device: Device policy ("cpu", "gpu", "npu", "auto"). ``"auto"``
-                and unknown strings act as no-op device filters.
-            ep_name: Full EP name (e.g., "DmlExecutionProvider"), or None
-                to skip EP-name filtering.
-
-        Returns:
-            The matching OrtEpDevice, or None if not found.
-        """
-        from ..sysinfo import get_device_ep_map, get_ep_device_map
-        from ..utils.constants import DEVICE_TO_DEVICE_TYPE
-
-        device_type = DEVICE_TO_DEVICE_TYPE.get(device.upper())
-
-        # No effective filter — refuse to pick an arbitrary ep_device.
-        if not ep_name and device_type is None:
-            return None
-
-        all_ep_devices = []
-        for ep_dev in ort.get_ep_devices():
-            if ep_name and ep_dev.ep_name != ep_name:
-                continue
-            if device_type is not None and ep_dev.device.type != device_type:
-                continue
-            all_ep_devices.append(ep_dev)
-
-        if not all_ep_devices:
-            return None
-
-        # Both filters set: any aggregated entry already satisfies both.
-        if ep_name and device_type is not None:
-            return all_ep_devices[0]
-
-        # ep_name set, device == "auto": pick by EP's preferred device order.
-        if ep_name:
-            preferred_devices = get_ep_device_map().get(ep_name, "").split("/")
-            for d in preferred_devices:
-                wanted = DEVICE_TO_DEVICE_TYPE.get(d.upper())
-                if wanted is None:
-                    continue
-                for ep_dev in all_ep_devices:
-                    if ep_dev.device.type == wanted:
-                        return ep_dev
-            return all_ep_devices[0]
-
-        # ep_name unset, device != "auto": pick by device's EP priority list.
-        ep_priority = get_device_ep_map().get(device.lower(), [])
-        for preferred_ep in ep_priority:
-            for ep_dev in all_ep_devices:
-                if ep_dev.ep_name == preferred_ep:
-                    return ep_dev
-        return all_ep_devices[0]
 
     def _validate_inputs(self, inputs: dict[str, Any]) -> None:
         """Validate inputs against model expectations.
@@ -816,7 +682,7 @@ class WinMLSession:
             test_model.ir_version = 8
 
             # 3. Try creating session with same device policy
-            sess_options = self._build_session_options(target_device)
+            sess_options, _, _ = self._build_session_options(target_device)
             sess_options.log_severity_level = 4  # Suppress ORT logs during probe
             ort.InferenceSession(
                 test_model.SerializeToString(),
--- a/src/winml/modelkit/session/session.py
+++ b/src/winml/modelkit/session/session.py
@@ -22,7 +22,7 @@ from .stats import PerfStats
 
 
 if TYPE_CHECKING:
-    from collections.abc import Generator
+    from collections.abc import Callable, Generator
 
     import onnx
 
@@ -135,7 +135,7 @@ class WinMLSession:
         ep_config: EPConfig | None = None,
         *,
         ep: EPNameOrAlias | None = None,
-        session_options: ort.SessionOptions | None = None,
+        session_options: Callable[[], ort.SessionOptions] | None = None,
     ) -> None:
         """Initialize WinMLSession.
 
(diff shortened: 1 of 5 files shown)


tests/unit/session/test_winml_session.py
```diff
@@ -366,188 +363,6 @@ def test_io_config_before_session_init(
...
-        c = helper.make_tensor_value_info("C", TensorProto.INT64, [1, 4])
-
-        identity = helper.make_node("Identity", ["A"], ["C"])
-        graph = helper.make_graph([identity], "no_signal", [a], [c])
-        model = helper.make_model(graph, opset_imports=[helper.make_opsetid("", 13)])
-        model.ir_version = 7
-        path = self._save(model, tmp_path / "no_signal.onnx")
-
-        session = WinMLSession(onnx_path=path, device="auto")
-        assert session.io_config["precision"] is None
```
xieofxie, line 548: previously when device='cpu' and ep was unset, the code
took a policy-based branch (set_provider_selection_policy) that
silently dropped provider_options. 

So although the format is incorrect, the test passes

xieofxie:
after https://github.com/microsoft/winml-cli/pull/712

src/winml/modelkit/session/session.py
```diff
@@ -364,165 +360,34 @@ def __del__(self) -> None:
...
-                        resolved,
-                    )
-                    return opts
-                logger.warning(
-                    "EP '%s' (%s) not found for device '%s'",
+        from ..sysinfo.device import resolve_device, resolve_eps
+        from ..utils.constants import DEVICE_TO_DEVICE_TYPE, normalize_ep_name
+
+        resolved_device, _ = resolve_device(device, ep=self._ep)
+        resolved_ep = normalize_ep_name(self._ep) if self._ep else resolve_eps(resolved_device)[0]
```
DingmaomaoBJTU, line 377: **Bug: `IndexError` when `resolve_eps` returns an empty list**

```python
resolved_ep = normalize_ep_name(self._ep) if self._ep else resolve_eps(resolved_device)[0]
```

`resolve_eps(resolved_device)` can return `[]` — for example when `resolve_device("auto")` falls through to `"cpu"` but `CPUExecutionProvider` isn't in the available-EP set (which happens on current `main` when ORT reports no providers at all). The `[0]` crashes with `IndexError`.

Suggested fix:

```python
if self._ep:
    resolved_ep = normalize_ep_name(self._ep)
else:
    eps = resolve_eps(resolved_device)
    if not eps:
        raise DeviceNotAvailableError(
            message=f"No EP available for resolved device '{resolved_device}'"
        )
    resolved_ep = eps[0]
```

(If this PR is meant to land after #712, the risk is lower because #712 makes `resolve_device("auto")` raise on zero EPs. But the defensive check costs nothing and protects against ordering surprises.)

src/winml/modelkit/session/session.py
```diff
@@ -364,165 +360,34 @@ def __del__(self) -> None:
...
-                opts = ort.SessionOptions()
-                opts.add_provider_for_devices([matched], self._provider_options)
-                logger.info("Discovered EP for %s: %s", device, matched.ep_name)
-                return opts
-
-        # Policy-based selection (last resort)
-        opts = self._session_options
-        policy = DEVICE_POLICY_MAP.get(
-            device.lower(), ort.OrtExecutionProviderDevicePolicy.PREFER_NPU
+        raise DeviceNotAvailableError(
```
DingmaomaoBJTU, line 392: **Loss of policy-based fallback — intentional?**

The old code had a three-level fallback:
1. Explicit EP binding via `add_provider_for_devices`
2. Device-based EP discovery
3. Policy-based selection via `set_provider_selection_policy` (last resort)

The new code only does (1) and raises `DeviceNotAvailableError` if no `ep_device` matches. This is a significant behavior change — any scenario where `resolve_device`/`resolve_eps` succeed (based on static mapping + `get_available_providers()`) but `ort.get_ep_devices()` disagrees (different API surface) will now hard-fail instead of falling through to policy.

If this is intentional (i.e., you want strict EP binding only), please add a brief comment explaining why the policy fallback was dropped. If not, consider:

```python
# Fallback: no matching ep_device found, use policy-based selection
logger.warning("No matching EP device for %s/%s, falling back to policy", resolved_ep, resolved_device)
opts.set_provider_selection_policy(DEVICE_POLICY_MAP.get(resolved_device, ort.OrtExecutionProviderDevicePolicy.PREFER_NPU))
return opts, resolved_device, resolved_ep
```

src/winml/modelkit/session/session.py
```diff
@@ -162,7 +162,7 @@ def __init__(
 
         # HF Pipeline may pass torch.device; coerce to string for downstream .lower() calls
         self._device = str(device) if not isinstance(device, str) else device
-        self._ep = ep if ep else None
+        self._ep = ep
```
DingmaomaoBJTU, line 168: **Subtle behavior change: empty string no longer coerced to `None`**

Old: `self._ep = ep if ep else None` — converts `""` to `None`
New: `self._ep = ep` — preserves `""`

This matters because `resolve_device(device, ep=self._ep)` treats `ep=""` differently from `ep=None`:
- `ep=None` → skips EP filtering entirely
- `ep=""` → enters the EP branch, `normalize_ep_name("")` returns `""`, fails with `ValueError("Unknown EP ''")`

If callers can pass `ep=""` (e.g., CLI with `--ep ""` or an empty config field), this is a regression. Consider keeping the coercion:

```python
self._ep = ep or None
```

src/winml/modelkit/session/session.py
```diff
@@ -364,165 +360,34 @@ def __del__(self) -> None:
...
-                    "EP '%s' (%s) not found for device '%s'",
+        from ..sysinfo.device import resolve_device, resolve_eps
+        from ..utils.constants import DEVICE_TO_DEVICE_TYPE, normalize_ep_name
+
+        resolved_device, _ = resolve_device(device, ep=self._ep)
+        resolved_ep = normalize_ep_name(self._ep) if self._ep else resolve_eps(resolved_device)[0]
+        device_type = DEVICE_TO_DEVICE_TYPE.get(resolved_device.upper())
+
+        opts = ort.SessionOptions()
+        for ep_dev in ort.get_ep_devices():
```
DingmaomaoBJTU: Nit: this calls `ort.get_ep_devices()` directly, bypassing `get_registered_ep_devices()` from `winml.py` which ensures WinML EPs are registered first. The class-level `_init_winml_eps_once()` in `__init__` handles registration, but it uses `WinMLEPRegistry` (a different registration path than `winml.register_execution_providers`). If these two registration paths can diverge, the `ep_devices` here might not include WinML-registered EPs.

Consider using `get_registered_ep_devices()` for consistency:

```python
from ..winml import get_registered_ep_devices
...
for ep_dev in get_registered_ep_devices():
```

src/winml/modelkit/session/session.py
```diff
@@ -275,12 +275,8 @@ def compile(self) -> None:
 
         # Resolve device label from the primary provider ORT actually selected
         if self._device == "auto" and actual_providers:
-            from ..sysinfo.device import get_ep_device_map
-
-            ep_map = get_ep_device_map()
-            resolved = ep_map.get(actual_providers[0])
-            if resolved and "/" not in resolved:
-                self._device = resolved
+            self._device = resolved_device
```
DingmaomaoBJTU, line 280: Good simplification — using `resolved_device` from `_build_session_options` directly instead of reverse-mapping from `get_ep_device_map()` is cleaner and avoids the old `"/" not in resolved` heuristic.

tests/unit/session/test_winml_session.py
```diff
@@ -176,20 +176,17 @@ def test_providers_are_valid_and_include_fallback(self, simple_matmul_onnx: Path
...
-
-        Note: WinML-registered EPs (like QNN) may be selected dynamically
-        via set_provider_selection_policy() even if not listed in
-        get_available_providers() beforehand.
         """
-        # Create session with NPU preference
         session = WinMLSession(
             onnx_path=simple_matmul_onnx,
-            device="npu",  # PREFER_NPU policy
+            device="auto",
```
DingmaomaoBJTU, line 189: This change from `device="npu"` to `device="auto"` is the right call since policy-based selection is gone, but the test name `test_providers_are_valid_and_include_fallback` and assertion `CPUExecutionProvider in actual_providers` now depend on the CI runner's hardware. On a machine with an NPU or GPU, `resolve_device("auto")` picks that device, binds its EP explicitly via `add_provider_for_devices`, and the resulting session may or may not include CPUExecutionProvider as a fallback.

Worth verifying: does `add_provider_for_devices` with a non-CPU ep_device always include CPUExecutionProvider in `session.get_providers()`? If not, the assertion on line 203 could fail on NPU/GPU hardware.

DingmaomaoBJTU (changes requested):
Good direction — unifying session creation through `resolve_device`/`resolve_eps` eliminates a lot of duplicated logic and makes the EP selection path consistent across all commands. The `_find_ep_device` removal is clean.

However, there are a few issues, one of which is a crash-level bug.

**Also note:** `DEVICE_POLICY_MAP` (lines 46–51) is now dead code — no remaining caller after the policy-based fallback was removed. Should be deleted (or kept only if you plan to re-introduce policy fallback). Similarly, the class docstring (lines 98–106) still says "Policy-based device selection" and "does NOT use explicit EP provider names" which is now the opposite of what the code does.

src/winml/modelkit/session/session.py
xieofxie: if resolve_device returns correctly, resolve_eps should work

src/winml/modelkit/session/session.py
xieofxie: yes, all are resolved via these two resolve functions. If we want to use policy, they should be handled inside resolve functions

timenick (commented):
probe

src/winml/modelkit/session/session.py
xieofxie: ep parameter is EPNameOrAlias  now, so no need to handle ""

src/winml/modelkit/session/session.py
```diff
@@ -364,165 +360,34 @@ def __del__(self) -> None:
...
-                logger.warning(
-                    "EP '%s' (%s) not found for device '%s'",
+        from ..sysinfo.device import resolve_device, resolve_eps
+        from ..utils.constants import DEVICE_TO_DEVICE_TYPE, normalize_ep_name
+
+        resolved_device, _ = resolve_device(device, ep=self._ep)
+        resolved_ep = normalize_ep_name(self._ep) if self._ep else resolve_eps(resolved_device)[0]
+        device_type = DEVICE_TO_DEVICE_TYPE.get(resolved_device.upper())
+
+        opts = ort.SessionOptions()
```
timenick: **Question: is `self._session_options` deliberately dropped?**

This function now always constructs a **fresh** `ort.SessionOptions()` and never reads `self._session_options`. But the constructor still accepts `session_options: ort.SessionOptions \| None = None` (line 139) and stores it (line 174), and this function’s own docstring still says "Build ORT SessionOptions from instance session_options and device".

The old code reused `self._session_options` for two of its four branches (the CPU fast path and the policy fallback). After this PR, anyone passing custom options:

```python
opts = ort.SessionOptions()
opts.log_severity_level = 3
opts.intra_op_num_threads = 4
opts.add_session_config_entry("session.disable_cpu_ep_fallback", "1")
session = WinMLSession(onnx_path=..., session_options=opts)
```

silently loses every setting on those `opts`.

Is this the intended behaviour? If so, the `session_options` parameter / `self._session_options` field / docstring should be removed (or marked deprecated) so the contract matches the implementation. If not, one option is:

```python
opts = self._session_options if self._session_options is not None else ort.SessionOptions()
```

(although that may interact with `add_provider_for_devices` if `opts` was already used for a previous call — the old code’s fresh-opts comment was "avoid 'already registered' errors from repeated calls.")

src/winml/modelkit/session/session.py
timenick, line 377: **Regression: `device="cpu"` with no explicit EP can silently bind OpenVINO-on-CPU.**

`resolve_eps("cpu")[0]` returns the first entry from `_DEVICE_EP_MAP["cpu"]` that is registered. That list is built from `EP_SUPPORTED_DEVICES` iteration order, where OpenVINO appears **before** CPUExecutionProvider:

```python
# utils/constants.py
EP_SUPPORTED_DEVICES = {
    ...
    "OpenVINOExecutionProvider": ("npu", "gpu", "cpu"),  # appears first for cpu
    "DmlExecutionProvider": ("gpu",),
    "CPUExecutionProvider": ("cpu",),                     # appears second
}
```

So on a machine with OpenVINO installed, `WinMLSession(onnx_path=..., device="cpu")` now binds OV-on-CPU instead of basic CPU. The old code explicitly guarded against this:

```python
if device.lower() == "cpu" and not self._ep:
    opts.set_provider_selection_policy(DEVICE_POLICY_MAP["cpu"])  # PREFER_CPU
    return opts
```

with the comment:

> `--ep cpu` is honoured here too so the CPUExecutionProvider gets bound explicitly; otherwise PREFER_CPU policy lets ORT prefer OV-on-CPU (or any other registered CPU-capable EP) over the basic CPU EP, silently ignoring the user's --ep choice.

This PR drops the safeguard without replacing it. Simple fix:

```python
if self._ep is None and resolved_device == "cpu":
    resolved_ep = "CPUExecutionProvider"
else:
    resolved_ep = (
        normalize_ep_name(self._ep) if self._ep
        else resolve_eps(resolved_device)[0]
    )
```

(Related: `resolve_eps(resolved_device)` can return `[]` if `_get_device_ep_map_from_ort()` ever reports an EP that’s not in the static `EP_SUPPORTED_DEVICES` table — same drift surface I flagged on #712. Indexing `[0]` would IndexError there. `next(iter(resolve_eps(resolved_device)), None)` would degrade gracefully.)

src/winml/modelkit/session/session.py
```diff
@@ -364,165 +360,34 @@ def __del__(self) -> None:
...
-                        "Explicit EP: %s (%s) device=%s -> %s",
-                        self._ep,
-                        target_name,
-                        device,
-                        resolved,
-                    )
-                    return opts
-                logger.warning(
-                    "EP '%s' (%s) not found for device '%s'",
+        from ..sysinfo.device import resolve_device, resolve_eps
```
timenick, line 372: **Startup-latency regression on the trivial CPU path.**

The old code had a fast path that returned without calling `resolve_device` when `device == "cpu"` and no explicit EP:

```python
if device.lower() == "cpu" and not self._ep:
    opts.set_provider_selection_policy(DEVICE_POLICY_MAP["cpu"])
    return opts
```

The new code always calls `resolve_device`, which transitively triggers `_get_device_ep_map_from_ort()` → `register_execution_providers(ort=True)` → `WinML().__init__()` → `EpCatalog()` → `provider.ensure_ready()` per provider. That’s significantly heavier than the old CPU fast path.

For `WinMLSession(onnx_path=..., device="cpu")` (common in tests, scripts, lightweight inference), startup gets noticeably slower. `resolve_device` itself is `lru_cache`d so the cost is one-time per process, but the one-time cost is now mandatory even when the caller knew up-front they wanted CPU.

Not necessarily a blocker, but worth either: (a) restoring a CPU fast path for the `device="cpu", ep=None` case, or (b) measuring and noting the latency delta in the PR description so reviewers can decide.

src/winml/modelkit/session/session.py
timenick, line 392: **Two small things on this error path:**

1. **`self._ep` may be `None`.** The f-string then renders as `"No available provider found for device 'cpu' and EP 'None'"` — reads awkwardly. Use `self._ep or '(auto)'`.

2. **The branch is effectively unreachable in normal operation.** If `resolve_device` succeeded, the `(resolved_device, resolved_ep)` pair came from `_get_device_ep_map_from_ort()`, which is itself built from `get_registered_ep_devices() == ort.get_ep_devices()`. So the for-loop above will always find a match unless the cache drifts between calls (race) or `DEVICE_TO_DEVICE_TYPE.get(...)` somehow returned `None` (it can’t for the 3 valid devices). A comment like `# Defensive: should not be reachable unless the EP enumeration races.` would make this clearer to future readers.

tests/unit/session/test_winml_session.py
```diff
@@ -893,99 +903,6 @@ def test_perf_stats_accessible_after_context(
...
-        """ep_name=None and device='auto' → None (no effective filter)."""
-        import onnxruntime as ort
-
-        devs = [self._ep_dev("QNNExecutionProvider", ort.OrtHardwareDeviceType.NPU)]
-        with self._patch_devices(devs):
-            assert WinMLSession._find_ep_device(device="auto") is None
-            assert WinMLSession._find_ep_device(device="auto", ep_name=None) is None
-
-
 @pytest.mark.ep("openvino")
```
timenick, line 963: **The deleted `TestFindEpDevice` class included a real regression guard that has no replacement.**

Deleting `_find_ep_device` and its tests is fine — the new for-loop in `_build_session_options` correctly checks `ep_name == resolved_ep and ep_dev.device.type == device_type` (AND semantics). But the deleted `test_ep_name_and_device_both_required` was specifically a regression guard, and the deletion comment in its docstring spells out the bug it caught:

> Regression guard: when both filters are set, both must match (AND logic). Previously these were two separate methods and the explicit-EP path ignored device, so `--ep qnn --device cpu` could return QNN-on-NPU.

Without that anchor, anyone refactoring the for-loop later can re-introduce the bug (e.g. flip the `and` to `or`, or drop the `device.type` check) and CI won’t notice. The behaviour also moved into a different class — the new owner is `WinMLSession`, so the test moves with it. A minimal replacement at the new layer:

```python
def test_ep_qnn_device_cpu_binds_qnn_on_cpu(self, simple_matmul_onnx):
    # Two QNN ep_devices: NPU first, then CPU. User asks for QNN-on-CPU.
    npu_dev = SimpleNamespace(ep_name="QNNExecutionProvider", device=SimpleNamespace(type=ort.OrtHardwareDeviceType.NPU))
    cpu_dev = SimpleNamespace(ep_name="QNNExecutionProvider", device=SimpleNamespace(type=ort.OrtHardwareDeviceType.CPU))
    captured = []
    real = ort.SessionOptions.add_provider_for_devices
    def spy(self_so, devs, opts):
        captured.append(devs[0].device.type)
        return real(self_so, devs, opts)
    with patch("winml.modelkit.session.session.ort.get_ep_devices", return_value=[npu_dev, cpu_dev]), \
         patch("winml.modelkit.sysinfo.device._get_device_ep_map_from_ort", return_value={"cpu": ("QNNExecutionProvider",), "npu": ("QNNExecutionProvider",)}), \
         patch.object(ort.SessionOptions, "add_provider_for_devices", spy):
        WinMLSession(onnx_path=simple_matmul_onnx, device="cpu", ep_config=EPConfig(provider="qnn"))
    assert captured == [ort.OrtHardwareDeviceType.CPU], "binding picked the wrong device"
```

The specific shape is negotiable — the key is that the regression guard exists somewhere for the new architecture.

timenick (commented):
Reviewed the resolve_device unification. The shape is right but I want to flag two behavior changes that may be unintentional, a missing regression-test anchor, and three minor / nits.

tests/unit/session/test_winml_session.py
xieofxie: there is no such assertion, so no bug

src/winml/modelkit/session/session.py
xieofxie: because we handle self._ep now, so the old comment is invalid

src/winml/modelkit/session/session.py
xieofxie: won't fix latency issue. The code logic should be like this now

src/winml/modelkit/session/session.py
xieofxie: - self._ep could be None
- sure, added

src/winml/modelkit/session/session.py
xieofxie: Good catch! I think I would use a lambda instead to workaround the current limitations

src/winml/modelkit/session/session.py
xieofxie: it is already called _init_winml_eps_once. So no error here. Use add_ep_for_device from winml.py

tests/unit/session/test_winml_session.py
xieofxie: added!

DingmaomaoBJTU (approved):

PR #709, opened 2026-05-22
fix: suppress ORT native stderr, fix HANDLE bug, clean up warnings

DingmaomaoBJTU:
## Summary

- ORT's pybind module writes `Init provider bridge failed.` directly to native stderr (fd 2 / Win32 `STD_ERROR_HANDLE`), bypassing Python's `logging`/`warnings` systems entirely
- Added `utils/native_stderr.py` — a dedicated module for capturing and replaying native stderr output from ORT/QNN
  - `suppress_ep_registration_stderr()` context manager redirects fd 2 via pipe, re-emits captured lines through Python logging
  - `replay_ort_startup_logs()` public API for deferred replay after logging is configured
- Fixed **64-bit HANDLE truncation bug**: set proper `argtypes`/`restype` via `ctypes.wintypes` for `GetStdHandle` and `SetStdHandle`
- Fixed **dup2 restore ordering**: UCRT's `dup2` for fds 0-2 internally calls `SetStdHandle`, so `GetStdHandle` must be captured before `dup2` and restored after
- **Platform-gated**: no-op on non-Windows — Linux/macOS CI gets a plain `import onnxruntime` with zero fd overhead
- `constants.py` restored to leaf-level constants-only module
- Downgraded HF/optimum/transformers logging noise to appropriate levels
- Added 10 unit tests covering fd capture, ANSI strip, Win32 HANDLE restore, and replay API

Reference: microsoft/winml-cli#477


timenick (approved):

vortex-captain (approved):

timenick (commented):
p

timenick (commented):
p

timenick (commented):
p

src/winml/modelkit/utils/constants.py
```diff
@@ -6,9 +6,78 @@
...
+    os.dup2(write_fd, 2)
+    os.close(write_fd)
+    old_w32 = None
+    if sys.platform == "win32":
+        import ctypes
+        import msvcrt
+
+        k32 = ctypes.WinDLL("kernel32")
+        _std_error_handle = ctypes.c_uint32(0xFFFFFFF4)
+        old_w32 = k32.GetStdHandle(_std_error_handle)
```
timenick: **Bug: `GetStdHandle` return is truncated to 32 bits on x64.**

`ctypes.WinDLL("kernel32").GetStdHandle` has no `restype` set, so ctypes defaults to `c_int` (4 bytes). On 64-bit Windows `HANDLE` is a `void*` (8 bytes). The high 32 bits of `old_w32` are silently truncated, so the `finally` branch calls `k32.SetStdHandle(_std_error_handle, old_w32)` with an **invalid handle** whenever the real HANDLE has any high bits set (very common — Windows hands out high-bit handles routinely).

Net effect: after `_suppress_ep_registration_stderr()` exits, Win32 `STD_ERROR_HANDLE` points to garbage. Any later Win32-API stderr write (including from ORT, QNN, or any native dependency) silently goes nowhere or crashes. The CRT `dup2` restore still works, so Python `sys.stderr` looks fine — which is exactly how this slips through manual testing.

Fix (and please also apply to `GetCurrentProcess` and `SetStdHandle`):

```python
from ctypes import wintypes

k32 = ctypes.WinDLL("kernel32", use_last_error=True)
k32.GetStdHandle.argtypes = [wintypes.DWORD]
k32.GetStdHandle.restype = wintypes.HANDLE
k32.SetStdHandle.argtypes = [wintypes.DWORD, wintypes.HANDLE]
k32.SetStdHandle.restype = wintypes.BOOL
k32.GetCurrentProcess.argtypes = []
k32.GetCurrentProcess.restype = wintypes.HANDLE
```

This is identical to the pattern that PR #705 just landed for `IsWow64Process2` after a similar review comment. A small test that asserts `GetStdHandle(STD_ERROR_HANDLE)` returns the same value before and after the wrap would pin this regression.

src/winml/modelkit/utils/constants.py
```diff
@@ -6,9 +6,78 @@
...
+# Matches ANSI SGR escape sequences (e.g. the colour codes ORT emits).
+_ANSI_RE = re.compile(r"\x1b\[[0-9;]*m")
+
+# ORT messages captured before any logging handler is configured are buffered
+# here so callers can replay them once the logging infrastructure is ready.
+_ort_startup_logs: list[str] = []
+
+
+@contextmanager
+def _suppress_ep_registration_stderr():
```
timenick: **Architecture: this doesn't belong in `utils/constants.py`.**

The existing comment in this file (around line 47 today on main) explicitly states:

> `utils.constants` stays leaf-level (no import dependency on sysinfo).

It's supposed to be a constants-only leaf. Landing 70+ lines of side-effecting fd manipulation, a Win32 HANDLE swap, an ANSI regex, a mutable module-level log buffer, and a context manager turns it into something very different. Downstream effects:

1. Every import of any constant from this file (`EPName`, `EP_ALIASES`, `SUPPORTED_EPS`, etc.) drags in `ctypes`, `msvcrt`, `os.pipe` machinery.
2. The module docstring (`"Shared constants for WinML CLI."`) no longer matches reality.
3. Future test isolation gets harder — importing constants triggers fd 2 manipulation in pytest.

Suggestion: extract to a dedicated module, e.g. `src/winml/modelkit/utils/native_stderr.py`:

```python
# native_stderr.py
from contextlib import contextmanager

@contextmanager
def suppress_ep_registration_stderr(): ...

def replay_ort_startup_logs(logger=None) -> None: ...
```

Then `constants.py` shrinks back to:

```python
from .native_stderr import suppress_ep_registration_stderr

with suppress_ep_registration_stderr():
    import onnxruntime as ort
```

Makes the helper testable in isolation, restores the leaf invariant, and gives `commands/sys.py` a public entry point for the replay (see separate comment about reaching into `_ort_startup_logs`).

src/winml/modelkit/utils/constants.py
```diff
@@ -6,9 +6,78 @@
...
+        captured = b"".join(chunks).decode("utf-8", errors="replace")
+        for line in captured.splitlines():
+            line = _ANSI_RE.sub("", line).strip()
+            if line:
+                _ort_startup_logs.append(line)
+                logger.debug("[ORT] %s", line)
+
+
+with _suppress_ep_registration_stderr():
+    import onnxruntime as ort
```
timenick, line 19: **Import-time side effect runs on every Python process that touches constants.**

Wrapping `import onnxruntime as ort` in `_suppress_ep_registration_stderr()` at module top-level means every importer of `winml.modelkit.utils.constants` pays the cost: `os.pipe()`, two `os.dup2()`s, a Win32 `GetStdHandle`/`SetStdHandle` swap, and a blocking `os.read()` loop — even when ORT writes nothing and even on non-Windows (where the symptom you're fixing doesn't exist).

Two concrete asks:

1. **Platform-gate the wrap.** The bug ("Init provider bridge failed.") is Windows-only. Add `if sys.platform != "win32": import onnxruntime as ort` first, fall through to the wrapped import only on Windows. That saves Linux/macOS CI runs the fd churn entirely.
2. **Consider lazy ORT initialization.** Many `winml.modelkit.utils.constants` consumers don't need ORT loaded — they just want `EPName`, `EP_ALIASES`, etc. Importing `onnxruntime` at module scope was already a heavy choice; you're now compounding it with fd manipulation. If a lazy-import refactor is too big for this PR, at minimum document the trade-off in the module docstring.

src/winml/modelkit/commands/sys.py
```diff
@@ -718,6 +718,12 @@ def sysinfo(
     pkg_logger.addHandler(rich_handler)
     pkg_logger.propagate = False
 
+    # Replay ORT messages that were captured before logging was configured.
+    from ..utils.constants import _ort_startup_logs
+
+    for _msg in _ort_startup_logs:
```
timenick: **Reaching into a private `_`-prefixed list from another module.**

`from ..utils.constants import _ort_startup_logs` then iterating it. The repo's import convention reserves `_`-prefixed symbols for test-only internal access (see `CLAUDE.md` import rules). Reaching into a sibling module's private mutable state from `commands/` couples the two modules through an undocumented protocol — any future refactor of how the buffer is stored, sized, or cleared silently breaks `sys.py`.

Suggested public API on the owning module (paired with the architecture comment elsewhere):

```python
# utils/native_stderr.py
def replay_ort_startup_logs(target_logger: logging.Logger | None = None) -> None:
    """Replay captured ORT startup lines and clear the buffer."""
    log = target_logger or logger
    for msg in _ort_startup_logs:
        log.debug("[ORT] %s", msg)
    _ort_startup_logs.clear()
```

Then `commands/sys.py` becomes:

```python
from ..utils.native_stderr import replay_ort_startup_logs
...
replay_ort_startup_logs()
```

No private access, no hard-coded logger name string, and the buffer-clearing behavior (which the current code is missing — the list grows indefinitely if anything calls the context manager again) becomes a single place to fix.

src/winml/modelkit/utils/constants.py
```diff
@@ -6,9 +6,78 @@
...
+            while chunk := os.read(read_fd, 4096):
+                chunks.append(chunk)
+        finally:
+            os.close(read_fd)
+        captured = b"".join(chunks).decode("utf-8", errors="replace")
+        for line in captured.splitlines():
+            line = _ANSI_RE.sub("", line).strip()
+            if line:
+                _ort_startup_logs.append(line)
+                logger.debug("[ORT] %s", line)
```
timenick: **No tests for fd manipulation — the most failure-prone code in the PR.**

This function does fd 2 dup/dup2 across CRT and Win32 layers in a try/finally, with manual ordering constraints documented only in a comment. That's exactly the surface where regressions are silent (see the separate HANDLE-truncation comment for one such regression that already lives in the PR). Today there's zero coverage that:

1. stderr written **inside** the context lands in `_ort_startup_logs`
2. fd 2 still prints to the real terminal **after** the context exits
3. on Windows, `GetStdHandle(STD_ERROR_HANDLE)` returns the same value before and after
4. multiline stderr is split into individual log lines
5. ANSI sequences are stripped

Minimum suggested coverage (pure pytest, no real ORT needed):

```python
# tests/unit/utils/test_native_stderr.py
import os, sys
import pytest
from winml.modelkit.utils.constants import (
    _suppress_ep_registration_stderr, _ort_startup_logs,
)

def test_captures_native_stderr():
    _ort_startup_logs.clear()
    with _suppress_ep_registration_stderr():
        os.write(2, b"hello\nworld\n")
    assert "hello" in _ort_startup_logs
    assert "world" in _ort_startup_logs

def test_strips_ansi():
    _ort_startup_logs.clear()
    with _suppress_ep_registration_stderr():
        os.write(2, b"\x1b[31mred message\x1b[0m\n")
    assert "red message" in _ort_startup_logs

def test_stderr_works_after_context(capfd):
    with _suppress_ep_registration_stderr():
        pass
    os.write(2, b"after\n")
    assert "after" in capfd.readouterr().err

@pytest.mark.skipif(sys.platform != "win32", reason="Win32 only")
def test_win32_std_error_handle_restored():
    import ctypes
    from ctypes import wintypes
    k32 = ctypes.WinDLL("kernel32")
    k32.GetStdHandle.argtypes = [wintypes.DWORD]
    k32.GetStdHandle.restype = wintypes.HANDLE
    STD_ERROR_HANDLE = wintypes.DWORD(-12 & 0xFFFFFFFF)
    before = k32.GetStdHandle(STD_ERROR_HANDLE)
    with _suppress_ep_registration_stderr():
        pass
    after = k32.GetStdHandle(STD_ERROR_HANDLE)
    assert before == after, "STD_ERROR_HANDLE not restored"
```

The last one would have caught the HANDLE-truncation bug flagged separately.

timenick (commented):
Reviewed the ORT stderr suppression. The intent is good, but I want to flag one real correctness bug (HANDLE truncation) plus four substantive design/coverage issues before this lands.

src/winml/modelkit/utils/constants.py
DingmaomaoBJTU: Fixed. Applied `ctypes.wintypes` for all Win32 API calls:

```python
k32 = ctypes.WinDLL("kernel32", use_last_error=True)
k32.GetStdHandle.argtypes = [wintypes.DWORD]
k32.GetStdHandle.restype = wintypes.HANDLE
k32.SetStdHandle.argtypes = [wintypes.DWORD, wintypes.HANDLE]
k32.SetStdHandle.restype = wintypes.BOOL
```

Also found and fixed a second ordering bug during testing: `old_w32` was captured **after** `os.dup2(write_fd, 2)`, but UCRT's `dup2` for fds 0-2 internally calls `SetStdHandle` — so `old_w32` was saving the pipe's handle, not the original. Fixed by capturing `GetStdHandle` before `dup2`, and restoring `SetStdHandle` after `dup2`. The `test_win32_std_error_handle_restored` test now pins this.

src/winml/modelkit/utils/constants.py
DingmaomaoBJTU: Done. Extracted to `utils/native_stderr.py` with public API:

```python
# native_stderr.py
suppress_ep_registration_stderr()   # context manager
replay_ort_startup_logs()           # drain startup buffer
```

`constants.py` is back to a leaf-level constants module — just a thin platform-gated import:

```python
if sys.platform == "win32":
    from .native_stderr import suppress_ep_registration_stderr
    with suppress_ep_registration_stderr():
        import onnxruntime as ort
else:
    import onnxruntime as ort
```

src/winml/modelkit/utils/constants.py
DingmaomaoBJTU: Done. The context manager now short-circuits on non-Windows:

```python
if sys.platform != "win32":
    yield
    return
```

And `constants.py` only wraps the import on Windows — Linux/macOS CI gets a plain `import onnxruntime as ort` with zero fd overhead.

src/winml/modelkit/commands/sys.py
DingmaomaoBJTU: Fixed. `native_stderr.py` now exposes a public `replay_ort_startup_logs()` that drains and clears the buffer. No more private `_ort_startup_logs` access from other modules:

```python
# utils/logging.py
from winml.modelkit.utils.native_stderr import replay_ort_startup_logs
replay_ort_startup_logs()
```

src/winml/modelkit/utils/constants.py
DingmaomaoBJTU: Added `tests/unit/utils/test_native_stderr.py` with 10 test cases covering all 5 suggested scenarios:

1. `test_captures_native_stderr` — stderr inside context lands in buffer
2. `test_stderr_works_after_context` — fd 2 still works after exit
3. `test_win32_std_error_handle_restored` — `GetStdHandle(STD_ERROR_HANDLE)` same before/after (Win32 only)
4. `test_skips_blank_lines` / `test_strips_ansi` — multiline split + ANSI strip
5. `test_custom_log_level` / `test_startup_buffer_disabled` — param coverage
6. `test_replays_and_clears_buffer` / `test_noop_when_empty` / `test_custom_target_logger` — replay API

The Win32 HANDLE test caught the `dup2`-ordering bug mentioned in the first comment.

timenick (approved):

pyproject.toml
```diff
@@ -4,7 +4,7 @@ requires = [ "setuptools>=61", "wheel" ]
 
 [project]
 name = "winml-cli"
-version = "0.0.4"
+version = "0.0.6"
```
timenick: Why are we changing the version in this PR?

timenick (approved):

PR #712, opened 2026-05-22
fix: resolve_device/resolve_eps uses get_registered_ep_devices

xieofxie:
For a device without openvino gpu but have openvino cpu, the follow command will crash before

` winml compile -m "resnet-cpu\model.onnx" -v --ep openvino `

Because it will be resolved to gpu device since dml has gpu device

### DIFF
--- a/src/winml/modelkit/sysinfo/device.py
+++ b/src/winml/modelkit/sysinfo/device.py
@@ -10,7 +10,7 @@ import functools
 import logging
 from typing import TYPE_CHECKING
 
-from ..utils.constants import EP_SUPPORTED_DEVICES, EPName, normalize_ep_name
+from ..utils.constants import DEVICE_TYPE_TO_DEVICE, EP_SUPPORTED_DEVICES, EPName, normalize_ep_name
 from ..winml import get_registered_ep_devices
 
 
@@ -173,58 +173,44 @@ def resolve_device(
     if device != "auto" and device not in _VALID_DEVICES:
         raise ValueError(f"Unknown device '{device}'. Expected 'auto', 'npu', 'gpu', or 'cpu'.")
 
-    # Materialize cached tuple as a fresh list per call so that any
-    # caller mutation (e.g. `available.append(...)`) cannot poison the cache.
-    available_devices = list(_get_available_devices())
-    available_eps = _get_available_eps()
+    device_ep_map: dict[str, list[EPName]] = {}
+    for ep_device in get_registered_ep_devices():
+        device_name = DEVICE_TYPE_TO_DEVICE.get(ep_device.device.type)
+        if device_name is not None:
+            device_ep_map.setdefault(device_name.lower(), []).append(ep_device.ep_name)
 
     if ep is not None:
         ep_full = normalize_ep_name(ep)
         if ep_full not in EP_SUPPORTED_DEVICES:
             raise ValueError(f"Unknown EP '{ep}'. Expected one of: {sorted(EP_SUPPORTED_DEVICES)}")
-        available_eps = available_eps & {ep_full}
-        if not available_eps:
+        device_ep_map = {dev: [ep_full] for dev, eps in device_ep_map.items() if ep_full in eps}
+        if not device_ep_map:
             raise ValueError(
                 f"Requested EP '{ep}' is not available on this system. "
                 f"Available EPs: {sorted(_get_available_eps())}."
             )
 
-        ep_compatible_devices = set(EP_SUPPORTED_DEVICES[ep_full])
-        available_devices = [d for d in available_devices if d in ep_compatible_devices]
-
-    if not available_eps:
-        logger.warning(
-            "No execution providers detected. Falling back to CPU. "
-            "Install onnxruntime or Windows App SDK for EP discovery."
-        )
-
     if device == "auto":
         # Walk priority list, pick first device with a matching EP
-        for dev in available_devices:
-            compatible_eps = _DEVICE_EP_MAP.get(dev, [])
-            if any(ep_name in available_eps for ep_name in compatible_eps):
+        for dev in _VALID_DEVICES:
+            if dev in device_ep_map:
                 logger.info(
                     "Auto-selected device '%s' with compatible EPs: %s for auto device",
                     dev,
-                    sorted(ep_name for ep_name in compatible_eps if ep_name in available_eps),
+                    sorted(device_ep_map[dev]),
                 )
-                return dev, available_devices
-        # Fallback: CPU is always valid
-        return "cpu", available_devices
+                return dev, [dev for dev in _VALID_DEVICES if dev in device_ep_map]
 
     # Explicit device requested -- raise if no compatible EP is available.
-    # Falling through with a warning would write an unusable config that fails
-    # later at compile/inference time. (issue #431)
-    compatible_eps = _DEVICE_EP_MAP.get(device, [])
-    if not any(ep_name in available_eps for ep_name in compatible_eps):
+    if device not in device_ep_map:
         raise ValueError(
-            f"Device '{device}' requested but no compatible EP is available. "
-            f"Compatible EPs: {compatible_eps}. "
-            f"Available EPs: {sorted(available_eps)}."
+            f"Device '{device}' requested but no EP is available. "
+            f"Available EPs: {sorted(_get_available_eps())}."
         )
-    return device, available_devices
+    return device, [dev for dev in _VALID_DEVICES if dev in device_ep_map]
 
 
+@functools.lru_cache(maxsize=3)
 def resolve_eps(resolved_device: str) -> list[EPName]:
     """Return list of available EPs compatible with the given device.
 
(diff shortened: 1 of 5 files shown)


src/winml/modelkit/sysinfo/device.py
```diff
@@ -194,54 +187,39 @@ def resolve_device(
...
         ep_full = normalize_ep_name(ep)
         if ep_full not in EP_SUPPORTED_DEVICES:
             raise ValueError(f"Unknown EP '{ep}'. Expected one of: {sorted(EP_SUPPORTED_DEVICES)}")
-        available_eps = available_eps & {ep_full}
-        if not available_eps:
+        device_ep_map = {dev: (ep_full,) for dev, eps in device_ep_map.items() if ep_full in eps}
+        if not device_ep_map:
             raise ValueError(
                 f"Requested EP '{ep}' is not available on this system. "
                 f"Available EPs: {sorted(_get_available_eps())}."
```
timenick, line 184: **Two sources of truth for "available EPs" can produce contradictory error messages.**

After this PR there are two parallel EP-discovery paths:

- `_get_available_eps()` → `WinMLEPRegistry.get_available_eps()` + `ort.get_available_providers()`
- `_get_device_ep_map_from_ort()` → `register_execution_providers()` + `ort.get_ep_devices()`

These call different ORT APIs and can legitimately disagree on whether a given EP is "available" (e.g. an EP that was registered but doesn’t have any matching `OrtEpDevice` will be in one set but not the other).

The error path here mixes the two:

```python
device_ep_map = {dev: (ep_full,) for dev, eps in device_ep_map.items() if ep_full in eps}
if not device_ep_map:
    raise ValueError(
        f"Requested EP '{ep}' is not available on this system. "
        f"Available EPs: {sorted(_get_available_eps())}."
    )
```

The **availability check** uses `_get_device_ep_map_from_ort()` (line above), but the **error message** quotes `_get_available_eps()`. So users can hit:

> Requested EP 'OpenVINOExecutionProvider' is not available on this system. Available EPs: [..., 'OpenVINOExecutionProvider', ...]

where the requested EP appears in the “Available EPs” list — a contradiction that wastes the user’s time.

Same pattern at the second `raise ValueError(...)` lower down (the explicit-device branch quotes `_get_available_eps()` for the same reason).

Suggested fix (single source of truth, no extra API calls):

```python
device_ep_map = dict(_get_device_ep_map_from_ort())
available_eps = sorted({ep for eps in device_ep_map.values() for ep in eps})

if ep is not None:
    ep_full = normalize_ep_name(ep)
    ...
    device_ep_map = {dev: (ep_full,) for dev, eps in device_ep_map.items() if ep_full in eps}
    if not device_ep_map:
        raise ValueError(
            f"Requested EP '{ep}' is not available on this system. "
            f"Available EPs: {available_eps}."
        )
```

And correspondingly in the explicit-device branch. After this, `_get_available_eps()` can either be removed (no remaining callers) or kept around with a docstring note saying "separate API; may disagree with `_get_device_ep_map_from_ort()` — prefer the latter for new code."

timenick (commented):
One concern: the new `_get_device_ep_map_from_ort()` and the existing `_get_available_eps()` now use different ORT APIs, and your error messages mix the two so users can see contradictory output.

src/winml/modelkit/winml.py
```diff
@@ -119,6 +121,7 @@ def register_execution_providers(ort: bool = True, ort_genai: bool = False) -> d
     return WinML().register_execution_providers(ort=ort, ort_genai=ort_genai)
 
 
+@functools.lru_cache(maxsize=1)
 def get_registered_ep_devices() -> list[Any]:
```
DingmaomaoBJTU: **Bug: `lru_cache` on a function returning `list` — cache poisoning risk**

`lru_cache` returns the **same object** on every call. Any caller that mutates the returned list (`.append()`, `.pop()`, slice assignment) silently corrupts the cache for every subsequent caller in the process.

`_get_available_devices` already solves this by returning a `tuple`. Apply the same pattern here:

```python
@functools.lru_cache(maxsize=1)
def get_registered_ep_devices() -> tuple[Any, ...]:
    ...
    return tuple(ort.get_ep_devices())
```

(Or keep returning `list` but drop the `@lru_cache` and let `_get_device_ep_map_from_ort`'s own cache handle the deduplication — the function is only called from one site now.)

src/winml/modelkit/sysinfo/device.py
```diff
@@ -87,50 +92,38 @@ def get_device_ep_map() -> dict[str, list[EPName]]:
...
+    do not change at runtime.
+    """
+    result: dict[str, list[EPName]] = {}
+    try:
+        for ep_device in get_registered_ep_devices():
+            device_name = DEVICE_TYPE_TO_DEVICE.get(ep_device.device.type)
+            if device_name is not None:
+                result.setdefault(device_name.lower(), []).append(ep_device.ep_name)
+    except Exception:
+        logger.debug("Failed to enumerate registered EP devices", exc_info=True)
```
DingmaomaoBJTU: **`except Exception` at DEBUG level hides actionable errors**

If ORT is installed but `get_ep_devices()` raises (driver bug, version mismatch, etc.), this swallows the error at DEBUG and returns an empty dict. The user then sees a cryptic `RuntimeError("No execution providers detected.")` from `resolve_device` with no clue **why** enumeration failed.

Consider logging at WARNING so the root cause is visible at default verbosity:

```python
except Exception:
    logger.warning("Failed to enumerate registered EP devices", exc_info=True)
```

(DEBUG is fine for "ORT not installed" — but that's an `ImportError` inside `get_registered_ep_devices`, not an `Exception` that reaches here.)

src/winml/modelkit/sysinfo/device.py
```diff
@@ -194,54 +187,39 @@ def resolve_device(
...
-            if any(ep_name in available_eps for ep_name in compatible_eps):
-                logger.info(
-                    "Auto-selected device '%s' with compatible EPs: %s for auto device",
-                    dev,
-                    sorted(ep_name for ep_name in compatible_eps if ep_name in available_eps),
-                )
-                return dev, available_devices
-        # Fallback: CPU is always valid
-        return "cpu", available_devices
+        chosen = available_devices[0]
```
DingmaomaoBJTU, line 193: **`available_devices[0]` — fragile IndexError path**

This is safe **today** because `_DEVICE_PRIORITY` covers all keys `DEVICE_TYPE_TO_DEVICE` can produce, and the `if not device_ep_map` guard above catches the empty case. But it's a latent `IndexError` if a future ORT device type is added that isn't in `_DEVICE_PRIORITY` — `device_ep_map` would be non-empty while `available_devices` is empty after filtering.

A defensive guard costs nothing:

```python
if device == "auto":
    if not available_devices:
        raise RuntimeError("No execution providers detected.")
    chosen = available_devices[0]
```

src/winml/modelkit/sysinfo/device.py
DingmaomaoBJTU, line 184: **Inconsistent EP source in error messages**

`resolve_device` now uses `_get_device_ep_map_from_ort()` for resolution logic but still calls `_get_available_eps()` here (and on line 222) for error messages. These two functions query different APIs (`get_ep_devices()` vs `get_available_providers()`) and can disagree — an EP could appear in `get_available_providers()` but not in `get_ep_devices()`, producing confusing messages like "EP X not available" while X is listed in the available set.

Consider deriving error-message EPs from the same source:

```python
all_registered_eps = {ep for eps in _get_device_ep_map_from_ort().values() for ep in eps}
```

src/winml/modelkit/winml.py
```diff
@@ -158,7 +161,7 @@ def add_ep_for_device(
     ep_devices = ort.get_ep_devices()
     for ep_device in ep_devices:
         if ep_device.ep_name == ep_name and ep_device.device.type == device_type:
-            print(f"Adding {ep_name} for {device_type}")
+            logger.info(f"Adding {ep_name} for {device_type}")
```
DingmaomaoBJTU: Nit: use lazy `%s` formatting instead of eager f-string — consistent with the `logger.info(...)` calls you wrote in `device.py` and avoids formatting cost when INFO is disabled:

```python
logger.info("Adding %s for %s", ep_name, device_type)
```

tests/unit/sysinfo/test_device.py
```diff
@@ -21,97 +20,105 @@
 from winml.modelkit.utils.constants import EP_NAMES
 
 
+def _make_ep_device(device_type) -> MagicMock:
```
DingmaomaoBJTU: Nit: `_make_ep_device` sets `.device.type` but not `.ep_name`, so when `_get_device_ep_map_from_ort()` iterates these mocks, it stores auto-generated `MagicMock` objects as EP names. Currently harmless because `_get_available_devices` only reads dict keys, but fragile if any future test inspects the EP names from `_get_device_ep_map_from_ort()` output.

Consider:

```python
def _make_ep_device(device_type, ep_name: str = "TestEP") -> MagicMock:
    mock = MagicMock()
    mock.device.type = device_type
    mock.ep_name = ep_name
    return mock
```

DingmaomaoBJTU (changes requested):
Overall the direction is right — using `get_registered_ep_devices()` (the authoritative ORT API) instead of WMI hardware probes eliminates the DML-GPU false positive that caused the original crash. Nice simplification of the test mocking too — single mock point via `_patch_device_ep_map` is much cleaner.

A few issues to address before merging (1 bug, 2 robustness, 2 nits).

src/winml/modelkit/sysinfo/device.py
xieofxie: Good point! I will also use ort.get_ep_devices() for _get_available_eps

src/winml/modelkit/sysinfo/device.py
xieofxie: already check against device_ep_map

src/winml/modelkit/winml.py
xieofxie: good catch !

src/winml/modelkit/sysinfo/device.py
xieofxie: updated _get_available_eps

timenick (approved):

DingmaomaoBJTU (approved):

PR #716, opened 2026-05-25
feat(monitor): query all NPU/GPU engines and take max utilization

xieofxie:
## Summary
- `PdhPoller` previously locked onto a single engine per adapter (first alphabetical `Compute*` for NPU, `3D` for GPU). On NPUs with multiple `Compute_*` engines, work scheduled to a non-monitored engine read as 0% utilization; on GPUs where DML lands on a Compute engine, the `3D`-only query missed the activity entirely.
- `build_adapter_query` now takes `engine_types: tuple[str, ...]` and registers one `util_<engtype>` + `running_time_<engtype>` counter per matching engine. `build_npu_query` covers all `Compute_*`; `build_gpu_query` covers both `3D` and `Compute_*`.
- `PdhPoller._poll_loop` aggregates samples with `max()` across `util_*` keys — each counter is a per-engine ratio over the same sample window, so max reports the most-loaded engine. Summing percentages would exceed 100% on multi-engine adapters and duplicate the signal that running-time already provides.
- `running_time_delta_ns` sums per-engine deltas — Running Time is wall-clock ns per engine, and engines are independent HW that can run in parallel, so total adapter compute time is additive.
- Adapter resolution (`resolve_adapter_luid`) is untouched; this only changes how counters are registered and aggregated for the resolved adapter.

### DIFF
--- a/src/winml/modelkit/session/monitor/_pdh.py
+++ b/src/winml/modelkit/session/monitor/_pdh.py
@@ -241,27 +241,32 @@ class PdhQuery:
 
 def build_adapter_query(
     luid: str,
-    engine_type: str = "Compute",
+    engine_types: tuple[str, ...] = ("Compute",),
     *,
     pid: int | None = None,
 ) -> PdhQuery:
     """Build a PdhQuery for any GPU/NPU adapter.
 
-    Generic builder that works for NPU (Compute engine), GPU (3D engine),
-    or any other adapter type.
+    Registers per-process Utilization Percentage and Running Time counters
+    for *every* engine on the adapter whose engtype starts with one of
+    *engine_types*. NPUs typically expose several ``Compute_*`` engines and
+    the runtime may schedule work on any of them; GPUs run DML on either the
+    ``3D`` engine or a ``Compute_*`` engine depending on the operator. Names
+    are suffixed with the engtype (``util_Compute_0``, ``running_time_3D``,
+    ...) so callers can aggregate (e.g. max utilization) across engines.
 
     Args:
         luid: Adapter LUID string (e.g. ``"0x00000000_0x00015A33"``).
-        engine_type: Engine type to monitor (e.g. ``"Compute"`` for NPU,
-            ``"3D"`` for GPU). Must match an engine type on the adapter.
+        engine_types: Engtype prefixes to register. An engine is included
+            when its engtype starts with any of these strings.
         pid: Process ID to monitor. Defaults to current process.
 
     Returns:
-        An opened PdhQuery with utilization, running time, and memory
-        counters registered.
+        An opened PdhQuery with util/running-time counters for every matched
+        engine plus the adapter's per-process memory counters.
 
     Raises:
-        ValueError: If the LUID or engine type is not found.
+        ValueError: If the LUID is unknown or no engtype matches.
     """
     if pid is None:
         pid = os.getpid()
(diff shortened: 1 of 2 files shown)


xieofxie:
Also tested vitisa, migraphx

timenick (commented):
probe

timenick (commented):
Two findings on the multi-engine PDH aggregation: the PR is *about* aggregation correctness, so I want it pinned by tests and documented for future readers.

src/winml/modelkit/session/monitor/_pdh.py
```diff
@@ -518,7 +534,15 @@ def _poll_loop(self) -> None:
...
-                util = values.get("utilization_pct")
+                # Take the max across all engines for this adapter — the
+                # workload can land on any of the registered engtypes and only
+                # the active one reports non-zero utilization.
+                util_vals = [
+                    v
+                    for k, v in values.items()
+                    if k.startswith("util_") and v is not None
+                ]
+                util = max(util_vals) if util_vals else None
```
timenick, line 546: **The aggregation logic this PR adds has zero unit-test coverage.**

The PR’s headline fix is the aggregation: `max(util)` here at line 545 and `sum(per-engine deltas)` at line 740 in `running_time_delta_ns`. But the only new assertion in `test_ep_monitor.py` is on `query.counter_names`:

```python
assert any(n.startswith("util_Compute") for n in names)
assert any(n.startswith("running_time_Compute") for n in names)
```

That verifies *which counters are registered*, not that the aggregation does the right thing across them. The next refactor that swaps `max` → `mean` (or `sum` → `max` on the other line) will not be caught by CI — which is the exact failure mode this PR was filed to fix in the first place ("locked onto a single engine… read as 0% utilization").

A small mock-based test would lock both aggregators in:

```python
def test_poll_loop_takes_max_util_across_engines(monkeypatch):
    poller = PdhPoller.__new__(PdhPoller)  # bypass __init__
    poller._query = MagicMock()
    poller._query._collect_once.return_value = {
        "util_Compute_0": 80.0,
        "util_Compute_1": 30.0,
        "util_3D": 10.0,
        "memory_local_bytes": None,
        "memory_shared_bytes": None,
        "cpu_pct_raw": None,
        "ram_working_set_bytes": None,
    }
    poller._stop_event = threading.Event()
    # … set up the minimal state _poll_loop reads …
    # one iteration:
    poller._stop_event.set()  # exit after first iter
    poller._poll_loop()
    assert poller._util_samples[-1] == 80.0


def test_running_time_delta_sums_across_engines():
    poller = PdhPoller.__new__(PdhPoller)
    poller._running_time_start_ns = {"running_time_Compute_0": 1000, "running_time_3D": 500}
    poller._running_time_end_ns = {"running_time_Compute_0": 1500, "running_time_3D": 800}
    assert poller.running_time_delta_ns == 800  # 500 + 300
```

Non-blocking, but these are <30 lines and they pin the exact contract the PR is establishing.

src/winml/modelkit/session/monitor/_pdh.py
```diff
@@ -702,10 +726,19 @@ def cpu_sample_count(self) -> int:
...
+        Sums per-engine deltas because the Running Time counter is
+        engine-scoped and a single workload can use several engines in
+        parallel — total adapter compute time is their sum, not the max.
+        """
+        total = 0
+        for key, end in self._running_time_end_ns.items():
+            start = self._running_time_start_ns.get(key)
+            if start is None:
+                continue
+            total += max(0, end - start)
```
timenick, line 742: **`max` for utilization vs `sum` for running-time can produce inconsistent stats.**

The two aggregators answer different questions, both legitimate in isolation:
- `max(util)` (line 545): *"is the adapter busy *right now*"* — useful for adapter saturation.
- `sum(running_time)` (here): *"total engine-time across the adapter"* — useful for throughput accounting.

The asymmetry surfaces when engines run in **true parallel**. Concrete example: two compute engines each at ~50% duty for 1s wall time → `max(util) ≈ 50%`, but `sum(running_time) / wall_time = (0.5s + 0.5s) / 1s = 100%`. Any consumer that derives a "% busy" from `running_time_delta_ns / wall_time` and compares against `mean_util_pct` will get numbers that disagree by 2×.

This is defensible — you almost certainly *want* sum for throughput accounting — but the existing docstring (`"total adapter compute time is their sum, not the max"`) is *why* you sum, not the consequence. Worth a one-line user-facing note in the report consumer or here:

```python
@property
def running_time_delta_ns(self) -> int:
    """Total adapter (NPU/GPU) running time delta in nanoseconds.

    Sums per-engine deltas because the Running Time counter is engine-scoped
    and a single workload can use several engines in parallel.

    Note: this is *total engine-time*, not adapter wall-time. On parallel
    workloads this can exceed wall-time, and `running_time_delta_ns /
    wall_time` may exceed mean_util_pct (which is max-across-engines, not
    sum). Prefer mean_util_pct for "how busy was the adapter" and this
    counter for "how much compute time did this run cost".
    """
```

Non-blocking — just want the next person reading the perf report to not be confused.

xieofxie: doc updated

src/winml/modelkit/session/monitor/_pdh.py
xieofxie: added!

timenick (approved):

PR #717, opened 2026-05-25
fix(inspect): make --list-tasks return in <500 ms (#544)

### LINKED ISSUE #544: [winml inspect] [P0] `--list-tasks` takes 12.6 s for what should be a static dict lookup
**Severity**: P1  
**Command**: `winml inspect`

---
**Category**: Local Command Latency — informational queries should return in <500 ms.

**Repro**:
```
Measure-Command { uv run winml inspect --list-tasks }
```

**Actual**: ~12.6 s warm. Output is a static list of ~22 task names.

**Expected**: <200 ms. Move the task list to a hand-coded constant in `loader/task.py`; do not import `optimum.exporters` to enumerate it at runtime.

**Why it matters**: This is a pure documentation-display call. A 12-second wait to read 22 strings is indefensible.

timenick:
## Summary

`winml inspect --list-tasks` is a pure documentation-display call that takes ~12 s warm. Fixes #544 by hand-coding the canonical task set and bypassing the resolver import chain that pulls in `transformers`/`optimum`.

## Root cause

[commands/inspect.py:150-155](../blob/main/src/winml/modelkit/commands/inspect.py#L150-L155) routed `--list-tasks` through [inspect.resolver.get_known_tasks](../blob/main/src/winml/modelkit/inspect/resolver.py#L59-L88). `resolver.py` imports from `..models`, which transitively imports `transformers`. The optimum branch inside `get_known_tasks` also imported `optimum.exporters.tasks.TasksManager` — and the legacy attributes it looked at (`_TASKS_TO_LIBRARY`, `_TASKS_TO_AUTOMODELS`) no longer exist in the pinned optimum, so the branch was paying ~10 s of imports to contribute zero tasks.

## Change

- Added `KNOWN_TASKS: frozenset[str]` to `src/winml/modelkit/loader/task.py` — a hand-coded snapshot of the union of optimum's `TasksManager.get_all_tasks()` and our own registry (35 tasks).
- Rewrote `--list-tasks` in `commands/inspect.py` to import `KNOWN_TASKS` directly from `loader.task`, bypassing `inspect.resolver` and `..models`.
- Simplified `get_known_tasks()` to a small union over the local registry plus `KNOWN_TASKS`, with no optimum import.

## Guard rails

- `tests/unit/loader/test_known_tasks.py` — asserts `KNOWN_TASKS` is a superset of `optimum.exporters.tasks.TasksManager.get_all_tasks()` and of every task in `HF_TASK_DEFAULTS` / `HF_MODEL_CLASS_MAPPING`. If a future optimum upgrade adds a task, CI fails with a clear message telling the maintainer to update `KNOWN_TASKS`.
- `tests/cli/test_import_time.py::TestCommandHelp::test_inspect_list_tasks_no_heavy_deps` — fails if `transformers`/`optimum`/`torch` ever sneak back into the `--list-tasks` path.

## Measurements (warm cache, this branch)

| Branch | run 1 | run 2 | run 3 |
|---|---|---|---|
| main | 11.19 s | 11.14 s | 11.68 s |
| this PR | 0.62 s | 0.45 s | 0.44 s |

~25× faster. Doesn't hit the <200 ms target from the issue (bare `winml --version` is ~450 ms because `winml.modelkit.utils.constants` imports `onnxruntime`), but the perceived-hang behavior is gone.

Closes #544.


DingmaomaoBJTU (commented):
很好的优化，12s → 450ms 效果显著。手写 KNOWN_TASKS + CI guard 的方案是合理的。几个 inline comments。

src/winml/modelkit/loader/task.py
```diff
@@ -89,6 +89,52 @@
...
+        "text-generation",
+        "text-to-audio",
+        "text-to-image",
+        "text2text-generation",
+        "time-series-forecasting",
+        "token-classification",
+        "visual-question-answering",
+        "zero-shot-image-classification",
+        "zero-shot-object-detection",
+    }
```
DingmaomaoBJTU, line 134: `KNOWN_TASKS` (35 tasks) and `TASK_ABBREV` (47 tasks, line 36-89) in the same file have a non-trivial difference. Tasks in `TASK_ABBREV` but missing from `KNOWN_TASKS` include: `instance-segmentation`, `universal-segmentation`, `video-classification`, `sequence-classification`, `text-encoding`, `summarization`, `translation`, `pretraining`, `table-question-answering`, `zero-shot-classification`, `audio-tokenization`, `zero-shot-audio-classification`, `any-to-any`, `multimodal-lm`, `backbone`, `time-series-prediction`, etc.

Two hand-maintained task lists in the same file are prone to drift. Suggestions:
1. Add a test assertion `KNOWN_TASKS >= set(TASK_ABBREV)` in `test_known_tasks.py` so CI catches it when someone adds a task to `TASK_ABBREV` but forgets `KNOWN_TASKS`.
2. Or derive `KNOWN_TASKS` from `TASK_ABBREV`: `KNOWN_TASKS = frozenset(TASK_ABBREV) | frozenset({...extra tasks...})`.

If the gap is intentional (e.g., `TASK_ABBREV` includes non-exportable tasks that shouldn't appear in `--list-tasks`), it would help to add a comment above `KNOWN_TASKS` clarifying the scoping rule.

src/winml/modelkit/inspect/resolver.py
```diff
@@ -57,34 +58,17 @@
...
-
-        if hasattr(TasksManager, "_TASKS_TO_LIBRARY"):
-            tasks.update(TasksManager._TASKS_TO_LIBRARY.keys())
-        if hasattr(TasksManager, "_TASKS_TO_AUTOMODELS"):
-            tasks.update(TasksManager._TASKS_TO_AUTOMODELS.keys())
-    except Exception:
-        pass
-
+    tasks.update(task for _, task in HF_MODEL_CLASS_MAPPING if task is not None)
     return tasks
```
DingmaomaoBJTU, line 90: `get_known_tasks()` unions `KNOWN_TASKS` + `HF_TASK_DEFAULTS` + `HF_MODEL_CLASS_MAPPING`, but `commands/inspect.py`'s `--list-tasks` now imports `KNOWN_TASKS` directly — bypassing `get_known_tasks()`.

This means if someone adds a new task to `HF_TASK_DEFAULTS` or `HF_MODEL_CLASS_MAPPING` without updating `KNOWN_TASKS`, `validate_task()` will accept it but `--list-tasks` won't show it — the user-visible task list and the actual accepted set diverge.

The CI tests (`test_covers_hf_task_defaults`, `test_covers_hf_model_class_mapping`) catch this drift, so it won't silently break. But consider either:
- Having `--list-tasks` also go through `get_known_tasks()` (if it can be made fast in the future via lazy import), or
- Adding a note in `get_known_tasks()`'s docstring explaining why `--list-tasks` doesn't use it and why the two sets may differ.

tests/unit/loader/test_known_tasks.py
```diff
@@ -0,0 +1,78 @@
...
+
+    def test_entries_are_canonical_task_names(self) -> None:
+        """Each entry is lowercase, hyphenated, no whitespace."""
+        import re
+
+        pattern = re.compile(r"^[a-z][a-z0-9-]*$")
+        bad = [t for t in KNOWN_TASKS if not pattern.match(t)]
+        assert not bad, f"Non-canonical task names in KNOWN_TASKS: {bad}"
+
+
```
DingmaomaoBJTU, line 40: Good structural invariant test. Consider adding an assertion for `KNOWN_TASKS >= set(TASK_ABBREV.keys())` to prevent drift between the two hand-maintained task lists in `task.py`. See the comment on `task.py` for details.

src/winml/modelkit/commands/inspect.py
```diff
@@ -150,11 +150,14 @@ def inspect(
...
+    # Handle --list-tasks (no model required).
+    # Import the hand-coded KNOWN_TASKS directly from loader.task to keep this
+    # branch fast — going through inspect.resolver pulls in ..models which
+    # transitively imports transformers and costs ~10s on a warm cache.
     if list_tasks:
-        from ..inspect.resolver import get_known_tasks
+        from ..loader.task import KNOWN_TASKS
 
-        for t in sorted(get_known_tasks()):
+        for t in sorted(KNOWN_TASKS):
```
DingmaomaoBJTU, line 160: Nit: `sorted()` on a `frozenset` is O(n log n) per call. Only runs once here so it's a non-issue — just noting for awareness.

DingmaomaoBJTU (commented):
Great optimization — 12s down to 450ms is a massive win. The hand-coded KNOWN_TASKS + CI guard approach is sound. A few inline comments.

src/winml/modelkit/loader/task.py
timenick: Thanks for flagging. I dug into the 18-task gap: 5 are aliases (`pretraining` → `feature-extraction`, `summarization` → `text2text-generation`, etc.) that legitimately should NOT be in `KNOWN_TASKS`. The other 13 (`video-classification`, `table-question-answering`, `keypoint-matching`, `any-to-any`, …) are a real pre-existing inconsistency — they have `TASK_ABBREV` entries and are referenced by `inference/tasks.py`, `models/hf/bart.py`, and the e2e testsets, but `validate_task()` rejects them.

Pre-dates this PR (`TASK_ABBREV` has been like this since the file was added). A naive `TASK_ABBREV ⊆ KNOWN_TASKS` test would either force the 13 dubious tasks into scope here or silently filter aliases — neither feels right for a perf PR.

Filed #724 to track the consolidation. Not addressing in this PR.

Generated with Claude Code

src/winml/modelkit/inspect/resolver.py
timenick: Good point — the dual path (`--list-tasks` reads `KNOWN_TASKS` directly; `validate_task` goes through `get_known_tasks()`) isn't obvious. Added a docstring note to `get_known_tasks` explaining the split and pointing at the CI tests in `test_known_tasks.py` that guard the invariant (commit 2fcbec4f).

Generated with Claude Code

tests/unit/loader/test_known_tasks.py
timenick: See reply on the `task.py` comment — same topic. tl;dr: a strict `TASK_ABBREV ⊆ KNOWN_TASKS` test fails today because of 5 aliases (legitimately not in KNOWN_TASKS) and 13 canonical tasks that ARE referenced elsewhere in the codebase but rejected by `validate_task`. Tracked in #724, not addressing in this PR.

Generated with Claude Code

src/winml/modelkit/commands/inspect.py
timenick: Noted, thanks.

Generated with Claude Code

DingmaomaoBJTU (approved):

PR #719, opened 2026-05-25
perf(inspect): dedupe AutoConfig + gate processor Auto* lookups (#543)

### LINKED ISSUE #543: [winml inspect] [P0] `winml inspect -m <hf_id>` takes 24 s end-to-end; first user-visible output silent for ~14 s
**Severity**: P1  
**Command**: `winml inspect`

---
**Category**: Time-to-First-Output — first byte to user must arrive within 500 ms of entry.

**Repro**:
```
Measure-Command { uv run winml inspect -m microsoft/resnet-50 }
```

**Actual**: ~24.3 s wall time on warm cache. The first ~14 s after pressing Enter shows nothing on screen, then the panels render almost instantly.

**Expected**: Print a banner immediately on entry (`Inspecting microsoft/resnet-50 …`), show a `rich.status` spinner during HF metadata fetch. `inspect` should never download model weights — only `config.json` is needed. With cached `config.json`, drop to <2 s.

**Why it matters**: A 14-second silence makes the user assume the command hung and Ctrl-C; then they re-run, get the same silence, and the issue compounds. A spinner inside the first 500 ms removes the ambiguity.

timenick:
## Summary

Follow-up to #718 (banner + spinner). The spinner removed the perceived-hang UX problem but the total wall time was still ~23 s. This PR cuts the resolver-side redundancy to bring that to ~18 s.

## Profile (warm cache, `microsoft/resnet-50`)

Before this PR (per-call counts inside `_inspect_model_v2`):

```
0.71s  AutoConfig.from_pretrained   ← explicit fetch in inspect.py
0.38s  AutoConfig.from_pretrained   ← duplicate inside resolve_loader_config
0.72s  resolve_io_specs
0.38s  hf_hub_download              ← preprocessor_config.json
0.31s  hf_hub_download              ← tokenizer_config.json (404)
0.70s  AutoTokenizer.from_pretrained
0.71s  AutoImageProcessor.from_pretrained
3.53s  AutoProcessor.from_pretrained
0.38s  AutoConfig.from_pretrained
0.70s  AutoTokenizer.from_pretrained
0.69s  AutoImageProcessor.from_pretrained
4.20s  AutoFeatureExtractor.from_pretrained
─────
12.33s total inspect work
```

After this PR:

```
… (same fast portion) …
0.69s  AutoImageProcessor.from_pretrained
3.49s  AutoProcessor.from_pretrained
0.70s  AutoTokenizer.from_pretrained
─────
9.29s total inspect work
```

`AutoConfig` calls 5 → 4, `AutoFeatureExtractor` entirely eliminated, duplicate `AutoImageProcessor` skipped.

End-to-end wall time (`uv run winml inspect -m microsoft/resnet-50 --format json`):

| Branch | run 1 | run 2 | run 3 |
|---|---|---|---|
| main | 23.4 s | 23.2 s | 23.0 s |
| this PR | 17.6 s | 17.7 s | 18.8 s |

### 1. Dedupe `AutoConfig.from_pretrained`

`_inspect_model_v2` already loads the parent `hf_config` (it needs the un-narrowed config for I/O introspection — `resolve_loader_config` may narrow it to a sub-config for multimodal models like CLIP). Then `resolve_loader_config` loaded it again.

Added an `hf_config=` kwarg to [`resolve_loader_config`](src/winml/modelkit/loader/config.py#L116) — when supplied, step 1's `AutoConfig.from_pretrained` is skipped. Pass-through from [`_inspect_model_v2`](src/winml/modelkit/commands/inspect.py#L315). Other callers default to `hf_config=None`, original behaviour preserved.

### 2. Gate processor Auto* lookups per-field

[`resolve_processor`](src/winml/modelkit/inspect/resolver.py#L820) ran every Auto* class instantiation in Strategy 2, even when Strategies 0+1 had already populated the corresponding field. Added `try_*` kwargs to [`_resolve_processor_from_auto_classes`](src/winml/modelkit/inspect/resolver.py#L1000) and skip each Auto* whose field is already known.

For ResNet, Strategies 0+1 set `image_processor_class` from the HF registry, so `try_image_processor=False` skips the redundant `AutoImageProcessor.from_pretrained`. `AutoProcessor` succeeds and supplies `feature_extractor_class` as a side effect, so `AutoFeatureExtractor` (the 4.2s call) is skipped via the existing inner gate.

If Strategies 0+1 already populate all four fields (CLIP-like models), Strategy 2 is skipped entirely.

## Why not further?

The remaining 9 s of resolver work is inside transformers' Auto* internals — `AutoProcessor.from_pretrained` does its own AutoConfig load and may construct sub-processors. Cutting deeper requires upstream restructuring, not a ModelKit-side fix.

## Tests

- `test_resolve_loader_config.py::test_hf_config_kwarg_skips_autoconfig_fetch` — pins the dedupe by patching `AutoConfig.from_pretrained` and asserting `call_count == 0` when `hf_config=` is supplied.
- `test_resolve_processor_gating.py` — three cases:
  - All four fields filled by Strategy 1 → `_resolve_processor_from_auto_classes` is not called at all
  - Partial fill → only the still-needed `try_*` flags are True
  - Nothing filled → every `try_*` flag is True

112 targeted tests pass.

Closes #543.

### DIFF
--- a/src/winml/modelkit/inspect/resolver.py
+++ b/src/winml/modelkit/inspect/resolver.py
@@ -1043,17 +1043,19 @@ def _resolve_processor_from_auto_classes(
     image_processor_class: str | None = None
     feature_extractor_class: str | None = None
 
-    # Try AutoProcessor first - for multimodal models.
-    # AutoProcessor's wrapper can supply tokenizer / image_processor /
-    # feature_extractor as a side effect, so we also run it when one of
-    # those is still needed even though the processor itself isn't.
-    if try_processor or try_tokenizer or try_image_processor or try_feature_extractor:
+    # Try AutoProcessor only when the processor class itself is needed.
+    # AutoProcessor.from_pretrained is the single most expensive Auto* call
+    # (~3.5s warm). When the caller only needs sub-pieces (tokenizer /
+    # image_processor / feature_extractor) we fall through to the standalone
+    # Auto* calls below, which are individually cheaper. AutoProcessor can
+    # still fill those sub-fields as a side effect on success — that's
+    # preserved here when ``try_processor`` is True.
+    if try_processor:
         try:
             from transformers import AutoProcessor
 
             processor = AutoProcessor.from_pretrained(model_id, use_fast=True)
-            if try_processor:
-                processor_class = type(processor).__name__
+            processor_class = type(processor).__name__
 
             # AutoProcessor may wrap tokenizer and image_processor
             if (
--- a/src/winml/modelkit/inspect/resolver.py
+++ b/src/winml/modelkit/inspect/resolver.py
@@ -885,31 +885,45 @@ def resolve_processor(
     except Exception as e:
         logger.debug("Failed to resolve processors from hub configs: %s", e)
 
-    # Strategy 2: Use Auto classes to fill in any missing information
-    # This approach actually loads the processors, so it's slower but more reliable
-    try:
-        (
-            auto_processor,
-            auto_tokenizer,
-            auto_image_processor,
-            auto_feature_extractor,
-        ) = _resolve_processor_from_auto_classes(model_id)
+    # Strategy 2: Use Auto classes to fill in any missing information.
+    # Skip entirely when Strategies 0 + 1 already populated every field —
+    # each Auto* instantiation does its own HF Hub I/O plus class init
+    # (AutoProcessor and AutoFeatureExtractor are several seconds each).
+    need_processor = processor_class is None
+    need_tokenizer = tokenizer_class is None
+    need_image_processor = image_processor_class is None
+    need_feature_extractor = feature_extractor_class is None
 
-        # Fill in missing values from auto classes
-        if processor_class is None and auto_processor:
-            processor_class = auto_processor
-            processor_source = "auto_class"
-        if tokenizer_class is None and auto_tokenizer:
-            tokenizer_class = auto_tokenizer
-            tokenizer_source = "auto_class"
-        if image_processor_class is None and auto_image_processor:
-            image_processor_class = auto_image_processor
-            image_processor_source = "auto_class"
-        if feature_extractor_class is None and auto_feature_extractor:
-            feature_extractor_class = auto_feature_extractor
-            feature_extractor_source = "auto_class"
-    except Exception as e:
-        logger.debug("Failed to resolve processors from auto classes: %s", e)
+    if need_processor or need_tokenizer or need_image_processor or need_feature_extractor:
+        try:
+            (
+                auto_processor,
+                auto_tokenizer,
+                auto_image_processor,
+                auto_feature_extractor,
+            ) = _resolve_processor_from_auto_classes(
+                model_id,
+                try_processor=need_processor,
+                try_tokenizer=need_tokenizer,
+                try_image_processor=need_image_processor,
+                try_feature_extractor=need_feature_extractor,
+            )
+
+            # Fill in missing values from auto classes
+            if need_processor and auto_processor:
+                processor_class = auto_processor
+                processor_source = "auto_class"
+            if need_tokenizer and auto_tokenizer:
+                tokenizer_class = auto_tokenizer
+                tokenizer_source = "auto_class"
+            if need_image_processor and auto_image_processor:
+                image_processor_class = auto_image_processor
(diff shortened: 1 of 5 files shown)


DingmaomaoBJTU (commented):
Overall this is a well-motivated performance optimization with clean implementation and good test coverage. Two comments inline.

src/winml/modelkit/inspect/resolver.py
```diff
@@ -1007,38 +1026,63 @@ def _resolve_processor_from_auto_classes(
 
     Args:
         model_id: HuggingFace model identifier
+        try_processor: When False, skip ``AutoProcessor.from_pretrained``.
+            AutoProcessor is the most expensive single call (several seconds
+            even on warm cache), so callers that already know all four classes
+            should pass False for every flag to skip Strategy 2 entirely.
+        try_tokenizer: When False, skip ``AutoTokenizer.from_pretrained``.
+        try_image_processor: When False, skip ``AutoImageProcessor.from_pretrained``.
+        try_feature_extractor: When False, skip ``AutoFeatureExtractor.from_pretrained``.
```
DingmaomaoBJTU, line 1035: The guard `if try_processor or try_tokenizer or try_image_processor or try_feature_extractor` is redundant with the outer guard in `resolve_processor` — when all four `need_*` flags are False the caller never invokes `_resolve_processor_from_auto_classes` at all, so at least one `try_*` is guaranteed True here. This guard will never short-circuit.

More importantly: when `try_processor=False` but other fields are still needed, the current code still runs `AutoProcessor.from_pretrained` (~3.5s) — it just skips assigning `processor_class`. The intent (AutoProcessor can fill tokenizer/image_processor/feature_extractor as a side effect) is sound, but there's a missed optimization opportunity:

```python
# Only call AutoProcessor when the processor itself is needed,
# or fall through to the cheaper standalone Auto* calls below.
if try_processor:
    # full AutoProcessor path (may fill sub-fields as side effect)
    ...
```

For example, if Strategy 0+1 already filled `image_processor` and `feature_extractor` and only `tokenizer` is still needed, skipping AutoProcessor and going straight to AutoTokenizer would save ~3.5s.

Not blocking — fine as a follow-up.

src/winml/modelkit/inspect/resolver.py
```diff
@@ -885,31 +885,45 @@ def resolve_processor(
...
+
+            # Fill in missing values from auto classes
+            if need_processor and auto_processor:
+                processor_class = auto_processor
+                processor_source = "auto_class"
+            if need_tokenizer and auto_tokenizer:
+                tokenizer_class = auto_tokenizer
+                tokenizer_source = "auto_class"
+            if need_image_processor and auto_image_processor:
+                image_processor_class = auto_image_processor
```
DingmaomaoBJTU, line 920: Nit: the four `need_*` booleans are only used inside the Strategy 2 block. Could be condensed with a dict + `any(...)`, but the current form reads just as clearly. No action needed.

DingmaomaoBJTU (commented):
Well-motivated perf optimization with clean implementation and solid test coverage. Two inline comments.

src/winml/modelkit/inspect/resolver.py
timenick: You're right on both counts. Applied in 663d1427:

- Dropped the dead inner OR-guard — the outer guard in `resolve_processor` already short-circuits Strategy 2.
- Gated `AutoProcessor.from_pretrained` on `try_processor` only. When `try_processor=False` we fall through to the cheaper standalone Auto* calls instead of paying the AutoProcessor cost just for side-effect sub-pieces.

For ResNet the path is unchanged (`processor_class` is still None after Strategy 0+1, so `try_processor=True`). The win lands on models that have `processor_class` in `preprocessor_config.json` but a missing `tokenizer_config.json` — there we now skip the ~3.5s AutoProcessor round-trip.

New test `test_try_processor_false_skips_autoprocessor` pins the gate by asserting `AutoProcessor.from_pretrained` is not called when `try_processor=False`.

Generated with Claude Code

src/winml/modelkit/inspect/resolver.py
timenick: Agreed, leaving as is.

Generated with Claude Code

DingmaomaoBJTU (approved):

PR #720, opened 2026-05-25
 Fix CLI option / config file value priority across commands

### LINKED ISSUE #721: Dataclass defaults silently override CLI option defaults when `-c/--config` is used
## Summary
 
 For several `winml` commands that accept both CLI options and a `-c/--config` JSON file, dataclass field defaults silently shadow the CLI option's `default=` value. The intended priority is:
 
 1. CLI option explicitly passed
 2. JSON key explicitly present in `-c` file
 3. CLI option's `default=` value
 4. Dataclass field default (must NOT shadow #3)
 
 In practice, #4 wins over #3 whenever the merge logic reads from `WinMLBuildConfig.from_dict(json)`, because that constructor populates missing keys with dataclass defaults. The merge code then treats those defaults as if they had come from JSON.
 
 ## Reproducible examples
 
 Each row below assumes the CLI option is NOT passed and the `-c` file has the contents shown.
 
 ### `winml eval` — `--ep`
 
 | `-c` file contents | Actual `cfg.ep` | Expected |
 |---|---|---|
 | *(no `-c`)* | `None` | `None` |
 | `{"compile": {}}` | **`"qnn"`** | `None` |
 | `{"compile": {"execution_provider": "dml"}}` | `"dml"` | `"dml"` |
 
 `"qnn"` comes from the `EPConfig.provider` dataclass default. On Linux or non-Qualcomm Windows hosts, this forces the session layer to try an EP that isn't installed and crashes.
 
 ### `winml eval` — `--samples` (eval dataset size)
 
 | `-c` file contents | Actual `cfg.dataset.samples` | Expected |
 |---|---|---|
 | *(no `-c`)* | `100` | `100` |
 | `{"quant": {}}` | **`10`** | `100` |
 | `{"quant": {"samples": 50}}` | **`50`** | `100` |
 
 Two distinct issues here:
 
 - `10` is the `WinMLQuantizationConfig.samples` dataclass default — a Tier-4 leak.
 - The current code also semantically conflates two different concepts: `quant.samples` is the **calibration** sample count used during model build (~10–100 samples is typical); `--samples` / `cfg.dataset.samples` is the **eval** sample count (~100–1000 is typical, often on a different dataset). They should be independent. Users wanting an eval-side override should write `{"eval": {"dataset": {"samples": N}}}`.
 
 ### `winml eval` — `--dataset-name`
 
 | `-c` file contents | Actual `cfg.dataset.name` | Expected |
 |---|---|---|
 | *(no `-c`)* | `None` | `None` |
 | `{"quant": {"dataset_name": "mrpc"}}` | **`"mrpc"`** | `None` |
 
 Same semantic conflation as `--samples`: `quant.dataset_name` is the calibration dataset, not the eval dataset.
 
 ### `winml compile`, `winml perf`, `winml analyze` — same shape
 
 Empty / partial JSON sections cause CLI defaults to be overridden by dataclass defaults of unrelated nested configs (e.g. `EPConfig.provider = "qnn"` overriding `--ep`'s default `None`).
 
 ### `winml export` — latent (no current observable defect)
 
 `_build_export_cfg.to_dict()` populates `config_kwargs` with every dataclass field default of `WinMLExportConfig`. Today the dataclass defaults match the CLI defaults, so behavior is correct by coincidence. Any future change to a `WinMLExportConfig` field default would silently change CLI behavior.
 
 ## Root cause
 
 `cli_utils.load_build_config(path)` returns a `WinMLBuildConfig` constructed via `from_dict`. For any field absent from JSON, `from_dict` substitutes the dataclass default. The downstream merge code reads `build_cfg.<section>.<field>` and copies it onto the CLI-bound variable — it has no way to tell whether the value came from explicit JSON or from a dataclass default.
 
 ## Suggested fix
 
 Have `load_build_config` also return the raw `dict` from `json.loads`, and have all command-side merge logic read from the raw dict using `"key" in section` instead of from dataclass attributes. Missing keys then stay missing, and only explicit JSON values override CLI defaults.
 
 For `winml eval`, additionally remove the existing "quant fields as fallback" block — quant-section keys must not bleed into eval-dataset configuration.
 
 ## Scope
 
 Affects 6 commands: `compile`, `perf`, `analyze`, `export`, `quantize`, `eval`

zhenchaoni:
Fixes #721 

 ## Problem
 
 Commands accepting both CLI options and a `-c/--config` JSON file were leaking dataclass defaults into CLI behavior. When a JSON section was empty (or absent), `WinMLBuildConfig.from_dict(json)` populated missing keys with dataclass field defaults, and the merge logic could not distinguish "user explicitly set this in JSON" from "dataclass picked the default". The dataclass default then silently overrode the CLI option's default.
 
 Concrete user-visible bugs:
 
 - `winml eval -c config.json` with `{"compile": {}}` forced `cfg.ep = "qnn"` regardless of the host's available EPs (would crash on Linux / non-Qualcomm Windows).
 - `winml eval -c config.json` with `{"quant": {}}` silently reduced the eval dataset to 10 samples (calibration default) instead of the CLI default 100.
 - Even with explicit `{"quant": {"samples": N, "dataset_name": ...}}`, those calibration values bled into the eval dataset config — two unrelated concepts conflated.
 
 ## Fix
 
 Establish a single uniform priority-merge pattern across all 6 commands (`compile`, `perf`, `analyze`, `export`, `quantize`, `eval`):
 
 ```python
 _, raw_cfg = cli_utils.load_build_config(config_file)
 section = raw_cfg.get("section") or {}
 if not cli_utils.is_cli_provided(ctx, "opt") and "key" in section:
     opt = section["key"]
 ```
 
 Reads only raw JSON keys, so missing keys are distinguishable from dataclass defaults by design.
 
 Changes per file:
 
 - `utils/cli.py` — `load_build_config` now returns `(build_cfg, raw_dict)`.
 - `commands/compile.py`, `perf.py`, `analyze.py`, `quantize.py` — switched merge blocks to raw-dict reads.
 - `commands/export.py` — same merge-block fix, plus replaced `_build_export_cfg.to_dict()` (which leaked dataclass defaults into `config_kwargs`) with a raw `export` section.
 - `commands/eval.py` — same merge-block fix for `cfg.task` and `cfg.ep`, plus **removed the "quant fields as fallback" block entirely**: `quant.samples` / `quant.dataset_name` are calibration knobs and must not bleed into the eval dataset config. Users wanting eval-side overrides should use `{"eval": {"dataset": {...}}}`.
 
 ## Tests
 
 New file `tests/unit/commands/test_config_value_priority.py` (replaces the older `test_cli_config_precedence.py`) defines a structured 4-tier priority contract and exercises it across all 6 commands:
 
 - T1 CLI explicit > T2 JSON explicit > T3 CLI default > T4 dataclass default (T4 must not shadow T3)
 - 32 `FieldCase`-parametrized tests + 7 targeted regression guards = **39 tests**
 - Adapters per command capture the effective `WinMLCompileConfig` / `BenchmarkConfig` / `WinMLExportConfig` / `WinMLQuantizationConfig` / `WinMLEvaluationConfig` at the boundary where the command body hands off to business logic.
 
 Each bug fix was driven TDD-style: tests written first, observed failing on pre-fix code, then turning green with the fix.
 
 ## Migration note
 
 `{"quant": {"samples": N}}` in a config file no longer affects `winml eval`'s dataset size. Use `{"eval": {"dataset": {"samples": N}}}` for eval-side overrides. (Calibration-only knobs in `quant` remain valid for build pipelines.)


ssss141414 (approved):

src/winml/modelkit/commands/perf.py
```diff
@@ -1245,13 +1243,16 @@ def perf(
...
     if config_file is not None:
-        build_cfg = cli_utils.load_build_config(config_file)
-        if build_cfg.loader and not cli_utils.is_cli_provided(ctx, "task"):
-            task = build_cfg.loader.task
-        if build_cfg.compile and not cli_utils.is_cli_provided(ctx, "ep"):
-            ep = build_cfg.compile.ep_config.provider
+        _, raw_cfg = cli_utils.load_build_config(config_file)
+        lc = raw_cfg.get("loader") or {}
+        cc = raw_cfg.get("compile") or {}
+        if not cli_utils.is_cli_provided(ctx, "task") and "task" in lc:
```
xieofxie, line 1254: what is the difference between cli_utils.is_cli_provided and check task is not None directly?

zhenchaoni: For `task`, they are equivalent here. But `is_cli_provided` is a safer method used in many places.

`is_cli_provided` can distinguish whether the option is user specified or CLI option default. For example, if an option's `task` default value is "a" but the user doesn't write it in command. `task` will not be NONE but `is_cli_provided` will be false.

Because we want CLI specified value > config file specified value > CLI option default value. So `is_cli_provided` is helpful for us to know if the option is specified or not.

tests/unit/eval/test_eval.py
```diff
@@ -874,23 +874,20 @@ def test_cli_ep_invalid_value_rejected(self):
...
-        fake_build_cfg.loader = None
-        fake_build_cfg.compile.ep_config.provider = "dml"
-        fake_build_cfg.quant = None
+        raw_cfg = {"compile": {"execution_provider": "dml"}}
 
         runner = CliRunner()
         with (
             patch("winml.modelkit.sysinfo.resolve_device", return_value=("gpu", ["gpu", "cpu"])),
             patch(
                 "winml.modelkit.utils.cli.load_build_config",
```
zhenchaoni, line 889: Update test_eval.py to change the way we mock `load_build_config`. The test case is identical to before and assert on the same outcome.

src/winml/modelkit/commands/perf.py
xieofxie: We should doc this `CLI specified value > config file specified value > CLI option default value` some where

src/winml/modelkit/commands/eval.py
```diff
@@ -266,31 +266,19 @@ def _build_eval_config(
...
-            cfg.task = build_cfg.loader.task
+        loader_section = raw.get("loader") or {}
+        if "task" in loader_section:
+            cfg.task = loader_section["task"]
 
         # Compile EP as fallback for --ep
-        if build_cfg.compile and build_cfg.compile.ep_config:
-            cfg.ep = build_cfg.compile.ep_config.provider
-
-        # Quant fields as fallback
```
xieofxie, line 279: no quant now?

xieofxie (approved):

src/winml/modelkit/commands/eval.py
zhenchaoni: quant will still be applied. This change is to avoid quant option changes the eval option. Earlier, if you use 10 samples for quant, the logic will also use 10 samples for eval. Since quant and eval doesn't necessary share the same config, so I avoid using quant configs as fallback for eval configs.

PR #723, opened 2026-05-25
Validate model task in config.

### LINKED ISSUE #520: [winml build] [P1] Config `loader.task` is not validated against `--model` (or against the model architecture); incompatible pair fails mid-build with a cryptic upstream error
**Severity**: P1  
**Command**: `winml build`

---
**Category**: Validate Inputs Up-Front — when two inputs are independently supplied (config file *and* `-m/--model`), their compatibility must be checked at command entry.

**Status**: Behavior derived from static code path inspection (`commands/build.py`, `loader/config.py`). `build` does not expose `--task` on the CLI (`uv run winml build --help | Select-String task` returns nothing — verified), so the only way to set the task is via `loader.task` in the config JSON.

**Repro**:
```
# config.json contains {"loader": {"task": "text-generation"}, ...}
uv run winml build -c config.json -m microsoft/resnet-50 -o temp\bld\
```

**Actual** (from code-path analysis and parallel to another audit finding):
- `resolve_loader_config()` accepts `task="text-generation"` from the config without checking it against `microsoft/resnet-50`'s architecture.
- `TasksManager.get_model_class_for_task("text-generation")` resolves to `AutoModelForCausalLM`.
- `AutoModelForCausalLM.from_pretrained("microsoft/resnet-50")` then raises the same `ValueError: Unrecognized configuration class ResNetConfig … AutoModelForCausalLM` + 100-line config-class dump observed in another audit finding, except now from inside the `build` pipeline (not `export`), after the Setup banner and Stages header have already printed (compounds another audit finding).
- No traceback de-duplication: the user sees the same wall of text emitted by `transformers` plus the build pipeline's own error wrapping.

**Expected**: At command entry, after loading both `--model` and the config, validate that `loader.task` is a supported task for the model architecture (using the same `TasksManager` map). On mismatch, exit 2 before the Setup banner prints, with one line:
```
Error: config.loader.task='text-generation' is not supported for --model microsoft/resnet-50 (architecture: resnet).
Supported tasks: image-classification, image-feature-extraction.
```
Do not download the model, do not print the Setup banner, do not let the upstream `ValueError` reach the user.

**Why it matters**: The combination is silent at the seam between config and CLI. A user who reuses a config from a text model with a vision model (or vice versa) sees a cryptic upstream error from `transformers` rather than a one-line message about their config field. Same root cause as another audit finding but on a different command surface; should be fixed by a single shared validator. Violates R4.1 (validate inputs at entry), R4.2 (reject unsupported combinations), R3.1 (actionable errors).

chinazhangchao:


timenick (commented):
probe

src/winml/modelkit/loader/config.py
```diff
@@ -259,6 +259,71 @@ def resolve_loader_config(
     return loader_config, resolved_hf_config, resolved_class
 
 
+def validate_task_supported_for_model(
```
timenick: **Reintroduces the duplicate `AutoConfig.from_pretrained` that PR #719 just deduped.**

The critical thing about [PR #719](https://github.com/microsoft/winml-cli/pull/719) was that `_inspect_model_v2` and `resolve_loader_config` were each running `AutoConfig.from_pretrained` once, paying ~0.7s on cold cache and ~0.4s on warm cache *per duplicate*. That PR added the `hf_config=` kwarg to `resolve_loader_config` so callers could pre-load once and pass it through.

This function fetches `AutoConfig.from_pretrained` again at line 285 — from a different entry point. Downstream, `_run_single_build → resolve_loader_config` (or `_inspect_model_v2` for non-build callers) will fetch it a second (or third) time, unless they're rewired to receive an `hf_config=` from this preflight.

The failure mode is invisible: on warm cache the user just pays ~0.4s extra; on cold cache it's an additional HF Hub round-trip + retry behavior. For `winml build` this matters because the same flag combination that triggers a preflight error also triggers two-three `AutoConfig` fetches in the happy path.

Two concrete options:

1. **Move the check into `resolve_loader_config`.** It already has `hf_config` in hand (either passed by the caller or just-loaded at line 184), so the validation is free.
2. **Pass `hf_config` through.** Have the preflight accept and return the loaded config, and have `_run_single_build` plumb it into `resolve_loader_config(..., hf_config=preflight_loaded_config)`. This preserves the CLI-entry preflight (good for UX) without paying for two fetches.

Without one of these, the perf-debt PR #719 paid down comes back on the build path.

src/winml/modelkit/commands/build.py
```diff
@@ -224,6 +225,37 @@ def _build_modules(
...
+) -> None:
+    """Validate config loader task(s) against --model architecture.
+
+    This runs at command entry before setup/stage output so incompatible
+    config/model combinations fail with an actionable one-line error.
+    """
+    if model_id is None:
+        return
+
+    from .config import _is_onnx_file
```
timenick: **Crossing sibling-module privacy boundary: `from .config import _is_onnx_file`.**

`_is_onnx_file` is a `_`-prefixed function in `commands/config.py`. The repo's import rules in `CLAUDE.md` allow private-symbol access **only in tests** — source code is supposed to either consume from `__init__.py` or stay within its own module:

> Never reach into internal submodules for symbols exported by `__init__.py`. **Exception**: private `_`-prefixed functions may use internal imports for testing implementation details.

A `commands/build.py` → `commands/config.py` private-symbol import is exactly the cross-coupling that bites later refactors: anyone moving / renaming / inlining `_is_onnx_file` in `config.py` silently breaks `build.py` without any `__init__.py` shielding to catch it.

The helper itself is generic and reusable ("is this string an .onnx file path?"). Two options:

1. **Promote to public `utils/`.** Move `_is_onnx_file` to `utils/cli.py` (or a new `utils/paths.py`) as `is_onnx_file_path`, re-export from `utils/__init__.py`, and import from both `commands/config.py` and `commands/build.py`.
2. **Duplicate the 3-line check.** `Path(model_input).suffix.lower() == ".onnx" and Path(model_input).exists()` (or whatever the actual body is) is small enough that two copies is fine if they’re likely to drift independently.

Option 1 is the canonical fix.

Same issue at line 682 in this file.

src/winml/modelkit/commands/build.py
```diff
@@ -224,6 +225,37 @@ def _build_modules(
     return results
 
 
+def _validate_loader_tasks_for_model(
```
timenick, line 351: **Coverage asymmetry: same UX problem can hit `winml config`, `winml perf`, `winml export`.**

`validate_task_supported_for_model` is generic and lives in `loader/__init__.py` as public API, but it's only wired into `winml build`. The same `(config-says-X, --model-says-Y)` mismatch can land on:

- `winml config -c X.json -m other.onnx --task ...` (config command itself merging conflicting sources)
- `winml perf -m ... --task ...`
- `winml export -m ... --task ...` (existing fallback through `resolve_cfg` catches some cases but the error path is different — see lines around `ONNXConfigNotFoundError` / `ValueError` in `commands/export.py`)

After this PR, the user gets a polished `config.loader.task='X' is not supported for --model Y (architecture: Z). Supported tasks: ...` from `winml build` and a different (or no) preflight error from the rest. That’s an inconsistent CLI surface for the same kind of user mistake.

Two paths:

1. **Wire the preflight into every CLI entrypoint that accepts task + model from independent sources.** The DRY work is already done — the helper is generic. The wiring is a few lines per command.
2. **Document why `build` is special** (the call out at the top of `_validate_loader_tasks_for_model`'s docstring would be the right place) and leave the others alone.

If you go with option 2, the helper probably shouldn’t be public API in `loader/__init__.py` — it should live where its only caller does (`commands/build.py`). Keeping it public signals "feel free to call this from other commands" while only one command does.

timenick (commented):
Three substantive findings. The preflight check works, but two of them are tied to recent changes elsewhere in the repo (#719 dedupe, CLAUDE.md import rules), and one is a coverage-asymmetry question that applies across CLI entrypoints.

timenick (approved):

zhenchaoni (changes requested):
A few concerns with the current acceptance semantics that I think are worth addressing before merge, because they interact with how the rest of the CLI resolves tasks.

### 1. `normalize_task` on both sides collapses modality

The check is:

```python
normalized_supported = {normalize_task(t) for t in supported_tasks}
if normalize_task(task) in normalized_supported:
    return hf_config
```

`normalize_task` wraps `TasksManager.map_from_synonym`, which treats `image-feature-extraction` as a synonym of `feature-extraction`. Collapsing both sides through it means the gate cannot distinguish text vs image FE. Two consequences:

- `--task image-feature-extraction` against a text-only arch (e.g. `bert-base-uncased`) **passes** — the normalized form `feature-extraction` is in BERT's supported list.
- `--task feature-extraction` against a vision-only arch (e.g. dinov2) **passes** — vision arches register `feature-extraction` (Optimum-canonical).

The PR title says "Validate model task" but the validator can't catch cross-modality mismatches. Worse, the second case silently produces wrong downstream behavior: our four task registries (`_EVALUATOR_REGISTRY`, `TASK_DATASET_MAPPING`, `TASK_SCHEMAS`, `TASK_TO_WINML_CLASS`) are keyed by **HF-pipeline task IDs**, where `feature-extraction` is text-only (`FeatureExtractionPipeline` requires a `PreTrainedTokenizer`) and `image-feature-extraction` is a separate pipeline (`ImageFeatureExtractionPipeline`). So a vision model with `loader.task = "feature-extraction"` routes through text logic in eval/quantize.

Suggested change: compare without normalizing, falling back to normalized only when the verbatim check fails. Something like:

```python
if task in supported_tasks:
    return hf_config
if normalize_task(task) in {normalize_task(t) for t in supported_tasks}:
    # Synonym match — log a hint about the canonical spelling instead of silently accepting
    logger.warning("Task %r matches via synonym; consider using canonical %r", task, normalize_task(task))
    return hf_config
```

### 2. HF-pipeline-only task names are falsely rejected

`normalize_task` doesn't consult `TASK_SYNONYM_EXTENSIONS` in io.py. Names handled there (`next-sentence-prediction` → `text-classification`, `mask-generation` preserved) won't be in `supported_tasks` and won't normalize to something in it. Same for `sentence-similarity` (HF-only, not in Optimum's synonym map). The `if not supported_tasks: return hf_config` escape only fires when Optimum knows nothing about the arch — it doesn't help for mainstream text models.

Repro: `winml build -m bert-base-uncased --task next-sentence-prediction` works pre-PR (handled in `export/io.py`) and will be rejected post-PR.

Suggested change: short-circuit on names that the rest of the codebase knows about:

```python
from ..loader.task import KNOWN_TASKS
from ..export.io import TASK_SYNONYM_EXTENSIONS

if task in TASK_SYNONYM_EXTENSIONS or task in KNOWN_TASKS:
    # Accept names that other CLI commands accept; let downstream resolution
    # raise ONNXConfigNotFoundError if the arch can't actually export it.
    return hf_config
```

### 3. Acceptance set now diverges from `winml config` / `winml export` / `winml perf`

The docstring notes that the other commands rely on `resolve_cfg` → `ONNXConfigNotFoundError`. Those paths use `resolve_task_and_model_class` + `KNOWN_TASKS`, which is broader and lists `image-feature-extraction` as a distinct entry. Result: a `config.json` produced by `winml config` (and accepted by `winml export`) can be rejected by `winml build`. That asymmetry will surface as bug reports.

Suggested change: either widen the acceptance set in this PR to match the rest of the CLI (combination of fixes 1 + 2), or move the validation up into `resolve_task_and_model_class` so all commands share one definition.

### 4. Test coverage

The current test `test_rejects_incompatible_config_task_and_model` only exercises the rejection path. Suggest adding:

- Vision-arch + `--task feature-extraction` → currently passes the gate (documenting the limitation).
- Text-arch + `--task image-feature-extraction` → currently passes (documenting the limitation, or fixing).
- Text-arch + `--task next-sentence-prediction` → currently rejected (regression vs. existing behavior).
- Same task accepted by `winml config -m ... --task ...` should also be accepted by `winml build`.

### Summary

The PR is solving a real problem, but the gate's notion of "valid" is Optimum's collapse, which is the wrong granularity for the four HF-pipeline-keyed registries that consume `loader.task` downstream. As-is, this introduces false rejections for some HF-pipeline names and silently accepts modality-mismatched inputs. Happy to pair on a follow-up if useful — there's overlapping work on the quantize/eval side where we're hitting the same canonical-name boundary.

zhenchaoni (approved):

PR #729, opened 2026-05-25
feat: merge windowsml usage using official example

xieofxie:
- remove class WinML as it is confusing
- also incidentally fixes a bug introduced by https://github.com/microsoft/winml-cli/pull/712 — the two independent singletons (WinML and WinMLEPRegistry) both call ort.register_execution_provider_library(...) but dedup separately, so after https://github.com/microsoft/winml-cli/pull/712 winml perf --device npu emits a library is already registered under QNNExecutionProvider traceback (benign but alarming, swallowed by except Exception)


src/winml/modelkit/session/ep_registry.py
```diff
@@ -52,9 +52,11 @@ def __init__(self) -> None:
         self._initialized = True
 
         self._ep_paths: dict[EPName, str] = {}
-        self._registered_eps: list[EPName] = []
+        self._registered_eps: dict[str, list[EPName]] = {
+            "onnxruntime": [],
```
timenick, line 56: Are these intended to differentiate between LLM and Non-LLM?

timenick:
Quick question on the switch to `with EpCatalog() as catalog:` — the old code at [ep_registry.py:92-104](https://github.com/microsoft/winml-modelkit/blob/main/src/winml/modelkit/session/ep_registry.py#L92-L104) had a long workaround comment explaining that `WinMLEpCatalogRelease` (called from `EpCatalog.close()` / `__del__`) crashes with `ACCESS_VIOLATION (0xC0000005)` on some QNN NPU driver configurations — a Windows SEH exception that Python `try/except` cannot catch. The workaround was to null out `self._catalog._handle = None` so `close()` becomes a no-op.

The new `with EpCatalog() as catalog:` still invokes `close()` via `__exit__`, which puts that crash path back — just relocated from interpreter shutdown to the return of `_load_ep_catalog`.

Two questions:
1. Has the underlying `WinMLEpCatalogRelease` issue been fixed upstream in windowsml? The original TODO said "Remove once windowsml fixes WinMLEpCatalogRelease to be safe during process teardown."
2. If not, has `winml perf --device npu` been verified on a QNN NPU machine after this change to confirm the crash doesn't return?

Side note: this PR also incidentally fixes a bug introduced by #712 — the two independent singletons (`WinML` and `WinMLEPRegistry`) both call `ort.register_execution_provider_library(...)` but dedup separately, so after #712 `winml perf --device npu` emits a `library is already registered under QNNExecutionProvider` traceback (benign but alarming, swallowed by `except Exception`). Worth calling out in the PR title/description so it shows up in `git log --grep` later.

Generated with Claude Code

src/winml/modelkit/session/ep_registry.py
xieofxie: yes, they are migrated from deleted WinML class

xieofxie:
> Quick question on the switch to `with EpCatalog() as catalog:` — the old code at [ep_registry.py:92-104](https://github.com/microsoft/winml-modelkit/blob/main/src/winml/modelkit/session/ep_registry.py#L92-L104) had a long workaround comment explaining that `WinMLEpCatalogRelease` (called from `EpCatalog.close()` / `__del__`) crashes with `ACCESS_VIOLATION (0xC0000005)` on some QNN NPU driver configurations — a Windows SEH exception that Python `try/except` cannot catch. The workaround was to null out `self._catalog._handle = None` so `close()` becomes a no-op.
> 
> The new `with EpCatalog() as catalog:` still invokes `close()` via `__exit__`, which puts that crash path back — just relocated from interpreter shutdown to the return of `_load_ep_catalog`.
> 
> Two questions:
> 
> 1. Has the underlying `WinMLEpCatalogRelease` issue been fixed upstream in windowsml? The original TODO said "Remove once windowsml fixes WinMLEpCatalogRelease to be safe during process teardown."
> 2. If not, has `winml perf --device npu` been verified on a QNN NPU machine after this change to confirm the crash doesn't return?
> 
> Side note: this PR also incidentally fixes a bug introduced by #712 — the two independent singletons (`WinML` and `WinMLEPRegistry`) both call `ort.register_execution_provider_library(...)` but dedup separately, so after #712 `winml perf --device npu` emits a `library is already registered under QNNExecutionProvider` traceback (benign but alarming, swallowed by `except Exception`). Worth calling out in the PR title/description so it shows up in `git log --grep` later.
> 
> Generated with Claude Code

from my understanding, this is the official usage of windowsml lib https://pypi.org/project/windowsml/ I feel it should be well tested so that access violation should not exist

timenick (approved):

PR #733, opened 2026-05-25
Add winml-cli skill

ssss141414:
## Summary
 Adds `skills/use-winml-cli/SKILL.md` — a skill that teaches an agent to drive the `winml` CLI for the
  Build-Your-Own-Model (BYOM) pipeline across NPU/GPU/CPU execution providers.


timenick:
winml-cli not winml-modelkit

timenick (commented):
probe

.claude/skills/winml-cli/SKILL.md
```diff
@@ -0,0 +1,190 @@
...
+
+The names above (`inspect`, `export`, `analyze`, `optimize`, `quantize`, `compile`, `perf`, plus the config/build pair) are stable concepts — they map to subcommands of `winml`. Confirm exact spelling and current flags via `winml --help` before you write any command.
+
+## Outputs are explicit; cache is opaque
+
+Every `winml` command has a **published output** — what `-o` writes or what it prints to stdout — and that is the only thing you can hand to a user or downstream command. What a command does *internally* (cached intermediate builds, temp graphs, EP context blobs in a private folder) is not a supported output. Don't try to fish artifacts out of cache, don't assume one command's internal byproducts can feed another.
+
+This is the easiest way to pick the right command for a goal: **map the goal to a published output, not to a side effect.**
+
+- `inspect`, `sys`, `analyze`, `hub` print to stdout (no `-o` artifact).
```
timenick: **`hub` is not a real `winml` command.**

Actual top-level commands in `src/winml/modelkit/commands/`: `analyze, build, catalog, compile, config, eval, export, inspect, optimize, perf, quantize, run, serve, sys`. No `hub`.

This line says:

> `inspect`, `sys`, `analyze`, `hub` print to stdout (no `-o` artifact).

Probably meant `catalog` (which lists hub models from the local registry). Same fix needed at line 178:

> `inspect`, `sys`, and `hub` print to stdout and don’t require an output path.

The skill itself flags this exact failure mode in the “Discover the CLI before doing anything else” section:

> Inventing plausible-sounding flags (a `--preset`, `--profile`, `--mode=fast`) is the most common way to waste the user’s time — the command will reject them and the user has to come back.

A misnamed command in the skill itself trains the agent to invent. Worth catching before merge.

timenick (commented):
Two findings + a meta question on the SKILL.md location.

**Meta concern: where should SKILL.md live?** This PR ships it at `.claude/skills/winml-modelkit/SKILL.md`, which is the Claude Code convention. That's fine if Claude Code is the team's primary agent. But the same skill won't be picked up by other agents whose loaders look elsewhere (GitHub Copilot CLI looks under `.github/skills/...` or `.github/extensions/...` for example). If WinML CLI wants this skill discoverable by both Claude Code and Copilot CLI, the simplest options are: (a) commit a copy at `.github/skills/winml-modelkit/SKILL.md` too, accepting the maintenance cost, or (b) standardize on one agent and document the choice in the repo's contributor docs. Worth at least adding a CONTRIBUTING/README note explaining which agent this targets so future contributors don't put a second skill in a different location.

.claude/skills/winml-cli/SKILL.md
```diff
@@ -0,0 +1,190 @@
...
+|---|---|
+| A latency / throughput number | `winml perf` (writes metrics JSON) |
+| A deployable `.onnx` they can ship or load from code | `winml config` then `winml build -o <dir>` (build writes the artifact) |
+| Both | `winml build -o <dir>` first, then `winml perf` against the built artifact |
+
+If the user's intent isn't clear from their prompt, ask one short question — "do you want a usable artifact, or just a latency number?" — before quoting commands.
+
+## Hardware and execution providers
+
+The right execution provider depends on the user's machine. Status as of 2026-05-20:
```
timenick: **Hardware/EP status table is dated `2026-05-20` — already stale on day one.**

The PR was opened on 5/25, so the table is 5 days stale before merge. README has its own `Supported Hardware` table that is more likely to be updated as EPs change status (e.g., when MIGraphX flips from Planned to Ready). Duplicating it here creates a drift surface: the skill table will silently fall out of sync, and an agent that quotes it confidently will be wrong.

Two cleaner options:

1. **Delete the table; point at the README and `winml sys --list-ep`.** The skill already gives the right advice (“trust `winml sys --list-ep` on the user’s machine for what’s actually registered”). A one-line pointer to README’s Supported Hardware table for the high-level matrix avoids the duplication entirely.

2. **Keep the table but anchor it to a single source.** E.g., add a comment `# Source: README.md Supported Hardware table as of YYYY-MM-DD; refresh on every winml-cli release.` and move the snapshot date into a more visible place. This still drifts but at least signals the refresh cadence.

Option 1 is lower-maintenance and matches the skill’s own "read it live" philosophy.

xieofxie:
do we need RAI for this?

ssss141414:
> do we need RAI for this?

Asked Qianhao, FoundryTK did't do RAI for releasing skills before...

DingmaomaoBJTU (approved):

DingmaomaoBJTU (approved):

PR #746, opened 2026-05-26
perf(inspect): skip redundant Auto* calls for text-only models

timenick:
## Summary

Two extra Strategy-2 gating fixes on top of #719. Profile on `cardiffnlp/twitter-roberta-base-sentiment-latest` (text-only) showed the inspect command was still taking ~16 s warm — most of it spent on Auto* calls that didn't need to run.

Targeting **`release/v0.1.0`** since #717 / #718 / #719 are already on that release; this is the natural follow-up.

## Profile (warm cache, `cardiffnlp/twitter-roberta-base-sentiment-latest`)

**Before this PR**:

```
[0]  AutoConfig             0.74s  (parent_hf_config — already deduped by #719)
[1]  AutoProcessor          4.27s  returns RobertaTokenizerFast
[7]  AutoTokenizer          2.22s  ← redundant, AutoProcessor already returned the tokenizer
[11] AutoImageProcessor     1.39s  FAIL (text model has no preprocessor_config.json)
[12] AutoFeatureExtractor   0.64s  FAIL (same)
─────
~16 s total
```

**After this PR**:

```
AutoConfig             5 calls (vs 8)
AutoProcessor          1 call
AutoTokenizer          1 call   (vs 2 — the redundant standalone load is gone)
AutoImageProcessor     0 calls  (skipped — no preprocessor_config.json)
AutoFeatureExtractor   0 calls  (skipped — same)
─────
~12 s total  (~25% faster)
```

### 1. Detect when `AutoProcessor` returns a leaf class

For single-modality models, `AutoProcessor.from_pretrained` returns the leaf class directly — e.g. RoBERTa → `RobertaTokenizerFast`. Such a return has no `.tokenizer` wrapper attribute, so the old code couldn't populate `tokenizer_class` and fell through to a redundant `AutoTokenizer.from_pretrained` (~2 s warm).

Pattern-match the returned class name (`*Tokenizer` / `*TokenizerFast`, `*ImageProcessor` / `*ImageProcessorFast`, `*FeatureExtractor`) and populate the corresponding field. The `.tokenizer` / `.image_processor` / `.feature_extractor` attribute path still wins for genuine multimodal `ProcessorMixin` returns (CLIP, etc.) — see the `test_autoprocessor_with_wrapped_pieces_uses_attributes` regression test.

### 2. `preprocessor_config.json` absence is authoritative

`_resolve_processor_from_hub_configs` already tries to download `preprocessor_config.json`. When the hub returns 404, the model has *no* image processor or feature extractor, period. Surface this as a `has_preprocessor_config` bool from the helper so the caller can skip the `AutoImageProcessor` / `AutoFeatureExtractor` round-trips (~2 s total wasted confirming 404s).

## Tests

`tests/unit/inspect/test_resolve_processor_gating.py`:

- `test_autoprocessor_returns_tokenizer_fills_tokenizer_class` — leaf-class detection populates `tokenizer_class` from class-name suffix and skips standalone `AutoTokenizer`
- `test_autoprocessor_returns_image_processor_fills_image_class` — same for `*ImageProcessor`
- `test_autoprocessor_returns_feature_extractor_fills_feature_class` — same for `*FeatureExtractor`
- `test_autoprocessor_with_wrapped_pieces_uses_attributes` — multimodal `ProcessorMixin` with `.tokenizer` attribute wins over name suffix
- `test_missing_preprocessor_config_skips_image_and_feature` — `has_preprocessor_config=False` skips `AutoImageProcessor` / `AutoFeatureExtractor`

55 targeted tests pass.

### DIFF
--- a/src/winml/modelkit/inspect/resolver.py
+++ b/src/winml/modelkit/inspect/resolver.py
@@ -10,7 +10,7 @@ Leverages existing loader, export, and models modules - NO NEW CONFIG LOGIC.
 from __future__ import annotations
 
 import logging
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, NamedTuple
 
 from ..loader.task import (
     HF_TASK_DEFAULTS,
@@ -871,26 +871,20 @@ def resolve_processor(
     # standard HuggingFace config conventions, not model-specific hardcoding.
     has_preprocessor_config = True
     try:
-        (
-            hub_proc,
-            hub_tok,
-            hub_img,
-            hub_fe,
-            has_preprocessor_config,
-            _has_tokenizer_config,
-        ) = _resolve_processor_from_hub_configs(model_id)
-        if hub_proc and processor_class is None:
-            processor_class = hub_proc
+        hub_result = _resolve_processor_from_hub_configs(model_id)
+        if hub_result.processor_class and processor_class is None:
+            processor_class = hub_result.processor_class
             processor_source = "hub_config"
-        if hub_tok and tokenizer_class is None:
-            tokenizer_class = hub_tok
+        if hub_result.tokenizer_class and tokenizer_class is None:
+            tokenizer_class = hub_result.tokenizer_class
             tokenizer_source = "hub_config"
-        if hub_img and image_processor_class is None:
-            image_processor_class = hub_img
+        if hub_result.image_processor_class and image_processor_class is None:
+            image_processor_class = hub_result.image_processor_class
             image_processor_source = "hub_config"
-        if hub_fe and feature_extractor_class is None:
-            feature_extractor_class = hub_fe
+        if hub_result.feature_extractor_class and feature_extractor_class is None:
+            feature_extractor_class = hub_result.feature_extractor_class
             feature_extractor_source = "hub_config"
+        has_preprocessor_config = hub_result.has_preprocessor_config
     except Exception as e:
         logger.debug("Failed to resolve processors from hub configs: %s", e)
 
@@ -950,9 +944,21 @@ def resolve_processor(
     )
 
 
-def _resolve_processor_from_hub_configs(
-    model_id: str,
-) -> tuple[str | None, str | None, str | None, str | None, bool, bool]:
+class _HubConfigResult(NamedTuple):
+    """Result of ``_resolve_processor_from_hub_configs``.
+
+    A NamedTuple rather than a plain tuple so the trailing boolean cannot be
+    silently swapped with the four ``str | None`` fields at the call site.
+    """
+
+    processor_class: str | None
+    tokenizer_class: str | None
+    image_processor_class: str | None
+    feature_extractor_class: str | None
+    has_preprocessor_config: bool
+
+
+def _resolve_processor_from_hub_configs(model_id: str) -> _HubConfigResult:
     """Resolve processor classes by fetching config files from HuggingFace Hub.
 
     This approach is fast because it only downloads small JSON config files,
@@ -962,14 +968,12 @@ def _resolve_processor_from_hub_configs(
         model_id: HuggingFace model identifier
 
     Returns:
-        Tuple of ``(processor_class, tokenizer_class, image_processor_class,
-        feature_extractor_class, has_preprocessor_config, has_tokenizer_config)``.
-        The two booleans report whether the respective config files actually
-        exist on the hub — a missing ``preprocessor_config.json`` is the
-        authoritative signal that the model has no image processor or
+        A ``_HubConfigResult`` whose ``has_preprocessor_config`` reports
+        whether ``preprocessor_config.json`` actually exists on the hub —
+        the authoritative signal that the model has no image processor or
         feature extractor, so the caller can skip the corresponding
         ``AutoImageProcessor`` / ``AutoFeatureExtractor`` round-trips
-        (which would each fail with a slow 404 lookup on text-only models).
+        (which would each spend ~1s confirming a 404 on text-only models).
     """
     import json
     from pathlib import Path
@@ -982,7 +986,6 @@ def _resolve_processor_from_hub_configs(
     image_processor_class: str | None = None
     feature_extractor_class: str | None = None
     has_preprocessor_config = False
-    has_tokenizer_config = False
 
     # Try to download and parse preprocessor_config.json
     # This file contains image_processor_type or processor_class
@@ -991,6 +994,10 @@ def _resolve_processor_from_hub_configs(
             repo_id=model_id,
             filename="preprocessor_config.json",
         )
+        # Set the flag as soon as the file exists on the hub, *before* parsing.
+        # A corrupt JSON is still proof that the model ships preprocessor
+        # config — fall back to Auto* lookups rather than declaring the model
+        # text-only and silently dropping its image/feature processor.
         has_preprocessor_config = True
         with Path(preprocessor_config_path).open(encoding="utf-8") as f:
             preprocessor_config = json.load(f)
@@ -1019,7 +1026,6 @@ def _resolve_processor_from_hub_configs(
             repo_id=model_id,
             filename="tokenizer_config.json",
         )
-        has_tokenizer_config = True
         with Path(tokenizer_config_path).open(encoding="utf-8") as f:
             tokenizer_config = json.load(f)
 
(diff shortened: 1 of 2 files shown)


DingmaomaoBJTU (commented):
## Review: `perf(inspect): skip redundant Auto* calls for text-only models`

Overall this is a well-motivated performance PR with solid profiling data. Both optimizations are sensible and the test coverage is good. A few items worth discussing:

### Design

**1. Growing positional tuple → consider a dataclass/NamedTuple**

`_resolve_processor_from_hub_configs` now returns a 6-element positional tuple. Every caller must destructure it by position, and future additions will make this increasingly fragile. A `NamedTuple` (or `@dataclass`) would make the return self-documenting and order-independent:

```python
class HubConfigResult(NamedTuple):
    processor_class: str | None
    tokenizer_class: str | None
    image_processor_class: str | None
    feature_extractor_class: str | None
    has_preprocessor_config: bool
    has_tokenizer_config: bool
```

Not blocking, but worth doing now while the change is fresh — the tuple was already borderline at 4 elements.

**2. `has_tokenizer_config` is returned but unused**

The caller destructures it as `_has_tokenizer_config` and discards it. If there's no planned consumer, consider not returning it yet (YAGNI). If it *is* planned, a TODO comment would help.

### Correctness

**3. `has_preprocessor_config` set `True` even on `json.JSONDecodeError`**

In `_resolve_processor_from_hub_configs`, the flag is set to `True` right after the `hf_hub_download` call but *before* `json.load`. If `preprocessor_config.json` exists on the hub but contains invalid JSON, `has_preprocessor_config` will be `True` even though no useful data was extracted. This is defensively fine (we don't skip Auto* calls when the file exists), but worth a comment noting this is intentional — a corrupt config file on the hub doesn't mean the model has no image processor.

**4. Leaf-class detection: ambiguous class names**

The `_is_tokenizer_class_name` heuristic checks `name.endswith(("Tokenizer", "TokenizerFast"))`. Today this works because HuggingFace naming is well-disciplined, but note that `endswith("Tokenizer")` already matches `*TokenizerFast` since "TokenizerFast" doesn't end with "Tokenizer". The tuple is correct — just flagging that the `"TokenizerFast"` entry is the one doing the real work for fast tokenizers, while `"Tokenizer"` covers the slow variant. This is fine.

### Tests

**5. Leaf-class tests don't verify the standalone Auto* calls are *actually* gated**

`test_autoprocessor_returns_tokenizer_fills_tokenizer_class` patches `AutoTokenizer.from_pretrained` and asserts `call_count == 0`, but it also passes `try_image_processor=False, try_feature_extractor=False`, so those Auto* classes are pre-gated. Consider adding a case where `try_tokenizer=True` *and* the leaf class is a tokenizer, but the model actually has no `.tokenizer` attr — this is the exact real-world scenario (text-only model) and is what the test claims to cover. The current tests *do* cover this via `_make_leaf_instance` (no attrs), so this is fine on closer reading — just wanted to confirm.

**6. Missing negative test: class name that *doesn't* match any heuristic**

What happens when `AutoProcessor` returns a leaf class with an unrecognized suffix (e.g. `SpeechT5Processor`)? The current code silently leaves `tokenizer_class`/`image_processor_class`/`feature_extractor_class` as `None` and the standalone Auto* calls below fill them in. This is the correct fallback, but a test for it would document the behavior.

### Nits

- The PR description says "~25% faster" but the numbers show 16s → 12s which is exactly 25%. Might as well say "25% faster" without the tilde.

Overall: solid perf improvement with good test coverage. The tuple→NamedTuple suggestion is the main structural feedback.

src/winml/modelkit/inspect/resolver.py
```diff
@@ -1009,7 +1032,35 @@ def _resolve_processor_from_hub_configs(
     except json.JSONDecodeError as e:
         logger.debug("Failed to parse tokenizer_config.json for %s: %s", model_id, e)
 
-    return processor_class, tokenizer_class, image_processor_class, feature_extractor_class
+    return (
+        processor_class,
```
DingmaomaoBJTU: `has_tokenizer_config` is computed and returned but the only caller discards it (`_has_tokenizer_config`). If there is no planned consumer yet, consider deferring this to keep the return shape minimal (YAGNI). If it is planned for a follow-up, a brief TODO would help.

src/winml/modelkit/inspect/resolver.py
```diff
@@ -940,7 +952,7 @@ def resolve_processor(
 
 def _resolve_processor_from_hub_configs(
     model_id: str,
-) -> tuple[str | None, str | None, str | None, str | None]:
+) -> tuple[str | None, str | None, str | None, str | None, bool, bool]:
```
DingmaomaoBJTU: Consider extracting this 6-element tuple into a `NamedTuple` or `@dataclass`. Positional destructuring was already borderline at 4 elements; at 6 it is easy to silently swap two fields (especially the two trailing booleans). A named type makes this self-documenting and resilient to future additions.

src/winml/modelkit/inspect/resolver.py
```diff
@@ -970,6 +991,7 @@ def _resolve_processor_from_hub_configs(
             repo_id=model_id,
             filename="preprocessor_config.json",
         )
+        has_preprocessor_config = True
```
DingmaomaoBJTU, line 1001: Note: if the file exists on the hub but contains invalid JSON, this flag is already `True` before the `json.JSONDecodeError` handler runs. This is defensively correct (corrupt config ≠ "no image processor"), but a brief comment would help future readers understand the ordering is intentional.

tests/unit/inspect/test_resolve_processor_gating.py
```diff
@@ -164,3 +212,119 @@ def test_try_processor_true_still_calls_autoprocessor(self) -> None:
...
+                "openai/clip-vit-base-patch32",
+                try_processor=True,
+                try_tokenizer=True,
+                try_image_processor=False,
+                try_feature_extractor=False,
+            )
+
+        assert proc == "CLIPProcessor"
+        assert tok == "CLIPTokenizer"
+        assert mock_at.call_count == 0
```
DingmaomaoBJTU, line 342: Consider adding a negative-path test where `AutoProcessor` returns a leaf class with an unrecognized suffix (e.g. `SpeechT5Processor`) — none of the `_is_*_class_name` heuristics match, so `tokenizer_class` etc. stay `None` and the standalone Auto* calls below fill them in. This would document the graceful-fallback behavior.

src/winml/modelkit/inspect/resolver.py
timenick: Agreed — dropped `has_tokenizer_config` entirely in c353f893. No consumer planned, so removing it now is cleaner than carrying a TODO.

Generated with Claude Code

src/winml/modelkit/inspect/resolver.py
timenick: Done in c353f893. The hub-config helper now returns a `_HubConfigResult` NamedTuple, so the call site uses attribute access (`hub_result.has_preprocessor_config`) instead of positional destructuring. Combined with dropping `has_tokenizer_config`, the surface is back down to 5 fields, all named.

Generated with Claude Code

src/winml/modelkit/inspect/resolver.py
timenick: Added the explanatory comment in c353f893:

```python
# Set the flag as soon as the file exists on the hub, *before* parsing.
# A corrupt JSON is still proof that the model ships preprocessor
# config — fall back to Auto* lookups rather than declaring the model
# text-only and silently dropping its image/feature processor.
has_preprocessor_config = True
```

Generated with Claude Code

tests/unit/inspect/test_resolve_processor_gating.py
timenick: Added `test_unrecognized_leaf_falls_through_to_standalone_autos` in c353f893. `AutoProcessor.from_pretrained` returns a leaf instance of `SpeechT5Processor` — none of the `_is_*_class_name` heuristics match the `Processor` suffix, so the test asserts the standalone `AutoTokenizer` / `AutoFeatureExtractor` calls below run and fill in `tokenizer_class` / `feature_extractor_class`.

Generated with Claude Code

DingmaomaoBJTU (approved):

PR #755, opened 2026-05-26
update eval scripts: add ONNX size tracking and output sanitization

DingmaomaoBJTU:
## Summary
- Add `_compute_onnx_size()` to measure combined ONNX + `.data` file sizes and include `onnx_size_bytes` in eval results
- Add `_sanitize_output()` to strip CLI chrome (Rich tables, device/IO banners) from `eval_result.json`, keeping only error-relevant content
- Minor formatting fixes in `reporter.py`

### DIFF
--- a/scripts/e2e_eval/run_eval.py
+++ b/scripts/e2e_eval/run_eval.py
@@ -1096,7 +1096,7 @@ def save_environment_info(path: Path) -> None:
         if result.returncode == 0:
             info["winml_sys"] = json.loads(result.stdout)
     except (subprocess.TimeoutExpired, FileNotFoundError, json.JSONDecodeError):
-        pass
+        pass  # winml sys unavailable or returned invalid JSON; skip optional field
 
     path.parent.mkdir(parents=True, exist_ok=True)
     path.write_text(json.dumps(info, indent=2), encoding="utf-8")
--- a/scripts/e2e_eval/run_eval.py
+++ b/scripts/e2e_eval/run_eval.py
@@ -31,6 +31,7 @@ from __future__ import annotations
 import argparse
 import contextlib
 import json
+import logging
 import os
 import platform
 import shutil
@@ -68,6 +69,8 @@ from utils.reporter import (
     write_summary_md,
 )
 
+
+logger = logging.getLogger(__name__)
 
 # ---------------------------------------------------------------------------
 # Constants
@@ -234,7 +237,11 @@ def _utc_now() -> str:
 
 
 def _compute_onnx_size(onnx_paths: dict[str, str]) -> int | None:
-    """Return combined size in bytes of all ONNX files + their .data companions.
+    """Return combined size in bytes of all ONNX files + their external data companions.
+
+    Parses the ONNX proto to discover all referenced external data files (not just
+    the conventional `.data` suffix). Falls back to the `.data` companion heuristic
+    if proto parsing is unavailable.
 
     Returns None if onnx_paths is empty or no files exist on disk.
     """
@@ -244,12 +251,24 @@ def _compute_onnx_size(onnx_paths: dict[str, str]) -> int | None:
     found_any = False
     for path_str in onnx_paths.values():
         p = Path(path_str)
-        if p.exists():
-            total += p.stat().st_size
-            found_any = True
-        data_p = p.with_suffix(p.suffix + ".data")
-        if data_p.exists():
-            total += data_p.stat().st_size
+        if not p.exists():
+            continue
+        total += p.stat().st_size
+        found_any = True
+        # Try to enumerate all external data files from the proto
+        try:
+            from winml.modelkit.onnx.external_data import get_external_data_files
+
+            ext_files = get_external_data_files(p)
+            for ext_name in ext_files:
+                ext_path = p.parent / ext_name
+                if ext_path.exists():
+                    total += ext_path.stat().st_size
+        except Exception:
+            # Fallback: check conventional .data companion
+            data_p = p.with_suffix(p.suffix + ".data")
+            if data_p.exists():
+                total += data_p.stat().st_size
     return total if found_any else None
 
 
@@ -1091,7 +1110,7 @@ def save_environment_info(path: Path) -> None:
     # `winml sys --format json` captures hardware details (devices, EPs,
     # backends) that the lightweight package-version probes above miss.
     try:
-        result = subprocess.run(
+        result = subprocess.run(  # noqa: S603
             [sys.executable, "-m", "winml", "sys", "--format", "json"],
             capture_output=True,
             text=True,
@@ -1099,8 +1118,8 @@ def save_environment_info(path: Path) -> None:
         )
         if result.returncode == 0:
             info["winml_sys"] = json.loads(result.stdout)
-    except (subprocess.TimeoutExpired, FileNotFoundError, json.JSONDecodeError):
-        pass  # winml sys unavailable or returned invalid JSON; skip optional field
+    except (subprocess.TimeoutExpired, FileNotFoundError, json.JSONDecodeError) as exc:
+        logger.debug("winml sys skipped: %s", exc)
 
     path.parent.mkdir(parents=True, exist_ok=True)
     path.write_text(json.dumps(info, indent=2), encoding="utf-8")
@@ -1213,6 +1232,11 @@ def parse_args() -> argparse.Namespace:
         help="Skip report generation (useful when running per-model in a pipeline loop)",
     )
     parser.add_argument("--verbose", "-v", action="store_true", help="Verbose output")
+    parser.add_argument(
+        "--raw-output",
+        action="store_true",
+        help="Keep raw subprocess output in eval_result.json without sanitization",
+    )
     parser.add_argument(
         "--continue",
         dest="continue_run",
@@ -1525,7 +1549,7 @@ def main() -> None:
             accuracy_result,
             ep=args.ep,
             onnx_size_bytes=onnx_size,
-            sanitize_fn=_sanitize_output,
+            sanitize_fn=None if args.raw_output else _sanitize_output,
         )
         results.append(result)
 
--- a/scripts/e2e_eval/run_eval.py
+++ b/scripts/e2e_eval/run_eval.py
@@ -34,6 +34,7 @@ import json
 import logging
 import os
 import platform
+import re
 import shutil
 import subprocess
 import sys
@@ -273,7 +274,7 @@ def _compute_onnx_size(onnx_paths: dict[str, str]) -> int | None:
 
 
 # Lines that carry no diagnostic value in eval_result.json.
-# Matching is case-insensitive, applied per-line.
+# Matching is case-insensitive, anchored at line start.
 _NOISE_PATTERNS = (
     "benchmarking onnx",
     "device:",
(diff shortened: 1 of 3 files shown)


DingmaomaoBJTU (commented):
Nice additions - ONNX size tracking and output sanitization are useful for keeping `eval_result.json` lean. A few observations below.

scripts/e2e_eval/run_eval.py
```diff
@@ -196,6 +196,66 @@ def _utc_now() -> str:
...
+
+# Lines that carry no diagnostic value in eval_result.json.
+# Matching is case-insensitive, applied per-line.
+_NOISE_PATTERNS = (
+    "benchmarking onnx",
+    "device:",
+    "task:",
+    "latency (ms)",
+    "throughput:",
+    "results saved to",
```
DingmaomaoBJTU, line 284: The `_compute_onnx_size` helper only looks for `<path>.data` companion files. Some ONNX models use external data stored in files with different naming patterns (e.g. `model.onnx_data`, or arbitrary names specified in the proto's `external_data` field). If exact size tracking matters, consider parsing the model proto's `external_data` entries or using a glob like `<stem>*` in the same directory.

scripts/e2e_eval/run_eval.py
```diff
@@ -196,6 +196,66 @@ def _utc_now() -> str:
...
+            continue
+        low = stripped.lower()
+        if any(pat in low for pat in _NOISE_PATTERNS):
+            continue
+        kept.append(stripped)
+    return "\n".join(kept)
+
+
 def _kill_process_tree(pid: int) -> None:
     """Kill a process and all its children.
```
DingmaomaoBJTU, line 318: The `_sanitize_output` function uses substring matching (`any(pat in low ...)`), which means a legitimate error line containing "device:" or "task:" as part of an error message would be stripped. Consider anchoring patterns (e.g. only match at line start) or using a positive-match approach for errors instead of negative-match for noise.

scripts/e2e_eval/utils/reporter.py
```diff
@@ -324,6 +337,7 @@ def generate_html_report(
     registry_path: Path | None = None,
 ) -> None:
     from .accuracy import format_delta
+
```
DingmaomaoBJTU, line 346: The docstring for `generate_html_report` ended up *after* the `from .accuracy import format_delta` import due to the blank line insertion. This moves the docstring outside the function definition from a reader's perspective (though Python still attaches it correctly). Consider placing the import after the docstring for clarity.

scripts/e2e_eval/run_eval.py
```diff
@@ -971,6 +1031,20 @@ def save_environment_info(path: Path) -> None:
...
         pass  # git not available or timed out; commit info stays empty
 
+    # `winml sys --format json` captures hardware details (devices, EPs,
+    # backends) that the lightweight package-version probes above miss.
+    try:
+        result = subprocess.run(  # noqa: S603
+            [sys.executable, "-m", "winml", "sys", "--format", "json"],
+            capture_output=True,
+            text=True,
+            timeout=30,
```
DingmaomaoBJTU, line 1118: Good addition. The `timeout=30` is reasonable for `winml sys`, but if this runs in CI environments with slow disk/network, it could occasionally timeout. The `pass` handling is fine, just calling it out.

scripts/e2e_eval/run_eval.py
```diff
@@ -196,6 +196,66 @@ def _utc_now() -> str:
...
+    Returns None if onnx_paths is empty or no files exist on disk.
+    """
+    if not onnx_paths:
+        return None
+    total = 0
+    found_any = False
+    for path_str in onnx_paths.values():
+        p = Path(path_str)
+        if p.exists():
+            total += p.stat().st_size
```
DingmaomaoBJTU: Consider using pathlib consistently here. You have `p = Path(path_str)` but then construct the .data companion as `Path(path_str + '.data')`. Using `p.with_suffix(p.suffix + '.data')` would be more idiomatic, though for the common case (.onnx -> .onnx.data) your approach actually works fine. Just noting for consistency.

scripts/e2e_eval/run_eval.py
```diff
@@ -196,6 +196,66 @@ def _utc_now() -> str:
...
+
+    Removes Rich benchmark tables, device/IO banners, and path lines that
+    bloat eval_result.json without aiding failure diagnosis. All classifier
+    patterns (see classifier.py) are error-related and survive this filter.
+    """
+    kept: list[str] = []
+    for line in text.splitlines():
+        stripped = line.strip()
+        if not stripped:
+            continue
```
DingmaomaoBJTU, line 306: The `_sanitize_output` function uses linear scanning (`any(pat in low for pat in _NOISE_PATTERNS)`) for each line. This is fine for typical output sizes, but if perf output ever gets large (thousands of lines), consider compiling the patterns into a single regex with `|` for better performance. Not blocking - just a future consideration.

DingmaomaoBJTU (commented):
Useful additions - ONNX size tracking and output sanitization will make eval results much cleaner. A few suggestions below.

scripts/e2e_eval/run_eval.py
```diff
@@ -971,6 +1031,20 @@ def save_environment_info(path: Path) -> None:
...
+    # backends) that the lightweight package-version probes above miss.
+    try:
+        result = subprocess.run(  # noqa: S603
+            [sys.executable, "-m", "winml", "sys", "--format", "json"],
+            capture_output=True,
+            text=True,
+            timeout=30,
+        )
+        if result.returncode == 0:
+            info["winml_sys"] = json.loads(result.stdout)
```
DingmaomaoBJTU, line 1121: Good addition capturing `winml sys` info. The `timeout=30` is reasonable but note that on cold starts or resource-constrained CI agents this could occasionally fail silently. Consider logging a debug-level message when the timeout fires so it's diagnosable: `logger.debug('winml sys timed out, skipping hardware info')`.

scripts/e2e_eval/utils/reporter.py
DingmaomaoBJTU, line 346: Bug: The docstring for `generate_html_report` got displaced below the `from .accuracy import format_delta` import. The docstring should be immediately after the `def` line (line 337) to be recognized by Python as the function's `__doc__`. As-is, it's just a string expression that does nothing.

scripts/e2e_eval/run_eval.py
```diff
@@ -233,6 +233,66 @@ def _utc_now() -> str:
...
+        data_p = p.with_suffix(p.suffix + ".data")
+        if data_p.exists():
+            total += data_p.stat().st_size
+    return total if found_any else None
+
+
+# Lines that carry no diagnostic value in eval_result.json.
+# Matching is case-insensitive, applied per-line.
+_NOISE_PATTERNS = (
+    "benchmarking onnx",
```
xieofxie, line 279: a little strange.. any way we add a parameter in eval command to just drop them?

scripts/e2e_eval/run_eval.py
```diff
@@ -233,6 +233,66 @@ def _utc_now() -> str:
...
+    "latency (ms)",
+    "throughput:",
+    "results saved to",
+    "inputs:",
+    "outputs:",
+    "samples/sec",
+)
+
+# Box-drawing characters used by Rich tables.
+_BOX_CHARS = frozenset("─│┌┐└┘├┤┬┴┼")
```
timenick, line 292: `_BOX_CHARS` only covers Unicode's LIGHT box-drawing style (`─│┌┐└┘├┤┬┴┼`), but `winml perf` uses Rich's default `Table()`, which renders with `box.HEAVY_HEAD`. I rendered a default Rich table locally and **3 of the 5 lines** start with chars **not** in this set:

| Row | First char | In `_BOX_CHARS`? |
|---|---|---|
| top border ``┏━━━━━┳━━━━━┓`` | ``┏`` (U+250F) | ❌ |
| header row ``┃ Avg ┃ P90 ┃`` | ``┃`` (U+2503) | ❌ |
| head separator ``┡━━━━━╇━━━━━┩`` | ``┡`` (U+2521) | ❌ |
| data row ``│ 12.5 │ 15.2 │`` | ``│`` | ✅ |
| bottom border ``└─────┴─────┘`` | ``└`` | ✅ |

Net result: `eval_result.json` keeps the **top border + header text + head separator** while stripping data rows and the bottom border — arguably uglier than no sanitization at all (orphaned half-table chrome).

Cheap fix — use a Unicode block range check instead of a hand-curated set:
```python
if 0x2500 <= ord(stripped[0]) <= 0x257F:  # Unicode "Box Drawing" block
    continue
```
That covers all four Rich styles (light, heavy, double, rounded) in one rule and won''t silently drift the next time someone changes the table style.

scripts/e2e_eval/run_eval.py
```diff
@@ -233,6 +233,66 @@ def _utc_now() -> str:
...
+        stripped = line.strip()
+        if not stripped:
+            continue
+        # Drop box-drawing table rows
+        if stripped[0] in _BOX_CHARS:
+            continue
+        low = stripped.lower()
+        if any(low.startswith(pat) for pat in _NOISE_PATTERNS):
+            continue
+        kept.append(stripped)
```
timenick, line 312: Appending `stripped` (post-`.strip()`) discards leading indentation, which destroys the structure of multi-line errors — the very content the docstring promises to preserve ("All classifier patterns are error-related and survive this filter"). For example, a traceback:
```
  File "foo.py", line 5, in bar
    raise RuntimeError("x")
```
becomes a visually-flat:
```
File "foo.py", line 5, in bar
raise RuntimeError("x")
```
which is noticeably harder to read.

Suggested fix — keep `stripped` only for the box/noise classifier checks, then append the original line (lightly trimmed):
```python
if not stripped:
    continue
if 0x2500 <= ord(stripped[0]) <= 0x257F:
    continue
low = stripped.lower()
if any(low.startswith(pat) for pat in _NOISE_PATTERNS):
    continue
kept.append(line.rstrip())
```

scripts/e2e_eval/run_eval.py
```diff
@@ -233,6 +233,66 @@ def _utc_now() -> str:
...
+# Lines that carry no diagnostic value in eval_result.json.
+# Matching is case-insensitive, applied per-line.
+_NOISE_PATTERNS = (
+    "benchmarking onnx",
+    "device:",
+    "task:",
+    "latency (ms)",
+    "throughput:",
+    "results saved to",
+    "inputs:",
```
timenick, line 285: A couple of patterns in `_NOISE_PATTERNS` can swallow legitimate diagnostic content with the current `low.startswith(pat)` matching:

- **`"inputs:"` / `"outputs:"`** — these silently strip lines like `Inputs: expected (1, 224, 224, 3), got (1, 3, 224, 224)` (a real shape-mismatch error), which is exactly the kind of "error-relevant" content a sanitizer is supposed to keep. Same concern for `"device:"` if a downstream error ever reads something like `Device: <name> is not available, falling back to CPU`.
- **`"samples/sec"`** — dead pattern under `startswith` semantics. `Throughput: 80 samples/sec` is already covered by `"throughput:"` above; no `winml perf` line literally starts with `samples/sec`.

Tightening options (cheapest first):
1. Just drop `inputs:` / `outputs:` / `samples/sec`. The remaining patterns are unambiguous CLI chrome.
2. Switch to exact-prefix-with-boundary: only strip when the line is `pat` followed by space or end, e.g. `low == pat or low.startswith(pat + " ")`, so error lines that happen to start with `Inputs:` but carry non-trivial content survive.

timenick (approved):

PR #756, opened 2026-05-26
ci: split scheduled pipelines into weekly Eval Report and daily E2E Test

KayMKM:
## Summary

Refactor the previous **Modelkit E2E Test** pipeline (which actually runs the full model registry and produces a markdown report) into two distinct pipelines with different cadences and scopes.

### Renamed (no behaviour change)

- `Modelkit E2E Test.yml` → `Modelkit Eval Report.yml`
- `templates/e2e-eval-jobs.yml` → `templates/eval-report-jobs.yml`
- Stage `displayName`s: `E2E Eval — {QNN, OV, AMD}` → `Eval Report — …`
- Continues to run **weekly on Friday 08:00 (UTC+8)** against the full model registry with sharding, `--list-json`, `--continue`, `--retry-failed`, and report generation.

### New: `Modelkit E2E Test.yml` (daily scheduled)

- **Schedule:** `0 16 * * *` UTC = 00:00 UTC+8 every day, staggered 8 h away from the weekly Eval Report cron.
- Three parallel stages (QNN / OV / AMD), each running on its dedicated self-hosted agent.
- Two phases per stage, both gated by queue-time parameters so a one-off run can be trimmed easily:

  1. **`winml perf` phase** — runs `winml perf` once per `(model × EP/device pair)` against an inline `models` parameter. Default list covers one small representative model per supported task (P0 first, P1/P2 filling the remainder).
  2. **pytest e2e phase** — runs a configurable list of `tests/e2e/test_<name>_e2e.py` suites (default: all 11). Tests use `require_ep()` to self-skip when the target EP is absent, so the same list is safe to run on all three agents.

- Each `winml perf` step uses `condition: always()` so every combination runs and the stage fails on any non-zero exit. No matrix sharding, no report generation.
- Reuses the eval-report setup helpers (parquet copy, `uv` venv, `PipAuthenticate`, `pip install -e .[dev]`).

## Why not PR-gating?

E2E runs on self-hosted hardware are too long and too flaky (driver / firmware variance) to gate every PR. The daily cadence keeps regressions surfaced within ~24 h without blocking developer throughput. Per-PR validation continues to rely on the existing unit / integration suites.

## Portal actions (not YAML-controllable)

- Repoint the existing pipeline definition to `Modelkit Eval Report.yml`.
- Create a new pipeline definition for `Modelkit E2E Test.yml`.
- **Do not** add the new pipeline as a required branch-policy check on `main` — it is informational only.

## Files

- `.pipelines/Modelkit Eval Report.yml` (renamed)
- `.pipelines/templates/eval-report-jobs.yml` (renamed)
- `.pipelines/Modelkit E2E Test.yml` (new, daily)
- `.pipelines/templates/e2e-test-jobs.yml` (new)


.pipelines/templates/e2e-test-jobs.yml
```diff
@@ -0,0 +1,218 @@
...
+  - name: pytestTargets
+    type: object
+    default: []
+
+  - name: pytestTimeout
+    type: number
+    default: 1000
+
+jobs:
+  - job: E2ETest_${{ parameters.agentSuffix }}
```
DingmaomaoBJTU, line 47: The `uv` version (`0.10.12`) is hardcoded both in the install URL and in `displayName`. Consider extracting it into a pipeline variable (or a parameter with a default) so bumping the version is a single-line change and won't silently drift between pipelines.

.pipelines/Modelkit E2E Test.yml
```diff
@@ -1,102 +1,203 @@
...
+  - cron: '0 16 * * *'
+    displayName: 'Daily run — 00:00 (UTC+8) every day'
+    branches:
+      include:
+        - main
+    always: true
+
+resources:
+  repositories:
+    - repository: ModelKitArtifacts
```
DingmaomaoBJTU, line 37: Cron math looks correct (`0 16 * * *` UTC = 00:00 UTC+8). Heads-up: Azure Pipelines rounds scheduled runs to the next available agent check-in, so consider mentioning that in the comment to set expectations for contributors who notice the run starting a few minutes late.

DingmaomaoBJTU (commented):
Nice refactoring — splitting scheduled eval-report from a fast PR-gating e2e test is a clear win. Well-documented YAML. A few suggestions below.

.pipelines/templates/e2e-test-jobs.yml
```diff
@@ -0,0 +1,218 @@
...
+
+      - powershell: |
+          $uvBin = "$env:USERPROFILE\.local\bin"
+          if (-not (Get-Command uv -ErrorAction SilentlyContinue)) {
+            Invoke-RestMethod https://astral.sh/uv/0.10.12/install.ps1 | Invoke-Expression
+            $env:PATH = "$uvBin;$env:PATH"
+          }
+          uv python install 3.11
+          Remove-Item -Recurse -Force "$(Build.SourcesDirectory)\.venv" -ErrorAction SilentlyContinue
+          uv venv $(Build.SourcesDirectory)\.venv --python 3.11
```
DingmaomaoBJTU, line 107: The `--no-sync` flag to `uv run` means the lockfile won't be checked. If `uv pip install -e .[dev]` already installed everything this is fine, but consider adding a brief comment explaining this choice — otherwise a future editor might remove the install step above thinking `uv run` handles it.

zhenchaoni (approved):

.pipelines/templates/e2e-test-jobs.yml
KayMKM: it is hardcoded because the latest uv version may have breaking change on pipeline. 0.10.12 is a stable version, so I   maintain it manually to avoid any breaking change

.pipelines/Modelkit E2E Test.yml
KayMKM: It is China time 8am, should be not many new commits..

.pipelines/templates/e2e-test-jobs.yml
KayMKM: added explaination

.pipelines/Modelkit E2E Test.yml
```diff
@@ -1,102 +1,208 @@
...
+    default: true
+
+  # Inline model list. Each entry: { hf_id: <str>, task: <str> }.
+  # Curated to cover one small representative model per supported task
+  # (P0 first, P1/P2 supplements to fill missing tasks). Edit at queue
+  # time (delete rows) to trim a one-off run.
+  - name: models
+    displayName: 'Models to test (inline list, used by runEval)'
+    type: object
+    default:
```
KayMKM, line 59: @DingmaomaoBJTU could you help to review the model lists? I pick one model for each task and try to choose P0 models

.pipelines/templates/e2e-test-jobs.yml
```diff
@@ -0,0 +1,243 @@
...
+                      $name = '${{ pairName }}'
+                      if (-not $catalog.ContainsKey($name)) {
+                          Write-Error "Unknown EP/device pair name: '$name'. Valid: $($catalog.Keys -join ', ')"
+                          exit 1
+                      }
+                      $pair = $catalog[$name]
+
+                      $hfModel = '${{ model.hf_id }}'
+                      $task    = '${{ model.task }}'
+                      $output  = "$(Agent.TempDirectory)/e2e-test/$name"
```
DingmaomaoBJTU, line 161: Output dir uses only the pair name (qnn_npu, dml_gpu, etc.) but not the model. With 13 models sharing the same directory per pair, if run_eval.py writes any fixed-name file (e.g. result.json, perf.csv), later models overwrite earlier results. Consider including the model slug in the path, e.g. e2e-test/$name/$($hfModel -replace '/','_').

.pipelines/templates/e2e-test-jobs.yml
```diff
@@ -0,0 +1,243 @@
...
+            Write-Error "Rules source not found at: $src"
+            Get-ChildItem $repoDir -Recurse -Depth 2 | Select-Object FullName
+            exit 1
+          }
+        displayName: 'Copy runtime check rules from ModelKitArtifacts'
+
+      - powershell: |
+          $uvBin = "$env:USERPROFILE\.local\bin"
+          if (-not (Get-Command uv -ErrorAction SilentlyContinue)) {
+            Invoke-RestMethod https://astral.sh/uv/0.10.12/install.ps1 | Invoke-Expression
```
DingmaomaoBJTU, line 102: The uv install script is fetched live from the Astral CDN on every run. If the CDN is down or rate-limits the agent, pipeline setup fails entirely. The version-pinned URL pattern (astral.sh/uv/0.10.12/install.ps1) is also non-standard — the documented way to pin is setting $env:UV_VERSION = '0.10.12' before running the generic installer. Since the agents already have network access to Azure Artifacts, consider caching uv there instead.

.pipelines/Modelkit E2E Test.yml
```diff
@@ -1,102 +1,208 @@
...
+      type: github
+      endpoint: github.com_yuesu_microsoft
+      name: gim-home/ModelKitArtifacts
+      ref: main
+
+parameters:
+  # --------------------------------------------------------------------
+  # winml perf section
+  # --------------------------------------------------------------------
+  - name: runEval
```
DingmaomaoBJTU, line 47: Minor naming nit: runEval is easy to confuse with the winml eval command (dataset accuracy evaluation). This parameter gates winml perf, so runPerf would be unambiguous — especially since the Eval Report pipeline already owns the eval terminology.

.pipelines/templates/e2e-test-jobs.yml
```diff
@@ -0,0 +1,243 @@
...
+    pool:
+      name: modelkit-selfhost-pool
+      demands:
+        - Agent.Name -equals ${{ parameters.agentName }}
+    timeoutInMinutes: 360
+    cancelTimeoutInMinutes: 2
+
+    steps:
+      - checkout: self
+        clean: false
```
DingmaomaoBJTU, line 58: Nit: clean: false is intentional to preserve HF model download cache between daily runs (avoids re-downloading multi-GB models each day). A short inline comment would prevent a future maintainer from flipping it to true to fix a stale-artifact issue without realizing the caching cost.

tests/e2e/test_config_e2e.py
```diff
@@ -46,19 +46,28 @@
...
 
     def _resolve_device_mock(
         device: str = "auto", *, ep: str | None = None
     ) -> tuple[str, list[str]]:
         # Keep tests deterministic while preserving explicit device requests.
+        ep_name = normalize_ep_name(ep)
         normalized = (device or "auto").lower()
+        if ep_name in EP_SUPPORTED_DEVICES:
+            supported = list(EP_SUPPORTED_DEVICES[ep_name])
+            chosen = normalized if normalized in supported else supported[0]
```
DingmaomaoBJTU: Silent fallback: if the caller requests a device not in EP_SUPPORTED_DEVICES[ep_name] (e.g. ep=qnn, device=cpu), the mock silently returns supported[0] instead of the requested device. This could mask tests that accidentally pass an unsupported EP/device combination — the mock succeeds while real hardware would fail. Consider asserting normalized in supported, or logging a warning.

.pipelines/templates/e2e-test-jobs.yml
KayMKM: run_eval.py result contains the model name

.pipelines/templates/e2e-test-jobs.yml
KayMKM: this is a very rare case.. and if we migrate the pipeline to onebranch we have to cache them, we can refactor this at that time

.pipelines/templates/e2e-test-jobs.yml
KayMKM: run_eval.py --clean-cache flag will do this

tests/e2e/test_config_e2e.py
KayMKM: this is not an available case, added a warning for it

.pipelines/Modelkit E2E Test.yml
KayMKM: the name indicates using run_eval script

zhenchaoni (approved):

PR #768, opened 2026-05-27
chore: enable checking types and fix analyze folder

xieofxie:


src/winml/modelkit/analyze/core/information_engine.py
```diff
@@ -639,6 +643,9 @@ def _query_doc_constraints(self, runtime_result: PatternRuntime, pattern_id: str
                 return None
 
             # Query doc checker
+            if self._doc_checker is None:
```
xieofxie, line 646: failed to initialize is allowed in constructor, so it is possible None.

Although callers checks it, to keep it consistent, add the check

DingmaomaoBJTU:
### Code review

Found 1 issue:

1. New absolute imports in `src/` files violate [CLAUDE.md](https://github.com/microsoft/winml-cli/blob/d541d50024fd731a899391ba459d55b38e2f3e71/CLAUDE.md#L30-L35) rule: "Never use absolute `from winml.modelkit.*` paths in source code". Both `config.py` and `match.py` add `from winml.modelkit.pattern.models import Pattern as PatternModel` inside `TYPE_CHECKING` blocks. These should use relative imports: `from .models import Pattern as PatternModel`.

https://github.com/microsoft/winml-cli/blob/d541d50024fd731a899391ba459d55b38e2f3e71/src/winml/modelkit/pattern/config.py#L22-L25

https://github.com/microsoft/winml-cli/blob/d541d50024fd731a899391ba459d55b38e2f3e71/src/winml/modelkit/pattern/match.py#L22-L25

DingmaomaoBJTU (approved):

PR #779, opened 2026-05-28
example: add readme and example.py for microsoft/table-transformer-detection

xieofxie:
Also update scripts/e2e_eval/run_pytorch_baseline.py to include pytorch model latency

### DIFF
--- a/examples/microsoft-table-transformer-detection/README.md
+++ b/examples/microsoft-table-transformer-detection/README.md
@@ -145,7 +145,8 @@ for WinML and from the latency JSON line on stdout for the PyTorch baseline.
 
 Result on CPU Intel(R) Core(TM) Ultra 7 258V:
 
-| Run | Device | Precision | mAP | mean latency (ms) | p50 latency (ms) |
+| Model | Device | Precision | mAP | mean latency (ms) | p50 latency (ms) |
 |---|---|---|---|---|---|
-| PyTorch baseline | CPU | fp32 | 0.988714 | _from step 3 latency line_ | _from step 3 latency line_ |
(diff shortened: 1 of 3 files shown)


examples/microsoft-table-transformer-detection/README.md
```diff
@@ -0,0 +1,152 @@
...
+The output JSON contains `latency_ms` (`mean`, `min`, `max`, `p50`, `p90`,
+`p95`, `p99`, `std`) and `throughput` (`samples_per_sec`, `batches_per_sec`).
+Mean and p50 latency are the headline numbers; report them alongside the
+device and precision used.
+
+---
+
+## 4. Evaluate the original PyTorch model
+
+`run_pytorch_baseline.py` loads the HuggingFace checkpoint with native PyTorch
```
zhenchaoni, line 105: Do we want to expose the functionality to evaluate PyTorch model? Then I would propose to update the `wimnl eval` command to have this function. The code change won't be much, you just need to use HF's "from_pretrained" instead of WinML's "from_pretrained" method.

zhenchaoni (changes requested):

examples/microsoft-table-transformer-detection/README.md
xieofxie: yes. It is both OK for me. Currently I just use what our e2e eval script uses.

Note that even `winml eval` supports it, `winml perf` may not.. because monitor or some functions rely on the ort session

xieofxie: To make things simple, I think it worth a change. We have a skip_build for WinMLAutoModel.from_onnx, maybe we just add it also to WinMLAutoModel.from_pretrained

zhenchaoni (approved):

PR #780, opened 2026-05-28
fix: config validates device/EP combination without system check

### LINKED ISSUE #765: config: should not use resolve_device
should only verify if the combination is valid, but don't check current device

xieofxie:
## Summary

- Adds `resolve_check_device_ep` helper that validates a (device, EP) combination without requiring the device/EP to actually exist on the system. Closes #765.
- `commands/config.py` and `config/build.py` now use `resolve_check_device_ep` instead of `resolve_device` so `winml config` no longer hard-fails on hosts where the requested EP isn't installed.
- When `device=auto` or `ep=None`, the helper delegates to the existing `resolve_device` + `resolve_eps` flow (system-aware behavior preserved). When both `device` and `ep` are explicit, it only validates against the static `EP_SUPPORTED_DEVICES` mapping.
- CLI cleanup: `-m/--model`, `-c/--config`, `--device` for the config command now use the shared `cli_utils.*_option` decorators.

## Tests

- New `TestResolveCheckDeviceEp` class in `tests/unit/sysinfo/test_device.py` covering both code paths (delegation and static-only) plus error cases (unknown EP, unsupported device, case-insensitivity).
- Existing config-test mocks updated from `resolve_device` to `resolve_check_device_ep` (`tests/unit/config/conftest.py`, `tests/unit/config/test_build.py`, `tests/unit/config/test_build_onnx.py`, `tests/unit/commands/test_config_cli.py`) so the lazy import in `config/build.py` is intercepted.

### DIFF
--- a/src/winml/modelkit/sysinfo/device.py
+++ b/src/winml/modelkit/sysinfo/device.py
@@ -240,12 +240,19 @@ def resolve_check_device_ep(
 
     Raises:
         ValueError: If the requested device or EP combination is not valid.
+
+    Returns:
+    Tuple of (resolved_device, available_devices, available_eps) where:
+    - resolved_device: The device that should be used based on the input parameters.
+    - available_devices: List of devices that are compatible with the first in available_eps
+    - available_eps: List of EPs that are compatible with the resolved device.
     """
     ep_name = normalize_ep_name(ep)
     if device == "auto" or ep_name is None:
-        resolved_device, available_devices = resolve_device(device=device, ep=ep_name)
+        resolved_device, _ = resolve_device(device=device, ep=ep_name)
         available_eps: list[EPName] = resolve_eps(resolved_device) if ep_name is None else [ep_name]
-        return resolved_device, available_devices, available_eps
+        supported_devices = EP_SUPPORTED_DEVICES[available_eps[0]]
+        return resolved_device, list(supported_devices), available_eps
 
     if ep_name not in EP_SUPPORTED_DEVICES:
         raise ValueError(f"Unknown EP '{ep}'. Expected one of: {sorted(EP_SUPPORTED_DEVICES)}")
--- a/src/winml/modelkit/sysinfo/device.py
+++ b/src/winml/modelkit/sysinfo/device.py
@@ -243,7 +243,7 @@ def resolve_check_device_ep(
     """
     ep_name = normalize_ep_name(ep)
     if device == "auto" or ep_name is None:
-        resolved_device, available_devices = resolve_device(device=device, ep=ep)
+        resolved_device, available_devices = resolve_device(device=device, ep=ep_name)
         available_eps: list[EPName] = resolve_eps(resolved_device) if ep_name is None else [ep_name]
         return resolved_device, available_devices, available_eps
 
--- a/src/winml/modelkit/sysinfo/device.py
+++ b/src/winml/modelkit/sysinfo/device.py
@@ -144,35 +144,6 @@ def _get_available_eps() -> frozenset[EPName]:
     return frozenset(ep for eps in _get_device_ep_map_from_ort().values() for ep in eps)
 
 
-def check_device(*, device: str = "auto", ep: EPNameOrAlias | None = None) -> None:
-    """Check that the requested device and/or EP combination is valid, raising if not.
-
-    Ideal for commands that do not need the device + ep actually exists on the system.
-
-    Args:
-        device: "auto", "npu", "gpu", or "cpu".
-        ep: Optional EP short name (e.g., "qnn", "dml"). When set,
-            availability is checked and an error is raised if no compatible EP
-            is found.
-
-    Raises:
-        ValueError: If the requested device or EP combination is not valid.
-    """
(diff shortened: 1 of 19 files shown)


DingmaomaoBJTU:
### Code review

Found 1 issue:

1. **`IndexError` on `_resolved_eps[0]` when no EPs are installed** -- In `commands/config.py`, the old code guarded EP display with `if ep:`, only printing when the user explicitly passed `--ep`. The new code unconditionally accesses `_resolved_eps[0]`. When `device="auto"` (default) and `ep=None` (not specified), `resolve_check_device_ep` delegates to `resolve_eps(resolved_device)`, which returns `[]` on systems with no EPs registered (CI runners, fresh machines, builds without the Windows ML ORT extensions). The unconditional `_resolved_eps[0]` then raises `IndexError: list index out of range`, crashing `winml config` on the most common invocation pattern. The test mock always returns `["QNNExecutionProvider"]`, so this path is untested.

https://github.com/microsoft/winml-cli/blob/dd79c41b14c0004546eea48131e302122d288897/src/winml/modelkit/commands/config.py#L441-L445

timenick (commented):
probe

src/winml/modelkit/sysinfo/device.py
```diff
@@ -223,3 +223,36 @@ def resolve_eps(resolved_device: str) -> list[EPName]:
...
+
+    if ep_name not in EP_SUPPORTED_DEVICES:
+        raise ValueError(f"Unknown EP '{ep}'. Expected one of: {sorted(EP_SUPPORTED_DEVICES)}")
+    supported_devices = EP_SUPPORTED_DEVICES[ep_name]
+    if device.lower() not in supported_devices:
+        raise ValueError(
+            f"EP '{ep}' does not support device '{device}'. "
+            f"Supported devices for {ep_name}: {', '.join(supported_devices)}."
+        )
+    return device.lower(), list(supported_devices), [ep_name]
```
timenick, line 265: **The middle return element (`available_devices`) has different meanings on the two paths.**

Path A (when `device == "auto"` or `ep_name is None`):
```python
resolved_device, available_devices = resolve_device(device=device, ep=ep_name)
# available_devices = devices actually registered on this system
```

Path B (explicit device + explicit ep):
```python
supported_devices = EP_SUPPORTED_DEVICES[ep_name]
# supported_devices = devices this EP can theoretically target (static lookup)
return device.lower(), list(supported_devices), [ep_name]
```

Same tuple shape, but the middle element’s meaning depends on which branch fired. Today’s only caller (`resolve_quant_compile_config` → `resolve_precision`) only reads `available_devices` in the `device="auto" + explicit precision` path, which doesn’t fire on Path B — so the divergence is dormant. But a future caller assuming “available_devices means what’s registered on this system” will get wrong data on Path B.

Two cheap options:

1. **Docstring note.** Add a `Returns:` section that explicitly calls out the per-path semantics, e.g.:

   ```
   Returns:
       Tuple of (resolved_device, devices, eps).
       - On the auto/None path: `devices` is the system-available list from resolve_device.
       - On the explicit-both path: `devices` is the EP's statically supported list,
         NOT what's installed on this machine. Use is_*_available() if you need a
         system check on the explicit path.
   ```

2. **Split the function.** `resolve_device_ep(...)` for Path A (system-aware) and `validate_device_ep_combo(...)` for Path B (static-only). Then each return tuple has a single clear semantics, and the call sites self-document which mode they want (`commands/config.py` calls the validator; everything else calls the resolver).

Not a blocker — just a doc/clarity ask before the second consumer lands.

timenick (commented):
One nit on the new helper. The design intent is clear and well-documented in the PR description; only flagging this because the function returns the same tuple shape from two paths with different semantics.

src/winml/modelkit/sysinfo/device.py
xieofxie: good point! let me update

xieofxie: now the middle value follows the same pattern

xieofxie:
> ### Code review
> Found 1 issue:
> 
> 1. **`IndexError` on `_resolved_eps[0]` when no EPs are installed** -- In `commands/config.py`, the old code guarded EP display with `if ep:`, only printing when the user explicitly passed `--ep`. The new code unconditionally accesses `_resolved_eps[0]`. When `device="auto"` (default) and `ep=None` (not specified), `resolve_check_device_ep` delegates to `resolve_eps(resolved_device)`, which returns `[]` on systems with no EPs registered (CI runners, fresh machines, builds without the Windows ML ORT extensions). The unconditional `_resolved_eps[0]` then raises `IndexError: list index out of range`, crashing `winml config` on the most common invocation pattern. The test mock always returns `["QNNExecutionProvider"]`, so this path is untested.
> 
> https://github.com/microsoft/winml-cli/blob/dd79c41b14c0004546eea48131e302122d288897/src/winml/modelkit/commands/config.py#L441-L445
> 
> 

for a resolved device, there should always be ep

timenick (approved):

PR #783, opened 2026-05-28
fix: run eval failed on amd

KayMKM:
## Summary

Fixes `scripts/e2e_eval/run_eval.py` crashing on VitisAI EP (AMD Ryzen AI NPU) and a latent bug in `winml build` that prevented the script's `--no-quant` workaround from actually taking effect.

The crash: VitisAI ships its own internal quantizer and runs it at session-create time. Layering winml's generic QDQ quantization pass on top produces a model VitisAI cannot consume, which manifests as `DpuKernelRunner.cpp:1920 DPU timeout` during `winml perf`. The fix is to tell winml to skip its own quantization when the selected EP quantizes natively.

### `src/winml/modelkit/commands/build.py` — root-cause fix (1 line)

When `--device` was passed to `winml build`, the internal `_patch_device` helper unconditionally re-populated `cfg.quant` with the device's default quantization config, silently undoing any prior `--no-quant`. The condition now respects `no_quant`:

```python
if no_quant or resolved_quant is None:
    cfg.quant = None
```

Without this, `winml build … --device npu --no-quant` still produced a `_quantized.onnx` artifact.

### `scripts/e2e_eval/run_eval.py` — script wiring

- New canonical-name set `_NATIVE_QUANT_EPS = {"VitisAIExecutionProvider"}` plus a helper `_ep_quantizes_natively(ep)` that funnels both canonical names and user aliases (e.g. `vitisai`) through `winml.modelkit.utils.constants.normalize_ep_name`. No hardcoded aliases.
- `_resolve_precision(...)` gained an `ep` parameter; for native-quant EPs it returns `None` so no precision flag is sent.
- `_run_build` now passes `--no-quant` to **both** `winml config` (so the persisted `build_config.json` has `quant: null` up-front) and `winml build` (defense in depth) when the EP quantizes natively.
- Call sites in `run_model` and `main` updated to thread `ep` through `_resolve_precision`.

## Why the earlier commits in this branch weren't enough

The first attempt (`fix(run_eval): skip quantize when VitisAI EP is selected`) wired `--no-quant` only into `winml build`. That didn't take effect because of the `_patch_device` bug above. The second attempt (`fix(vitisai): resolve auto-precision to w8a8 for VitisAI NPU`) tried to switch precision instead of skipping — also wrong, since VitisAI wants an fp32 input and quantizes it itself. The final state keeps the script clean (`--no-quant`, no precision override) and fixes the actual `winml build` bug.

## Verification

Manual end-to-end on AMD Ryzen AI (VitisAI NPU), with a clean `~/.cache/winml/artifacts/...` and output dir:

```pwsh
uv run --no-sync python scripts/e2e_eval/run_eval.py `
  --hf-model facebook/convnext-tiny-224 `
  --task image-classification `
  --device npu --ep vitisai `
  --eval-type perf --no-report --verbose --timeout 1800 `
  --output-dir e2e-test\vitisai_npu
```

Before: `winml perf` crashed with `DpuKernelRunner.cpp:1920 DPU timeout`.
After:
- Cached `imgcls_*_winml_build_config.json` has `"quant": null`.
- No `_quantized.onnx` artifact produced.
- Perf step: **PASS** in ~120 s.

### DIFF
--- a/scripts/e2e_eval/run_eval.py
+++ b/scripts/e2e_eval/run_eval.py
@@ -81,26 +81,28 @@ TIMEOUT_SKIP_LIST_PATH = Path(__file__).parent / "cache" / "timeout_skip_list.js
 _DEFAULT_SAMPLES = 1000
 _DEFAULT_PRECISION_NPU = "w8a16"
 
-# EPs that ship their own internal quantizer (run after the ONNX model is
-# loaded into the EP).  Layering winml's generic QDQ quantizer on top of
-# them noticeably degrades accuracy, so we pass ``--no-quant`` to
-# ``winml build`` and let the EP quantize natively.
+# EPs whose eval track keeps the model unquantized (the "fp" variant)
+# rather than running winml's QDQ pass on top.  This is an eval-setup
+# choice -- e.g. VitisAI / AMD Ryzen AI is benchmarked on the fp32/fp16
+# model -- not a claim about the EP's internal pipeline.  For these EPs
+# the harness passes ``--no-quant`` to both ``winml config`` and
+# ``winml build`` (see :func:`_run_build` and :func:`run_model`).
 #
 # Entries are canonical ``EPName`` values (the ``*ExecutionProvider`` form);
 # user-facing aliases like ``vitisai`` are normalised via
-# ``normalize_ep_name`` in :func:`_ep_quantizes_natively` so we only have to
-# list each EP once here.
-_NATIVE_QUANT_EPS = frozenset({"VitisAIExecutionProvider"})
+# ``normalize_ep_name`` in :func:`_should_skip_winml_quant` so each EP only
+# needs to be listed once.
+_EPS_SKIP_WINML_QUANT = frozenset({"VitisAIExecutionProvider"})
 
 
-def _ep_quantizes_natively(ep: str | None) -> bool:
-    """True if the EP ships its own internal quantizer."""
+def _should_skip_winml_quant(ep: str | None) -> bool:
+    """True if the eval harness should run this EP on the unquantized model."""
     # Lazy import: keeps ``scripts/e2e_eval`` cheap to load (winml.modelkit
     # transitively imports onnxruntime) and matches the existing in-function
     # import pattern used elsewhere in this script.
     from winml.modelkit.utils.constants import normalize_ep_name
 
-    return normalize_ep_name(ep) in _NATIVE_QUANT_EPS
+    return normalize_ep_name(ep) in _EPS_SKIP_WINML_QUANT
 
 
 def _resolve_precision(device: str, explicit: str | None, ep: str | None = None) -> str | None:
@@ -112,16 +114,24 @@ def _resolve_precision(device: str, explicit: str | None, ep: str | None = None)
     (NHWC layout transformer inserts Conv nodes that QNN GPU's GetCapability
     does not claim).
 
-    For EPs with native quantizers (e.g. VitisAI) the flag is omitted; the
-    actual "skip winml quantization" instruction is conveyed via
-    ``winml build --no-quant`` in :func:`_run_build`.
+    For EPs in :data:`_EPS_SKIP_WINML_QUANT` (e.g. VitisAI) the flag is forced
+    off regardless of ``explicit``: the harness pairs these EPs with
+    ``--no-quant`` at config/build time, so a non-empty ``--precision`` would
+    produce a config that says "quantize to X" while the build says "skip
+    quantization" -- a contradiction.  An explicit value is dropped with a
+    one-line warning so the override is visible in the log.
 
-    An explicit per-model precision always takes precedence.
+    Otherwise an explicit per-model precision always takes precedence.
     """
+    if _should_skip_winml_quant(ep):
+        if explicit:
+            safe_print(
+                f"  [precision] Ignoring explicit precision={explicit!r} for EP {ep!r}: "
+                "this EP is run on the unquantized variant (--no-quant)."
+            )
+        return None
     if explicit:
         return explicit
-    if _ep_quantizes_natively(ep):
-        return None
     return _DEFAULT_PRECISION_NPU if device == "npu" else None
 
 
@@ -433,11 +443,12 @@ def _run_build(
         config_args += ["--task", entry.task]
     if ep:
         config_args += ["--ep", ep]
-    # Native-quant EPs: also pass --no-quant to winml config so the generated
-    # build_config.json is written with quant=None up-front. Otherwise on NPU
-    # the config command would still apply its default precision (w8a16) and
-    # we'd be relying on --no-quant at build time to override it.
-    if _ep_quantizes_natively(ep):
+    # EPs in _EPS_SKIP_WINML_QUANT are evaluated on the unquantized variant.
+    # Pass --no-quant to winml config so the generated build_config.json is
+    # written with quant=None up-front; otherwise on NPU the config command
+    # would still apply its default precision (w8a16) and we'd be relying on
+    # --no-quant at build time alone to override it.
+    if _should_skip_winml_quant(ep):
         config_args += ["--no-quant"]
 
     config_proc = _run_subprocess(config_args, timeout)
@@ -479,11 +490,10 @@ def _run_build(
         ]
         if ep:
             build_args += ["--ep", ep]
-        # EPs with native quantizers (e.g. VitisAI / AMD Ryzen AI) quantize
-        # internally at session-create time; running winml's generic QDQ
-        # pass on top noticeably degrades accuracy.  --no-quant tells
-        # winml build to keep the model fp32 and lets the EP do its own.
-        if _ep_quantizes_natively(ep):
+        # Mirror the --no-quant passed to winml config above so the build
+        # stage also skips QDQ regardless of what the config carries (defence
+        # in depth; see _EPS_SKIP_WINML_QUANT for the rationale).
+        if _should_skip_winml_quant(ep):
             build_args += ["--no-quant"]
 
         build_proc = _run_subprocess(build_args, timeout)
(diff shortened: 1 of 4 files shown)


src/winml/modelkit/config/precision.py
```diff
@@ -35,6 +35,14 @@
     "cpu": "fp16",
 }
 
+# EP-specific auto-precision overrides (applied when precision="auto" and
+# the EP appears in this map).  VitisAI (AMD Ryzen AI NPU) requires INT8
+# weights + INT8 activations; the generic NPU w8a16 default produces a model
+# the EP loads but crashes on first inference.
+_EP_AUTO_PRECISION: dict[EPName, str] = {
+    "VitisAIExecutionProvider": "w8a8",
```
DingmaomaoBJTU: w8a8 will cause accuracy drop. By default no quant for amd?

KayMKM: updated

scripts/e2e_eval/run_eval.py
```diff
@@ -81,8 +81,19 @@
 _DEFAULT_SAMPLES = 1000
 _DEFAULT_PRECISION_NPU = "w8a16"
 
+# EPs that ship their own internal quantizer (run after the ONNX model is
+# loaded into the EP).  Layering winml's generic QDQ quantizer on top of
+# them noticeably degrades accuracy, so we pass ``--no-quant`` to
+# ``winml build`` and let the EP quantize natively.
+_NATIVE_QUANT_EP_NAMES = frozenset({"vitisai", "vitisaiexecutionprovider"})
```
DingmaomaoBJTU: Why are both vitisai and vitisaiexecutionprovider in this set? The two entries seem to cover a short-form alias vs. the full ORT provider name (case is already handled by .lower()). A brief inline comment clarifying the distinction would help future readers who wonder what string to pass to --ep.

scripts/e2e_eval/run_eval.py
```diff
@@ -91,10 +102,16 @@ def _resolve_precision(device: str, explicit: str | None) -> str | None:
...
+    For EPs with native quantizers (e.g. VitisAI) the flag is omitted; the
+    actual "skip winml quantization" instruction is conveyed via
+    ``winml build --no-quant`` in :func:`_run_build`.
+
     An explicit per-model precision always takes precedence.
     """
     if explicit:
         return explicit
+    if _ep_quantizes_natively(ep):
+        return None
```
DingmaomaoBJTU: When _ep_quantizes_natively is true and device == npu, returning None means winml config runs without --precision. On NPU the tool may still apply its own default (w8a16), so build_config.json could still carry quant settings that --no-quant in _run_build must then suppress.

Worth either documenting that --no-quant fully overrides the config precision, or also passing --precision fp32 to winml config so the config is written clean in the first place.

scripts/e2e_eval/run_eval.py
```diff
@@ -446,6 +463,12 @@ def _run_build(
         ]
         if ep:
             build_args += ["--ep", ep]
+        # EPs with native quantizers (e.g. VitisAI / AMD Ryzen AI) quantize
+        # internally at session-create time; running winml's generic w8a8
```
DingmaomaoBJTU: Nit: comment says "generic w8a8 QDQ pass" but _DEFAULT_PRECISION_NPU = "w8a16" -- the NPU default is w8a16, not w8a8. Should read "w8a16 QDQ pass" (or just "generic QDQ pass") to stay consistent with the constant above.

scripts/e2e_eval/run_eval.py
DingmaomaoBJTU: Why are both `"vitisai"` and `"vitisaiexecutionprovider"` needed here? If `.lower()` already normalises case (so `"VitisAI"` and `"vitisai"` both match), it looks like the two entries cover a *short-form* alias vs. the full ORT provider name. A brief inline comment explaining that would help future readers know what string a caller should actually pass and which form `--ep` accepts.

scripts/e2e_eval/run_eval.py
DingmaomaoBJTU: When `_ep_quantizes_natively` is true and `device == "npu"`, returning `None` causes `winml config` to run **without `--precision`**. On NPU, `winml config` may still pick its own default precision (w8a16), which means the generated `build_config.json` could still carry quantisation settings. The `--no-quant` in `_run_build` should then override that, but the correctness of this path depends on `--no-quant` fully suppressing the config's quant directives.

Worth either:
- verifying (and documenting here) that `--no-quant` takes full precedence over the config's precision field, or
- also passing `--precision fp32` (or a dedicated `--no-quant` to `winml config` if it supports one) so the config itself is written clean, making the two-step behaviour more self-evident.

scripts/e2e_eval/run_eval.py
DingmaomaoBJTU: Nit: the comment says "generic **w8a8** QDQ pass" but `_DEFAULT_PRECISION_NPU = "w8a16"` ? the default is actually w8a16, not w8a8. Should say "w8a16 QDQ pass" (or just "generic QDQ pass") to stay consistent with the constant above.

scripts/e2e_eval/run_eval.py
KayMKM: reuse winml.modelkit.utils.constants to resolve the ep names

scripts/e2e_eval/run_eval.py
KayMKM: fixed

scripts/e2e_eval/run_eval.py
KayMKM: fixed

KayMKM: fixed

scripts/e2e_eval/run_eval.py
KayMKM: fixed

scripts/e2e_eval/run_eval.py
KayMKM: fixed

scripts/e2e_eval/run_eval.py
```diff
@@ -549,7 +572,7 @@ def run_model(
     """
     if not onnx_paths:
         # No pre-built paths: fall back to HF model ID (single model only)
-        precision = _resolve_precision(device, None)
+        precision = _resolve_precision(device, None, ep=ep)
```
DingmaomaoBJTU, line 603: **Latent bug: missing `--no-quant` in the `run_model` fallback path.**

When `onnx_paths` is `None`/empty, this branch calls `winml perf -m {hf_id} --ep vitisai` directly (winml perf does its own internal build). `_resolve_precision` correctly returns `None` so `--precision` is omitted, but `--no-quant` is never appended here -- only `_run_build` adds it. If `winml perf` triggers an internal quantization step when given an HF model ID, the model will still be quantized.

In the current `main()` flow this branch is unreachable for VitisAI (build always runs first and populates `onnx_paths`), but `run_model` is a public function. Consider appending `--no-quant` after the `--ep` block (around line 589) to keep the function self-consistent regardless of call site.

scripts/e2e_eval/run_eval.py
DingmaomaoBJTU: **Explicit precision + native-quant EP conflict.**

If `entry.precision` is set (e.g. `"w8a8"`) for a VitisAI model, `_resolve_precision` returns it and it gets forwarded to `winml config --precision w8a8`. Meanwhile `_run_build` unconditionally appends `--no-quant` to `winml build` for any VitisAI EP. The result is a config that says 'quantize to w8a8' but a build that says 'skip quantization' -- two conflicting instructions whose resolution depends on undocumented `winml build` precedence rules.

Suggestion: either emit a warning and ignore `explicit` for native-quant EPs, or assert/document that `--no-quant` always wins over the config precision field.

scripts/e2e_eval/run_eval.py
```diff
@@ -81,8 +81,19 @@
...
+# loaded into the EP).  Layering winml's generic QDQ quantizer on top of
+# them noticeably degrades accuracy, so we pass ``--no-quant`` to
+# ``winml build`` and let the EP quantize natively.
+_NATIVE_QUANT_EP_NAMES = frozenset({"vitisai", "vitisaiexecutionprovider"})
 
-def _resolve_precision(device: str, explicit: str | None) -> str | None:
+
+def _ep_quantizes_natively(ep: str | None) -> bool:
+    """True if the EP ships its own internal quantizer."""
+    return (ep or "").lower() in _NATIVE_QUANT_EP_NAMES
```
DingmaomaoBJTU: **No tests for the new logic.** The three new code paths (`_ep_quantizes_natively`, the `_resolve_precision(ep=...)` branch, and the `--no-quant` injection in `_run_build`) have no unit tests. Since a wrong value here silently degrades accuracy rather than crashing, a few pytest parametrised cases would be valuable -- e.g. assert `_resolve_precision('npu', None, ep='vitisai') is None`, `_resolve_precision('npu', None, ep=None) == 'w8a16'`, and a `_run_build` mock that verifies `--no-quant` is present/absent based on ep.

scripts/e2e_eval/run_eval.py
```diff
@@ -81,8 +81,29 @@
...
+# Entries are canonical ``EPName`` values (the ``*ExecutionProvider`` form);
+# user-facing aliases like ``vitisai`` are normalised via
+# ``normalize_ep_name`` in :func:`_ep_quantizes_natively` so we only have to
+# list each EP once here.
+_NATIVE_QUANT_EPS = frozenset({"VitisAIExecutionProvider"})
 
-def _resolve_precision(device: str, explicit: str | None) -> str | None:
+
+def _ep_quantizes_natively(ep: str | None) -> bool:
+    """True if the EP ships its own internal quantizer."""
```
DingmaomaoBJTU: This is not true, we skip it not because amd has internal quantizer. It just because we'd like to keep the fp16 as variant.

KayMKM: renamed function name and updated comments

scripts/e2e_eval/run_eval.py
KayMKM: fixed

scripts/e2e_eval/run_eval.py
KayMKM: fixed

scripts/e2e_eval/run_eval.py
KayMKM: added tests

DingmaomaoBJTU (approved):

PR #785, opened 2026-05-28
examples: add 12 builtin model recipes (fp16 + w8a8 + w8a16, 36 configs)

ssss141414:
Adds curated recipe configs for the 12 builtin models — those that pass fp16 eval on all 9 (EP, device) buckets.


DingmaomaoBJTU (approved):

timenick (changes requested):
Perhaps we put these in examples/recipes, as we already have other samples created by the #779

timenick (approved):

Issue #782, opened 2026-05-28
`winml inspect -m facebook/dinov2-base --task image-feature-extraction` returns unsupported


zhenchaoni:
Repro

```
winml inspect -m facebook/dinov2-base --task image-feature-extraction
```

It says
 Model Type         dinov2                                                                                         │
│   Task               image-feature-extraction (via TasksManager)                                                    │
│   Architectures      Dinov2Model                                                                                    │
│   Overall Support    Unsupported    

The expected behavior is the model should be exported.

DingmaomaoBJTU:
**Triage summary**

- **Component**: Inspect
- **Priority**: P1

`winml inspect` incorrectly reports dinov2 model with `image-feature-extraction` task as "Unsupported". The model should be exportable — likely the resolver doesn't recognize the dinov2 architecture or the `image-feature-extraction` task mapping is missing. Suggested next step: check the inspect resolver's architecture/task support table and add dinov2 support.

DingmaomaoBJTU:
## Triage (2026-06-03)

**Priority:** P2
**Component:** inspect
**Owner:** @DingmaomaoBJTU
**Root Cause:** `winml inspect` incorrectly reports dinov2 (image-feature-extraction) as unsupported. Likely the support check does not account for the dinov2 architecture or the normalized task mapping from optimum.
**Action:** Fix the inspect support matrix to recognise dinov2/image-feature-extraction. Related to #778 (task resolution gap for image embedding models).

DingmaomaoBJTU:
## Triage (2026-06-05)

**Priority**: P1
**Component**: inspect / loader / task detection
**Owner**: @DingmaomaoBJTU (already assigned)

**Assessment**: winml inspect returns 'Unsupported' for acebook/dinov2-base --task image-feature-extraction. This is related to #800 (modality-aware task detection). The model is valid and exportable — the issue is that the inspect pathway fails to recognize image-feature-extraction as a supported task for DINOv2 architecture because it goes through Optimum's TasksManager which collapses it to \eature-extraction\.

**Action**: Will be resolved by #800's refactor (unified modality-aware task detection). In the interim, mark as blocked-on #800 or apply a targeted fix in inspect to handle the modality-aware mapping.

Issue #778, opened 2026-05-28
winml build and eval command fails on image embedding model like facebook/dinov2-base


zhenchaoni:
**Repro**

```
winml build -m facebook/dinov2-base -o dino
```

or

```
winml eval -m facebook/dinov2-base
```

**Actual:**

The commands failed.

Traceback (most recent call last):
  File "C:\Users\zhenni\repos\winml-cli\src\winml\modelkit\quant\quantizer.py", line 112, in quantize_onnx
    data_reader = DatasetCalibrationReader(
                  ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\zhenni\repos\winml-cli\src\winml\modelkit\datasets\__init__.py", line 192, in __init__
    self._dataset = universal_calib_dataset(
                    ^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\zhenni\repos\winml-cli\src\winml\modelkit\datasets\__init__.py", line 124, in universal_calib_dataset
    raise RuntimeError(f"Failed to create {task} dataset: {e}") from e
RuntimeError: Failed to create feature-extraction dataset: <class 'transformers.models.dinov2.configuration_dinov2.Dinov2Config'>

**Expected:**

The commands should success.

**Root cause:**

The model task resolution is by calling optimum. Optimum only resolves to `normalized` task name. In this case, the image embedding model and text embedding model are both resolved to `feature-extraction`. However, in quantization and evaluation step, we use HF task name to pick the calibration dataset or evaluator. In HF task, feature-extraction is for text model and image-feature-extraction is for image model. As a result, we use text dataset to quantize or evaluate an image model, causing the failure;.

DingmaomaoBJTU:
**Triage summary**

- **Component**: Build
- **Priority**: P0

`winml build` and `winml eval` fail on all image embedding models (e.g. dinov2) because optimum normalizes `image-feature-extraction` to `feature-extraction`, then the pipeline uses a text calibration dataset for an image model. Suggested next step: add modality-aware task disambiguation so that image models get routed to `image-feature-extraction` datasets and evaluators.

Issue #724, opened 2026-05-25
[task] Source-of-truth inconsistency: 13 task names in TASK_ABBREV are rejected by validate_task


timenick:
**Component**: Loader / task
**Severity**: P2 — correctness, not user-facing crash

`src/winml/modelkit/loader/task.py` defines `TASK_ABBREV` (used for cache-key generation) and — after #717 — `KNOWN_TASKS` (used by `--list-tasks` / `validate_task`). 18 task names live in `TASK_ABBREV` but not in `KNOWN_TASKS`:

- **5 are aliases** that `normalize_task()` collapses to canonical forms — these legitimately don't belong in `KNOWN_TASKS`:
  `pretraining→feature-extraction`, `sequence-classification→text-classification`,
  `summarization→text2text-generation`, `translation→text2text-generation`,
  `zero-shot-classification→text-classification`

- **13 are canonical task names that are actively referenced elsewhere in the codebase** but `validate_task()` rejects them:
  `instance-segmentation`, `universal-segmentation`, `keypoint-matching`,
  `masked-image-modeling`, `video-classification`, `text-encoding`,
  `table-question-answering`, `audio-tokenization`,
  `zero-shot-audio-classification`, `any-to-any`, `multimodal-lm`,
  `backbone`, `time-series-prediction`

**Evidence the 13 are real**:
- `src/winml/modelkit/inference/tasks.py` defines `TaskInputSpec` entries for `video-classification`, `keypoint-matching`, `table-question-answering`, `zero-shot-audio-classification`
- `src/winml/modelkit/models/hf/bart.py:484` — `@register_composite_model("bart", "table-question-answering")`
- `scripts/e2e_eval/testsets/models_all.json` tags real models with `video-classification`, `any-to-any`, `table-question-answering`
- `tests/unit/commands/test_run_task_combos.py:342` runs `keypoint-matching`
- `tests/unit/serve/test_predict_file_inputs.py` covers `keypoint-matching`, `zero-shot-audio-classification`

**Symptom**: `winml inspect -m <id> --task video-classification` raises `Unknown task 'video-classification'`, even though `inference/tasks.py` understands the task and `cache/path.py` would happily abbreviate it.

**Surfaced by**: review on #717. The inconsistency pre-dates that PR — `TASK_ABBREV` had these entries since the file was added — but #717 made it visible by introducing `KNOWN_TASKS`. Out of scope for #717.

**Proposed fix** (one of):
[...]

PR #786, opened 2026-05-29
Fix model-task inconsistency for vision feature-extraction models

### LINKED ISSUE #777: Model task handling is not consistent between config and build
**Repro step:**

Generate a config with --task image-feature-extraction
```
winml config -m facebook/dinov2-base --task image-feature-extraction -o dino.json
```

Use the the config json file to build

```
winml build -m facebook/dinov2-base --config dino.json -o dino
```

**Actual:**
There is an error message showing `image-feature-extraction` is not supported for the model. 

[05/28/26 11:33:36] WARNING  Output names count mismatch: config has 1 ['last_hidden_state'], model trace 
                             inferred 2 ['last_hidden_state', 'pooler_output']. Keeping config names (from
                             Optimum OnnxConfig).                                                         
                    WARNING  Optimum model patcher failed for model_type='dinov2',                        
                             task='image-feature-extraction'. Export may produce incorrect results for    
                             models requiring Transformers 4.53+ tracing patches.                         
                             Traceback (most recent call last):                                           
                               File "C:\Users\zhenni\repos\wmk\src\winml\modelkit\export\htp\exporter.py",
                             line 474, in _get_optimum_patcher                                            
                                 cfg_cls = TasksManager.get_exporter_config_constructor(                  
                                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                  
                               File                                                                       
                             "C:\Users\zhenni\repos\wmk\.venv\Lib\site-packages\optimum\exporters\tasks.py
                             ", line 1300, in get_exporter_config_constructor                             
                                 raise ValueError(                                                        
                             ValueError: dinov2 doesn't support task image-feature-extraction for the onnx
                             backend. Supported tasks are: feature-extraction, image-classification. 

**Expected:**

`facebook/dinov2-base` is an image embedding model, image-feature-extraction should be supported

**Root cause:**

get_exporter_config_constructor expects an optimum normalized task name. In this case, both text embedding model and image embedding model  have the same normalized task name "feature-extraction", so the code expects to use feature-extraction. There is an inconsistency inside our code where some places expect normalized task name while other places expect HF model task name.

zhenchaoni:
## Fix model-task inconsistency for vision feature-extraction models

Fixes #777, #778, #782.

### Principle

`winml inspect` is the source of truth for valid `(model_id, task)` pairs. Both `feature-extraction` and `image-feature-extraction` are valid ways to address an image-embedding model like `facebook/dinov2-base`. Downstream commands must accept whichever name `winml inspect` accepts, then use `(model_id, task)` to locate the concrete class to act on.

### Root cause

Optimum's `TasksManager.get_exporter_config_constructor` only knows canonical Optimum task names. Several call sites passed the raw user-supplied task straight through, so HF aliases like `image-feature-extraction` were rejected with "Unsupported". The evaluator additionally needs to know which HF pipeline name to dispatch on, which the canonical Optimum task name doesn't carry by itself for bimodal tasks like `feature-extraction`.

### Fix

- **Inspect / export / HTP exporter**: normalize via `_map_task_synonym(task)` (in `export/io.py`) before any `TasksManager` lookup because it requires normalized task input. This is a single function reused at each `TasksManager` boundary — no new global table.
- **Quantize**: `_resolve_dataset_class(task, io_config)` in `datasets/__init__.py` dispatches to `TextDataset` / `ImageDataset` based on the actual ONNX input names. No `AutoConfig.from_pretrained` round-trip. Bimodal io_configs fall back to `RandomDataset` with a warning.
- **Evaluate**: Because HF pipeline and evaluate library have their task name convention, `to_hf_pipeline_task(task, model_id)` in `eval/evaluate.py` translates to the HF pipeline name the underlying `evaluate` library expects. Uses `OnnxConfig.inputs` (no weights loaded) to pick the modality. Bimodal models (e.g. CLIP combined: both `pixel_values` and `input_ids`) keep the task unchanged via a `len(hits) == 1` guard, preserving the explicit user task.

### Validation

`facebook/dinov2-base`:

| Command | Before | After |
|---|---|---|
| `winml inspect -m facebook/dinov2-base --task image-feature-extraction` | "Unsupported" | Resolves via `Dinov2OnnxConfig` |
| `winml export -m facebook/dinov2-base -t image-feature-extraction` | KeyError on TasksManager | Valid ONNX with `last_hidden_state` |
| `winml eval -m facebook/dinov2-base --task feature-extraction` | `RuntimeError: Failed to create feature-extraction dataset` | kNN metrics on mini-imagenet |
| `winml quantize <onnx> --task feature-extraction -m facebook/dinov2-small` | Failure by using TextDataset | Routes to `ImageDataset` |

`openai/clip-vit-base-patch32` (bimodal, regression check):

- `winml eval -m openai/clip-vit-base-patch32 --task feature-extraction` → stays `feature-extraction` (text STS evaluator); not silently rerouted to image.
- `winml eval -m openai/clip-vit-base-patch32` (auto-detect) → resolves to `feature-extraction` (text).

### Tests

Unit:
- `tests/unit/eval/test_eval.py::TestResolveTask` — auto-detect, explicit task, bimodal guard, HF pipeline translation.
- test_random_dataset.py — `TASK_DATASET_MAPPING` covers all registered tasks, including bimodal dict-of-dict.

E2E (`-m e2e`, dinov2 chosen because it isn't in `MODEL_BUILD_CONFIGS` and so actually exercises the `TasksManager` path):
- `tests/e2e/test_inspect_e2e.py::TestInspectDinoV2` — both `image-feature-extraction` and `feature-extraction` resolve.
- `tests/e2e/test_export_e2e.py::TestExportDinoV2::test_image_feature_extraction`.
- `tests/e2e/test_eval_e2e.py::TestEvalPerTask::test_image_feature_extraction` parameterized over both task names.
- `tests/e2e/test_quantize_e2e.py::test_feature_extraction_with_pixel_values_uses_image_dataset`.


timenick:
Thanks for tackling these three — the e2e coverage on `dinov2` end-to-end across inspect/export/eval/quantize is exactly what was missing. Two concrete concerns and a proposal before this lands.

### Concrete regression: CLIP `feature-extraction` will swap dataset class

`canonical_task_to_known_task` iterates `_OPTIMUM_CANONICAL_TO_KNOWN_TASK["feature-extraction"]` and returns on first input-name match. For CLIP, `CLIPOnnxConfig` for `feature-extraction` declares **both** `pixel_values` and `input_ids` (verified via `TasksManager.get_exporter_config_constructor("onnx", model_type="clip", task="feature-extraction")` — inputs are `['pixel_values', 'input_ids', 'attention_mask']`). With current dict insertion order, `pixel_values` matches first → routes to `image-feature-extraction`.

But `src/winml/modelkit/models/hf/clip.py:53` already encodes the right answer:

```python
("clip", "feature-extraction"): CLIPTextModelWithProjection,        # text encoder
("clip", "image-feature-extraction"): CLIPVisionModelWithProjection, # vision encoder
```

So `winml quantize -m openai/clip-vit-base-patch32 --task feature-extraction`:

- Before this PR: loader picks `CLIPTextModelWithProjection`, datasets picks `TextDataset` → consistent
- After this PR: loader still picks `CLIPTextModelWithProjection`, datasets translated to `image-feature-extraction` → `ImageDataset` → dtype/shape mismatch at calibration

The input-name heuristic doesn't have enough information to disambiguate multi-modal architectures; `MODEL_CLASS_MAPPING` already holds the truth and disagrees with what the heuristic returns.

### `DatasetCalibrationReader.__init__` quietly rewrites `self.task`

Caller passes `task="feature-extraction"`, but after the new lines `self.task` becomes `"image-feature-extraction"`. Any downstream consumer of `reader.task` expecting "what the caller asked for" gets something different. Either name the new attribute (`self.resolved_task`?) or document the semantic shift explicitly on the class.

### Architectural direction — two patches vs. one source fix

The pattern we're ending up with: three places inline-call `_map_task_synonym` (forward), plus a recovery function (reverse) that has to re-derive what was lost. Both directions exist because `_detect_task_from_config` drops modality at the source by calling `TasksManager.infer_task_from_model`, which is Optimum-canonical.

Prevention alternative: have `_detect_task_from_config` return the HF (modality-aware) task to begin with (HF Hub `pipeline_tag` first, then a `(model_type, config)`→HF modality fallback). Then:

- the three inline forward-direction calls collapse into one boundary helper (we'd want this regardless — a lint rule against direct `TasksManager` imports keeps new code honest),
- `canonical_task_to_known_task` and `_OPTIMUM_CANONICAL_TO_KNOWN_TASK` go away,
- the CLIP-style ambiguity above doesn't exist because we never need to recover modality from input names.

### Proposed path

1. Fix the CLIP regression in this PR — it's a marquee architecture and the regression is real.
2. Resolve the `self.task` semantic shift (rename or document).
3. Merge — the e2e fix value for #777/#778/#782 is real and shouldn't wait.
4. I'll open a follow-up doing the prevention refactor: `_detect_task_from_config` → HF task, centralize the forward-normalize into a boundary helper with a lint rule, delete the canonical-to-known table. Your e2e tests stay green throughout.

Happy to pair on the follow-up — there's overlap with the #724 / `KNOWN_TASKS` cleanup I'm queuing up separately.

---

(Edit: removed accidental `#1`/`#2`/`#3` section markers that GitHub auto-linked to unrelated issues.)

DingmaomaoBJTU (commented):
Solid fix for the task inconsistency between Optimum canonical names and HF-pipeline names. The approach of static disambiguation via OnnxConfig inputs is clean and avoids loading model weights. Good test coverage across unit and e2e. Two minor suggestions left inline.

DingmaomaoBJTU (commented):
Solid fix addressing a real gap in the task translation pipeline. The bidirectional approach (HF->Optimum forward + Optimum->HF reverse) is well thought out. Good test coverage spanning unit and e2e. A few comments below.

DingmaomaoBJTU (commented):
Solid fix for the Optimum canonical-vs-HF-pipeline task name inconsistency. The approach of using static OnnxConfig inputs for disambiguation is clever and avoids loading model weights. Thorough test coverage. A few observations inline.

src/winml/modelkit/datasets/__init__.py
```diff
@@ -188,6 +188,12 @@ def __init__(
             # Pass model_path for RandomDataset fallback on unsupported tasks
             dataset_kwargs["model_path"] = model_path
 
+        # Convert to known task before looking up the dataset registry.
+        from ..loader.task import canonical_task_to_known_task
+
+        task = canonical_task_to_known_task(task, model_name)
```
DingmaomaoBJTU: Consider adding a debug log here (e.g. logger.debug('Mapped canonical task %r -> %r', original_task, task)) so users with -vv can see the task remapping in action during calibration. This would help debugging dataset selection issues.

src/winml/modelkit/eval/evaluate.py
```diff
@@ -209,22 +209,28 @@ def _load_model(config: WinMLEvaluationConfig) -> WinMLPreTrainedModel:
...
+
     if config.task is not None:
         task = config.task
     else:
         if config.model_id is None:
             raise ValueError("Cannot infer task without model_id. Provide --task.")
 
-        console = Console()
-        console.print("[bold]Detecting model task...[/bold]")
-
```
DingmaomaoBJTU: Good catch moving console.print outside the if/else so the detected task is always displayed. However, now when config.task is not None (user explicitly passed --task), the message says 'Detected task' which is slightly misleading since it was not auto-detected. Consider using 'Resolved task' or 'Using task' as the label to cover both cases.

DingmaomaoBJTU (commented):
Good PR overall. The two-direction fix (HF->Optimum forward via _map_task_synonym and Optimum->HF reverse via canonical_task_to_known_task) is well-structured and the test coverage is thorough. A few inline suggestions below.

src/winml/modelkit/inspect/resolver.py
```diff
@@ -337,12 +337,15 @@ def resolve_exporter(
         import optimum.exporters.onnx.model_configs  # noqa: F401
         from optimum.exporters.tasks import TasksManager
 
+        # TasksManager expects normalized task names
```
DingmaomaoBJTU: `_map_task_synonym` is a private function (leading underscore) from `..export.io` being imported across module boundaries in 4+ places. Consider either making it a public utility (e.g., export it from `winml.modelkit.export` or move to `utils/`) or re-exporting from `loader/task.py` where the related `canonical_task_to_known_task` lives.

DingmaomaoBJTU (commented):
Nice fix for the Optimum canonical-vs-HF-pipeline task name mismatch. Thorough test coverage. A few observations:

DingmaomaoBJTU (commented):
Solid fix for task inconsistency between Optimum canonical names and HF pipeline names. The approach of disambiguating via OnnxConfig inputs is clever and avoids maintaining a separate per-model mapping. Some performance and maintainability notes inline.

src/winml/modelkit/eval/evaluate.py
```diff
@@ -207,24 +207,62 @@ def _load_model(config: WinMLEvaluationConfig) -> WinMLPreTrainedModel:
...
+    """Convert task name to an HF-pipeline-recognizable name."""
+    mapping = HF_TASK_NAME_MAPPING.get(task)
+    if mapping is None or model_id is None:
+        return task
+
+    try:
+        from transformers import AutoConfig
+
+        from ..export.io import _get_onnx_config
+
```
DingmaomaoBJTU, line 232: `to_hf_pipeline_task` calls `AutoConfig.from_pretrained` which hits the network. If `_resolve_task` already loaded the config (the `else` branch above), this duplicates the hub call. Consider threading the already-loaded `hf_config` through to avoid the extra network roundtrip when `config.task` was not provided.

src/winml/modelkit/datasets/__init__.py
```diff
@@ -51,6 +51,23 @@
 }
```
DingmaomaoBJTU, line 52: The `_resolve_dataset_class` function logs a warning and falls back to `RandomDataset` when the io_config doesn't match any key in the dispatch dict OR matches multiple. The multiple-match case (`len(hits) != 1`) conflates 'no match' with 'ambiguous match' into the same fallback. Consider distinguishing them in the log message (e.g. 'no known input found' vs 'ambiguous inputs') for debuggability.

src/winml/modelkit/inspect/resolver.py
```diff
@@ -337,12 +337,15 @@ def resolve_exporter(
...
+        from ..export.io import _map_task_synonym
```
DingmaomaoBJTU: nit: importing a private function (`_map_task_synonym`) from `export.io` in multiple places (here, `commands/inspect.py`, `export/htp/exporter.py`) suggests it should be promoted to a public utility — perhaps in `loader/task.py` alongside `normalize_task`. This would also remove the cross-layer dependency (inspect -> export).

DingmaomaoBJTU (commented):
Good fix for the task-naming inconsistency between HF pipeline aliases and Optimum's canonical names. The approach of normalizing via `_map_task_synonym` at each TasksManager call site is correct. The `to_hf_pipeline_task` evaluator mapper and the bimodal `feature-extraction` dataset dispatch are both well-thought-out. A few observations:

DingmaomaoBJTU (commented):
Solid work handling the bimodal `feature-extraction` task and normalizing task synonyms before TasksManager lookups. Inline observations:

1. **Bimodal dispatch** (`datasets/__init__.py`): The `_resolve_dataset_class` approach of checking ONNX input names (`pixel_values` vs `input_ids`) to pick the dataset class is pragmatic. However, the `hits` logic silently falls back to `RandomDataset` when the model has **both** `pixel_values` and `input_ids` (multimodal models like CLIP). Consider logging at INFO when `len(hits) > 1` with the specific hit names, so users know why random calibration was chosen.

2. **`to_hf_pipeline_task` robustness** (`eval/evaluate.py`): The function downloads a full `AutoConfig` just to probe OnnxConfig inputs. If the model is local-only or the network is down, this silently returns the original task (good). But the `_get_onnx_config` call can also raise for unsupported model types - the broad `except Exception` is correct here, but consider logging at WARNING (not just DEBUG) since an eval mismatch can produce confusing silent failures downstream.

3. **Task synonym normalization** (`inspect/resolver.py`, `export/htp/exporter.py`, `commands/inspect.py`): Importing `_map_task_synonym` from `..export.io` in multiple places is a cross-layer import (inspect -> export). Consider promoting `_map_task_synonym` to the loader layer (`loader/task.py`) alongside the other task utilities, then having both export and inspect import from there.

4. **E2E test parametrize** (`test_eval_e2e.py`): Good - testing both `image-feature-extraction` and `feature-extraction` paths ensures the mapping doesn't regress.

5. **Potential overlap with PR #790**: This PR and #790 (timm library routing) both touch `inspect/resolver.py` and `loader/task.py`. Coordinate merge order to avoid conflicts - #790 adds `resolve_optimum_library` calls in the same `resolve_exporter` function.

src/winml/modelkit/eval/evaluate.py
zhenchaoni: HF already have cache mechanism. The second call will be cheap. We use this pattern across the project, no need to fix.

src/winml/modelkit/datasets/__init__.py
zhenchaoni: The message is very general, saying the model + task isn't supported by our quantization dataset. So no need to be granular for exact reason.

src/winml/modelkit/inspect/resolver.py
zhenchaoni: This makes sense. `_map_task_synonym` is already (even before this PR) across multiple modules. I have made it as public.

DingmaomaoBJTU (approved):

DingmaomaoBJTU (approved):

PR #788, opened 2026-05-29
feat: show tqdm for _ensure_provider_ready

### LINKED ISSUE #725: show progress bar for EP downloading

xieofxie:
Closes https://github.com/microsoft/winml-cli/issues/725

winml sys <

image

Failed

image


DingmaomaoBJTU (commented):
Good addition! The tqdm progress bar for EP downloads is a nice UX improvement. A few inline suggestions below.

DingmaomaoBJTU (commented):
Good feature addition - the tqdm progress bar and timeout-with-cancel pattern are clean. A few items to address below.

DingmaomaoBJTU (commented):
Clean implementation with excellent test coverage (6 tests covering all paths: fast-path, progress, timeout, error surfacing). The threading.Event + tqdm pattern is appropriate. One minor observation inline.

DingmaomaoBJTU (commented):
Nice UX improvement - users will appreciate seeing download progress instead of a silent hang. The threading.Event + timeout pattern is solid. A few inline notes.

DingmaomaoBJTU (commented):
Good implementation! The tqdm progress bar for EP downloads is a nice UX improvement. The timeout/cancel logic is solid and the test coverage is thorough. A few suggestions below.

DingmaomaoBJTU (commented):
## Code Review: PR #788 - feat: show tqdm for _ensure_provider_ready

**Overall**: Clean implementation with good test coverage. The async-with-progress pattern is well-structured, and the timeout/cancel logic is solid. A few suggestions:

### Thread safety of tqdm bar updates
The `_on_progress` callback is likely invoked from a different thread (the async provider's download thread). While tqdm is generally thread-safe for `update()`, the read-modify pattern `current - bar.n` followed by `bar.update(delta)` is not atomic. If two rapid `on_progress` calls race, you could see a negative delta (though your `if delta > 0` guard prevents that from becoming a bug). Just noting this is safe due to the guard -- nice.

### `logger.warning` vs `logger.info` for download start
Using `logger.warning` for "Downloading execution provider" means it shows even in default (WARNING) mode. This seems intentional (user should know why it's slow), but it could be noisy if many EPs need downloads. Consider `logger.info` + the tqdm bar as the user signal, so `-v` is needed for the text log. Alternatively, keep WARNING but only for the first download in a session.

### Log level bump from DEBUG to INFO for failures
Changing `logger.debug` to `logger.info` for "Failed to ensure EP is ready" is a good call -- these are actionable user-facing messages. However, the `continue` means an EP silently disappears from the catalog. Consider adding the exception type to the message (e.g., `type(e).__name__`) so users know if it's a network issue vs. a missing dependency.

### Test: monkeypatching module-level constant
`monkeypatch.setattr(ep_registry, 'EP_DOWNLOAD_TIMEOUT_SECONDS', 0.05)` -- nice pattern for testing timeouts without slowing CI. The 50ms timeout is tight enough that a slow CI machine could race (unlikely but possible if the thread scheduler is slow to wake). 0.1s might be safer with negligible test slowdown.

### Missing: what happens if tqdm isn't installed?
`from tqdm import tqdm` is inside the non-Ready branch, so it'll raise ImportError if tqdm isn't available. Is tqdm a required dependency? If optional, consider a try/except with a no-op fallback bar.

DingmaomaoBJTU (commented):
Good UX improvement. Showing a tqdm progress bar for EP downloads is much better than a silent hang. The timeout/cancel logic and test coverage are solid.

DingmaomaoBJTU (commented):
Good UX improvement. The tqdm progress bar for EP downloads is well-implemented with proper timeout/cancellation handling. Tests cover all the edge cases nicely.

DingmaomaoBJTU (commented):
Clean implementation of tqdm progress for EP downloads. Good use of threading.Event for timeout + cancel pattern. A couple of thread-safety and UX notes inline.

DingmaomaoBJTU (commented):
Good feature addition — the tqdm progress bar for EP downloads significantly improves UX. The implementation is clean overall with proper timeout handling and cleanup. I have a few inline suggestions below regarding thread-safety and edge-case handling.

src/winml/modelkit/session/ep_registry.py
```diff
@@ -20,6 +20,68 @@
...
+        desc=f"Downloading {provider.name}",
+        unit="%",
+        leave=True,
+    )
+    done = threading.Event()
+
+    def _on_progress(fraction: float) -> None:
+        current = max(0, min(100, int(fraction * 100)))
+        delta = current - bar.n
+        if delta > 0:
```
xieofxie: Short answer: no, I don't think we need it. Three reasons:

  1. windowsml already swallows exceptions in on_progress. Looking at the binding:
  @AsyncProgressCallback
  def _on_progress(async_ptr, value):
      try:
          _user_progress(value)
      except Exception:
          pass
  1. Even if our callback raised on a closed bar, it would be caught inside windowsml and never reach native code.
  2. tqdm's update() after close() is a defined no-op, not undefined behavior. close() sets self.disable = True, and update()
  early-returns when disable is true. Worst case: the late call increments self.n on an already-closed bar instance and produces no
  output. Nothing crashes, nothing renders.
  3. op.close() in finally blocks until native completion anyway. From the windowsml AsyncOperation.close():
  WinMLAsyncGetStatus(ctypes.byref(self._block), True)  # waits for native completion
  3. So by the time _ensure_provider_ready returns to its caller, the native thread is guaranteed to be done — no callbacks can fire after
   that. The only "race window" is between bar.close() (line 81) and op.close() (line 82) finishing, and inside that window the worst that
   happens is the harmless no-op above.

  The proposed flag would also need to be wrapped in a list / nonlocal (Python closures), and it would protect only against a benign
  condition. I'd leave the code as-is.

  If we ever switched bars (e.g., to Rich Progress, which is less tolerant of post-close updates), I'd revisit — but for tqdm
  specifically, this isn't worth the noise.

DingmaomaoBJTU (commented):
Good PR overall - clean separation between sync fast-path and async download flow, solid timeout/cancellation handling, and thorough test coverage. A few suggestions below.

src/winml/modelkit/session/ep_registry.py
```diff
@@ -20,6 +20,68 @@
...
+    logger.warning(
+        "Downloading execution provider %r. This may take several minutes "
+        "depending on network speed (timeout: %ds).",
+        provider.name,
+        EP_DOWNLOAD_TIMEOUT_SECONDS,
+    )
+
+    bar = tqdm(
+        total=100,
+        desc=f"Downloading {provider.name}",
```
xieofxie: no need. update bar.n directly

src/winml/modelkit/session/ep_registry.py
```diff
@@ -20,6 +20,68 @@
 
 logger = logging.getLogger(__name__)
 
+
```
xieofxie, line 24: add WINMLCLI_EP_DOWNLOAD_TIMEOUT

src/winml/modelkit/session/ep_registry.py
```diff
@@ -20,6 +20,68 @@
...
+EP_DOWNLOAD_TIMEOUT_SECONDS = 5 * 60
+
+
+def _ensure_provider_ready(provider: Any) -> None:
+    """Ensure an EP is ready, showing a tqdm progress bar when downloading.
```
xieofxie, line 130: make optional

src/winml/modelkit/session/ep_registry.py
```diff
@@ -20,6 +21,108 @@
...
+        provider.name,
+        EP_DOWNLOAD_TIMEOUT_SECONDS,
+    )
+
+    bar = _make_progress_bar(provider.name)
+    done = threading.Event()
+
+    def _on_progress(fraction: float) -> None:
+        bar.n = max(0, min(100, int(fraction * 100)))
+        bar.refresh()
```
DingmaomaoBJTU, line 165: `done.wait(timeout=...)` blocks the calling thread. If `_load_ep_catalog` is called on the main thread (e.g. during CLI startup), and multiple providers need downloading, they're serialized. Is that intentional? For a first-run scenario with N providers needing download, this could mean N * timeout worst-case. Consider noting this in the docstring or adding a max-total-timeout guard.

src/winml/modelkit/session/ep_registry.py
```diff
@@ -20,6 +21,108 @@
...
+        return int(raw)
+    except ValueError:
+        logger.warning("Invalid WINMLCLI_EP_DOWNLOAD_TIMEOUT=%r; using default 300s.", raw)
+        return 5 * 60
+
+
+EP_DOWNLOAD_TIMEOUT_SECONDS = _ep_download_timeout_default()
+
+
+class _NoopBar:
```
DingmaomaoBJTU, line 48: `EP_DOWNLOAD_TIMEOUT_SECONDS` is computed at module import time. If a test monkeypatches the env var after import (which the tests do via `monkeypatch.setattr`), the module-level constant won't pick it up. The tests work because they patch the attribute directly, but documenting this (or making it a function call at point-of-use) would avoid future confusion.

DingmaomaoBJTU (commented):
Solid implementation. The async-with-timeout pattern is well-structured, the _NoopBar fallback keeps tqdm optional, and test coverage is thorough (timeout, get_status error, no-tqdm). One concern and a nit:

src/winml/modelkit/session/ep_registry.py
```diff
@@ -20,6 +21,108 @@
...
+        bar.n = max(0, min(100, int(fraction * 100)))
+        bar.refresh()
+
+    op = provider.ensure_ready_async(on_complete=done.set, on_progress=_on_progress)
+    try:
+        if not done.wait(timeout=EP_DOWNLOAD_TIMEOUT_SECONDS):
+            op.cancel()
+            raise TimeoutError(
+                f"EP {provider.name!r} download did not complete within "
+                f"{EP_DOWNLOAD_TIMEOUT_SECONDS}s; cancelled."
```
DingmaomaoBJTU, line 175: Minor: if `ensure_ready_async()` itself raises (e.g. native layer throws), `bar.close()` is never called because the assignment happens before the `try` block. Consider wrapping the full sequence in try/finally or assigning `op = None` as a sentinel:```python
op = provider.ensure_ready_async(on_complete=done.set, on_progress=_on_progress)
try:
    ...
finally:
    bar.close()
    if op is not None:
        op.close()
```Not critical (the bar is cosmetic and _load_ep_catalog catches exceptions upstream), but worth a defensive tweak.

File: src/winml/modelkit/session/ep_registry.py [hunk 8]
```diff
@@ -20,6 +21,108 @@
...
+        "Downloading execution provider %r. This may take several minutes "
+        "depending on network speed (timeout: %ds).",
+        provider.name,
+        EP_DOWNLOAD_TIMEOUT_SECONDS,
+    )
+
+    bar = _make_progress_bar(provider.name)
+    done = threading.Event()
+
+    def _on_progress(fraction: float) -> None:
```
DingmaomaoBJTU, line 157: `_on_progress` is called from whichever thread the native async op uses. tqdm's `bar.n` assignment + `bar.refresh()` is generally safe from a background thread, but worth a brief comment noting this is intentional, in case a future maintainer questions it.

DingmaomaoBJTU (commented):
Nice implementation. The async flow with threading.Event is clean, the timeout is configurable via env var, and tests are thorough. Two minor observations below.

src/winml/modelkit/session/ep_registry.py
xieofxie: no need

src/winml/modelkit/session/ep_registry.py
xieofxie: fixed!

src/winml/modelkit/session/ep_registry.py
```diff
@@ -20,6 +21,137 @@
...
+        if not success:
+            # Failure-path notice — kept in finally so it fires for every
+            # non-success exit (launch failure, timeout, get_status OSError).
+            # Printed after bar.close() so it appears below the bar's last frame.
+            console.print(f"[red]❌ Failed to download {provider.name} EP[/red]")
+            console.print("Try:")
+            console.print("  1. Check your internet connection")
+            console.print(
+                "  2. Troubleshoot: "
+                "https://learn.microsoft.com/en-us/windows/ai/new-windows-ml/execution-provider-errors",
```
xieofxie: should we use aka.ms ?

xieofxie: use https://aka.ms/winmlcli/ep-errors

src/winml/modelkit/session/ep_registry.py
```diff
@@ -20,6 +21,187 @@
...
+
+    if provider.ready_state == EpReadyState.Ready:
+        provider.ensure_ready()
+        return
+
+    # Lazy-import to keep ep_registry import cheap (rich pulls in pygments etc.);
+    # this branch only runs on the cold "EP needs download" path.
+    from ..utils.console import get_console
+
+    console = get_console()
```
xieofxie, line 151: Add a new issue to suppress this https://github.com/microsoft/winml-cli/issues/811

src/winml/modelkit/session/ep_registry.py
```diff
@@ -20,6 +21,187 @@
...
+
+    op = None
+    success = False
+    try:
+        op = provider.ensure_ready_async(on_complete=done.set, on_progress=_on_progress)
+        if not done.wait(timeout=EP_DOWNLOAD_TIMEOUT_SECONDS):
+            op.cancel()
+            raise TimeoutError(
+                f"EP {provider.name!r} download did not complete within "
+                f"{EP_DOWNLOAD_TIMEOUT_SECONDS}s; cancelled."
```
DingmaomaoBJTU, line 175: Potential late-callback race: `done.wait()` returns as soon as the native op calls `on_complete`, but a native implementation could fire a stale `on_progress` callback from its own thread after `on_complete`. At that point the main thread has already set `bar.n = 100` and is heading into the `finally` block. The late callback would write a smaller value (e.g. 62) back to `bar.n`, causing the final refresh to show a non-100% bar, or worse, writing to a bar that has already been `close()`'d. Consider either: (a) setting a threading flag that `_on_progress` checks before mutating `bar.n`, or (b) protecting `bar.n` assignment with a lock that the success path also acquires before forcing `bar.n = 100`.

src/winml/modelkit/session/ep_registry.py
```diff
@@ -20,6 +21,187 @@
...
+    ``package_family_name`` even after the EP is Ready. When the EP is delivered
+    as an MSIX package its ``library_path`` lives under ``WindowsApps`` in a
+    folder named with the full package identity::
+
+        ...\\WindowsApps\\<Name>_<Version>_<Arch>_<ResourceId>_<PublisherId>\\...
+
+    e.g. ``MicrosoftCorporationII.WinML.Intel.OpenVINO.EP.1.8_1.8.79.0_x64__8wekyb3d8bbwe``
+    yields version ``1.8.79.0`` and package family name
+    ``MicrosoftCorporationII.WinML.Intel.OpenVINO.EP.1.8_8wekyb3d8bbwe`` (the
+    package family name is ``<Name>_<PublisherId>``).
```
DingmaomaoBJTU, line 99: `itertools.pairwise` was added in Python 3.10. If the project's `python_requires` allows 3.9, this will raise `ImportError` at runtime on older interpreters. Either bump the minimum Python version in pyproject.toml or replace with a manual sliding-window: `zip(parts, parts[1:])`.

src/winml/modelkit/session/ep_registry.py
```diff
@@ -20,6 +21,187 @@
...
+
+    Lets users on slow networks raise the cap without code changes. Falls back
+    to the default when the env var is unset, empty, or non-integer.
+    """
+    raw = os.environ.get("WINMLCLI_EP_DOWNLOAD_TIMEOUT")
+    if not raw:
+        return 5 * 60
+    try:
+        return int(raw)
+    except ValueError:
```
DingmaomaoBJTU, line 36: `EP_DOWNLOAD_TIMEOUT_SECONDS` is evaluated once at module import time. If a test (or a caller) sets `WINMLCLI_EP_DOWNLOAD_TIMEOUT` after the module is already imported, the constant won't reflect the change. The timeout tests work around this correctly by patching `ep_registry.EP_DOWNLOAD_TIMEOUT_SECONDS` directly — which is the right approach. Just worth documenting on the constant itself (e.g., `# Evaluated once at import; patch ep_registry.EP_DOWNLOAD_TIMEOUT_SECONDS in tests`) so future callers don't expect `os.environ` changes to take effect mid-run.

src/winml/modelkit/session/ep_registry.py
```diff
@@ -80,9 +262,19 @@ def _load_ep_catalog(self) -> None:
...
+                    logger.info(
+                        "Failed to ensure EP %s is ready: %s (winerror=%s)",
+                        provider.name,
+                        e,
+                        getattr(e, "winerror", None),
+                    )
+                    continue
                 except Exception as e:
-                    logger.debug("Failed to ensure EP %s is ready: %s", provider.name, e)
+                    logger.info("Failed to ensure EP %s is ready: %s", provider.name, e)
```
DingmaomaoBJTU, line 287: The OSError case above is elevated to `logger.info` with `winerror` included — that makes sense for surfacing HRESULT codes. But the generic `except Exception` branch here is also elevated from `debug` to `info`. In a scan that iterates all providers, a non-fatal glitch on one EP (e.g., an EP that's installed but momentarily unavailable) will now print an INFO line for every provider that fails transiently. Consider keeping the generic branch at `debug` and only elevating `OSError` (where winerror is meaningful).

src/winml/modelkit/session/ep_registry.py
xieofxie: we needs 3.11

src/winml/modelkit/session/ep_registry.py
xieofxie: this makes -v shows it, it will not show by default

DingmaomaoBJTU (approved):

src/winml/modelkit/session/ep_registry.py
xieofxie: fixed

src/winml/modelkit/session/ep_registry.py
xieofxie: added

PR #789, opened 2026-05-29
chore: typechecking build, commands, compiler folder

xieofxie:
Table for stub

```
  ┌──────────────┬──────────┬────────────────────────────────┬────────────────────────────────────────────┐
  │     Lib      │ py.typed │            Reality             │              Override status               │
  ├──────────────┼──────────┼────────────────────────────────┼────────────────────────────────────────────┤
  │ torch        │ yes      │ Has inline types (v2.11)       │ Override is a no-op — mypy uses real types │
  ├──────────────┼──────────┼────────────────────────────────┼────────────────────────────────────────────┤
  │ torchvision  │ no       │ No types, no community stubs   │ Genuinely needed                           │
  ├──────────────┼──────────┼────────────────────────────────┼────────────────────────────────────────────┤
  │ onnx         │ yes      │ Has inline types (v1.18)       │ Override is a no-op                        │
  ├──────────────┼──────────┼────────────────────────────────┼────────────────────────────────────────────┤
  │ onnxruntime  │ no       │ Untyped; no community stubs    │ Genuinely needed                           │
  ├──────────────┼──────────┼────────────────────────────────┼────────────────────────────────────────────┤
  │ transformers │ yes      │ Inline types but partial/loose │ Override is a no-op — types ARE used       │
  ├──────────────┼──────────┼────────────────────────────────┼────────────────────────────────────────────┤
  │ datasets     │ no       │ Untyped                        │ Genuinely needed                           │
  ├──────────────┼──────────┼────────────────────────────────┼────────────────────────────────────────────┤
  │ optimum      │ no       │ Untyped                        │ Genuinely needed                           │
  ├──────────────┼──────────┼────────────────────────────────┼────────────────────────────────────────────┤
  │ timm         │ yes      │ Has inline types (v1.0.26)     │ Override is a no-op                        │
  ├──────────────┼──────────┼────────────────────────────────┼────────────────────────────────────────────┤
  │ onnxscript   │ yes      │ Has inline types (v0.7)        │ Override is a no-op                        │
  ├──────────────┼──────────┼────────────────────────────────┼────────────────────────────────────────────┤
  │ snakemd      │ no       │ Untyped                        │ Genuinely needed                           │
  ├──────────────┼──────────┼────────────────────────────────┼────────────────────────────────────────────┤
  │ openvino     │ n/a      │ Not installed locally          │ n/a                                        │
  └──────────────┴──────────┴────────────────────────────────┴────────────────────────────────────────────┘
```

plotext added to ignore_missing_imports (no community stubs, untyped library)


DingmaomaoBJTU (commented):
Overall this is a solid typing hygiene pass - adding return annotations to `__getattr__`, narrowing `object` to concrete types, and fixing closure-scoped variables. A few comments below on spots that could use attention.

DingmaomaoBJTU (commented):
**Type-checking improvements look good overall** — removing the no-op override stubs, tightening `object` -> concrete types in eval/inspect, and the `EPName` narrowing in analyze are all welcome.

However, there's a **closure scoping bug** in `commands/build.py` that will cause `_on_node_result` to always see `_current_ep = None` (details in inline comment). Please fix before merging.

DingmaomaoBJTU (commented):
Good incremental tightening. The stub cleanup (removing no-op overrides for typed libraries) and the `object` -> concrete type upgrades are solid improvements. The deprecated import fix (`huggingface_hub.utils` -> `huggingface_hub.errors`) is a bonus. Two observations inline.

DingmaomaoBJTU (commented):
Good progress on typechecking. The casts are well-annotated with comments explaining why they're needed (e.g. Click's Choice validation guaranteeing the literal types).

**Question**: The `build/__init__.py` diff removes the public API (BuildResult, build_hf_model, build_onnx_model, __all__) and replaces the module docstring. Any downstream code doing `from winml.modelkit.build import build_hf_model` would break. Is this intentional (moving to lazy imports only) or a rebase artifact?

**Minor**: In `commands/quantize.py`, the cast pattern (`cast('Literal[...]', method)`) is clean. Since these come from Click's Choice validator the cast is safe.

**Suggestion**: In `optracing/base.py`, the broad `dict[EPName, Any]` for `_ep_monitors` loses the type safety you're adding elsewhere. Consider a Protocol or TypeVar bound to the monitor interface if one exists.

src/winml/modelkit/commands/build.py
```diff
@@ -937,13 +946,13 @@ def _show_io(sl: Any, config: WinMLBuildConfig) -> None:
...
+    for i, in_spec in enumerate(inputs):
+        name = in_spec.name or "(unnamed)"
+        shape = str(list(in_spec.shape)) if in_spec.shape else "dynamic"
+        dtype = getattr(in_spec, "dtype", None) or "?"
         sl.io_input(name, shape, dtype, first=(i == 0))
-    for i, t in enumerate(outputs):
-        name = t.name or "(unnamed)"
+    for i, out_spec in enumerate(outputs):
+        name = out_spec.name or "(unnamed)"
         # OutputTensorSpec has name only — show name, no shape/dtype
```
DingmaomaoBJTU, line 956: **Bug (closure scoping)**: `_current_ep = ep_name` creates a *local* variable inside `_on_ep_start` because there's no `nonlocal _current_ep` declaration. The original code used `_current_ep[0] = ep_name` (list-mutation trick) specifically to work around this Python closure limitation.

Since `_on_node_result` reads `_current_ep` from the enclosing scope, it will always see `None` (the initial value) and return early, so the progress bars will never update.

Fix: add `nonlocal _current_ep` at the top of `_on_ep_start`, or revert to the list-mutation pattern.

src/winml/modelkit/commands/eval.py
```diff
@@ -244,24 +245,12 @@ def _build_eval_config(
     from ..eval import DatasetConfig, WinMLEvaluationConfig
     from ..utils.config_utils import merge_config
 
-    # Initialize config object from CLI ctx params.
-    p = ctx.params
-    cfg = WinMLEvaluationConfig(
-        task=p.get("task"),
-        device=p.get("device"),
-        precision=p.get("precision"),
```
DingmaomaoBJTU, line 252: Nice refactor: `collect_cli_overrides` + dataclass instantiation is much cleaner than the manual `p.get()` soup. This pattern should become the standard for other commands too.

src/winml/modelkit/commands/perf.py
```diff
@@ -373,6 +375,8 @@ def _run_benchmark(self) -> PerfStats:
 
     def _run_benchmark_simple(self) -> PerfStats:
         """Execute benchmark without live monitoring."""
+        assert self._model is not None
+        assert self._inputs is not None
         session = self._model._session
         total_iterations = self.config.warmup + self.config.iterations
```
DingmaomaoBJTU, line 382: The old code used `_ep_monitors = {"vitisai": VitisAIMonitor}` keyed by EP *alias*, but the new code keys by the full provider name `"VitisAIExecutionProvider"` and looks up via `session.ep_name`. Confirm `session.ep_name` actually returns the full provider string (not the alias) at this point — otherwise this will silently skip the VitisAI monitor.

src/winml/modelkit/commands/optimize.py
```diff
@@ -35,6 +35,8 @@
 if TYPE_CHECKING:
     from collections.abc import Callable
 
+F = TypeVar("F", bound="Callable[..., Any]")
+
```
DingmaomaoBJTU, line 39: Adding positional-only `/` here is a behavioral change — any existing test or plugin that calls `optimize.callback(ctx=..., list_capabilities=...)` by keyword will now break. Is this intentional? If it's only to satisfy mypy about `ctx` not being passed by keyword, a less risky approach is `# type: ignore[misc]` on just that line or an `@overload`.

DingmaomaoBJTU (changes requested):
Good type-tightening overall. The removal of overly-broad `ignore_missing_imports` for torch/onnx/transformers is a real win, and the `TypeVar`-based Click decorator signatures are the correct pattern. However, there's a closure scoping bug introduced in `build.py` that will break the analyze progress display at runtime.

src/winml/modelkit/compiler/__init__.py
```diff
@@ -1,81 +1,83 @@
...
+            QFormatConvertStage=QFormatConvertStage,
+        )
+        return globals()[name]
+
+    raise AttributeError(f"module {__name__!r} has no attribute {name!r}")
+
+
+__all__ = [
+    "QDQ_OP_TYPES",
+    "CompileContext",
```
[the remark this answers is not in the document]
xieofxie: fixed

src/winml/modelkit/commands/quantize.py
```diff
@@ -29,6 +30,10 @@
 from ..utils.logging import configure_logging
 
 
+if TYPE_CHECKING:
+    from typing import Literal
```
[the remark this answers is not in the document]
xieofxie: used in cast

src/winml/modelkit/analyze/analyzer.py
```diff
@@ -838,14 +838,6 @@ def has_errors(self) -> bool:
         """True if blocking errors (unsupported patterns) exist."""
         return self.lint.errors > 0
 
-    @property
-    def autoconf(self) -> WinMLOptimizationConfig | None:
```
xieofxie, line 842: remove as it is unused and just a plain attribute access

src/winml/modelkit/commands/config.py
```diff
@@ -326,35 +328,43 @@ def config(
...
-                ep=ep,
-            )
-
-            # Handle output format
+            # Generate config(s) - module parameter selects overload:
+            # module=str → list[WinMLBuildConfig], module=None → WinMLBuildConfig.
             if module:
-                # Module mode: result is list[WinMLBuildConfig]
-                configs = result
+                configs = generate_hf_build_config(
```
DingmaomaoBJTU: The two overloaded branches now repeat all 10 keyword arguments. If a new parameter is added to generate_hf_build_config (e.g. quantization_config), it must be added in both places, which is easy to miss.```python
# Alternative: build shared kwargs once
_shared_kwargs = dict(
    model_id=hf_model,
    task=task,
    ...
)
configs = generate_hf_build_config(module=module, **_shared_kwargs)
# vs
config_obj = generate_hf_build_config(**_shared_kwargs)
```
This preserves the overload dispatch while keeping the args DRY.

File: src/winml/modelkit/commands/perf.py [hunk 9]
```diff
@@ -416,9 +422,9 @@ def _run_benchmark_monitored(self) -> PerfStats:
 
         # EP-specific proof-of-execution monitor.
         # When QNN/OpenVINO monitors become real, add entries here.
-        _ep_monitors = {"vitisai": VitisAIMonitor}
-        ep = self.config.ep
-        monitor_cls = _ep_monitors.get(ep)
+        _ep_monitors: dict[EPName, Any] = {"VitisAIExecutionProvider": VitisAIMonitor}
```
DingmaomaoBJTU, line 425: Behavior change: the old key was 'vitisai' (an alias), so _ep_monitors.get(ep) matched when the user passed --ep vitisai. The new key is canonical 'VitisAIExecutionProvider' and the lookup uses session.ep_name. This is correct, but worth a short comment: # keyed by canonical EPName, not by alias.

src/winml/modelkit/commands/inspect.py
```diff
@@ -332,7 +332,7 @@ def _inspect_model_v2(
     # =========================================================================
     # STEP 2: Shared loader resolution (same call as config command)
     # =========================================================================
-    from huggingface_hub.utils import RepositoryNotFoundError
+    from huggingface_hub.errors import RepositoryNotFoundError
```
DingmaomaoBJTU, line 335: Import moved from huggingface_hub.utils to huggingface_hub.errors. This is the correct modern location (errors module introduced in huggingface_hub 0.20). Worth verifying pyproject.toml minimum version covers this, otherwise older envs get an ImportError at runtime.

src/winml/modelkit/commands/eval.py
```diff
@@ -244,24 +245,12 @@ def _build_eval_config(
...
-            samples=p.get("samples"),
-            shuffle=p.get("shuffle"),
-            streaming=p.get("streaming"),
-            build_script=p.get("dataset_script"),
-        ),
-    )
+    # Initialize config object from CLI ctx params. ``collect_cli_overrides``
+    # filters to user-provided values and applies the cli_name → field_name
+    # renames declared on the dataclass fields (e.g. output → output_path).
+    eval_kwargs = cli_utils.collect_cli_overrides(ctx, WinMLEvaluationConfig)
```
DingmaomaoBJTU, line 260: collect_cli_overrides is cleaner, but it must silently drop Click params that are NOT WinMLEvaluationConfig fields (config_file, trust_remote_code, column, label_mapping are Click-only). Worth a quick sanity check that extra keys don't leak into the dataclass constructor as unexpected kwargs.

src/winml/modelkit/build/common.py
```diff
@@ -214,10 +215,11 @@ def _run_analyze_loop(
         config.optim.update(discovered_optim)
         logger.info("  [autoconf] final config: %s", discovered_optim)
 
-    if analysis.autoconf:
+    final_optim_config = analysis.optimization_config if analysis else None
```
DingmaomaoBJTU, line 220: Nit: analysis can only be None when max_analyze_iterations==0 (loop body never ran). A brief comment like '# analysis is None only when max_analyze_iterations=0' would make the guard's intent clear.

src/winml/modelkit/optracing/base.py
```diff
@@ -24,12 +25,22 @@ class OpTracer(ABC):
     at construction time, then call ``run()`` to execute profiling.
     """
 
+    def __init__(self, onnx_path: Path, *, output_dir: Path, level: str = "basic") -> None:
```
DingmaomaoBJTU, line 31: Good consolidation. QNNProfiler is correctly updated to call super().__init__(). Worth noting in the class docstring that future subclasses must do the same — otherwise onnx_path/output_dir/level won't be initialized.

src/winml/modelkit/commands/perf.py
xieofxie: no need

src/winml/modelkit/commands/inspect.py
xieofxie: transformers>=4.57 (a direct dep) pulls in huggingface_hub>=0.34.0

src/winml/modelkit/commands/config.py
xieofxie: good point!

src/winml/modelkit/commands/eval.py
xieofxie: checked and change one name `label_mapping` -> `label_mapping_path`

DingmaomaoBJTU (approved):

PR #790, opened 2026-05-29
feat(timm): enable timm image-classification models via library routing

vortex-captain:
## Summary

timm checkpoints load through transformers'' generic `TimmWrapper` (`model_type="timm_wrapper"`) and previously failed in **every** `winml` command with *"Cannot detect task: config has no ''architectures'' field"*. Two gaps:

1. **Task/class detection** — timm repos load as `TimmWrapperConfig` with `architectures=None`, so auto-detection could not resolve a task or class.
2. **OnnxConfig location** — Optimum registers timm''s config (`TimmDefaultOnnxConfig`) only under `library_name="timm"`, but every `winml` lookup defaults to `transformers`.

`timm_wrapper` is transformers'' generic bridge for the whole timm library — not a model architecture — so it is resolved at the **shared resolution layer**, not as a per-model config. Only the library is recorded; the task is derived from Optimum.

## Changes (no `models/hf/` entry)

- **`loader/task.py`** — `WRAPPED_LIBRARY_MODEL_TYPES` (`model_type -> optimum_library`) + `resolve_optimum_library()`. When a config has no `architectures`, `_detect_task_and_class_from_config` derives the task from Optimum''s task list for the library (`get_supported_tasks("timm_wrapper", "timm")` -> `["image-classification"]`) and the class from `get_model_class_for_task` (generic `AutoModelForImageClassification`, which transformers dispatches to `TimmWrapper` at load). The task is not hardcoded; the branch imports `optimum.exporters.onnx.model_configs` first to populate Optimum''s registry (scoped so normal model loading never pays for it).
- **`export/io.py`** — `_get_onnx_config` routes the library via `resolve_optimum_library`, so `timm_wrapper` resolves Optimum''s `TimmDefaultOnnxConfig` from every call site (config/build/export/inspect) with no `--library` flag.
- **`commands/inspect.py`** + **`inspect/resolver.py`** — route both the CLI inspect path and the public `inspect_model` path the same way: library routing for the OnnxConfig lookup, plus wrapped-library task detection so the task is not mislabeled.
- Tests: `resolve_optimum_library` + wrapped-library architectures fallback with task derivation (loader); timm library routing for `resolve_io_specs` / `_get_onnx_config` (export); public inspect path `detect_task` / `resolve_exporter` for timm (inspect).

## Validation

**Functional (end-to-end)** on a timm image-classification model:

| Command | Before | After |
|---|---|---|
| `winml config` | exit 2 — *no ''architectures'' field* | task=image-classification, 1 input |
| `winml export` | exit 2 — same | `model.onnx` (pixel_values to logits) |
| `winml inspect` | exit 1 — same | `AutoModelForImageClassification` + `TimmDefaultOnnxConfig`, full I/O table |

`config` -> `export` -> `optimize` -> `model.onnx` validated end-to-end for multiple timm CNN classifiers. Also resolves on a timm ViT backbone (`num_labels=0`) -> task=image-classification, matching Optimum''s own `infer_task_from_model`, so it generalizes across timm architectures (CNN + ViT).

**No impact on existing models** — scanned all 439 entries / 401 unique models in `scripts/e2e_eval/testsets/models_all.json`: **0** are `timm_wrapper` (by JSON metadata and by loaded config; 330 loadable). Since `timm_wrapper` is the only trigger of the new branch, no existing model changes behavior. (71 fail to load a config — custom/GGUF/tabular types that fail at `AutoConfig` regardless; 7 have empty `architectures` but are not timm — a pre-existing "Cannot detect task", identical before and after the PR.)

**No overhead for normal (non-timm) models** — `winml config` on a standard non-timm model: this branch vs base, min ~12.6s vs ~12.5s (within run-to-run noise). Non-timm configs have `architectures`, so they skip the new branch; the only added cost is one dict lookup.

**Unit tests** — `tests/unit/loader` + `tests/unit/export` + `tests/unit/inspect`: green.


src/winml/modelkit/loader/task.py
```diff
@@ -314,8 +344,33 @@ def _detect_task_and_class_from_config(config: PretrainedConfig) -> tuple[str, t
...
+        library = WRAPPED_LIBRARY_MODEL_TYPES.get(getattr(config, "model_type", None) or "")
+        if library is not None:
+            # Populate Optimum's exporter registry (incl. the wrapped library's
+            # task list) before querying it; scoped to this rare branch so normal
+            # model loading never pays for the import.
+            import optimum.exporters.onnx.model_configs  # noqa: F401
+
+            supported = get_supported_tasks(config.model_type, library_name=library)
+            if supported:
+                task = supported[0]
```
vortex-captain, line 373: Fragile `[0]`: the export task is selected as the first element of the Optimum task list by registry/dict insertion order. For `timm_wrapper` this is unambiguous today (I verified `get_supported_tasks("timm_wrapper","timm") == ["image-classification"]` — a single element). But `WRAPPED_LIBRARY_MODEL_TYPES` is presented as a general extension point ("a few transformers model_types are thin wrappers"), and for any future wrapped library that exposes >1 task, `supported[0]` would silently pick an arbitrary/order-dependent task with no signal. Suggest either (a) assert `len(supported) == 1` for wrapper types and raise a clear error otherwise, or (b) document explicitly that wrapper entries are only valid when the wrapped library exposes exactly one task. As-is the generality of the abstraction over-promises relative to the `[0]` implementation.

src/winml/modelkit/loader/task.py
```diff
@@ -154,6 +154,36 @@
...
+
+
+def resolve_optimum_library(model_type: str | None, library_name: str = "transformers") -> str:
+    """Route a transformers model_type to the Optimum library that owns its export.
+
+    Most models export under the library they were requested with. A few
+    transformers model_types are thin wrappers whose Optimum OnnxConfig lives in
+    another library (see :data:`WRAPPED_LIBRARY_MODEL_TYPES`); route those so the
+    OnnxConfig lookup succeeds without an explicit ``--library`` flag. Only the
+    default ``"transformers"`` library is remapped, so an explicit choice always
```
vortex-captain: Docstring slightly overstates the guarantee. "Only the default `transformers` library is remapped, so an explicit choice always wins" is true for an explicit *non-transformers* `--library`, but an explicit `--library transformers` is indistinguishable from the default and **is** rerouted (the predicate is `library_name == "transformers"`). For timm this is actually the desired outcome, but the wording implies user intent is always preserved, which isn't the case for an explicit `transformers`. Recommend rewording to "any explicit non-`transformers` library wins".

src/winml/modelkit/loader/task.py
```diff
@@ -314,8 +344,33 @@ def _detect_task_and_class_from_config(config: PretrainedConfig) -> tuple[str, t
...
+    # Auto* class that transformers dispatches to the wrapper at load).
+    if not getattr(config, "architectures", None):
+        library = WRAPPED_LIBRARY_MODEL_TYPES.get(getattr(config, "model_type", None) or "")
+        if library is not None:
+            # Populate Optimum's exporter registry (incl. the wrapped library's
+            # task list) before querying it; scoped to this rare branch so normal
+            # model loading never pays for the import.
+            import optimum.exporters.onnx.model_configs  # noqa: F401
+
+            supported = get_supported_tasks(config.model_type, library_name=library)
```
vortex-captain: Minor: `getattr(config, "model_type", None) or ""` is used to compute `library` (line 358), but `config.model_type` is then dereferenced directly here and at lines 363/366. It's safe in practice (reaching this point requires `model_type` to be a real key, so the attribute exists), but mixing the defensive `getattr` with three bare attribute accesses reads inconsistently. Consider binding `model_type = config.model_type` once after the guard and reusing it.

src/winml/modelkit/loader/task.py
```diff
@@ -154,6 +154,36 @@
...
+# Only the library is recorded here -- it is the irreducible Optimum-taxonomy
+# fact. The export task is derived from Optimum's task list for that library
+# (get_supported_tasks), not hardcoded.
+# model_type -> optimum_library
+WRAPPED_LIBRARY_MODEL_TYPES: dict[str, str] = {
+    "timm_wrapper": "timm",
+}
+
+
+def resolve_optimum_library(model_type: str | None, library_name: str = "transformers") -> str:
```
vortex-captain, line 174: Public-API placement/export convention. `resolve_optimum_library` is a new **public** function consumed cross-package (`export/io.py`, `commands/inspect.py`), yet it is imported via the submodule path `from ..loader.task import resolve_optimum_library` and is not added to `loader/__init__.py` `__all__`, whereas its siblings (`get_supported_tasks`, `normalize_task`, `resolve_task_and_model_class`) are exported there. Per CLAUDE.md ("Every symbol that external code needs must be exported in the package's `__init__.py`"), consider adding `resolve_optimum_library` (and, if intended as public, `WRAPPED_LIBRARY_MODEL_TYPES`) to `loader/__init__.py`. Also: the module-level docstring "Public API:" list at the top of task.py was not updated to mention the new public function. (Separately, this is placed under the `# Model Class Resolution` section even though it's a library-routing concern — the file has a dedicated `# Public API` section lower down.)

src/winml/modelkit/loader/task.py
```diff
@@ -314,8 +344,33 @@ def _detect_task_and_class_from_config(config: PretrainedConfig) -> tuple[str, t
         return resolve_task_and_model_class(config, task=override_task)
 
     # [1] Resolve architecture class from config.
-    # If config.architectures is missing/empty, this raises ValueError and the
-    # caller should provide task explicitly.
+    # Some model_types (e.g. timm via "timm_wrapper") are generic library
```
vortex-captain, line 352: Edge case for the `--model-class` path. This wrapper fallback lives only in `_detect_task_and_class_from_config` (Case 1: `task is None and model_class is None`). If a user runs a timm model with `--model-class X` but no `--task`, `resolve_task_and_model_class` Case 3 calls `_detect_task_from_config()` → `_resolve_model_class_from_config()`, which still raises `ValueError: ... no 'architectures' field` (no wrapper fallback there). So `--model-class X` alone fails on timm; the workaround is to also pass `--task image-classification`. Probably acceptable (advanced path), but worth either handling or documenting the required `--task` companion flag.

tests/unit/loader/test_detect_task_and_class.py
```diff
@@ -148,3 +152,55 @@ def test_no_override_for_unrelated_model(self):
...
+        config = MagicMock()
+        config.architectures = None
+        config.model_type = "timm_wrapper"
+        config._name_or_path = ""
+
+        task, resolved_class = _detect_task_and_class_from_config(config)
+
+        # Task is derived from Optimum's task list for the timm library, not hardcoded.
+        assert WRAPPED_LIBRARY_MODEL_TYPES["timm_wrapper"] == "timm"
+        assert task == "image-classification"
```
vortex-captain, line 195: Good — this test genuinely exercises the load-bearing scoped import: it does not pre-import `optimum.exporters.onnx.model_configs` nor mock `get_supported_tasks`, so it depends on the production branch performing that import (I confirmed that without it `get_supported_tasks` returns `[]` and this test would instead hit the ValueError). Two small suggestions: (1) the assertions hard-code `task == "image-classification"` and class name `AutoModelForImageClassification`, which couple the test to today's Optimum task ordering — fine as a guard, but a comment noting this is intentional (it IS the contract) would help future maintainers; (2) consider one negative assertion that the routing (not just any resolution) is what's under test — e.g. that the same config under an explicit non-timm library does not resolve — to guard against a future change that accidentally resolves timm via the normal `transformers` path.

src/winml/modelkit/export/io.py
```diff
@@ -210,11 +210,19 @@ def _get_onnx_config(
 
     normalized_task = _map_task_synonym(task)
 
+    # Route model_types whose Optimum OnnxConfig is registered under another
+    # library (e.g. timm via "timm_wrapper" -> "timm") so the lookup succeeds
+    # from every call site without an explicit --library flag.
+    from ..loader.task import resolve_optimum_library
+
+    library_name = resolve_optimum_library(model_type, library_name)
```
vortex-captain, line 218: Centralizing the reroute here (so config/build/export/inspect all benefit from one place) is the right call and I verified all the cross-package callers (`export/config.py::build_export_config_from_model`, `resolve_io_specs`, `generate_dummy_inputs`, inspect resolver) flow through `_get_onnx_config`. Note the same caveat as the loader docstring: because the reroute predicate is `library_name == "transformers"`, a user-supplied `--library transformers` is also rerouted (config command's `--library` default is `transformers`, so user-explicit vs default are indistinguishable here). Desired for timm, but worth a one-line comment so a future reader doesn't assume `--library` is always honored verbatim.

src/winml/modelkit/commands/inspect.py
```diff
@@ -426,11 +426,13 @@ def _inspect_model_v2(
...
             from optimum.exporters.tasks import TasksManager
 
+            from ..loader.task import resolve_optimum_library
+
             onnx_config_cls = TasksManager.get_exporter_config_constructor(
                 exporter="onnx",
                 model_type=model_type,
                 task=task,
-                library_name="transformers",
+                library_name=resolve_optimum_library(model_type),
```
vortex-captain, line 433: This updates the v2 (live CLI) inspect path correctly, and `resolve_optimum_library` is a pure dict lookup that cannot raise, so there's no regression risk to the existing display-label lookup. For symmetry, note the legacy `inspect/resolver.py::resolve_exporter` (line 346) was left with `library_name="transformers"` — see the overview comment; if that public path is meant to stay supported it should be routed the same way.

vortex-captain (commented):
## Review: feat(timm) — enable timm image-classification via library routing

I reviewed this fresh (treating nothing as known-good) and verified the core Optimum taxonomy facts by read-only introspection of the installed optimum 2.1.0 / optimum-onnx 0.1.0 / transformers 4.57.6 stack:

- `get_supported_tasks_for_model_type("timm_wrapper", exporter="onnx", library_name="timm")` → `["image-classification"]` (exactly one task).
- The same lookup under `library_name="transformers"` raises `KeyError("timm_wrapper is not supported yet for transformers")`, so `get_supported_tasks` (which swallows exceptions → `[]`) correctly does **not** false-trigger the new branch for non-timm/default paths.
- The scoped `import optimum.exporters.onnx.model_configs` is genuinely load-bearing in the loader branch: without it, `get_supported_tasks("timm_wrapper","timm")` raises `KeyError('default-timm-config')` → returns `[]` → branch skipped. Good that the import is placed there.
- `TimmWrapperConfig(num_labels=10)` really does have `model_type="timm_wrapper"` and `architectures=None`, so the test fixtures are realistic.
- Both `timm_wrapper` and `default-timm-config` resolve to `TimmDefaultOnnxConfig` under `library="timm"`.
- `export → loader` import direction already exists on `main` (export/config.py, export/htp/exporter.py), so the new `from ..loader.task import …` introduces no new layering violation.

**Verdict:** the mechanism is sound and the centralization in `_get_onnx_config` is a clean choice. Below are correctness edge-cases, one consistency gap, and several minor/convention items. Comment-only — no blocking objection.

### Notable item not on a changed line (so flagged here)

`src/winml/modelkit/inspect/resolver.py:346` — the **parallel** inspect path `resolve_exporter()` still hardcodes `library_name="transformers"` in its `TasksManager.get_exporter_config_constructor(...)` call. This function is exported in the `inspect` package public API (`inspect/__init__.py` `__all__` → `inspect_model` → `resolve_exporter`) but is **not** the path the `winml inspect` CLI uses (the CLI calls `commands/inspect.py::_inspect_model_v2`, which you did update). Still, for a timm model the legacy `inspect_model()` would (a) mislabel the OnnxConfig (the constructor lookup raises KeyError under transformers and is swallowed) and (b) `resolver.py::detect_task` would fall through its `except ValueError` to `HF_TASK_DEFAULTS` and return `next-sentence-prediction` as the task. Since `resolve_io_specs`/`_discover_io_attrs_from_onnx_config` now self-route via `_get_onnx_config`, this leaves that public path half-routed. Either route it the same way for consistency, or note that `inspect_model()`/`resolve_exporter` are legacy and slated for removal.

src/winml/modelkit/loader/task.py
vortex-captain: Fixed in 7ff36609: corrected the docstring - an explicit `--library transformers` is still rerouted for wrapped types (the check keys on the default transformers library), so only an explicit non-transformers library is returned unchanged.

src/winml/modelkit/loader/task.py
vortex-captain: Fixed in 7ff36609: `resolve_optimum_library` is now exported via `loader/__init__.py` (`__all__` + module Public API docstring), and `export/io.py` / `commands/inspect.py` import it from `..loader`.

src/winml/modelkit/loader/task.py
vortex-captain: Fixed in 84fb6963: dropped the redundant `or ""` (`WRAPPED_LIBRARY_MODEL_TYPES.get(...)` returns None for a missing/None key anyway), and `model_type` is now computed once and used consistently in the branch - no more direct `config.model_type` dereferences.

DingmaomaoBJTU (commented):
Good work routing timm models through the correct Optimum library. The centralization in `resolve_optimum_library` is clean, and the test coverage is solid. A few suggestions below -- nothing blocking.

DingmaomaoBJTU (commented):
Review with inline comments

DingmaomaoBJTU (commented):
Well-designed feature with strong test coverage and minimal surface area. The wrapped-library routing is clean and extensible. A couple of minor observations inline.

DingmaomaoBJTU (commented):
Good, well-structured PR. The library-routing design is clean - recording only the library (not the task) in WRAPPED_LIBRARY_MODEL_TYPES is the right abstraction. A few observations below.

DingmaomaoBJTU (commented):
Well-structured feature. The WRAPPED_LIBRARY_MODEL_TYPES dict is nicely extensible, and the tests cover both the happy path and the fallback-still-raises case. A few suggestions inline.

src/winml/modelkit/loader/task.py
```diff
@@ -154,6 +155,40 @@
...
+def resolve_optimum_library(model_type: str | None, library_name: str = "transformers") -> str:
+    """Route a transformers model_type to the Optimum library that owns its export.
+
+    Most models export under the library they were requested with. A few
+    transformers model_types are thin wrappers whose Optimum OnnxConfig lives in
+    another library (see :data:`WRAPPED_LIBRARY_MODEL_TYPES`); route those so the
+    OnnxConfig lookup succeeds without an explicit ``--library`` flag.
+
+    Only the ``"transformers"`` library is rerouted, so an explicit
+    non-``"transformers"`` library is returned unchanged. (An explicit
```
DingmaomaoBJTU: Good design: only storing the library in the registry and deriving the task from Optimum at runtime. This avoids hardcoding timm tasks and automatically picks up any new tasks Optimum adds. One concern: get_supported_tasks may return tasks in non-deterministic order depending on Optimum version. Consider sorting or explicitly preferring image-classification if multiple tasks are supported, to avoid surprising behavior if Optimum adds a second task later.

src/winml/modelkit/loader/task.py
```diff
@@ -314,8 +349,34 @@ def _detect_task_and_class_from_config(config: PretrainedConfig) -> tuple[str, t
         return resolve_task_and_model_class(config, task=override_task)
 
     # [1] Resolve architecture class from config.
-    # If config.architectures is missing/empty, this raises ValueError and the
-    # caller should provide task explicitly.
+    # Some model_types (e.g. timm via "timm_wrapper") are generic library
+    # wrappers that carry no `architectures` field. Resolve those through their
```
DingmaomaoBJTU: Using supported[0] as the task assumes the first task in the list is the primary one. This works today since timm only registers image-classification, but could silently change if Optimum adds more tasks. Consider adding a brief comment or an assertion/warning if len(supported) > 1.

src/winml/modelkit/inspect/resolver.py
```diff
@@ -121,6 +124,16 @@ def detect_task(config: PretrainedConfig) -> tuple[str, str]:
         if mt == model_type_normalized:
             return task, "HF_MODEL_CLASS_MAPPING"
```
DingmaomaoBJTU: Importing the private _detect_task_and_class_from_config in resolver is pragmatic but creates coupling to an internal API. If this function signature changes, resolver breaks silently. Consider whether a thin public wrapper in the loader package would be cleaner long-term.

DingmaomaoBJTU (commented):
Clean implementation of timm library routing. The WRAPPED_LIBRARY_MODEL_TYPES registry is well-designed and extensible. Test coverage is thorough. A few inline observations.

DingmaomaoBJTU (commented):
Good work! Clean implementation of the timm library routing. The approach of deriving the task from Optimum's registry rather than hardcoding is the right call. A few inline suggestions below.

tests/unit/export/test_timm_library_routing.py
```diff
@@ -0,0 +1,56 @@
...
+    """Minimal offline TimmWrapperConfig (no hub download)."""
+    from transformers import TimmWrapperConfig
+
+    return TimmWrapperConfig(num_labels=10)
+
+
+class TestTimmLibraryRouting:
+    """timm_wrapper resolves to Optimum's TimmDefaultOnnxConfig via library routing."""
+
+    def test_get_onnx_config_routes_to_timm_default(self, timm_wrapper_config) -> None:
```
DingmaomaoBJTU: Good use of `scope="module"` to avoid repeated config construction. Consider adding a comment noting that `num_labels=10` is arbitrary -- it just needs to be nonzero to simulate a real classification config (vs. a backbone with `num_labels=0`).

DingmaomaoBJTU (commented):
Solid implementation of library routing for timm models. The WRAPPED_LIBRARY_MODEL_TYPES registry pattern is extensible and the test coverage is thorough across all resolution paths (loader, export, inspect).

DingmaomaoBJTU (commented):
Well-structured feature addition. The library-routing approach via `WRAPPED_LIBRARY_MODEL_TYPES` is clean and extensible. Good test coverage across export, inspect, and loader paths. A few observations:

DingmaomaoBJTU (commented):
Good feature addition — the library-routing approach for timm (and future wrapped libraries) is clean and well-tested. A few observations below.

DingmaomaoBJTU (commented):
Good feature addition. The library-routing approach for timm (and future wrapped libraries) is clean and well-tested. A few observations inline.

DingmaomaoBJTU (commented):
## Code Review: feat(timm) - enable timm image-classification models via library routing

**Overall**: Clean, well-motivated feature. The layered approach (shared resolution + per-site routing) is well-designed and the test coverage is thorough. A few observations below.

DingmaomaoBJTU (commented):
## Code Review: feat(timm) - enable timm image-classification models via library routing

**Overall**: Clean, well-motivated feature with thorough test coverage. The layered resolution approach is well-designed. See inline comments for specific observations.

DingmaomaoBJTU (commented):
Good feature addition! The library-routing approach is clean and the test coverage is solid. A few suggestions below.

DingmaomaoBJTU (commented):
Good feature addition — the library-routing approach is well-designed. A few suggestions inline.

DingmaomaoBJTU (commented):
Good feature overall — timm library routing is well-motivated and the test coverage is solid. A few suggestions below around robustness and maintainability.

DingmaomaoBJTU (commented):
Well-structured feature. The `WRAPPED_LIBRARY_MODEL_TYPES` dict is a clean extension point, and test coverage is thorough. A few inline observations.

DingmaomaoBJTU (commented):
Thanks for this — clean routing layer for timm/wrapped-library models. A few suggestions inline.

DingmaomaoBJTU (commented):
Good work routing timm models through the Optimum library layer. A few inline suggestions below.

DingmaomaoBJTU (commented):
## Code Review: PR #790 - feat(timm): enable timm image-classification models via library routing

**Overall**: Well-structured feature addition. The approach of routing through WRAPPED_LIBRARY_MODEL_TYPES is clean and extensible. Good test coverage across the loader, inspector, and export paths. A few observations:

### src/winml/modelkit/loader/task.py - \
esolve_optimum_library\ signature
The function signature \
esolve_optimum_library(model_type: str | None, library_name: str = 'transformers')\ has a default for \library_name\, but in \inspect.py\ line 427 it's called with only one arg: \
esolve_optimum_library(model_type)\. This is fine, but in \xport/io.py\ it's called as \
esolve_optimum_library(model_type, library_name)\ which re-passes the caller's library_name. If \library_name\ is already \None\ upstream (unlikely but possible), the \== 'transformers'\ check would silently skip routing. Consider guarding: \if (library_name or 'transformers') == 'transformers'\.

### src/winml/modelkit/loader/task.py - \supported[0]\ task selection
At line ~194: \	ask = supported[0]\ picks the first task from Optimum's task list. This is an implicit ordering dependency — if Optimum ever reorders or adds tasks for timm, the default task changes silently. Consider asserting or logging a warning if \len(supported) > 1\, or documenting that the first entry is intentionally the preferred task.

### src/winml/modelkit/inspect/resolver.py - Bare \xcept Exception\
At line ~83: \xcept Exception: logger.debug(...)\ swallows all errors during wrapped-library detection. This is fine for graceful fallback, but consider catching a narrower exception set (e.g., \(ImportError, KeyError, ValueError)\) so truly unexpected errors (e.g., \RuntimeError\ from a corrupted Optimum registry) still surface during development.

### tests/unit/export/test_timm_library_routing.py - Fixture hardcodes \
um_labels=10\
The \	imm_wrapper_config\ fixture uses \TimmWrapperConfig(num_labels=10)\. Since the tests verify I/O shapes and routing (not label count), this is fine, but a comment explaining why 10 (or that the value is arbitrary) would prevent future confusion.

### General: Extensibility
The \WRAPPED_LIBRARY_MODEL_TYPES\ dict is a clean extension point. When adding future wrapped libraries (e.g., diffusers wrappers), only this dict needs updating. Nice design.

DingmaomaoBJTU (commented):
Good feature - the library-routing approach via WRAPPED_LIBRARY_MODEL_TYPES is clean and extensible. A few suggestions inline.

DingmaomaoBJTU (commented):
Well-structured addition. The library-routing approach via WRAPPED_LIBRARY_MODEL_TYPES is clean and extensible. Tests are comprehensive covering both the export and inspect paths.

DingmaomaoBJTU (commented):
Clean feature addition. The `resolve_optimum_library` router is well-scoped and the test coverage is solid across all three resolution paths (loader, export/io, inspect/resolver). A couple of minor observations inline.

src/winml/modelkit/loader/task.py
```diff
@@ -314,8 +349,34 @@ def _detect_task_and_class_from_config(config: PretrainedConfig) -> tuple[str, t
...
+        if library is not None:
+            # Populate Optimum's exporter registry (incl. the wrapped library's
+            # task list) before querying it; scoped to this rare branch so normal
+            # model loading never pays for the import.
+            import optimum.exporters.onnx.model_configs  # noqa: F401
+
+            supported = get_supported_tasks(model_type, library_name=library)
+            if supported:
+                task = supported[0]
+                model_class = TasksManager.get_model_class_for_task(task, framework="pt")
```
vortex-captain, line 384: taking first as best effort, since task is not specified by user

src/winml/modelkit/loader/task.py
```diff
@@ -314,8 +349,34 @@ def _detect_task_and_class_from_config(config: PretrainedConfig) -> tuple[str, t
...
+                logger.info(
+                    "config has no 'architectures'; resolved %s via %s library (task=%s, class=%s)",
+                    model_type,
+                    library,
+                    task,
+                    model_class.__name__,
+                )
```
vortex-captain, line 391: taking first as best effort, since task is not specified by user

DingmaomaoBJTU (commented):
Clean implementation of timm library routing. The approach of using a single lookup table (WRAPPED_LIBRARY_MODEL_TYPES) with a pure routing function is well-designed. A few inline comments.

src/winml/modelkit/loader/task.py
vortex-captain: taking first as best effort, since task is not specified by user

src/winml/modelkit/loader/task.py
```diff
@@ -314,8 +349,34 @@ def _detect_task_and_class_from_config(config: PretrainedConfig) -> tuple[str, t
...
+                return task, model_class
+    # If config.architectures is still missing/empty, this raises ValueError and
+    # the caller should provide task explicitly.
```
vortex-captain, line 394: taking first as best effort, since task is not specified by user

src/winml/modelkit/loader/task.py
vortex-captain, line 391: taking first as best effort, since task is not specified by user

DingmaomaoBJTU (commented):
Well-structured feature addition. The library routing is minimal and placed at exactly the right abstraction layer. Good test coverage with both unit and integration tests. A few observations inline.

DingmaomaoBJTU (commented):
Good feature addition for timm library routing. The approach of having a centralized WRAPPED_LIBRARY_MODEL_TYPES mapping is clean and extensible. A few inline observations below.

src/winml/modelkit/loader/task.py
```diff
@@ -314,8 +349,34 @@ def _detect_task_and_class_from_config(config: PretrainedConfig) -> tuple[str, t
...
+        library = WRAPPED_LIBRARY_MODEL_TYPES.get(model_type) if model_type else None
+        if library is not None:
+            # Populate Optimum's exporter registry (incl. the wrapped library's
+            # task list) before querying it; scoped to this rare branch so normal
+            # model loading never pays for the import.
+            import optimum.exporters.onnx.model_configs  # noqa: F401
+
+            supported = get_supported_tasks(model_type, library_name=library)
+            if supported:
+                task = supported[0]
```
zhenchaoni, line 373: Add a comment here. Since we don't have architecture here, if there will be other architectures for timm_wrapper in the future, the logic will be wrong. It is OK to hardcode today because image-classification seems to be the only task for timm wrapper.

src/winml/modelkit/inspect/resolver.py
```diff
@@ -121,6 +124,16 @@ def detect_task(config: PretrainedConfig) -> tuple[str, str]:
...
 
+    # Wrapped-library model types (e.g. timm via "timm_wrapper") carry no
+    # `architectures`; reuse the loader's resolution to derive the real task
+    # instead of falling through to the HF_TASK_DEFAULTS mislabel below.
+    if model_type in WRAPPED_LIBRARY_MODEL_TYPES and not getattr(config, "architectures", None):
+        try:
+            task, _ = _detect_task_and_class_from_config(config)
+            return task, "wrapped-library"
```
zhenchaoni: What is this detection source used? Can we define a constant?

src/winml/modelkit/commands/inspect.py
```diff
@@ -426,11 +424,13 @@ def _inspect_model_v2(
             import optimum.exporters.onnx.model_configs  # noqa: F401
             from optimum.exporters.tasks import TasksManager
 
+            from ..loader import resolve_optimum_library
```
zhenchaoni, line 427: Try to see if import loader will be slow and load a lot of modules. Since `resolve_optimum_library` is a very lightweight one, if importing the entire "loader" module take multiple seconds, you can consider move resolve_optimum_library to a different place.

vortex-captain: marginal cost is ~0.13s, should be fine.

src/winml/modelkit/inspect/resolver.py
vortex-captain: It's a display tag. Added the constant

src/winml/modelkit/loader/task.py
vortex-captain: added comment

DingmaomaoBJTU (commented):
Well-designed feature. The `WRAPPED_LIBRARY_MODEL_TYPES` approach is clean and extensible. Good test coverage across loader, export, and inspect paths. A couple of observations inline.

DingmaomaoBJTU (commented):
Clean design — routing by model_type via a single dict avoids scattering timm-specific logic. The scoped import of `optimum.exporters.onnx.model_configs` is well-documented, and the `supported[0]` caveat comment is good forward-looking design. Tests cover all three resolution paths (loader, export, inspect). Two minor comments below.

DingmaomaoBJTU (commented):
Well-designed feature. The library-routing approach (centralizing into `resolve_optimum_library` + `WRAPPED_LIBRARY_MODEL_TYPES`) is clean and extensible. Tests cover both the happy path and fallback-raises. A couple of notes:

src/winml/modelkit/loader/task.py
```diff
@@ -154,6 +155,40 @@
...
+# (get_supported_tasks), not hardcoded.
+# model_type -> optimum_library
+WRAPPED_LIBRARY_MODEL_TYPES: dict[str, str] = {
+    "timm_wrapper": "timm",
+}
+
+
+def resolve_optimum_library(model_type: str | None, library_name: str = "transformers") -> str:
+    """Route a transformers model_type to the Optimum library that owns its export.
+
```
DingmaomaoBJTU, line 176: Nit (defensive): resolve_optimum_library accepts model_type: str | None but the in check will evaluate to False for None since None is not a dict key. This works correctly but consider an early if not model_type: return library_name guard to make the contract explicit for future readers.

src/winml/modelkit/loader/task.py
```diff
@@ -314,8 +349,41 @@ def _detect_task_and_class_from_config(config: PretrainedConfig) -> tuple[str, t
...
+            supported = get_supported_tasks(model_type, library_name=library)
+            if supported:
+                # Take the first task. A wrapped library currently exposes a
+                # single ONNX export task, so the first is the only/right default
+                # -- for timm that is "image-classification". This holds only
+                # while the library has one task: if a wrapped library ever
+                # exposes multiple tasks, supported[0] becomes an arbitrary pick
+                # and this needs explicit per-model-type selection (or --task).
+                # Revisit then.
```
DingmaomaoBJTU: Fragility note: Taking supported[0] works today because timm exposes only image-classification, but if a wrapped library ever advertises multiple tasks this silently picks an arbitrary one. Consider emitting a logger.warning when len(supported) > 1 so you get an early signal rather than a silent mislabel.

src/winml/modelkit/inspect/resolver.py
```diff
@@ -15,8 +15,11 @@
 from ..loader.task import (
     HF_TASK_DEFAULTS,
     KNOWN_TASKS,
+    WRAPPED_LIBRARY_MODEL_TYPES,
+    _detect_task_and_class_from_config,
```
DingmaomaoBJTU, line 19: Cross-module private API: _detect_task_and_class_from_config is imported with its underscore prefix from loader.task into the inspect package. Since two separate packages now depend on it, consider promoting it to public API (drop the underscore) or adding a thin public wrapper to signal this is a supported contract.

src/winml/modelkit/export/io.py
```diff
@@ -210,11 +210,19 @@ def _get_onnx_config(
 
     normalized_task = _map_task_synonym(task)
 
+    # Route model_types whose Optimum OnnxConfig is registered under another
+    # library (e.g. timm via "timm_wrapper" -> "timm") so the lookup succeeds
+    # from every call site without an explicit --library flag.
+    from ..loader import resolve_optimum_library
```
DingmaomaoBJTU, line 216: Minor: The lazy import is consistent with this files style, but since resolve_optimum_library is just a dict lookup with no heavy transitive imports, a module-level import would also work and avoids repeated import overhead per call.

tests/unit/export/test_timm_library_routing.py
```diff
@@ -0,0 +1,56 @@
...
+        """resolve_io_specs yields the timm image-classifier I/O without a --library flag."""
+        specs = resolve_io_specs("timm_wrapper", "image-classification", timm_wrapper_config)
+        assert specs["input_names"] == ["pixel_values"]
+        assert "logits" in specs["output_names"]
+
+    def test_pixel_values_is_4d_nchw(self, timm_wrapper_config) -> None:
+        specs = resolve_io_specs("timm_wrapper", "image-classification", timm_wrapper_config)
+        shape = specs["input_shapes"][0]
+        assert len(shape) == 4, f"pixel_values should be 4D NCHW, got {shape}"
+        assert shape[1] == 3, f"expected 3 channels, got {shape[1]}"
```
DingmaomaoBJTU, line 56: Suggestion: Consider adding a negative test that verifies a non-timm model type (e.g. bert) with library_name=transformers is NOT rerouted through this path, to guard against accidental widening of the routing logic in the future.

DingmaomaoBJTU (commented):
Good PR overall. The approach of routing wrapped-library model types through a shared resolution layer is clean and extensible. A few observations inline.

DingmaomaoBJTU (commented):
Clean implementation of the timm library routing. The architecture is sound - a single `WRAPPED_LIBRARY_MODEL_TYPES` dict as the source of truth, with `resolve_optimum_library` as the routing function. Comments:

1. **Single-task assumption** (`loader/task.py`): The comment on `supported[0]` acknowledges the fragility if a wrapped library ever exposes multiple tasks. This is fine today (timm = image-classification only). Suggestion: add an assert/warning when `len(supported) > 1` so this doesn't silently pick wrong in the future:
   \\\python
   if len(supported) > 1:
       logger.warning('Wrapped library %s exposes %d tasks; picking first: %s', library, len(supported), supported[0])
   \\\

2. **Import inside function** (`inspect/resolver.py:134`): `_detect_task_and_class_from_config` is imported at the top of the file but also called here. Good - avoids circular imports since it's from the same package. However, importing `WRAPPED_LIBRARY_MODEL_TYPES` at the top and using it in the guard condition is cleaner than doing both checks inline.

3. **Test coverage**: The unit tests are thorough - `TestResolveOptimumLibrary` covers the routing table, `TestWrappedLibraryArchitecturesFallback` covers the detection path. The `test_timm_library_routing.py` integration tests ensure the full stack works. Well structured.

4. **Potential conflict with PR #786**: Both PRs modify `inspect/resolver.py`'s `resolve_exporter` function and `loader/task.py`. The changes are complementary (this PR routes library, #786 normalizes task names) but will conflict textually. Recommend merging one first and rebasing the other.

5. **`resolve_optimum_library` default parameter**: The signature `resolve_optimum_library(model_type, library_name='transformers')` means callers that already have a library variable must always pass it explicitly. The `_get_onnx_config` callsite in `export/io.py` correctly passes `library_name` through. Good.

DingmaomaoBJTU (commented):
Solid feature addition. The library-routing approach is clean and well-isolated — it only triggers for wrapped model types and has zero overhead for normal models. Tests cover the key paths (loader, export, inspect). A few inline observations below.

src/winml/modelkit/loader/task.py
DingmaomaoBJTU, line 176: The comment here is good — acknowledging this is fragile if a wrapped library ever exposes multiple tasks. Consider adding a guard: if len(supported) > 1, log a warning and/or fall through to require --task explicitly, rather than silently picking supported[0].

src/winml/modelkit/loader/task.py
```diff
@@ -154,6 +155,40 @@
...
+# library through a single type (e.g. timm via "timm_wrapper"). Such configs
+# carry no `architectures` field, and their Optimum ONNX export config is
+# registered under the wrapped library, not "transformers". This is a
+# library-routing concern handled at the common resolution layer (the loader
+# below and export.io._get_onnx_config), not a per-model OnnxConfig.
+#
+# Only the library is recorded here -- it is the irreducible Optimum-taxonomy
+# fact. The export task is derived from Optimum's task list for that library
+# (get_supported_tasks), not hardcoded.
+# model_type -> optimum_library
```
DingmaomaoBJTU, line 168: Nit: The default parameter library_name: str = "transformers" could be slightly confusing since callers in xport/io.py pass a variable that might already be the correct library. The docstring clarifies this well, but consider naming it default_library or documenting at the call site that this is the "fallback" rather than a forced override.

vortex-captain: That would be a lot of renaming. An alternative is to make `None` the default value for `library_name`, which requires basically as many places to change. Would be better in a dedicated refactor PR

src/winml/modelkit/loader/task.py
vortex-captain: added a warning

DingmaomaoBJTU (approved):

zhenchaoni (approved):

PR #793, opened 2026-05-29
fix: unify verbosity declaration and stderr-only logger output (#566)

### LINKED ISSUE #566: [cross-cutting] [P2] Verbosity is declared inconsistently; logger output mixes with UI
**Severity**: P2  
**Command**: `cross-cutting`

---
**Category**: Verbosity Convention — one declaration, one log format, one stream.

**Repro**:
```
uv run winml export -m microsoft/resnet-50 --verbose -o temp\ux\rn.onnx
uv run winml export -m microsoft/resnet-50 -vv -o temp\ux\rn.onnx     # error: extra argument
uv run winml --quiet export -m microsoft/resnet-50 -o temp\ux\rn.onnx # works
uv run winml export --quiet -m microsoft/resnet-50 -o temp\ux\rn.onnx # error: no --quiet on export
```

**Actual**: Mix:
- Top-level group declares `-v/--verbose` (count) and `-q/--quiet`.
- 12 of 13 subcommands **redeclare** `--verbose` as `is_flag=True` (so `-vv` after subcommand fails).
- No subcommand exposes `-q/--quiet` (parent-level only).
- Only `analyze` uses the shared `verbosity_options` decorator.
- DEBUG / INFO log lines are interleaved with Rich tables on stdout.

1. **Declaration**: top-level group declares `-v/--verbose` (count, `-v` = INFO, `-vv` = DEBUG) and `-q/--quiet` (errors only). **No subcommand redeclares either flag.** Apply via shared `verbosity_options` decorator.

2. **Default level**: `WARNING` and above. `-q` raises to `ERROR`. `-v` lowers to `INFO`. `-vv` lowers to `DEBUG`.

3. **Log format** (stderr only):
   ```python
   logging.Formatter(
       "[%(asctime)s %(levelname)-7s %(name)s] %(message)s",
       datefmt="%H:%M:%S",
   )
   ```
   Sample line: `[14:32:11 INFO    winml.export] Loaded config.json (cached)`

4. **Streams**:
   - **stdout** = user-facing UI (Rich panels, success messages, JSON results, `--help`).
   - **stderr** = log lines (DEBUG/INFO/WARNING/ERROR), library warnings, progress bars, banners.
   - This way `winml export ... > out.onnx 2> log.txt` does the right thing.

5. **Compliance** — current state vs. the proposal:

   | Command | Inherits top-level `-v/-q` | No subcommand redeclare | Logger → stderr only |
   |---|---|---|---|
   | `analyze` | ✅ | ✅ | ⚠ (some `print` to stdout) |
   | `build` | ✅ | ❌ redeclares `--verbose` | ❌ |
   | `compile` | ✅ | ❌ | ❌ |
   | `config` | ✅ | ❌ | ❌ |
   | `eval` | ✅ | ❌ | ❌ |
   | `export` | ✅ | ❌ | ❌ |
   | `hub` | ✅ | ❌ | ❌ |
   | `inspect` | ✅ | ❌ | ❌ |
   | `optimize` | ✅ | ❌ | ❌ |
   | `perf` | ✅ | ❌ | ❌ |
   | `quantize` | ✅ | ❌ | ❌ |
   | `sys` | ✅ | ❌ | ❌ (DEBUG inside tables — another audit finding) |
   | `expand_rules` | ✅ | ✅ | ✅ |

   So today only `analyze` and `expand_rules` partially comply. The rest must remove their local `--verbose` declarations and route loggers through `logging.basicConfig(stream=sys.stderr)`.

**Why it matters**: Without a single convention, users learn the flags 13 times. Mixed streams make piping (`> file 2> log`) impossible and break automation.

xieofxie:
Fixes #566.

## Problem

- Top-level group declared ``-v/--verbose`` (count) and ``-q/--quiet``, but 12 of 13 subcommands redeclared ``--verbose`` as ``is_flag=True``, so ``winml export -vv …`` errored with ``extra argument``.
- No subcommand exposed ``-q/--quiet``, so ``winml export --quiet …`` failed with ``no such option``.
- Each command wired logging differently; DEBUG/INFO lines interleaved with Rich tables on stdout, breaking ``cmd > out 2> log.txt``.

## Changes

- ``utils/cli.py``: ``verbosity_options`` decorator (``-v`` count, ``-q`` flag) + new ``resolve_verbosity(ctx, verbose, quiet)`` helper that merges top-level and subcommand-level values (max of verbose, OR of quiet). Honors the legacy ``ctx.obj[""debug""]`` so tests that bypass ``main()`` still raise the verbosity floor.
- ``utils/logging.py``: format ``[%(asctime)s %(levelname)-7s %(name)s] %(message)s`` with ``datefmt=%H:%M:%S``, ``stream=sys.stderr``. Idempotent — re-creates the WinML handler bound to the current ``sys.stderr`` on each call so Click ``CliRunner`` stream redirection keeps working, and leaves non-WinML handlers (notably pytest ``caplog``) intact.
- ``cli.py``: top-level group uses ``@verbosity_options`` (replaces inline declarations); ``--debug`` alias preserved.
- 12 subcommands (``build``, ``compile``, ``config``, ``eval``, ``export``, ``inspect``, ``optimize``, ``perf``, ``quantize``, ``sys``, plus ``analyze`` cleanup): replace ad-hoc ``--verbose`` (``is_flag=True``) with ``@cli_utils.verbosity_options``, add ``quiet: bool`` param, call ``configure_logging(verbosity=verbose, quiet=quiet)`` after ``resolve_verbosity``. Removes the legacy ``if ctx.obj.get(""debug""): verbose = True`` blocks (folded into the helper).
- ``serve/app.py``: pre-existing latent bug — module-level ``logging.getLogger(""winml.modelkit"").setLevel(INFO)`` ran at import, which muted DEBUG capture in unrelated tests that got collected alongside the serve test module. Split into ``_attach_log_handler()`` (idempotent, called from ``_register_routes``) and a paired ``_ensure_log_capture_level`` / ``_restore_log_capture_level`` invoked from the production lifespan. Tests that build the app via ``_register_routes`` + a mock lifespan no longer leak global logger state.

## Behavior

Both flag positions work; subcommand value wins when both are passed (max/OR merge):

```text
winml -v export -m … -o …            # top-level: works
winml export -vv -m … -o …            # subcommand: now works (was: extra argument)
winml --quiet export -m … -o …        # top-level: works
winml export --quiet -m … -o …        # subcommand: now works (was: no such option)
winml inspect -vv -m … --format json > out 2> log.txt   # clean stdout/stderr split
```

## Tests

- ``tests/cli/`` (23): pass
- ``tests/unit/`` (5061 collected): **5058 pass**, 3 fail — all 3 pre-existing on main and unrelated to this change:
  - ``test_winml_session.py::TestOpenVINODeviceRouting::test_compile_openvino_cpu_device_succeeds``
  - ``test_winml_session.py::TestOpenVINODeviceRouting::test_compile_openvino_cpu_provider_not_npu`` (both env: no OpenVINO EP installed)
  - ``test_config_utils.py::TestMergeConfigNoneHandling::test_none_to_value_transition`` (test isolation, passes alone)


DingmaomaoBJTU (commented):
Good refactoring overall — unifying verbosity across all subcommands is a solid improvement. The 
esolve_verbosity helper and idempotent configure_logging approach are well-designed. A few observations below.

DingmaomaoBJTU (commented):
Good refactoring overall - the unified verbosity contract is a clear improvement over the scattered ad-hoc --verbose flags. A few observations below, mostly around scope of root logger manipulation and minor readability.

DingmaomaoBJTU (commented):
Review with inline comments

src/winml/modelkit/commands/inspect.py
```diff
@@ -75,13 +78,6 @@ def _looks_like_local_path(model_id: str) -> bool:
     default="table",
     help="Output format (default: table)",
 )
-@click.option(
-    "-v",
-    "--verbose",
-    is_flag=True,
-    default=False,
```
DingmaomaoBJTU: The early OR-merge of `quiet` before `resolve_verbosity()` is called again below adds subtle complexity. Consider restructuring to call `resolve_verbosity()` once early (before the banner) to eliminate the duplicated quiet-merge and the 'idempotent OR' caveat.

src/winml/modelkit/utils/logging.py
```diff
@@ -35,6 +44,11 @@ def configure_logging(
 ) -> None:
     """Configure root logger based on verbosity level.
```
DingmaomaoBJTU: Setting `root.setLevel(log_level)` will lower the root logger level globally, which could surface unexpected log records from third-party libraries (requests, urllib3, etc.) when `-vv` is used. Consider targeting only the `winml.modelkit` logger or setting the handler level without changing the root level.

src/winml/modelkit/serve/app.py
```diff
@@ -92,9 +92,51 @@ def since(self, after_seq: int) -> list[dict]:
...
+        pkg_logger.addHandler(_log_handler)
+
+
+def _ensure_log_capture_level() -> int | None:
+    """Raise the package logger level to INFO if needed; return prior level.
+
+    Returns ``None`` when no change was made (so the caller knows there is
+    nothing to restore). Pair with :func:`_restore_log_capture_level`.
+    """
+    pkg_logger = logging.getLogger("winml.modelkit")
```
DingmaomaoBJTU: Storing internal state on `app.state._log_capture_prior` (underscore-prefixed) is slightly fragile — if FastAPI ever cleans or validates state attributes differently. A module-level variable (guarded by the lifespan context) might be safer, or at minimum document that this is intentional internal state.

DingmaomaoBJTU (commented):
Solid refactor that correctly centralizes verbosity handling. The decorator-to-factory change, resolve_verbosity helper, and idempotent configure_logging are all well-motivated. A few observations inline, nothing blocking.

DingmaomaoBJTU (commented):
Good refactor overall - the centralized `verbosity_options()` + `resolve_verbosity()` pattern is a solid improvement over the per-command ad-hoc approach. A few concerns noted in inline comments, mostly around the root-logger level side-effect and a subtle quiet-vs-verbose interaction for non-logging uses of `bool(verbose)`.

DingmaomaoBJTU (commented):
Solid refactoring — unifying verbosity across all 12 subcommands is a big DX improvement. The `resolve_verbosity` helper with max(verbose) + OR(quiet) semantics is clean and intuitive. A few inline notes on edge cases.

src/winml/modelkit/utils/cli.py
```diff
@@ -191,26 +191,59 @@ def verbosity_options(f):
...
+    ``winml -v export …`` and ``winml export -v …`` should behave the same.
+    This helper takes the max verbosity and OR of quiet so users can supply
+    the flag at either level (or both).
+
+    Precedence: ``-q``/``--quiet`` always wins over verbosity, including the
+    ``--debug`` alias — ``winml --debug export -q …`` runs at ERROR. ``-q`` is
+    an explicit "shut up" signal and trumps any verbosity raise, so the user
+    is never surprised by debug spam after they asked for quiet.
+
     Args:
```
DingmaomaoBJTU: The `resolve_verbosity` docstring says quiet always wins over verbosity, but the implementation returns both values independently and leaves the precedence decision to `configure_logging`. Consider adding an explicit guard here (e.g. `if quiet: verbose = 0`) to match the documented contract, or adjust the docstring to say 'quiet takes precedence as enforced by configure_logging'.

src/winml/modelkit/commands/inspect.py
```diff
@@ -189,7 +187,12 @@ def inspect(
...
     # and in JSON mode (Click 8.4 mixes stderr into CliRunner.result.output,
     # and JSON consumers expect clean stdout regardless).
-    quiet = bool(ctx.obj and ctx.obj.get("quiet"))
+    #
+    # NOTE: this is an EARLY OR-merge of top-level --quiet, done before the
+    # banner so `winml --quiet inspect …` actually suppresses it. The full
+    # resolve_verbosity() merge happens below, before configure_logging() —
+    # the duplication is idempotent (OR), not a subtle ordering dependency.
+    quiet = quiet or bool(ctx.obj and ctx.obj.get("quiet"))
     json_mode = output_format.lower() == "json"
```
DingmaomaoBJTU: This early `quiet` OR-merge before `resolve_verbosity()` is called a few lines later creates a subtle double-merge pattern. The inline comment acknowledges it's idempotent, but it makes the flow harder to follow. Consider extracting a tiny helper like `_is_quiet(ctx, quiet)` or at minimum adding a forward reference comment at the `resolve_verbosity` call site noting that `quiet` was already partially merged above.

src/winml/modelkit/serve/app.py
```diff
@@ -92,9 +92,51 @@ def since(self, after_seq: int) -> list[dict]:
...
+    if pkg_logger.level == logging.NOTSET or pkg_logger.level > logging.INFO:
+        prior = pkg_logger.level
+        pkg_logger.setLevel(logging.INFO)
+        return prior
+    return None
+
+
+def _restore_log_capture_level(prior: int | None) -> None:
```
DingmaomaoBJTU: Nice fix for the test-isolation leak. One concern: `_ensure_log_capture_level` checks `pkg_logger.level > logging.INFO` but if the logger is at DEBUG (from a `-vv` invocation), this would overwrite it to INFO, silently downgrading verbosity. Should this only raise the level (i.e., skip the setLevel if already at DEBUG)?

DingmaomaoBJTU (commented):
Thorough refactoring that properly unifies verbosity across all subcommands. The resolve_verbosity helper and idempotent configure_logging are well-designed. A few inline comments on edge cases.

src/winml/modelkit/utils/cli.py
```diff
@@ -191,26 +191,59 @@ def verbosity_options(f):
...
+            help="Quiet mode - errors only to stderr",
+        )(f)
+        return click.option(
+            "--verbose",
+            "-v",
+            count=True,
+            help="Increase verbosity (-v=INFO, -vv=DEBUG)",
+        )(f)
+
+    return decorator
```
DingmaomaoBJTU: Nit: the decorator inner function does not return a value explicitly. It relies on `click.option(...)` decorating `f` in-place. While this works with Click's current implementation, returning the result would be more explicit and defensive:
```python
def decorator(f):
    f = click.option(...)(f)
    f = click.option(...)(f)
    return f
return decorator
```
Currently you apply `--quiet` first, then return the result of the `--verbose` application. The `--quiet` result is dropped. This works because Click mutates `f`, but it is fragile if Click ever changes behavior.

src/winml/modelkit/utils/logging.py
```diff
@@ -49,13 +63,21 @@ def configure_logging(
...
+    # invocation would write to a stream the test no longer captures.
+    # We leave non-WinML handlers (notably pytest's caplog handler) alone.
+    for h in list(root.handlers):
+        if getattr(h, _HANDLER_MARKER, False):
+            root.removeHandler(h)
+    own_handler = logging.StreamHandler(sys.stderr)
+    own_handler.setFormatter(logging.Formatter(_LOG_FORMAT, datefmt=_DATE_FORMAT))
+    setattr(own_handler, _HANDLER_MARKER, True)
+    own_handler.setLevel(log_level)
+    root.addHandler(own_handler)
```
DingmaomaoBJTU: Setting `root.setLevel(log_level)` on the **root** logger could lower the threshold for all third-party loggers (urllib3, transformers, onnxruntime). Consider scoping the level to `logging.getLogger('winml.modelkit')` instead, and leaving root at its default (WARNING). The handler-level filter will still gate what gets emitted, but propagated records from noisy libraries won't slip through.

src/winml/modelkit/serve/app.py
```diff
@@ -126,6 +168,11 @@ def create_app(
     @asynccontextmanager
     async def lifespan(app: FastAPI):
         app.state.start_time = time.time()
+        # Raise the modelkit logger to INFO so the ring handler receives
+        # operational records during `winml serve`. Tests that build the app
+        # via ``_register_routes`` + their own mock lifespan never reach this
+        # branch, so the global level stays unchanged for them.
+        app.state._log_capture_prior = _ensure_log_capture_level()
```
DingmaomaoBJTU: Storing internal state as `app.state._log_capture_prior` uses a leading-underscore attribute on a Starlette `State` object. This is fine functionally, but `State.__getattr__` raises `AttributeError` for missing attrs. The `getattr(..., None)` fallback in the shutdown path handles this correctly. Might be cleaner to initialize it to `None` unconditionally at the top of `lifespan` (before the `if mode` branch) so the shutdown restore is always symmetric.

DingmaomaoBJTU (commented):
Good refactoring - unifying verbosity across all subcommands is a solid DX improvement. The `resolve_verbosity` merge logic (max verbose, OR quiet) is clean and the idempotent handler management in `configure_logging` is well-thought-out for test isolation.

A few items to consider (see inline comments):
1. Root logger level setting may expose noisy third-party logs
2. Build-config `verbose` key support was dropped in compile.py
3. Minor decorator return-value hygiene in `verbosity_options()`

DingmaomaoBJTU (commented):
Overall: solid refactoring that unifies verbosity across all subcommands. The idempotent handler approach in configure_logging is well-thought-out for test isolation. A few inline suggestions below.

DingmaomaoBJTU (commented):
Good refactoring overall. The unified verbosity approach is clean and the idempotent handler management in configure_logging is well-designed for test isolation. A few inline notes.

DingmaomaoBJTU (commented):
Good refactor - the factory pattern for verbosity_options() and the resolve_verbosity() helper cleanly eliminate the 12x duplicated --verbose declarations. A few observations inline.

DingmaomaoBJTU (commented):
Well-structured refactor that unifies verbosity handling across all subcommands. The approach of merging parent/child -v/-q via resolve_verbosity() is clean and ensures positional equivalence. A few inline notes below.

DingmaomaoBJTU (commented):
Solid refactoring — unifying verbosity across all subcommands is a big usability win. The `resolve_verbosity()` merge logic is well-documented. A few comments below.

DingmaomaoBJTU (commented):
Solid refactor - the unified verbosity_options() decorator and resolve_verbosity() helper eliminate repetition and fix the winml -v subcmd vs winml subcmd -v inconsistency. The idempotent configure_logging with tagged handlers is a nice touch for test isolation.

**Observations:**

1. **src/winml/modelkit/utils/logging.py** - Setting root.setLevel(log_level) on the root logger will lower the threshold for all third-party library loggers (urllib3, httpx, onnxruntime) when the user passes -vv. Not a regression (old basicConfig did the same), but since you are refactoring: consider scoping to logging.getLogger('winml.modelkit').setLevel(log_level) and leaving the root logger untouched. That way -vv surfaces winml debug logs without flooding stderr with noisy library internals.

2. **src/winml/modelkit/commands/inspect.py** - The early quiet OR-merge before the banner followed by resolve_verbosity() again later is subtle. A simpler pattern: call resolve_verbosity once at the top and use it everywhere. Removes the duplicated OR logic and the need for the explanatory paragraph. Minor - current code is correct.

3. **src/winml/modelkit/serve/app.py** - Nice defensive design - bounding the ring handler level independently prevents test-pollution. The attach/ensure/restore trio is well-documented. Nit: consider storing log_capture_prior on app.extra or a module-level var rather than app.state with a leading underscore.

DingmaomaoBJTU (commented):
Good refactoring to unify verbosity handling across all subcommands. The `resolve_verbosity` merge logic is clean and the idempotent `configure_logging` approach fixes the CliRunner stderr-swap issue nicely. A few observations below.

DingmaomaoBJTU (commented):
Clean refactoring. Unifying verbosity under a shared decorator and making configure_logging idempotent is a solid architectural improvement. The serve/app.py changes to avoid side effects at import time are well-thought-out. A few comments below.

DingmaomaoBJTU (commented):
Good PR - clean unification of verbosity across all subcommands. The resolve_verbosity + configure_logging pattern is consistent and well-documented. A few inline suggestions below.

DingmaomaoBJTU (commented):
Good refactoring! The unified verbosity pattern is much cleaner. A few concerns below.

DingmaomaoBJTU (commented):
Good refactoring overall — the unified `verbosity_options()` + `resolve_verbosity()` pattern cleanly solves the duplicate-declaration problem and the stderr split is a clear improvement. A few observations below.

DingmaomaoBJTU (commented):
Thorough refactoring that properly unifies verbosity handling across all subcommands. The merge-semantics (max verbose, OR quiet) are well-thought-out, and the idempotent logger setup for CliRunner compatibility is a nice detail. A few comments inline.

DingmaomaoBJTU (commented):
Good refactoring to unify verbosity handling across all subcommands. The 
esolve_verbosity merge pattern (max verbose, OR quiet) is sound and the configure_logging idempotency fix avoids handler accumulation. A few inline comments below.

DingmaomaoBJTU (commented):
Reviewed - see inline comments

DingmaomaoBJTU (commented):
Overall: Clean unification of verbosity handling across all subcommands. The 
esolve_verbosity() helper and factory-style erbosity_options() simplify the codebase well. A few suggestions below.

DingmaomaoBJTU (commented):
## Code Review: PR #793 - Unify verbosity declaration and stderr-only logger output

**Overall**: Clean unification of verbosity handling across all subcommands. The 
esolve_verbosity() helper and factory-style erbosity_options() simplify the codebase well. A few suggestions:

### src/winml/modelkit/utils/cli.py - 
esolve_verbosity() pairing
The 
esolve_verbosity() helper nicely centralizes the merge logic. One concern: if a future subcommand forgets to call 
esolve_verbosity() after applying @verbosity_options(), the top-level -v flag silently gets ignored. Consider documenting this pairing requirement in the erbosity_options() docstring, or having the decorator itself inject a callback that auto-resolves (e.g., via click.pass_context + eager default).

### src/winml/modelkit/commands/inspect.py - Early quiet merge
The early OR-merge of quiet before the full 
esolve_verbosity() call (line ~187) is pragmatic but makes the flow harder to follow. Consider extracting this into a tiny helper (e.g., _early_quiet(ctx, quiet)) or adding a one-line comment referencing why it can't wait for 
esolve_verbosity().

### src/winml/modelkit/utils/logging.py - Root logger level
Setting both the handler level AND root logger level to log_level (line ~71) means other handlers on the root logger will also be affected. Since you already set the handler level, consider only lowering 
oot.setLevel(logging.DEBUG) (the minimum) and letting individual handlers control their own threshold. This gives future handlers (file logging, telemetry) flexibility.

### src/winml/modelkit/serve/app.py - _log_capture_prior on app.state
Storing _log_capture_prior on pp.state (a private-prefix attr on a public namespace) works but is fragile. If middleware or tests poke pp.state, this could collide. Consider a module-level variable or a dedicated dataclass on pp.state.log_state.

### Repeated ool(verbose) casts across commands
The conversion ool(verbose) is scattered across export, perf, compile, sys, inspect. This suggests downstream APIs still expect a boolean. Consider adding a helper property (e.g., is_verbose(verbose) returning erbose > 0) in cli_utils rather than repeating the cast at every call site — cleaner and self-documenting.

DingmaomaoBJTU (commented):
## AI Code Review (PR #793)

Overall this is a solid refactor that unifies verbosity handling across all subcommands. The approach is clean — centralizing `verbosity_options()` as a decorator factory and `resolve_verbosity()` for merging parent/child flags.

### Key Observations:

**1. Potential key mismatch in `resolve_verbosity` (src/winml/modelkit/utils/cli.py)**
`resolve_verbosity` reads `ctx.obj.get('verbosity', 0)` but the top-level group callback parameter is named `verbose`. If `ctx.obj` stores under key `verbose` (not `verbosity`), this will silently always read 0 and top-level `-v` will be ignored. Please verify the key name matches what `cli()` stores in `ctx.ensure_object(dict)`.

**2. Root logger level side-effect (src/winml/modelkit/utils/logging.py)**
`configure_logging` sets `root.setLevel(log_level)` on the root logger. When a subcommand uses `-vv` (DEBUG), this enables DEBUG for all loggers in the process (urllib3, asyncio, etc.). Consider targeting `winml.modelkit` logger instead of root, or adding a filter to the handler.

**3. Early quiet merge in inspect.py**
The early `quiet = quiet or bool(ctx.obj and ctx.obj.get('quiet'))` before `resolve_verbosity()` is acknowledged as idempotent, but creates a subtle window where quiet is resolved but verbose is not. Consider moving both into a single `resolve_verbosity()` call.

**4. Silent config.json breakage in compile**
Previously compile read `verbose` from build-config JSON. That path is removed — users with `\"verbose\": true` in config.json will now have it silently ignored. Worth a changelog note.

**5. Good: serve/app.py ring handler decoupling** 
Solid fix for test isolation. One edge case worth documenting: if `create_app` is called twice, `_ensure_log_capture_level` returns `None` on the second call, so the first caller's restore becomes a no-op.

### Summary
The refactor is well-motivated and clean. Item #1 (key mismatch) is the most critical to verify — the rest are minor improvements.

DingmaomaoBJTU (commented):
Solid refactoring. Unifying verbosity handling across all subcommands is well-motivated and the resolve_verbosity() merge logic (max for -v, OR for -q) is clean. The idempotent configure_logging approach that preserves caplog handlers is thoughtful.

DingmaomaoBJTU (commented):
Thorough refactor that consolidates a lot of duplicated verbosity boilerplate. The 
esolve_verbosity + configure_logging pattern is clean. A few observations left as inline comments — mostly around edge cases in the serve app's log-level lifecycle and the bool/int type boundary for config-file values. No blocking issues.

DingmaomaoBJTU (commented):
Solid refactoring. Unifying verbosity into a single decorator + resolve_verbosity() helper eliminates the fragmented logging setup across subcommands. The idempotent configure_logging with handler marker is a nice touch for test isolation.

src/winml/modelkit/utils/logging.py
DingmaomaoBJTU: Setting 
oot.setLevel(log_level) on the root logger is a fairly aggressive change -- it could suppress third-party library logs that were previously visible. Consider scoping this to the 'winml.modelkit' logger instead, and only setting the root level if no package-scoped configuration suffices.

src/winml/modelkit/serve/app.py
```diff
@@ -126,6 +168,11 @@ def create_app(
     @asynccontextmanager
     async def lifespan(app: FastAPI):
         app.state.start_time = time.time()
+        # Raise the modelkit logger to INFO so the ring handler receives
+        # operational records during `winml serve`. Tests that build the app
```
DingmaomaoBJTU: Storing _log_capture_prior on pp.state is pragmatic, but the underscore prefix signals 'private' -- if any middleware or test introspects app.state, this could conflict. Minor: consider a more namespaced key like _winml_log_capture_prior.

DingmaomaoBJTU (commented):
Good refactoring overall — the decorator-factory pattern, the 
esolve_verbosity merge helper, and the idempotent handler management are solid. A few observations below (inline).

DingmaomaoBJTU (commented):
## Code Review: PR #793 — Unify verbosity declaration and stderr-only logger output

**Overall Assessment:** This is a well-structured refactoring that consolidates duplicated verbosity boilerplate across 10+ subcommands into a shared decorator + merge helper. The approach is sound and the test-isolation improvements in `configure_logging` are welcome.

A few items to address below (inline comments), mostly around root-logger side effects and a minor race in the serve handler lifecycle.

DingmaomaoBJTU (commented):
Good refactoring overall. A few items to consider:

DingmaomaoBJTU (commented):
Overall this is a well-structured refactoring that correctly unifies verbosity handling across all subcommands. The decorator-factory pattern, the resolve_verbosity merge logic, and the idempotent handler management in configure_logging are all sound. A few observations below - mostly minor, one potentially impactful (root logger scope).

DingmaomaoBJTU (commented):
Good refactoring - unifying verbosity across all subcommands removes a lot of inconsistency. A few inline suggestions below.

DingmaomaoBJTU (commented):
Thorough refactoring that unifies the verbosity story across all subcommands. The decorator-factory pattern and `resolve_verbosity` helper are clean. A few comments below.

DingmaomaoBJTU (commented):
Solid refactoring - this properly unifies the verbosity story across all 12+ subcommands. The `resolve_verbosity()` merge logic (max verbosity, OR quiet) is clean and the stderr-only logging fixes the stdout corruption issue. A few notes inline.

DingmaomaoBJTU (commented):
Solid refactoring that unifies the verbosity story across all subcommands. The `resolve_verbosity` max/OR merge is the right approach. A few observations inline.

DingmaomaoBJTU (commented):
Solid refactor that unifies verbosity across all subcommands. The decorator-to-factory change for `verbosity_options()` and the new `resolve_verbosity()` helper make it trivial to add new subcommands with consistent behavior. A few observations:

1. **resolve_verbosity edge case**: If a user passes both `-q` and `-vv` at the same level (e.g. `winml export -q -vv`), quiet wins unconditionally. This is documented, but consider emitting a brief stderr warning that `-q` is suppressing `-v` flags since it's likely unintentional.

2. **serve/app.py - _ensure_log_capture_level**: The restore logic in `_restore_log_capture_level` only restores if the current level is still `logging.INFO`. This is a nice defensive check, but if another component sets it to INFO *after* your set (coincidentally same value), you'll skip the restore. Low risk, just noting.

3. **configure_logging idempotency**: The handler replacement strategy (mark with `_HANDLER_MARKER`, remove old, add new) is clean. One subtlety: each call does `root.setLevel(log_level)` which could lower the root logger level if a different subsystem set it higher. Consider using `root.setLevel(min(root.level, log_level))` or only adjusting if the new level is more permissive — though for a CLI tool this is likely fine as-is.

4. **inspect.py early quiet merge**: The comment explains the duplication well. The early OR-merge before the banner followed by the full `resolve_verbosity()` later is idempotent. No issue here, just confirming the pattern reads correctly.

Overall: clean, well-tested refactor. Ship it. 🚢

DingmaomaoBJTU (commented):
Solid refactor - unifying verbosity across all subcommands is a much-needed cleanup. The resolve_verbosity() merge logic is well-thought-out (max verbose, OR quiet, quiet-wins-over-debug). A few observations below.

DingmaomaoBJTU (commented):
Solid unification — this eliminates a whole class of UX bugs where -v/-q only worked in one position. The idempotent handler logic in configure_logging and the careful test-isolation fix in serve/app.py are well thought through. A few minor observations below.

DingmaomaoBJTU (commented):
**Overall**: Solid refactoring that unifies a fragmented verbosity story across 12+ subcommands. The `resolve_verbosity` merge strategy (max verbose, OR quiet) is well-designed and the stderr-only logging avoids stdout corruption. A few items to address below.

DingmaomaoBJTU (commented):
Clean, thorough refactor. The resolve_verbosity merging strategy (max verbose, OR quiet) is sound and the idempotent handler management in logging.py nicely fixes the CliRunner test isolation issue. A few observations below.

DingmaomaoBJTU (commented):
Good refactoring - the unified verbosity contract is a clear win. A few observations:

1. **Breaking change** (`utils/cli.py`): Converting `verbosity_options` from a bare decorator to a factory (`verbosity_options()`) is cleaner and matches Click convention. However, any external scripts calling `@cli_utils.verbosity_options` (without parens) will break. The internal callsites are all fixed, but worth noting in the PR description for downstream consumers.

2. **Root logger scope** (`utils/logging.py`): `root.setLevel(log_level)` affects **all** loggers in the process. Fine for the CLI (single-purpose process), but could expose noisy third-party logs (urllib3, transformers) at DEBUG in the serve module or embedded usage. Consider scoping to `logging.getLogger('winml.modelkit')` instead.

3. **Inspect double-quiet** (`commands/inspect.py`): The pre-merge `quiet = quiet or bool(ctx.obj.get('quiet'))` before `resolve_verbosity` is pragmatic but smelly. Consider having `resolve_verbosity` return an early-quiet flag, or extract an `is_quiet(ctx)` helper for the banner check.

4. **Serve module** (`serve/app.py`): Good separation - deferring `_attach_log_handler()` to `_register_routes` prevents test pollution. The `_ensure/_restore_log_capture_level` pair is clean.

5. **Minor**: In `compile.py` line 176, `config.verbose = bool(verbose)` coerces the count to bool. If any downstream code checks verbosity level (>1 for extra detail), this loses information. Same pattern in `export.py` and `perf.py`. Consider keeping the int or documenting that config.verbose is always bool.

DingmaomaoBJTU (commented):
Well-motivated refactoring — the unified verbosity contract and stderr-only output is a clear improvement. The serve/app.py changes for test isolation are thoughtful. A few inline observations.

DingmaomaoBJTU (commented):
Solid refactor — unifying verbosity across 12+ commands eliminates a genuine usability bug. The resolve_verbosity helper, idempotent configure_logging, and test-safe serve/app.py changes are all well thought out. A few inline comments below on potential edge cases.

src/winml/modelkit/commands/compile.py
```diff
@@ -157,10 +151,8 @@ def compile(
             embed = cc["embed_context"]
```
DingmaomaoBJTU, line 151: The removed lines here mean that a `verbose` key in the build config JSON is no longer honored for the compile command. Previously `if not cli_utils.is_cli_provided(ctx, 'verbose') and 'verbose' in cc: verbose = cc['verbose']` allowed config-file-driven verbosity. Was this intentional? If so, worth a note in the PR description that config-file verbosity is deprecated in favor of CLI flags only.

src/winml/modelkit/commands/inspect.py
```diff
@@ -198,13 +201,9 @@ def inspect(
     from ..inspect import InspectError, ModelNotFoundError, NetworkError
```
DingmaomaoBJTU, line 221: Nit: The early OR-merge of `quiet` before the banner (line ~191) followed by a second `resolve_verbosity()` merge (line ~201) is a bit subtle. The comment explains it well, but consider extracting the early quiet check into a tiny helper like `_is_quiet(ctx, quiet)` to make the intent even clearer at the call site and avoid readers wondering if it's a bug.

src/winml/modelkit/serve/app.py
```diff
@@ -126,6 +168,11 @@ def create_app(
     @asynccontextmanager
```
DingmaomaoBJTU, line 168: Storing state on `app.state._log_capture_prior` with a leading underscore is fine for internal use, but note that if a test creates the app with a custom lifespan that skips this branch, the `_restore_log_capture_level(getattr(app.state, '_log_capture_prior', None))` in shutdown gracefully handles the missing attr via `getattr(..., None)` — nice defensive coding.

src/winml/modelkit/utils/logging.py
```diff
@@ -49,13 +63,21 @@ def configure_logging(
...
-        datefmt="%Y-%m-%dT%H:%M:%S",
-        stream=sys.stderr,
-        force=True,
-    )
+    root = logging.getLogger()
+    # Drop any prior WinML handler and install a fresh one bound to the
+    # *current* ``sys.stderr``. Click's ``CliRunner.invoke()`` swaps the
+    # process stderr for each test, so a cached handler from an earlier
+    # invocation would write to a stream the test no longer captures.
+    # We leave non-WinML handlers (notably pytest's caplog handler) alone.
```
DingmaomaoBJTU, line 71: Setting `root.setLevel(log_level)` here affects ALL loggers in the process (not just winml). If a user runs `winml export -vv`, this sets the root logger to DEBUG, which may cause noisy output from third-party libraries (transformers, onnxruntime, urllib3, etc.). Consider scoping this to `logging.getLogger('winml')` or `logging.getLogger('winml.modelkit')` instead, and only setting root level to DEBUG when truly needed.

src/winml/modelkit/utils/cli.py
```diff
@@ -191,26 +191,59 @@ def verbosity_options(f):
...
-    f = click.option(
-        "--verbose",
-        "-v",
-        count=True,
-        help="Increase verbosity (-v=INFO, -vv=DEBUG)",
-    )(f)
-    return f  # noqa: RET504
+    if ctx.obj:
+        verbose = max(verbose, int(ctx.obj.get("verbosity", 0)))
+        # ``debug`` is the historical backward-compat alias for ``-vv``; keep
```
DingmaomaoBJTU, line 251: **Nit:** The docstring says `-q` always wins over verbosity, but `resolve_verbosity` doesn't enforce this - it returns both values independently and relies on `configure_logging` to honor the precedence. Consider adding an inline comment or defensively zeroing `verbose` when `quiet=True` to make the contract self-contained.

src/winml/modelkit/utils/logging.py
```diff
@@ -49,13 +63,21 @@ def configure_logging(
...
+    for h in list(root.handlers):
+        if getattr(h, _HANDLER_MARKER, False):
+            root.removeHandler(h)
```
DingmaomaoBJTU, line 74: Using `setattr(own_handler, _HANDLER_MARKER, True)` is pragmatic but invisible to type checkers. A lightweight alternative: subclass `logging.StreamHandler` (e.g., `class _WinMLHandler(logging.StreamHandler): pass`) and filter by `isinstance` - same effect, zero dynamic attrs, mypy-friendly.

src/winml/modelkit/serve/app.py
```diff
@@ -92,9 +92,51 @@ def since(self, after_seq: int) -> list[dict]:
...
+def _restore_log_capture_level(prior: int | None) -> None:
+    if prior is None:
+        return
+    pkg_logger = logging.getLogger("winml.modelkit")
+    # Only restore if the value still matches what we set. If anyone else
+    # (middleware, debug hook, a nested ``configure_logging`` from a CLI
+    # invocation routed through /v1/cli) has since changed it, defer to them
+    # rather than silently rolling their change back.
+    if pkg_logger.level == logging.INFO:
+        pkg_logger.setLevel(prior)
```
DingmaomaoBJTU, line 138: Nice defensive guard - only restoring if the current level still matches what we set. Worth a one-line comment noting: if middleware lowers the level below INFO, the check correctly skips restore (deferring to whoever changed it). Makes the intent explicit for future readers.

DingmaomaoBJTU (commented):
Good refactoring! The unification of verbosity across all subcommands is a solid improvement. A few observations below.

src/winml/modelkit/serve/app.py
```diff
@@ -92,9 +92,51 @@ def since(self, after_seq: int) -> list[dict]:
 
 _log_handler = _RingHandler()
 _log_handler.setFormatter(logging.Formatter("%(message)s"))
-# Attach to modelkit root logger so all sub-loggers feed into the ring
-logging.getLogger("winml.modelkit").addHandler(_log_handler)
-logging.getLogger("winml.modelkit").setLevel(logging.INFO)
+# Bound the records the ring receives independently of the package logger's
```
xieofxie, line 95: Explanation from CC:
```
What main did (the problem)

  On main, the module did this at import time:

  logging.getLogger("winml.modelkit").addHandler(_log_handler)
  logging.getLogger("winml.modelkit").setLevel(logging.INFO)   # ← global side effect

  The second line is the troublemaker. Raising the winml.modelkit package logger to INFO is a process-wide mutation that fires the moment
  anything imports app.py. Under pytest, when the serve test module gets collected, every other test in that session inherits an
  INFO-level package logger — so any test that asserts on DEBUG-level capture (e.g. caplog) silently sees nothing, because DEBUG is now
  being filtered out. That cross-test contamination is exactly what this hualxie/unify_verbosity branch is about. (The original commit
  message is literally "I don't understand" — it was the author debugging that surprise.)

  What the branch does (the fix)

  It splits one blunt import-time side effect into three scoped pieces:

  1. _log_handler.setLevel(logging.INFO) — bound what the ring buffer accepts at the handler, not the logger. The ring still only collects
  INFO+, but without touching any logger's level.
  2. _attach_log_handler() (called from _register_routes) — idempotently attaches the handler, and only the handler. Tests that wire up
  routes get the ring without any level change.
  3. _ensure_log_capture_level() / _restore_log_capture_level() — the actual INFO level bump now happens only inside the lifespan startup
  hook — i.e. during a real winml serve run — and is restored on shutdown. Tests that build the app with a mock lifespan never hit this
  branch, so the global level stays untouched for them.

  The restore is also defensive: it only rolls back if the level still equals the INFO it set, so it won't clobber a concurrent change
  made by middleware, a debug hook, or a nested configure_logging.

  Net intent

  Preserve the runtime behavior of winml serve (operational logs flow into the /v1/logs ring at INFO), while removing the import-time
  global logger-level mutation that was leaking into and breaking unrelated tests' DEBUG capture. The second commit (address review
  feedback) just clarified the intent comments on those three points.

  One thing worth noting: this only changes the level handling — the handler attach moved from import-time to _register_routes. So make
  sure every app-construction path (including any that bypasses _register_routes) still attaches the ring, otherwise /v1/logs would come
  back empty in those paths. Want me to check that all code paths into the app go through _register_routes?
```
```
Confirmed. I traced every app-construction path. Here's the result:

  There are two distinct apps

  ┌─────────────────────────────┬─────────────────────────────────────────────────────────┬───────────────────────────────────────┐
  │                             │                  Phase 0 — cli_api.py                   │          Phase 1/3 — app.py           │         ├─────────────────────────────┼─────────────────────────────────────────────────────────┼───────────────────────────────────────┤
  │ When used                   │ winml serve with no model and no --multi (serve.py:135) │ model given or --multi (serve.py:148) │
  ├─────────────────────────────┼─────────────────────────────────────────────────────────┼───────────────────────────────────────┤
  │ Has /v1/logs?               │ No                                                      │ Yes (app.py:712)                      │
  ├─────────────────────────────┼─────────────────────────────────────────────────────────┼───────────────────────────────────────┤
  │ Uses the ring _log_handler? │ No — never, not even on main                            │ Yes                                   │
  └─────────────────────────────┴─────────────────────────────────────────────────────────┴───────────────────────────────────────┘

  The Phase 0 app has no /v1/logs endpoint and has never referenced the ring handler — git grep for
  v1/logs|_log_handler|RingHandler|addHandler in cli_api.py returns nothing. So it neither needs nor lost anything.

  The Phase 1 app is safe

  The endpoint and the attach are now co-located in the same function, so they cannot diverge:

  - /v1/logs (app.py:712) is registered inside _register_routes.
  - _register_routes calls _attach_log_handler() as its first line (app.py:234).
  - _register_routes has exactly one call site — create_app at app.py:229.
  - create_app has exactly one call site — serve.py:155.

  So any app that exposes /v1/logs is guaranteed to have attached the ring handler in the same call. There's no path where the endpoint
  exists but the handler is missing — /v1/logs can't come back empty due to the refactor.

  This is actually more robust than main: on main, the attach was an import-time side effect decoupled from route registration; now it's
  bound to the route that consumes it.

  Two nice properties worth noting

  1. _attach_log_handler is idempotent (if _log_handler not in pkg_logger.handlers), so constructing multiple Phase-1 apps in one process
  — common in tests — won't double-attach and duplicate every log line. The main version was also effectively idempotent only by luck
  (re-running the module-level addHandler would have duplicated); the explicit guard is an improvement.
  2. Level timing is fine: the INFO bump happens in lifespan startup, which completes before uvicorn serves any request, so by the time a
  client can poll /v1/logs the package logger is already at INFO and records are flowing.

  Net: the refactor is sound — no path loses log capture, and it removed the import-time global level mutation that was the actual bug.
  Nothing further to fix here.
```

src/winml/modelkit/commands/compile.py
xieofxie: Create an issue as a new feature https://github.com/microsoft/winml-cli/issues/799

  ▎ Intentional. compile was the only command that honored a config-file verbose key, and it was never a real WinMLBuildConfig field
  ▎ (absent from to_dict/from_dict, so winml config never wrote it) — just an undocumented top-level raw-JSON peek. This branch unifies
  ▎ verbosity on the -v/-q CLI flags across all commands, so compile now matches everything else. Added a note to the PR description.

src/winml/modelkit/utils/logging.py
```diff
@@ -19,13 +19,22 @@
     Quiet:   level = ERROR (40)
 
 All log output goes to stderr so stdout stays clean for structured data
-(JSON, compact output, piped commands).
+(JSON, compact output, piped commands). Format:
```
xieofxie, line 22: explan by CC:
```
 Here's what changed in logging.py and why. There are three distinct changes, but they all serve one goal: making configure_logging safe
  to call multiple times and safe under pytest.

  1. New log format

  - format="[%(asctime)s] %(levelname)s: %(message)s"
  - datefmt="%Y-%m-%dT%H:%M:%S"
  + "[%(asctime)s %(levelname)-7s %(name)s] %(message)s"
  + datefmt="%H:%M:%S"

  Three cosmetic improvements, hoisted into module constants (_LOG_FORMAT, _DATE_FORMAT):
  - Adds %(name)s — the logger name (e.g. winml.modelkit.export), so you can tell which subsystem emitted a line.
  - %(levelname)-7s — left-pads the level to 7 chars so messages align in a column (INFO   , WARNING, DEBUG  ).
  - Shorter timestamp — 14:32:11 instead of a full ISO datetime, since CLI runs are interactive and the date is just noise.

  2. The real change: hand-managed handler instead of basicConfig(force=True)

  This is the substantive one. Before:

  logging.basicConfig(..., stream=sys.stderr, force=True)

  After: manually walk the root handlers, remove only WinML's own handler, and add a fresh one.

  force=True is a blunt instrument — it removes and closes every existing handler on the root logger. That caused two problems, both of
  which this branch is built to fix:

  (a) It destroyed pytest's caplog handler. Pytest attaches its own handler to the root logger to capture log records in tests. force=True
  would rip it out, so any test asserting on captured logs would see nothing. The new code only removes handlers tagged with the
  _HANDLER_MARKER attribute (_winml_cli_handler) — i.e. handlers this function itself installed on a prior call — and leaves caplog (and
  any other foreign handler) untouched. The comment calls this out explicitly. This is the same class of test-contamination problem you
  saw in the app.py change.

  (b) Stale stderr binding under CliRunner. Click's CliRunner.invoke() swaps sys.stderr for a fresh capture buffer on every invocation. A
  handler cached from an earlier call would still point at the previous test's (now-dead) stream, so its output would vanish. By removing
  its own old handler and binding a new StreamHandler(sys.stderr) to the current stderr on every call, each invocation's logs land in the
  stream that invocation is actually capturing.

  The marker pattern is the key trick:
  setattr(own_handler, _HANDLER_MARKER, True)   # tag it as ours
  ...
  if getattr(h, _HANDLER_MARKER, False):         # only remove ours
      root.removeHandler(h)

  3. Idempotency — why it needs to be called repeatedly now

  The new docstring says it's idempotent, and that's directly tied to the branch's resolve_verbosity design. Verbosity can be supplied at
  two positions (winml -v compile … and winml compile -v …), and resolve_verbosity merges them inside the subcommand. So the subcommand
  calls configure_logging after the top-level group may have already called it. Without the "remove-only-my-handler, then re-add" logic,
  repeated calls would either stack duplicate handlers (every line printed twice, three times…) or — with the old force=True — keep nuking
  caplog. The new version converges: first call installs the handler, subsequent calls just swap in a freshly-bound handler at the
  updated level.

  What stayed the same

  - The verbosity→level math (WARNING - verbosity*10, clamped to DEBUG, ERROR when quiet) is unchanged.
  - The verbose: bool backward-compat shim (lines 59–61) is unchanged — old callers passing verbose=True still map to verbosity=1.
  - Output still goes to stderr.

  Net: same levels and same stderr destination, but the handler is now installed surgically so it (a) survives being called multiple times
  per process, (b) re-binds to the live stderr each time, and (c) never clobbers pytest's log capture.
```

src/winml/modelkit/utils/logging.py
xieofxie: Thanks for flagging this. To clarify: setting root.setLevel(log_level) here isn't a new behavior introduced by this change — it
  preserves exactly what main did.

  Previously this function used:

  logging.basicConfig(level=log_level, ..., force=True)

  and logging.basicConfig(level=...) calls root.setLevel(level) under the hood. So on main, winml export -vv already set the root logger
  to DEBUG, and third-party libraries (transformers, onnxruntime, urllib3, …) — which are NOTSET and inherit the root level — were already
  affected by -v/-vv. This refactor only swapped basicConfig(force=True) for explicit handler management (so we stop clobbering pytest's
  caplog handler and re-bind to the live sys.stderr per CliRunner invocation); the root-level semantics are intentionally unchanged.

  So there's no regression here — the third-party verbosity behavior is identical to before.

  That said, your point about debug noise from third-party libs is a fair one on its own merits. If we want to improve it, I'd treat it as
  a separate, deliberate change rather than folding it into this refactor — e.g. scope -v/-vv to winml.modelkit and only lower the root
  level at -vvv (pip-style), so deep third-party debugging is still reachable when truly needed. Happy to open a follow-up for that if you
  think it's worth it.

xieofxie: https://github.com/microsoft/winml-cli/issues/802

src/winml/modelkit/commands/inspect.py
xieofxie: ▎ Good catch — agreed the double-merge reads like a possible bug. Rather than wrap it in a _is_quiet helper (which keeps two merge
  ▎ sites), I removed the duplication entirely: resolve_verbosity now runs once before the banner, so the merged --quiet has a single
  ▎ source of truth and the early OR-merge + its NOTE comment are gone. Banner suppression for both winml --quiet inspect … and winml
  ▎ inspect -q still works (covered by the inspect CLI tests).

src/winml/modelkit/commands/sys.py
```diff
@@ -720,33 +717,24 @@ def sysinfo(
...
-    # Restore logger state on exit so tests using caplog are not affected.
-    #
-    # For --format json, send log records to stderr so DEBUG/WARNING lines do
-    # not corrupt the JSON payload on stdout (verbose+json was unparseable).
-    from rich.console import Console as _RichConsole
-    from rich.logging import RichHandler
+    # Merge top-level -v/-q with subcommand-level flags so either position works.
+    verbose, quiet = cli_utils.resolve_verbosity(ctx, verbose, quiet)
+
+    # Standard verbosity contract: stderr-only logs in the shared format.
```
DingmaomaoBJTU, line 723: The `finally` block (lines 823-826) saves/restores `pkg_logger` (`winml.modelkit`), but `configure_logging` now adds its handler to the **root** logger and calls `root.setLevel()`, not `pkg_logger`. After `sysinfo` exits, the root logger's handler and level are not cleaned up, so the test-isolation guarantee stated in the comment above is no longer effective. Either update the snapshot to cover `logging.getLogger()` (root) instead, or rely on `configure_logging`'s own idempotent handler-replacement to keep things clean and remove this save/restore entirely.

src/winml/modelkit/commands/compile.py
```diff
@@ -157,10 +151,8 @@ def compile(
...
         if not cli_utils.is_cli_provided(ctx, "validate") and "validate" in cc:
             validate = cc["validate"]
-        if not cli_utils.is_cli_provided(ctx, "verbose") and "verbose" in cc:
-            verbose = cc["verbose"]
 
-    configure_logging(verbose=verbose)
+    configure_logging(verbosity=verbose, quiet=quiet)
```
DingmaomaoBJTU, line 164: The old code applied `verbose` from the build config file when the user didn't pass `-v` on the CLI (`if not is_cli_provided(ctx, 'verbose') and 'verbose' in cc: verbose = cc['verbose']`). That block was removed here. Users who rely on `verbose: true` in their build config JSON/YAML to get debug output will silently lose it — the field is now loaded into `raw_cfg` but never read back for verbosity. Is this intentional (config-file verbosity being deprecated in favor of CLI-only flags)? If so, it's worth a changelog note.

src/winml/modelkit/utils/logging.py
DingmaomaoBJTU, line 71: Both `own_handler.setLevel(log_level)` and `root.setLevel(log_level)` are set to the same value. The root level already gates every record before it reaches any handler, so `own_handler.setLevel` is redundant here. If the intent is to let a future second root-handler operate at a different level, consider setting `own_handler.setLevel(logging.DEBUG)` and leaving `root.setLevel` as the sole gate — that's the conventional pattern and avoids the double-filter surprise.

xieofxie: ▎ Agreed — fixed. Dropped own_handler.setLevel(log_level) and left the handler at NOTSET, so the root level is the single gate. This
  ▎ also restores the exact behavior of the previous logging.basicConfig, which never set a handler level. Added a comment noting root is
  ▎ the sole gate so the redundant filter doesn't get reintroduced.

src/winml/modelkit/commands/sys.py
xieofxie: ▎ You're right — that save/restore was a leftover from the old approach where sys installed its own RichHandler on winml.modelkit with
  ▎ propagate=False. Now that it uses the shared configure_logging (which targets root, idempotently), the pkg_logger snapshot guards
  ▎ nothing it changes. I removed the save/restore entirely (option b) and rely on configure_logging's idempotent handler replacement —
  ▎ same contract every other command uses for test isolation. Verified with the sysinfo + CLI test suites (106 passing).

src/winml/modelkit/commands/compile.py
xieofxie: ▎ Good catch — restored. winml compile again honors verbose from the build config, but only when no -v/-q was passed on either CLI
  ▎ position (CLI always wins). I had to adapt the guard: since resolve_verbosity now merges the top-level and subcommand -v, the old
  ▎ is_cli_provided(ctx, "verbose") check alone would let the config override a top-level winml -v compile …, so the condition is now the
  ▎ merged verbose == 0 and not quiet. int(cc["verbose"]) keeps the old semantics (true→INFO, 2→DEBUG).
  ▎
  ▎ On your broader point: config-file verbosity is currently only wired up for compile. Rather than expand it ad hoc here, I filed #799
  ▎ to track supporting it consistently across all commands: https://github.com/microsoft/winml-cli/issues/799

DingmaomaoBJTU (approved):

Issue #800, opened 2026-06-02
[task] Unify modality-aware task detection across CLI commands (single source of truth)


timenick:
**Component**: Loader / task · CLI commands
**Type**: Refactor / correctness

## Problem

The same model yields different `task` strings across commands because task *detection* is implemented three different ways and resolves to Optimum's lossy canonical name. Concretely, for an image-embedding model:

```bash
winml inspect -m facebook/dinov2-small        # -> feature-extraction  (lossy, verified on main)
winml inspect -m facebook/dinov2-small --task image-feature-extraction  # -> image-feature-extraction
```

Auto-detection collapses `image-feature-extraction` -> `feature-extraction` because detection goes through Optimum's `TasksManager`, whose vocabulary is modality-blind. Downstream consumers (calibration dataset, evaluator) then pick a *text* asset for an *image* model.

## Root cause

- Detection has 3 separate implementations: `loader/task.py::_detect_task_from_config`, `inspect/resolver.py::detect_task`, `eval/evaluate.py::_resolve_task` (+ reverse reconstruction `to_hf_pipeline_task`).
- All resolve to Optimum-canonical (`feature-extraction`), losing modality.
- Two collapse helpers (`normalize_task`, `export/io.py::map_task_synonym`) are both lossy; there is no single modality-aware detector.

## Scope

**In scope (7 surfaces):** `inspect`, `config`, `export`, `eval`, `perf`, `quantize`, and `build` (config-driven `loader.task`).

**Out of scope:** `run` / `serve` (runtime/inference pipeline), `catalog` (data filter), and **all validation** (`KNOWN_TASKS` / `validate_task` / `--list-tasks`).

## Design

- **New** `loader/task.py::detect_task(config) -> (WinMLTask, Source)` — single, offline, config-only detector. Modality upgrade (vision: top-level `image_size`/`patch_size` -> `image-feature-extraction`) applied to the **returned** task only; internal model-class resolution keeps the Optimum-canonical task.
- **Rename/relocate** `map_task_synonym` -> `loader/task.py::to_optimum_task` — the single Optimum-boundary collapse, called only at `TasksManager.*` sites.
- `normalize_task` is **left untouched** (keeps `build`'s model-aware validation byte-identical). Cross-command consistency comes from *surfacing the preserved task*, not from changing any collapse helper.

## PR plan

- [ ] **PR1** — relocate `map_task_synonym` -> `to_optimum_task` (+ shim). Zero behavior change.
[...]

PR #801, opened 2026-06-02
refactor(task): relocate map_task_synonym to loader.task as to_optimum_task

timenick:
## What

PR1 of #800. Relocate `map_task_synonym` -> `loader/task.py::to_optimum_task` to establish a single WinML->Optimum task-collapse boundary.

## Changes

- `loader/task.py`: add `to_optimum_task` + `TASK_SYNONYM_EXTENSIONS` (moved from `export/io.py`); exported via `loader/__init__.py`.
- `export/io.py`: local implementation removed; `map_task_synonym` kept as a backward-compatible alias (`= to_optimum_task`); internal use repointed.
- Optimum-boundary call sites repointed to `to_optimum_task`: `commands/inspect.py`, `export/htp/exporter.py`, `inspect/resolver.py`.
- `commands/build.py`: `TASK_SYNONYM_EXTENSIONS` now imported from `loader`.
- New `tests/unit/loader/test_task_boundary.py` pins the collapse contract.

## Behavior

No behavior change. `map_task_synonym` stays importable from `export.io`; the collapse semantics (`image-feature-extraction` -> `feature-extraction`, WinML extensions preserved) are byte-identical. Existing synonym and #777/#782 regression tests stay green.

Sets up PR2 (#800), which adds the modality-aware `detect_task` and relies on this single collapse boundary.

### DIFF
--- a/src/winml/modelkit/export/htp/exporter.py
+++ b/src/winml/modelkit/export/htp/exporter.py
@@ -470,14 +470,14 @@ class HTPExporter:
             logger.debug("Model has no config.model_type; skipping Optimum patcher.")
             return contextlib.nullcontext()
 
-        # TasksManager expects normalized task names
-        from ..io import map_task_synonym
+        # TasksManager expects Optimum-canonical task names
+        from ...loader import to_optimum_task
 
         try:
             cfg_cls = TasksManager.get_exporter_config_constructor(
                 "onnx",
                 model_type=model_type,
-                task=map_task_synonym(task),
+                task=to_optimum_task(task),
                 library_name="transformers",
             )
             return cfg_cls(model_config).patch_model_for_export(model)
--- a/src/winml/modelkit/commands/build.py
+++ b/src/winml/modelkit/commands/build.py
@@ -273,8 +273,8 @@ def _validate_task_supported_for_model(
     Raises:
         ValueError: If the task is not supported for the model architecture.
     """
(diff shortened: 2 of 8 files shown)


src/winml/modelkit/export/htp/exporter.py
```diff
@@ -470,14 +470,14 @@ def _get_optimum_patcher(model: nn.Module, task: str | None) -> Any:
             logger.debug("Model has no config.model_type; skipping Optimum patcher.")
             return contextlib.nullcontext()
 
-        # TasksManager expects normalized task names
-        from ..io import map_task_synonym
+        # TasksManager expects Optimum-canonical task names
+        from ...loader import to_optimum_task
```
zhenchaoni, line 474: Code is good to me. Please also double check that the time to import loader is similar to import io, so it will not make the code slower.

zhenchaoni (approved):

src/winml/modelkit/export/htp/exporter.py
timenick: Verified — importing `loader` is **lighter** than importing `export.io`, so it cannot slow this path down:

- `import winml.modelkit.export.io` → ~7.5s (warm), and **eagerly** loads `optimum.exporters.tasks`.
- `import winml.modelkit.loader` → ~100ms, and does **not** load `optimum` (loader uses lazy optimum imports).

`to_optimum_task` keeps the same lazy `TasksManager` import inside the function, so runtime is unchanged too. Net: neutral-to-faster.

Generated with Claude Code

PR #805, opened 2026-06-03
Implement tensor similarity evaluator

### LINKED ISSUE #804: feat: support tensor level comparison for model evaluation
---

### Problem

`winml eval` today reports task-level metrics (top-1, mIoU, BLEU, ...) on a labeled dataset. That tells us *whether* an optimized model still works, but not *how much* the optimize / quantize / compile passes perturbed the raw output tensors relative to the source HF PyTorch model.

When investigating a build-pipeline regression today we have to:

- Stand up the HF reference by hand.
- Feed both models the same inputs by hand.
- Compute SQNR / PSNR / cosine / MSE / max-abs-diff on the outputs by hand.
- Re-do all of the above per output tensor, per sample, per config we want to sweep.

There is no first-class, label-free, dataset-free signal in `winml eval` for "did this quantization config drift the tensors?" — which is the question we ask most often when picking quantization configs or chasing accuracy regressions.

### Proposal

Add a new evaluation mode to `winml eval`:

```bash
winml eval --mode compare -m <onnx_or_hf_id> --task <task> --samples 100
```

`--mode compare` should:

1. Load the ONNX candidate (existing `WinMLPreTrainedModel` path).
2. Load the matching HF PyTorch reference on CPU/fp32, resolved generically from the task (no per-task mapping).
3. Draw inputs from `RandomDataset` over the candidate's ONNX I/O spec — no labeled dataset, no HF `pipeline`, no preprocessor in the loop, so any divergence reflects the build pipeline only.
4. Run both backends on identical inputs per sample.
5. Report per-output-tensor parity metrics: **SQNR (dB), PSNR (dB), cosine similarity, MSE, max absolute diff**, aggregated as **mean / std / min / max** across samples.

Default mode stays `onnx` (today's behavior); compare is fully opt-in.

### Acceptance criteria

- `winml eval --mode compare -m microsoft/resnet-50 --task image-classification --precision fp16 --samples 100` runs end-to-end and writes a `metrics` block with the 5 metrics × 4 stats (= 20 keys) per output tensor.
- Per-sample math matches the team-wide `eval_tensors` reference library bit-for-bit on the same `.npy` pair.
- Architecture-agnostic: no model-specific names, layer patterns, or hardcoded outputs in metric or evaluator code.
- Composite models (e.g. BLIP) fail fast with a clear message instructing the user to run `compare` per sub-component.
- ONNX/HF output-name divergence is handled gracefully (compute on the intersection, warn on drift) rather than failing.
- Output integrates with the existing eval report renderer — no custom renderer required for this issue.
- Unit tests cover metric numerics + evaluator dispatch + composite guard; one e2e test exercises the full CLI on a small real model.

### Out of scope (separate issues)

- `--mode hf` (run the HF `pipeline` on a labeled dataset as the reference for task-level metric parity).
- A dedicated compare-mode renderer / richer per-output presentation.
- Multi-component (composite) compare flow.

### Notes

Per-sample metric math is fixed and well-known; the design space is mostly around the evaluator wiring (model loading, input generation, output pairing) and the output shape that the eval renderer consumes.

zhenchaoni:
Fixes #804

### Summary

Adds a new `compare` mode to `winml eval` that compares an ONNX candidate against its HF PyTorch reference on identical random inputs and reports per-output tensor-parity metrics (SQNR, PSNR, cosine similarity, MSE, max absolute diff). This isolates divergence introduced by the build pipeline (optimize / quantize / compile) from data- or pipeline-related differences — there is no labeled dataset, no HF `pipeline`, and no preprocessor in the loop.

### Motivation

Task-level metrics (top-1, mIoU, BLEU, ...) tell us *whether* an optimized model still works, but not *how much* the optimize/quantize/compile passes perturbed the raw tensors. Tensor-similarity gives a fast, label-free, dataset-free signal for build-pipeline regressions and for picking quantization configs.

### Usage

```bash
winml eval --mode compare -m microsoft/resnet-50 --task image-classification --precision fp16 --samples 100
```

### What's new

- **`winml eval --mode {onnx,compare}`** — new Click option on `winml eval`. `onnx` (default) is the existing dataset-driven flow; `compare` activates the new evaluator.
- **`TensorSimilarityEvaluator`** (tensor_similarity_evaluator.py) — loads the HF reference on CPU/fp32 via `resolve_task_and_model_class`, draws inputs from `RandomDataset` over the candidate's ONNX I/O spec, runs both backends per sample, and aggregates per-output metrics.
- **`TensorSimilarityMetric`** (tensor_similarity.py) — stateful `update` / `compute` / `reset` metric mirroring `MeanIoUMetric`. Per-sample math is bit-equivalent to the team-wide `eval_tensors` reference library on the same `.npy` pair.
- **Dispatch** — `evaluate.py` registers `"compare-tensor"` and `get_evaluator_class` routes to it when `config.mode == "compare"`; compare mode bypasses default-dataset resolution and the dataset section of `print_config`.
- **Config** — `WinMLEvaluationConfig.mode: str = "onnx"`; `to_dict` only emits `mode` when non-default.

### Output shape

`compute()` returns display-ready flat dict so the existing generic eval report renders without a custom renderer:

```python
{
    f"{metric}_{stat}": {output_name: float},  # 5 metrics × 4 stats = 20 keys
    ...
}
```

Stats are `mean / std / min / max`. The renderer prints one row per `{metric}_{stat}` with `output_name=value` cells joined across outputs.

### Notable design choices

- **Output-name overlap, not strict equality.** ONNX and HF output sets can differ (HF often exposes auxiliary tensors). We compute on the intersection and warn on divergence rather than failing.
- **Composite-model guard.** Multi-component models (e.g. BLIP) raise a `TypeError` with guidance to run `compare` per sub-component — there is no canonical "one HF reference" for the composite.
- **int dtype normalization.** Narrow int tensors are upcast to int64 before inference so HF embeddings accept them; `WinMLSession` down-casts to the ORT graph's declared dtype on its side. The same input dict feeds both backends.
- **Architecture-agnostic.** No model-specific names, layer patterns, or hardcoded outputs anywhere in metric or evaluator code.

### Tests

- test_tensor_similarity_metric.py — 10 unit tests for the metric (numerics, identity, stat shape, reset, empty-state error).
- test_tensor_similarity_evaluator.py — 4 unit tests (composite-model guard, output-name overlap, dispatch).
- test_eval.py — `get_evaluator_class` updated to take `WinMLEvaluationConfig`; `"compare-tensor"` exempted from the task-schema set check.
- `tests/e2e/test_eval_e2e.py::test_compare_mode_image_classification` — full CLI run on `microsoft/resnet-50` fp16: asserts the 20-key flat shape, per-output cosine bounds `[-1, 1]`, and (QNN host only) `cosine_similarity_mean >= 0.95`.

53 unit tests pass; ruff clean.

### Out of scope (follow-ups)

- `--mode hf` (run the HF pipeline on a labeled dataset as the reference for task-level metrics) — placeholder removed from this PR; will land as a separate change.
- A custom renderer for compare mode (currently uses the generic table).

### DIFF
--- a/src/winml/modelkit/commands/eval.py
+++ b/src/winml/modelkit/commands/eval.py
@@ -16,7 +16,7 @@ import click
 from rich.console import Console
 
 from ..utils import cli as cli_utils
-from ..utils.eval_utils import TASK_SCHEMAS, TaskSchema
+from ..utils.eval_utils import EVAL_MODES, TASK_SCHEMAS, EvalMode, TaskSchema
 
 
 if TYPE_CHECKING:
@@ -149,7 +149,7 @@ logger = logging.getLogger(__name__)
 )
 @click.option(
     "--mode",
-    type=click.Choice(["onnx", "compare"], case_sensitive=False),
+    type=click.Choice(EVAL_MODES, case_sensitive=False),
     default="onnx",
     show_default=True,
     help=(
@@ -182,7 +182,7 @@ def eval(
     dataset_script: str | None,
     trust_remote_code: bool,
     show_schema: bool,
-    mode: str,
+    mode: EvalMode,
     config_file: Path | None,
 ) -> None:
     r"""Evaluate a model for a task.
--- a/src/winml/modelkit/commands/eval.py
+++ b/src/winml/modelkit/commands/eval.py
@@ -147,6 +147,18 @@ logger = logging.getLogger(__name__)
     default=False,
     help="Print expected dataset schema for the given --task and exit.",
 )
+@click.option(
+    "--mode",
+    type=click.Choice(["onnx", "compare"], case_sensitive=False),
+    default="onnx",
+    show_default=True,
+    help=(
+        "Evaluation mode. "
+        "'onnx' (default): evaluate the ONNX candidate on the dataset. "
+        "'compare': compare ONNX vs HF reference output tensors on identical "
+        "random inputs and report tensor-similarity metrics per output tensor."
+    ),
+)
 @cli_utils.build_config_option()
 @click.pass_context
 def eval(
@@ -170,6 +182,7 @@ def eval(
     dataset_script: str | None,
     trust_remote_code: bool,
     show_schema: bool,
+    mode: str,
     config_file: Path | None,
 ) -> None:
     r"""Evaluate a model for a task.
@@ -505,7 +518,8 @@ def display_eval_report(result: EvalResult, console: Console) -> None:
     console.print()
     console.print(f"[dim]Task:[/dim]       {cfg.task}")
     console.print(f"[dim]Device:[/dim]     {cfg.device}")
-    console.print(f"[dim]Dataset:[/dim]    {ds.path}")
+    if ds.path:
+        console.print(f"[dim]Dataset:[/dim]    {ds.path}")
     console.print(f"[dim]Samples:[/dim]    {ds.samples}")
     if cfg.model_path:
         console.print(f"[dim]ONNX:[/dim]       {cfg.model_path}")
--- a/src/winml/modelkit/eval/config.py
+++ b/src/winml/modelkit/eval/config.py
@@ -87,6 +87,13 @@ class WinMLEvaluationConfig:
             device-to-provider mapping when provided.
         dataset: Dataset configuration.
         output_path: Path to write JSON results.
+        mode: Evaluation mode.
+
+            - ``"onnx"`` (default): evaluate the ONNX candidate on the
+              labeled dataset.
+            - ``"compare"``: compare ONNX vs HF reference output tensors
+              on identical random inputs and report tensor-similarity
+              metrics per output tensor.
 
     Usage:
         config = WinMLEvaluationConfig(
@@ -103,6 +110,7 @@ class WinMLEvaluationConfig:
     ep: EPNameOrAlias | None = None
     dataset: DatasetConfig = field(default_factory=DatasetConfig)
     output_path: Path | None = field(default=None, metadata={"cli_name": "output"})
+    mode: str = "onnx"
 
     def to_dict(self) -> dict:
         """Convert to dictionary for serialization."""
@@ -121,6 +129,8 @@ class WinMLEvaluationConfig:
         result["dataset"] = self.dataset.to_dict()
         if self.output_path is not None:
             result["output_path"] = str(self.output_path)
+        if self.mode != "onnx":
+            result["mode"] = self.mode
         return result
 
     @classmethod
@@ -148,4 +158,5 @@ class WinMLEvaluationConfig:
             ep=data.get("ep"),
             dataset=dataset,
             output_path=(Path(data["output_path"]) if data.get("output_path") else None),
+            mode=data.get("mode", "onnx"),
         )
--- a/src/winml/modelkit/eval/config.py
+++ b/src/winml/modelkit/eval/config.py
@@ -12,6 +12,7 @@ from pathlib import Path
 from typing import Any
 
 from ..utils.constants import EPNameOrAlias
+from ..utils.eval_utils import EvalMode
 
 
 @dataclass
@@ -87,7 +88,7 @@ class WinMLEvaluationConfig:
             device-to-provider mapping when provided.
         dataset: Dataset configuration.
         output_path: Path to write JSON results.
-        mode: Evaluation mode.
+        mode: Evaluation mode (see :data:`EvalMode`).
 
             - ``"onnx"`` (default): evaluate the ONNX candidate on the
               labeled dataset.
@@ -110,7 +111,7 @@ class WinMLEvaluationConfig:
     ep: EPNameOrAlias | None = None
     dataset: DatasetConfig = field(default_factory=DatasetConfig)
     output_path: Path | None = field(default=None, metadata={"cli_name": "output"})
-    mode: str = "onnx"
+    mode: EvalMode = "onnx"
 
     def to_dict(self) -> dict:
         """Convert to dictionary for serialization."""
--- a/src/winml/modelkit/eval/evaluate.py
+++ b/src/winml/modelkit/eval/evaluate.py
@@ -289,7 +289,16 @@ def evaluate(config: WinMLEvaluationConfig) -> EvalResult:
     copies via ``dataclasses.replace`` and ``deepcopy`` so the original
     config and any module-level defaults remain untouched.
     """
-    config = replace(config, task=_resolve_task(config), dataset=deepcopy(config.dataset))
+    from ..utils.eval_utils import EVAL_MODES
+
+    mode = config.mode if config.mode is not None else "onnx"
+    if mode not in EVAL_MODES:
+        raise ValueError(
+            f"Invalid mode {mode!r}; expected one of {EVAL_MODES} or None."
+        )
+    config = replace(
+        config, mode=mode, task=_resolve_task(config), dataset=deepcopy(config.dataset)
+    )
     if config.mode != "compare" and config.dataset.path is None:
         default = _DEFAULT_DATASETS.get(config.task)
         if default is None:
--- a/src/winml/modelkit/eval/evaluate.py
+++ b/src/winml/modelkit/eval/evaluate.py
@@ -57,16 +57,19 @@ _EVALUATOR_REGISTRY: dict[str, str] = {
         "winml.modelkit.eval.zero_shot_classification_evaluator:WinMLZeroShotClassificationEvaluator",
     "zero-shot-image-classification":
         "winml.modelkit.eval.zero_shot_image_classification_evaluator:WinMLZeroShotImageClassificationEvaluator",
+    "compare-tensor":
+        "winml.modelkit.eval.tensor_similarity_evaluator:TensorSimilarityEvaluator",
 }
 
 
-def get_evaluator_class(task: str) -> type[WinMLEvaluator]:
+def get_evaluator_class(config: WinMLEvaluationConfig) -> type[WinMLEvaluator]:
     """Return the evaluator class for *task*, or raise ValueError if unsupported."""
-    spec = _EVALUATOR_REGISTRY.get(task)
+    key = "compare-tensor" if config.mode == "compare" else config.task
+    spec = _EVALUATOR_REGISTRY.get(key)
     if spec is None:
         supported = ", ".join(sorted(_EVALUATOR_REGISTRY))
         raise ValueError(
-            f"Task '{task}' is not supported by `winml eval`. "
+            f"Task '{key}' is not supported by `winml eval`. "
             f"Supported tasks: {supported}."
         )
     module_path, class_name = spec.rsplit(":", 1)
@@ -278,7 +281,7 @@ def evaluate(config: WinMLEvaluationConfig) -> EvalResult:
     config and any module-level defaults remain untouched.
     """
     config = replace(config, task=_resolve_task(config), dataset=deepcopy(config.dataset))
-    if config.dataset.path is None:
+    if config.mode != "compare" and config.dataset.path is None:
         default = _DEFAULT_DATASETS.get(config.task)
         if default is None:
             raise ValueError(
@@ -309,7 +312,7 @@ def evaluate(config: WinMLEvaluationConfig) -> EvalResult:
 
     from ..utils.eval_utils import DatasetValidationError
 
-    cls = get_evaluator_class(config.task)
+    cls = get_evaluator_class(config)
     try:
         console.print("[bold]Loading dataset and evaluating...[/bold]")
         task_evaluator = cls(config, model)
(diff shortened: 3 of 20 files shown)


src/winml/modelkit/commands/eval.py
```diff
@@ -170,6 +182,7 @@ def eval(
     dataset_script: str | None,
     trust_remote_code: bool,
     show_schema: bool,
+    mode: str,
```
xieofxie: use a type instead of str

src/winml/modelkit/eval/metrics/tensor_similarity.py
```diff
@@ -0,0 +1,133 @@
...
+    Exactly one input all-zero -> ``0.0`` (a dead vector against a live
+    one is NOT a perfect match, even though the angle is undefined).
+    """
+    norm_ref = float(np.linalg.norm(ref))
+    norm_test = float(np.linalg.norm(test))
+    if norm_ref == 0.0 and norm_test == 0.0:
+        return 1.0
+    if norm_ref == 0.0 or norm_test == 0.0:
+        return 0.0
+    return float(np.dot(ref, test) / (norm_ref * norm_test))
```
vortex-captain, line 74: Just being curious: cosine similarity seems to be a robust measure when mean/max of abs diff can be quite large due to numerical errors of both models being compares; on the other hand, the norms of 2 vectors are lost in similarity computation, would it be possible to compare norms jointly with cosine similarity?

vortex-captain (approved):

src/winml/modelkit/eval/metrics/tensor_similarity.py
zhenchaoni: You are right, cosine similarity is 1 as long as 2 vectors are parallel. But we have other metrics that can catch the problem, such as SNR, MSE, ABS diff.

src/winml/modelkit/commands/eval.py
zhenchaoni: fixed

vortex-captain (approved):

src/winml/modelkit/eval/config.py
```diff
@@ -109,6 +117,7 @@ class WinMLEvaluationConfig:
     ep: EPNameOrAlias | None = None
     dataset: DatasetConfig = field(default_factory=DatasetConfig)
     output_path: Path | None = field(default=None, metadata={"cli_name": "output"})
+    mode: EvalMode = "onnx"
```
xieofxie, line 120: `mode` is typed as `Literal["onnx", "compare"]` but that's not enforced at runtime. Click's `click.Choice` validates the CLI surface, but `WinMLEvaluationConfig(mode="hf")` or `from_dict({"mode": "hf"})` constructs an instance silently. In `evaluate.get_evaluator_class` (line 69):

```python
key = "compare-tensor" if config.mode == "compare" else config.task
```

any unknown mode falls through to the `task` lookup — so a typo or the future `"hf"` placeholder behaves like `onnx`. Suggest adding a `__post_init__` validator:

```suggestion
    mode: EvalMode = "onnx"

    def __post_init__(self) -> None:
        if self.mode not in EVAL_MODES:
            raise ValueError(
                f"Invalid mode {self.mode!r}; expected one of {EVAL_MODES}."
            )
```

src/winml/modelkit/eval/config.py
```diff
@@ -155,4 +166,5 @@ def from_dict(cls, data: dict) -> WinMLEvaluationConfig:
             ep=data.get("ep"),
             dataset=dataset,
             output_path=(Path(data["output_path"]) if data.get("output_path") else None),
+            mode=data.get("mode", "onnx"),
```
xieofxie, line 169: Same concern as the `mode` field default — `from_dict` accepts any string under the `mode` key without checking it against `EVAL_MODES`. With the `__post_init__` validator suggested above this becomes free (the dataclass constructor enforces it); without it, callers loading a hand-edited JSON can get silent onnx-mode fallback.

src/winml/modelkit/eval/tensor_similarity_evaluator.py
```diff
@@ -0,0 +1,170 @@
...
+
+    def __init__(
+        self,
+        config: WinMLEvaluationConfig,
+        model: WinMLPreTrainedModel,
+    ) -> None:
+        from ..models.winml.composite_model import WinMLCompositeModel
+
+        if isinstance(model, WinMLCompositeModel):
+            sub_tasks = list(type(model)._SUB_MODEL_CONFIG.values())
```
xieofxie: `type(model)._SUB_MODEL_CONFIG` reaches into a private class attribute. If it's ever renamed or moved, the friendly error message degrades to a bare `AttributeError` *before* the helpful guidance can be raised. Make it defensive:

```suggestion
            sub_tasks = list(getattr(type(model), "_SUB_MODEL_CONFIG", {}).values())
```

src/winml/modelkit/eval/tensor_similarity_evaluator.py
```diff
@@ -0,0 +1,170 @@
...
+            self.config.model_id, dtype=torch.float32
+        ).eval()
+
+    def prepare_data(self) -> Any:
+        """Build a RandomDataset over the candidate ONNX's I/O spec."""
+        from ..datasets.random_dataset import RandomDataset
+
+        ds = self.config.dataset
+        return RandomDataset(
+            model_path=str(self.model._onnx_path),
```
xieofxie: Accessing `self.model._onnx_path` reaches into a private attribute of `WinMLPreTrainedModel`. Same concern at line 101 for `self.data._io_config`. If either internal field is renamed, this breaks silently for compare mode while every other code path stays green. Suggest either:

- promoting them to public properties (`WinMLPreTrainedModel.onnx_path`, `RandomDataset.io_config`) and using those here, **or**
- documenting the contract on both classes with a comment so future refactors know this is load-bearing.

src/winml/modelkit/eval/tensor_similarity_evaluator.py
```diff
@@ -0,0 +1,170 @@
...
+            max_samples=int(ds.samples or 100),
```
xieofxie: `int(ds.samples or 100)` swallows an explicit `0`. A user passing `--samples 0` (or a config with `samples: 0`) likely means "no samples" — they get 100 instead. Distinguish unset from zero:

```suggestion
            max_samples=int(ds.samples if ds.samples is not None else 100),
            seed=int(ds.seed if ds.seed is not None else 42),
```

src/winml/modelkit/eval/tensor_similarity_evaluator.py
```diff
@@ -0,0 +1,170 @@
...
+
+    def compute(self) -> dict[str, dict[str, float]]:
+        """Run paired inference and return display-ready per-metric per-output values.
+
+        Returns ``{f"{metric}_{stat}": {output_name: float}}`` — the flat shape
+        the generic eval report renderer prints as one row per ``{metric}_{stat}``
+        with ``output_name=value`` cells joined across outputs.
+        """
+        import torch
+        from tqdm import tqdm
```
xieofxie, line 97: `tqdm` is in `optional-dependencies.dev`, not the runtime dependencies (see the discussion on #788 and the `_make_progress_bar` fallback in `src/winml/modelkit/session/ep_registry.py`). A production install without the `[dev]` extra will hit `ImportError` here mid-evaluation. Suggest mirroring the no-op fallback pattern:

```python
try:
    from tqdm import tqdm
except ImportError:
    def tqdm(it, **_kw):  # silent passthrough
        return it
```

or factor `_make_progress_bar` into a shared util so both call sites share one source of truth.

src/winml/modelkit/eval/tensor_similarity_evaluator.py
```diff
@@ -0,0 +1,170 @@
...
+
+        from .metrics.tensor_similarity import TensorSimilarityMetric
+
+        input_names = list(self.data._io_config["input_names"])
```
xieofxie: Same private-attribute concern as `model._onnx_path` above. `RandomDataset` exposes `_io_config` as an implementation detail. If you only need `input_names`, consider a public accessor on the dataset (`RandomDataset.input_names`) so the evaluator doesn't double as a coupling point between two internals.

xieofxie (commented):
Drive-by review focused on correctness, conventions in this repo (e.g. tqdm-is-optional from #788), and a few edge cases worth tightening before merge. Individual suggestions on the lines below.

src/winml/modelkit/eval/metrics/tensor_similarity.py
```diff
@@ -0,0 +1,133 @@
...
+        self._per_sample["mse"].append(mse_val)
+        self._per_sample["max_abs_diff"].append(_max_abs_diff(ref, test))
+
+    def compute(self) -> dict[str, float]:
+        """Return ``{f"{metric}_{stat}": float}`` for stats mean/std/min/max."""
+        result: dict[str, float] = {}
+        for metric, values in self._per_sample.items():
+            if not values:
+                continue
+            finite = [v for v in values if math.isfinite(v)]
```
xieofxie, line 116: NaN propagation isn't handled distinctly from `+inf`. `math.isfinite(nan) == False`, so a metric run where *every* sample produced NaN (e.g. the candidate output had a NaN that polluted MSE/cosine) takes the empty-`finite` branch and falls back to `mean_val = +inf`, misreporting as "perfect." Suggest distinguishing the cases:

```python
finite = [v for v in values if math.isfinite(v)]
if finite:
    ...
elif all(math.isinf(v) and v > 0 for v in values):
    mean_val, std_val = math.inf, 0.0
else:
    mean_val, std_val = math.nan, math.nan
```

Also worth adding a unit test that updates with `nan`-containing tensors and asserts the aggregate is NaN, not `+inf`.

src/winml/modelkit/eval/evaluate.py
```diff
@@ -59,16 +59,19 @@
...
+    "compare-tensor":
+        "winml.modelkit.eval.tensor_similarity_evaluator:TensorSimilarityEvaluator",
 }
 
 
-def get_evaluator_class(task: str) -> type[WinMLEvaluator]:
+def get_evaluator_class(config: WinMLEvaluationConfig) -> type[WinMLEvaluator]:
     """Return the evaluator class for *task*, or raise ValueError if unsupported."""
-    spec = _EVALUATOR_REGISTRY.get(task)
+    key = "compare-tensor" if config.mode == "compare" else config.task
```
xieofxie, line 69: This is the silent-fallthrough site referenced from the `config.py` comment: any `config.mode` value other than the literal string `"compare"` resolves as a regular task lookup. If you add the `__post_init__` validator on `WinMLEvaluationConfig`, this stays correct-by-construction. Without it, consider an explicit guard:

```python
if config.mode not in EVAL_MODES:
    raise ValueError(f"Unsupported mode {config.mode!r}")
key = "compare-tensor" if config.mode == "compare" else config.task
```

tests/e2e/test_eval_e2e.py
```diff
@@ -766,6 +766,76 @@ def test_dataset_script_without_trust_remote_code(
...
+        for output_name in per_output_names or ():
+            lo, hi = cos_min[output_name], cos_max[output_name]
+            assert -1.0 <= lo <= hi <= 1.0, (
+                f"cosine outside [-1, 1] for {output_name}: min={lo}, max={hi}"
+            )
+
+        # fp16 parity on QNN should be near-perfect (>= 0.95). VitisAI/CPU
+        # paths can degrade more, so gate the magnitude check on QNN.
+        if is_host("qnn"):
+            cos_mean = metrics["cosine_similarity_mean"]
```
xieofxie: The `>= 0.95` magnitude check is gated on QNN only. On CPU / VitisAI runners the assertion is structurally skipped, so a regression that drops fp16 parity to e.g. 0.2 would still pass the test on those hosts. Consider a looser CPU-host bound (e.g. `>= 0.5`) as a sanity floor — that way the test detects total breakage even on the non-QNN matrix.

src/winml/modelkit/eval/config.py
zhenchaoni: Fixed in evaluate.py to validate the input.

src/winml/modelkit/eval/config.py
zhenchaoni: Fixed in evaluate.py to validate the input.

src/winml/modelkit/eval/tensor_similarity_evaluator.py
zhenchaoni: fixed

src/winml/modelkit/eval/tensor_similarity_evaluator.py
zhenchaoni: fixed by using public methods

src/winml/modelkit/eval/tensor_similarity_evaluator.py
zhenchaoni: fixed

src/winml/modelkit/eval/tensor_similarity_evaluator.py
zhenchaoni: tqdm is currently already indirectly installed by other mandatory dependencies. I resolved this comment by explicitly put tqdm in mandatory dependencies to avoid confusion.

src/winml/modelkit/eval/tensor_similarity_evaluator.py
zhenchaoni: fixed by using public properties

src/winml/modelkit/eval/metrics/tensor_similarity.py
zhenchaoni: fixed according to your suggestion

src/winml/modelkit/eval/evaluate.py
zhenchaoni: fixed by adding the validation logic

tests/e2e/test_eval_e2e.py
zhenchaoni: fixed according to your suggestion.

xieofxie (approved):

PR #806, opened 2026-06-03
fix(export): resolve timm image_size from pretrained_cfg

vortex-captain:
## Summary

- Fix `winml config` emitting `[1, 3, 64, 64]` instead of `[1, 3, 224, 224]` for timm image-classification models (e.g. `timm/repghostnet_200.in1k`).
- Root cause: `TimmWrapperConfig` stores shape info in `pretrained_cfg["input_size"]` (a plain dict). Optimum's `NormalizedConfig` only walks `PretrainedConfig` children, so the value is invisible to `DummyVisionInputGenerator`, which then defaults to 64x64. timm models also have no `preprocessor_config.json` on the hub, so winml's existing fallback misses too.
- Fix: when `preprocessor_config.json` is unavailable, synthesize a preprocessor-style dict from `hf_config.pretrained_cfg.input_size`. The existing size-parsing block (`size` int / dict-with-height-width / shortest_edge) is unchanged — the timm concern is isolated at the data-fetch boundary in `_get_preprocessor_dict` / `_synthesize_preprocessor_dict`.
- Registry: added `timm/mobilenetv3_small_100.lamb_in1k` and `timm/repghostnet_200.in1k` to `models_all.json`; both PASS perf on CPU. Perf output confirms `pixel_values [1, 3, 224, 224] float32`.


timenick:
### CI failure here is unrelated to this PR — it's network-dependent unit tests

The failing job ([run 26930137132](https://github.com/microsoft/winml-cli/actions/runs/26930137132/job/79448041756?pr=806)) is erroring on `tests/unit/export/test_io.py`, not on the timm change in this PR. All 9 errors are the same `OSError`:

```
OSError: We couldn't connect to 'https://huggingface.co' to load the files,
and couldn't find them in the cached files.
```

**Root cause:** two module-scoped fixtures pull configs from the HF Hub at collection time:

- `t5_config` → `AutoConfig.from_pretrained("google-t5/t5-small")` ([test_io.py#L700-L705](https://github.com/microsoft/winml-cli/blob/main/tests/unit/export/test_io.py#L700-L705))
- `qwen_config` → `AutoConfig.from_pretrained("Qwen/Qwen3-0.6B")` ([test_io.py#L708-L713](https://github.com/microsoft/winml-cli/blob/main/tests/unit/export/test_io.py#L708-L713))

These fixtures and the T5/Qwen test classes were introduced in #334 (`feat: multi-model support with KV cache (T5, Qwen, Mu2)`, commit `38775de8`).

A unit test shouldn't require network access — it makes the suite fail on offline runners, behind firewalls, or whenever the Hub is unavailable, which is exactly what hit this PR.

**Suggested fix:** build these configs synthetically, the way the Marian/BART tests in the *same file* already do (e.g. [`marian_config`](https://github.com/microsoft/winml-cli/blob/main/tests/unit/export/test_io.py#L1133-L1146) constructs `MarianConfig(...)` directly with small dims, no network). T5/Qwen can do the same. If a real `from_pretrained` download is genuinely needed, those cases belong under `tests/integration/`, not `tests/unit/`.

Generated with Claude Code

vortex-captain:
> ### CI failure here is unrelated to this PR — it's network-dependent unit tests
> The failing job ([run 26930137132](https://github.com/microsoft/winml-cli/actions/runs/26930137132/job/79448041756?pr=806)) is erroring on `tests/unit/export/test_io.py`, not on the timm change in this PR. All 9 errors are the same `OSError`:
> 
> ```
> OSError: We couldn't connect to 'https://huggingface.co' to load the files,
> and couldn't find them in the cached files.
> ```
> 
> **Root cause:** two module-scoped fixtures pull configs from the HF Hub at collection time:
> 
> * `t5_config` → `AutoConfig.from_pretrained("google-t5/t5-small")` ([test_io.py#L700-L705](https://github.com/microsoft/winml-cli/blob/main/tests/unit/export/test_io.py#L700-L705))
> * `qwen_config` → `AutoConfig.from_pretrained("Qwen/Qwen3-0.6B")` ([test_io.py#L708-L713](https://github.com/microsoft/winml-cli/blob/main/tests/unit/export/test_io.py#L708-L713))
> 
> These fixtures and the T5/Qwen test classes were introduced in #334 (`feat: multi-model support with KV cache (T5, Qwen, Mu2)`, commit `38775de8`).
> 
> A unit test shouldn't require network access — it makes the suite fail on offline runners, behind firewalls, or whenever the Hub is unavailable, which is exactly what hit this PR.
> 
> **Suggested fix:** build these configs synthetically, the way the Marian/BART tests in the _same file_ already do (e.g. [`marian_config`](https://github.com/microsoft/winml-cli/blob/main/tests/unit/export/test_io.py#L1133-L1146) constructs `MarianConfig(...)` directly with small dims, no network). T5/Qwen can do the same. If a real `from_pretrained` download is genuinely needed, those cases belong under `tests/integration/`, not `tests/unit/`.
> 
> Generated with Claude Code

fixed

src/winml/modelkit/export/io.py
```diff
@@ -206,48 +206,105 @@ def _get_onnx_config(
...
+            preprocessor dict when ``preprocessor_config.json`` is missing.
     """
     if not model_id:
         return
 
     if "height" in shape_kwargs or "width" in shape_kwargs:
         return
 
+    config = _get_preprocessor_dict(model_id, hf_config)
+    size = config.get("size")
```
timenick, line 233: Robustness gap: `_synthesize_preprocessor_dict` only runs when `get_image_processor_dict` **raises**. If it succeeds and returns a dict that just doesn''t contain `"size"` (e.g. a partial `preprocessor_config.json` carrying only `mean` / `std` / `interpolation`), `size` here is `None`, the size-parsing block below is skipped, and we silently fall back to Optimum''s 64×64 default — exactly the bug this PR set out to fix, just via a different code path.

The PR description scopes this to "`preprocessor_config.json` is unavailable", so this is one step past the stated bug — but cheap to close at the same time, since some timm wrapper uploads (and a handful of HF community uploads) do ship partial preprocessor configs without a size key.

Cheapest fix — retry through the synthesizer when the resolved size is `None`:
```python
config = _get_preprocessor_dict(model_id, hf_config)
size = config.get("size")
if size is None and hf_config is not None:
    size = _synthesize_preprocessor_dict(hf_config).get("size")
```
Alternative: push the check into `_get_preprocessor_dict` itself (`if "size" not in config:` → fall through to synthesizer), which keeps `_populate_image_size_from_preprocessor` ignorant of the resolution order — matching the docstring claim that it returns a dict "so downstream parsing logic does not need to know which source it came from".

zhenchaoni (approved):

src/winml/modelkit/export/io.py
vortex-captain: fixed

timenick (approved):

PR #807, opened 2026-06-03
feat(task): modality-aware detect_task unifies detection across commands

timenick:
## What

PR2 of #800. Add a single, offline, modality-aware task detector and route every detection consumer through it, so the same model resolves to the same WinMLTask everywhere. Builds on #801 (the `to_optimum_task` boundary), now on `main`.

## Changes

- **`loader/task.py::detect_task(config) -> (WinMLTask, source)`** — new single detector. Config-only dispatch (HF_MODEL_CLASS_MAPPING -> wrapped-library -> TasksManager -> HF_TASK_DEFAULTS), then a D2 vision heuristic (top-level `image_size`/`patch_size` upgrades `feature-extraction` -> `image-feature-extraction`) applied to the **returned** task only. Internal model-class resolution keeps the Optimum-canonical task, so no Optimum API receives `image-feature-extraction`.
- **`inspect/resolver.py::detect_task`** delegates to the loader detector (drops its own dispatch + now-dead imports).
- **`eval/evaluate.py::_resolve_task`** uses `detect_task` and drops the reverse io_config reconstruction (`to_hf_pipeline_task` / `HF_TASK_NAME_MAPPING`); an explicit `--task` is surfaced verbatim.
- `config` / `perf` / `build` inherit modality-aware detection via `resolve_task_and_model_class` Case 1 — no code change.
- New `tests/unit/loader/test_detect_task.py` and `tests/integration/test_task_consistency.py` (cross-resolver agreement on real configs); `tests/unit/eval/test_eval.py` updated.

## Behavior

- Image-embedding models (DINOv2) auto-detect `image-feature-extraction` across inspect / config / export / eval / perf / build — source-level fix for #778, so calibration datasets and evaluators pick the image asset directly.
- CLIP / SAM2 / timm / text models unchanged (CLIP's `image_size` is nested -> D2 does not fire).
- **Accepted behavior change:** explicit `winml eval --task feature-extraction` on a vision model no longer auto-upgrades; per the canonical rule it is out-of-domain and surfaced as-is (fails downstream rather than being silently rewritten). A clean early rejection would need model-aware validation (out of scope).


src/winml/modelkit/eval/evaluate.py
```diff
@@ -26,6 +26,9 @@
 
 # Map task -> "module_path:ClassName"; modules are imported lazily by
 # get_evaluator_class() to improve command latency.
+# Keep the key/value-per-line layout: collapsing each entry onto one line (the
```
zhenchaoni, line 29: what is the purpose of adding this comment?

src/winml/modelkit/inspect/resolver.py
```diff
@@ -113,45 +105,14 @@ def validate_task(task: str) -> None:
 
 
 def detect_task(config: PretrainedConfig) -> tuple[str, str]:
```
zhenchaoni: Why not just remove this? We can make other callers to use the loader detect task

src/winml/modelkit/loader/task.py
```diff
@@ -275,6 +275,94 @@ def _detect_task_from_config(config: PretrainedConfig) -> str:
...
+    modality for the otherwise-lossy ``feature-extraction`` task (D2 heuristic).
+    """
+    try:
+        keys = config.to_dict().keys()
+    except Exception:
+        keys = vars(config).keys()
+    return "image_size" in keys or "patch_size" in keys
+
+
+def _apply_vision_modality(config: PretrainedConfig, task: str) -> str:
```
zhenchaoni: I think vision modality is one case. Maybe in the future, there will be more cases. I think you can call this method as "disambiguate_model_task" or "resolve_task_modality"

src/winml/modelkit/loader/task.py
```diff
@@ -275,6 +275,94 @@ def _detect_task_from_config(config: PretrainedConfig) -> str:
     return task
 
 
+def _is_top_level_vision_config(config: PretrainedConfig) -> bool:
```
zhenchaoni: I suggest you to have a register based solution.
```
feature-extraction: {
   "image-feature-extraction": ["image_size", "patch_size" ]
}
```

There is audio feature extraction, although we don't support yet. The registry based approach will be more extensible

src/winml/modelkit/eval/evaluate.py
```diff
@@ -255,13 +231,11 @@ def _resolve_task(config: WinMLEvaluationConfig) -> str:
 
         from transformers import AutoConfig
 
-        from ..loader.task import _detect_task_from_config
+        from ..loader import detect_task
```
zhenchaoni, line 246: Since we decide to use a unique and Unambiguous task name, I think you can also undo some changes in src\winml\modelkit\datasets\__init__.py

src/winml/modelkit/eval/evaluate.py
timenick: The registry's `module:ClassName` values exceed 100 chars on a single line, so the formatter (which wants one entry per line) and the linter (E501) disagree. `# fmt: off` keeps the readable key/value-per-line layout that passes both.

The cleaner fix — dropping the repeated `winml.modelkit.eval.` prefix so entries fit on one line naturally — is tracked in #808 and will land in a follow-up PR.

Generated with Claude Code

src/winml/modelkit/inspect/resolver.py
timenick: Done — removed `resolver.detect_task` and import `detect_task` from `..loader` directly in `inspect/__init__.py`, so there is a single detector with no shim. (f285a891)

Generated with Claude Code

src/winml/modelkit/loader/task.py
timenick: Done — renamed to `_resolve_task_modality`. (f285a891)

Generated with Claude Code

src/winml/modelkit/loader/task.py
timenick: Done — replaced the hard-coded check with a data-driven `_TASK_MODALITY_DISAMBIGUATION` table (a modality-blind task -> its modality-aware variants, each keyed by the signal config fields). Behavior is unchanged and `audio-feature-extraction` can be added as data, not code. (bc36bc15)

Generated with Claude Code

src/winml/modelkit/eval/evaluate.py
timenick: Done — `feature-extraction` now maps directly to `TextDataset` and `_resolve_dataset_class` drops the io_config reverse reconstruction (the datasets-side twin of the eval `to_hf_pipeline_task` removal). (bc36bc15)

Generated with Claude Code

zhenchaoni (approved):

PR #810, opened 2026-06-04
Enable telemetry in shipped wheel and reword consent as "unlinked pseudonymized"

timenick:
## Summary

- **Re-enable telemetry in the shipped wheel.** Restore the `InstrumentationKey` injection + post-build verify steps in `modelkit-official-build.yml` (reverting the pipeline portion of #728). The PyPI **sdist** still ships the empty placeholder; only the **wheel** carries the injected key, gated on the `INSTRUMENTATION_KEY` build secret defined in the ADO pipeline.
- **Reword the consent notice** from "anonymous" to "unlinked pseudonymized" in the first-run prompt (`consent.py`) and `docs/Privacy.md`. This is the more accurate classification — the persisted per-machine device-id hash is pseudonymized, not anonymous.
- **Bump `_CONSENT_VERSION` 1 → 2** so already-consented users see and re-accept the updated notice.

## Notes

- Source `constants.py` keeps the empty iKey placeholder, so dev/source installs (`pip install -e .`) still never emit.
- Telemetry remains off in CI and non-TTY contexts regardless of stored consent.
- ⚠️ **Ops dependency:** the injection step reads the `INSTRUMENTATION_KEY` ADO pipeline secret. #728 removed only the consuming steps, so the secret should still exist — please confirm it's present/valid before the next official build, or the build fails closed with `INSTRUMENTATION_KEY env var is empty or missing`.

Verified: telemetry unit + integration suites pass (185 + 4); pipeline block confirmed byte-identical to the proven pre-#728 version.

### DIFF
--- a/README.md
+++ b/README.md
@@ -272,6 +272,30 @@ Each arrow is a WinML CLI command. You can enter the pipeline at any stage (for 
 
 ---
 
+## :lock: Data / Telemetry
+
+Official WinML CLI releases can collect **unlinked pseudonymized** usage telemetry
+to help improve the product. Telemetry is classified as **Optional**. A one-time
+prompt on your first run asks for consent (default: accept — press **Enter** to
+enable, type `n` to decline).
+
+Dev installs (`pip install -e .` or running from a source checkout) never send
+telemetry.
+
+**Control** — edit `%USERPROFILE%\.winml\config.json`:
+
+- Set `telemetry.consent` to `"disabled"` to opt out
+- Set `telemetry.consent` to `"enabled"` to opt in
+- Delete the file to re-show the first-run prompt on the next run
+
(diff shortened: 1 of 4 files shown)


xieofxie (approved):

README.md
```diff
@@ -272,6 +272,24 @@ Each arrow is a WinML CLI command. You can enter the pipeline at any stage (for
 
 ---
 
+## :lock: Data / Telemetry
+
+Official WinML CLI releases can collect **unlinked pseudonymized** usage telemetry
+to help improve the product. Telemetry is classified as **Optional**. A one-time
+prompt on your first run asks for consent (default: accept — press **Enter** to
```
xieofxie, line 279: never tried it..

PR #817, opened 2026-06-05
fix(optim): untie batched constant MatMul for OpenVINO GPU

xieofxie:
## Problem

`winml perf -m cross-encoder/nli-deberta-v3-small --task zero-shot-classification --ep openvino --device gpu` fails to compile:

```
[GPU] Failed to select implementation for
name:matmul:/deberta/encoder/layer.5/attention/self/MatMul_1
type: gemm
... compile_graph.cpp:59  (shape_type == dynamic_shape || node->selected_impl != nullptr)
```

## Root cause

OpenVINO GPU's oneDNN gemm **cannot select an implementation for a batched (rank ≥ 3) MatMul where an operand is a compile-time constant.** Verified by isolation against the real OV-GPU EP:

| case | result |
|---|---|
| 3D dynamic @ 3D dynamic (content q·kᵀ) | ✅ compiles |
| 3D dynamic @ 3D **constant** (position terms) | ❌ fails |
| 3D @ 2D constant | ✅ compiles |
| operand converted to runtime input | ✅ compiles |

For a static-shaped node `selected_impl` must be non-null; impl-selection returns nothing for the batched-constant gemm, so the assert fires. DeBERTa hits this because its disentangled-attention position key/query depend only on weights and **fold to 3D constants** during export (12 such MatMuls — 2 per layer). Disabling torch constant-folding doesn't help: OV folds the all-constant subgraph itself.

## Fix

A new EP-gated surgery transform, **`untie-constant-batched-matmul`**, routes each constant operand through `Add(const, zero)` where `zero` is a data-dependent runtime `[1]` tensor (`Cast → Reshape(-1) → Slice[0:1] → Sub`). This makes the operand runtime-valued so OV's constant folder can't repack it into a gemm weight, while:
- keeping the **single batched MatMul** (no perf regression — a 2D per-head decomposition also works but explodes into 144 tiny matmuls),
- leaving **numerics unchanged** (`+0`).

Wired via **autoconf**: `BatchedConstMatMulValidator` detects the pattern and, gated to **Intel IHV + GPU**, emits a `GraphOptimization` opportunity the existing autoconf loop auto-applies. Pattern-based and architecture-agnostic (no model-name hardcoding). The detector doesn't re-fire after surgery, so autoconf converges.

Two incidental bugs fixed:
- Model-validator device filter was **case-sensitive** (`"gpu"` ≠ `"GPU"`) → made case-insensitive.
- First construction used `ReduceMin` (no axes), which **crashed the static analyzer's** reduction input-generator → replaced with ubiquitous analyzer-safe ops.

## Verification

- Original failing command now **compiles on OV-GPU and benchmarks** (~15.5 ms avg, ~64 samples/sec).
- GPU output matches CPU reference (argmax matches; diff `6e-4`, normal fp16/fp32).
- Detector gates correctly (openvino+GPU on; NPU / CPU / DML off).
- New unit tests pass; full optim + analyze unit suites (1923 tests) pass — no regressions.

> Note: artifacts are cached by build-config hash, so an existing stale cache needs `--rebuild` / `--ignore-cache` to pick up the fix.


xieofxie:
Source code here https://github.com/openvinotoolkit/openvino/blob/master/src/plugins/intel_gpu/src/graph/graph_optimizer/compile_graph.cpp

Also issue created https://github.com/openvinotoolkit/openvino/issues/36272

xieofxie:
Add an issue to allow user to opt out https://github.com/microsoft/winml-cli/issues/835

DingmaomaoBJTU (commented):
Overall the design is sound: the surgery correctly routes constant operands through a data-dependent `+0` to defeat OpenVINO GPU's constant folder, and the validator correctly gates on Intel IHV + GPU. A few correctness issues and documentation gaps are worth addressing before merge.

src/winml/modelkit/optim/pipes/surgery.py
```diff
@@ -319,3 +331,137 @@ def _remove_isnan_in_attention_mask(
...
+                zero_by_dtype[dtype] = name
+            return name
+
+        untied = 0
+        for node, idx in targets:
+            const_name = node.input[idx]
+            dtype = initializers[const_name].data_type
+            if dtype not in (TensorProto.FLOAT, TensorProto.FLOAT16, TensorProto.DOUBLE):
+                continue
+            dyn = f"{prefix}_{node.name}_in{idx}".replace("/", "_")
```
DingmaomaoBJTU: **Bug: node-name collision when names are empty or duplicated**

`dyn` is derived solely from `node.name`, which is optional in ONNX — real exported models (PyTorch, TF, DeBERTa etc.) routinely leave node names blank, and even named models may have duplicate names. When two target MatMuls share the same name or are both unnamed, this loop produces two `Add` nodes with identical output names, which makes the transformed graph invalid (`onnx.checker.check_model` rejects it and ORT fails at load time).

Example: two unnamed nodes both produce `dyn = 'winml_ovgpu_untie__in1'`.

Fix — use a loop index instead of the node name:
```python
for untie_idx, (node, idx) in enumerate(targets):
    dyn = f"{prefix}_untied{untie_idx}_in{idx}"
```

src/winml/modelkit/optim/pipes/surgery.py
```diff
@@ -319,3 +331,137 @@ def _remove_isnan_in_attention_mask(
...
+        """Make a batched MatMul's constant operand runtime-valued.
+
+        OpenVINO GPU's oneDNN gemm cannot select an implementation for a batched
+        (rank >= 3) MatMul where an operand is a compile-time constant: the same
+        gemm with a dynamic operand, and 2D constant gemm, both compile fine.
+        Transformer disentangled-attention position terms depend only on weights,
+        so they fold into 3D constants and hit this case.
+
+        Fix: route each such constant operand through ``Add(const, zero)`` where
+        ``zero = Sub(s, s)`` and ``s = ReduceMin(Cast(first_input -> float))``.
```
DingmaomaoBJTU: **Docstring describes `ReduceMin` but the implementation uses `Reshape + Slice`**

The docstring says `s = ReduceMin(Cast(first_input -> float))`, but the actual code does `Cast -> Reshape([-1]) -> Slice([0:1]) -> Sub(elem, elem)`. The ReduceMin strategy was apparently abandoned in favour of the slice approach, but the docstring was not updated. Please revise the docstring to match the current graph construction.

src/winml/modelkit/optim/pipes/surgery.py
```diff
@@ -319,3 +331,137 @@ def _remove_isnan_in_attention_mask(
...
+        )
+        flat = f"{prefix}_flat"
+        new_inits.append(numpy_helper.from_array(np.array([-1], dtype=np.int64), f"{prefix}_m1"))
+        new_nodes.append(helper.make_node("Reshape", [xf, f"{prefix}_m1"], [flat], name=flat))
+        elem = f"{prefix}_elem"
+        new_inits.append(numpy_helper.from_array(np.array([0], dtype=np.int64), f"{prefix}_0"))
+        new_inits.append(numpy_helper.from_array(np.array([1], dtype=np.int64), f"{prefix}_1"))
+        new_nodes.append(
+            helper.make_node(
+                "Slice", [flat, f"{prefix}_0", f"{prefix}_1", f"{prefix}_0"], [elem], name=elem
```
DingmaomaoBJTU: **`{prefix}_0` reused as both `starts` and `axes` — correct but easy to misread**

The Slice node receives `f"{prefix}_0"` (value `[0]`) for both the `starts` and `axes` inputs. This is numerically correct because both happen to be `[0]`, but it looks like a copy-paste error and the two roles are conceptually distinct. Consider using separate tensors with descriptive names (e.g. `{prefix}_starts` and `{prefix}_axis`) to make the intent explicit and guard against future edits silently breaking one slot.

src/winml/modelkit/optim/pipes/surgery.py
```diff
@@ -319,3 +331,137 @@ def _remove_isnan_in_attention_mask(
...
+            if len(const_idx) != 1:
+                continue
+            idx = const_idx[0]
+            if len(initializers[node.input[idx]].dims) >= 3:
+                targets.append((node, idx))
+
+        if not targets:
+            return model
+
+        if not graph.input:
```
DingmaomaoBJTU, line 387: **Guard checks graph input *existence* but not element count**

`if not graph.input` only ensures at least one input is declared; it does not prevent a runtime failure when the first input is empty (e.g. batch size = 0 in a model that supports dynamic/zero batch). If `flat` has 0 elements the subsequent `Slice([0:1])` is out-of-bounds, producing a runtime error rather than a zero. This is a rare edge case but worth at least documenting in the method docstring so callers are aware of the assumption.

src/winml/modelkit/analyze/core/model_validators/batched_const_matmul_validator.py
```diff
@@ -0,0 +1,133 @@
...
+            return infer_ihv_from_ep_name(self.ep) == IHVType.INTEL
+        except Exception:  # pragma: no cover - defensive
+            return False
+
+    def validate(self) -> Information | None:
+        """Detect batched MatMul with a single constant rank>=3 operand."""
+        if not self._is_enabled():
+            return None
+
+        initializers = {init.name for init in self.graph.initializer}
```
DingmaomaoBJTU, line 74: **Validator misses constants expressed as `Constant` nodes**

`initializers = {init.name for init in self.graph.initializer}` only covers the initializer list. ONNX also allows constant values via `Constant` op nodes, which appear after some ORT preprocessing passes and in models produced by certain exporters. The surgery in `surgery.py` has the same limitation, so both sides are consistent — but a comment here noting this as a known gap would help future contributors avoid confusion when those cases go undetected.

tests/unit/optim/pipes/test_pipe_surgery.py
```diff
@@ -342,3 +342,120 @@ def test_surgery_pipe_runs_last(self) -> None:
...
+        add_nodes = [n for n in result.graph.node if n.op_type == "Add"]
+        assert len(add_nodes) == 1
+        assert add_nodes[0].output[0] in matmul.input
+        # Graph remains structurally valid.
+        onnx.checker.check_model(result)
+
+    def test_numerics_unchanged(self) -> None:
+        """+0 tie leaves outputs bit-for-bit identical on ORT CPU."""
+        import onnxruntime as ort
+
```
DingmaomaoBJTU, line 428: **Missing test: two target MatMul nodes with empty or duplicate names**

The existing tests use a model with exactly one MatMul node, so the name-collision bug on line 436 of `surgery.py` is not exercised. Please add a test that builds a model with two batched-const MatMul nodes that share the same name (or have empty names) and asserts that the result passes `onnx.checker.check_model` and produces correct numerics.

src/winml/modelkit/optim/pipes/surgery.py
xieofxie: Fixed in 79bcc3f5. The loop now indexes targets (`for untie_idx, (node, idx) in enumerate(targets)`) and builds `dyn = f"{prefix}_untied{untie_idx}_in{idx}"`, so names are unique regardless of whether the source MatMul is named, unnamed, or duplicated. Added `test_duplicate_node_names_do_not_collide` which builds two unnamed batched-const MatMuls and asserts unique Add outputs + `check_model` validity.

src/winml/modelkit/optim/pipes/surgery.py
xieofxie: Fixed in 79bcc3f5. Updated the docstring to describe the actual construction: `Cast(first_input -> float) -> Reshape([-1]) -> Slice([0:1])` yields a single element `elem`, and `zero = Sub(elem, elem)`. Dropped the stale ReduceMin wording.

src/winml/modelkit/optim/pipes/surgery.py
xieofxie: Fixed in 79bcc3f5. Split into distinct `_slice_starts`, `_slice_ends`, and `_slice_axis` initializers (each `[0]`/`[1]` as appropriate) with a comment, so the two roles are explicit and a future edit to one cannot silently break the other.

src/winml/modelkit/optim/pipes/surgery.py
xieofxie: Documented in 79bcc3f5. Added an explicit "Assumption" note to the docstring: the first graph input must have at least one element at runtime; for a zero-sized input the `Slice([0:1])` is out of bounds and would raise at inference time rather than produce a zero.

src/winml/modelkit/analyze/core/model_validators/batched_const_matmul_validator.py
xieofxie: Documented in 79bcc3f5. Added a "Known gap" comment noting that constants expressed as `Constant` op nodes (rather than initializers) are not detected, that the surgery in surgery.py shares the same limitation so the two stay consistent, and that handling `Constant`-node weights would require changes on both sides.

tests/unit/optim/pipes/test_pipe_surgery.py
xieofxie: Added in 79bcc3f5: `test_duplicate_node_names_do_not_collide` builds a model with two batched-const MatMuls that both have empty names, then asserts the result has two distinct Add outputs, passes `onnx.checker.check_model`, and matches the original model numerically on ORT CPU.

src/winml/modelkit/analyze/core/model_validators/batched_const_matmul_validator.py
xieofxie: because constant node will be simplified to inititializer, so not handled

src/winml/modelkit/analyze/core/model_validators/model_validator_manager.py
```diff
@@ -64,6 +65,11 @@ class ModelValidatorManager:
             "class": PatternMatchingValidator,
             "enabled_devices": None,  # All devices
         },
+        "batched_const_matmul": {
```
DingmaomaoBJTU, line 69: Please use static rules

src/winml/modelkit/optim/pipes/surgery.py
```diff
@@ -57,13 +58,16 @@ class SurgeryPipeConfig(PipeConfig):
...
+        untie_constant_batched_matmul: Make a batched MatMul's constant operand
+            runtime-valued so OpenVINO GPU can select a gemm implementation
         verbose: Enable verbose logging
     """
 
     clamp_constant_values: bool = False
     clamp_min: float = -1e3
     clamp_max: float = 1e3
     remove_isnan_in_attention_mask: bool = False
+    untie_constant_batched_matmul: bool = False
```
DingmaomaoBJTU, line 70: Change to rewrite pipe？

DingmaomaoBJTU (approved):
LGTM! All 6 previously flagged issues resolved. The ModelValidatorManager API change (device/ep now mandatory kwargs, needs_context removed) is a clean improvement.

src/winml/modelkit/optim/pipes/surgery.py
DingmaomaoBJTU: ✅ Fixed — loop now uses `enumerate(targets)`, producing `dyn = f"{prefix}_untied{untie_idx}_in{idx}"`. Unique names are guaranteed regardless of whether node names are blank or duplicated.

src/winml/modelkit/optim/pipes/surgery.py
DingmaomaoBJTU: ✅ Fixed — docstring updated to accurately describe the actual construction: `Cast(first_input → float) → Reshape([-1]) → Slice([0:1]) → Sub(elem, elem)`.

src/winml/modelkit/optim/pipes/surgery.py
DingmaomaoBJTU: ✅ Fixed — split into three distinct initializers: `_slice_starts` ([0]), `_slice_ends` ([1]), and `_slice_axis` ([0]), each with a descriptive name so neither role can be silently corrupted by a future edit.

src/winml/modelkit/optim/pipes/surgery.py
DingmaomaoBJTU: ✅ Fixed — explicit "Assumption" note added to the docstring: the first graph input must have at least one element at runtime; a zero-batch feed would cause an out-of-bounds Slice error.

src/winml/modelkit/analyze/core/model_validators/batched_const_matmul_validator.py
DingmaomaoBJTU: ✅ Fixed — "Known gap" comment added to the validator documenting that `Constant` op nodes are not detected, and that this is consistent with the surgery side. Most exporters and ORT preprocessing emit weights as initializers, so the fix covers the target case in practice.

tests/unit/optim/pipes/test_pipe_surgery.py
DingmaomaoBJTU: ✅ Fixed — `test_duplicate_node_names_do_not_collide` added: builds two unnamed MatMul nodes (both `name=""`), runs the surgery, asserts two distinct Add nodes, passes `onnx.checker.check_model`, and verifies numerics via ORT.

PR #821, opened 2026-06-08
feat: add --allow-unsupported-nodes to perf/build/eval/run

xieofxie:
## Problem

`winml perf -m depth-anything/Depth-Anything-V2-Small-hf --task depth-estimation --ep openvino --device gpu` aborts at the optimize/analyze loop:

```
RuntimeError: Unsupported nodes persist after N analyze pass(es): ['OP/ai.onnx/Resize']
```

The static analyzer flags nodes the target EP doesn't fully support (e.g. `Resize` on OpenVINO GPU), so the build raises — even though the EP can still run the model (CPU fallback for those nodes).

## Change

Add an opt-in `--allow-unsupported-nodes` flag that turns the hard failure into a warning and lets the build proceed. The gate lives at the single `raise` in `build/common.py`:

```python
if analysis is not None and analysis.has_errors:
    message = f"Unsupported nodes persist after {analyze_iterations} analyze pass(es): ..."
    if allow_unsupported_nodes:
        logger.warning("%s. Continuing anyway ...", message)
    else:
        raise RuntimeError(message)
```

The flag is a **shared CLI option** (`cli_utils.allow_unsupported_nodes_option()`) applied to every command that builds a model via `WinMLAutoModel.from_*`:

| Command | Path | Status |
|---|---|---|
| `winml perf` | `BenchmarkConfig` → `from_pretrained`/`from_onnx`; `--module` → `build_hf_model` | ✅ verified e2e |
| `winml build` | `_run_optimize_stage` → `run_optimize_analyze_loop`; module mode → `build_hf_model` | ✅ verified e2e |
| `winml eval` | `WinMLEvaluationConfig.allow_unsupported_nodes` → `from_pretrained` | ✅ |
| `winml run` | `InferenceEngine.load` → `_load_from_hf` → `from_pretrained` | plumbed (command currently disabled) |

ONNX / `skip_build` load paths don't run the analyzer, so they're untouched. Default is `False` everywhere — strictly opt-in. Named `--allow-unsupported-nodes` (not a generic `--force`) since it gates exactly this analyzer check.

## Verification

- `winml perf ... --ep openvino --device gpu --allow-unsupported-nodes` → builds, compiles on OV GPU, benchmarks (~15.9 ms avg).
- `winml build ... --ep openvino --device gpu --allow-unsupported-nodes` → logs the warning (`['OP/ai.onnx/Resize']`) and completes.
- New/updated unit tests cover the gate (`build/test_common.py`), and flag passthrough for build/eval/run; full commands + inference + build + eval suites pass (871 tests).

### DIFF
--- a/src/winml/modelkit/commands/build.py
+++ b/src/winml/modelkit/commands/build.py
@@ -490,13 +490,7 @@ def _validate_loader_tasks_for_model(
     default=None,
     help="Maximum autoconf re-optimization rounds (default: 3). --no-analyze sets this to 0.",
 )
-@click.option(
-    "--allow-unsupported-nodes",
-    is_flag=True,
-    default=False,
-    help="Continue the build instead of failing when the analyzer reports "
-    "unsupported nodes (the EP may fall back to another device for them).",
-)
+@cli_utils.allow_unsupported_nodes_option()
 @cli_utils.trust_remote_code_option(
     optional_message="Trust remote code for custom model architectures (e.g., Mu2)."
 )
--- a/src/winml/modelkit/commands/build.py
+++ b/src/winml/modelkit/commands/build.py
@@ -669,8 +669,10 @@ def build(
             extra_kwargs["hack_max_optim_iterations"] = max_optim_iterations
         if trust_remote_code:
             extra_kwargs["trust_remote_code"] = True
-        if allow_unsupported_nodes:
-            extra_kwargs["allow_unsupported_nodes"] = True
+        # Always set (even when False) so downstream pipeline functions can rely
+        # on the key being present, matching the module-mode path which passes
+        # allow_unsupported_nodes explicitly regardless of its value.
+        extra_kwargs["allow_unsupported_nodes"] = allow_unsupported_nodes
 
         if isinstance(config_or_configs, list):
             # ---- MODULE MODE: array config, one build per submodule ----
--- a/src/winml/modelkit/commands/build.py
+++ b/src/winml/modelkit/commands/build.py
@@ -166,6 +166,7 @@ def _build_modules(
     rebuild: bool = False,
     ep: EPNameOrAlias | None = None,
     device: str | None = None,
+    allow_unsupported_nodes: bool = False,
 ) -> list[BuildResult]:
     """Build each module config using init-weight parent for submodule extraction.
 
@@ -182,6 +183,8 @@ def _build_modules(
         rebuild: If True, overwrite existing artifacts.
         ep: Target execution provider for analyzer.
         device: Target device for analyzer.
+        allow_unsupported_nodes: If True, warn instead of failing the build when
+            the analyzer reports unsupported nodes that persist.
 
     Returns:
         List of BuildResult in the same order as *configs*.
@@ -222,6 +225,7 @@ def _build_modules(
             rebuild=rebuild,
             ep=ep,
             device=device,
+            allow_unsupported_nodes=allow_unsupported_nodes,
         )
         results.append(result)
 
@@ -486,6 +490,13 @@ def _validate_loader_tasks_for_model(
     default=None,
     help="Maximum autoconf re-optimization rounds (default: 3). --no-analyze sets this to 0.",
 )
+@click.option(
+    "--allow-unsupported-nodes",
+    is_flag=True,
+    default=False,
+    help="Continue the build instead of failing when the analyzer reports "
+    "unsupported nodes (the EP may fall back to another device for them).",
+)
 @cli_utils.trust_remote_code_option(
     optional_message="Trust remote code for custom model architectures (e.g., Mu2)."
 )
@@ -505,6 +516,7 @@ def build(
     device: str,
     no_analyze: bool,
     max_optim_iterations: int | None,
+    allow_unsupported_nodes: bool,
     trust_remote_code: bool,
     verbose: int,
     quiet: bool,
@@ -663,6 +675,8 @@ def build(
             extra_kwargs["hack_max_optim_iterations"] = max_optim_iterations
         if trust_remote_code:
             extra_kwargs["trust_remote_code"] = True
+        if allow_unsupported_nodes:
+            extra_kwargs["allow_unsupported_nodes"] = True
 
         if isinstance(config_or_configs, list):
             # ---- MODULE MODE: array config, one build per submodule ----
@@ -697,6 +711,7 @@ def build(
                 rebuild=rebuild,
                 ep=ep,
                 device=device,
+                allow_unsupported_nodes=allow_unsupported_nodes,
             )
 
             # Report per-module results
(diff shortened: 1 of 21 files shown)


xieofxie:
Note: although AI says ` (CPU fallback for those nodes)` , the op profiler find that gpu supports it as interpolate type

src/winml/modelkit/inference/engine.py
```diff
@@ -330,13 +333,25 @@ def load(
...
+                        model_id,
+                        task=task,
+                        device=device,
+                        ep=ep,
+                        allow_unsupported_nodes=allow_unsupported_nodes,
+                    )
                 else:
                     raise
         elif path.suffix == ".onnx" and path.exists():
             self._load_from_onnx(path, task=task, device=device, ep=ep)
```
DingmaomaoBJTU, line 348: `allow_unsupported_nodes` is also silently dropped in the `_load_from_build_dir` success path (line 322). When the build dir already contains a valid cached ONNX, the flag has no effect — which is correct since there is no build/analyze step to gate. However the docstring only says ''during an HF build'' without calling out the dir-cache and .onnx cases. A reader who passes `--allow-unsupported-nodes` when pointing at a pre-built cache dir may wonder why no warning is ever emitted. Suggest adding: `Note: has no effect when loading from a pre-built ONNX file or a cached build directory (no build/analyze step runs in those paths).`

src/winml/modelkit/commands/build.py
```diff
@@ -663,6 +669,8 @@ def _patch_device(cfg: WinMLBuildConfig) -> None:
             extra_kwargs["hack_max_optim_iterations"] = max_optim_iterations
         if trust_remote_code:
             extra_kwargs["trust_remote_code"] = True
+        if allow_unsupported_nodes:
+            extra_kwargs["allow_unsupported_nodes"] = True
```
DingmaomaoBJTU: Only inserting `allow_unsupported_nodes` into `extra_kwargs` when `True` creates an implicit contract: every downstream pipeline function that consumes it must use `.pop('allow_unsupported_nodes', False)` (with the default). The current code is correct, but a future pipeline function that uses a plain `.pop('allow_unsupported_nodes')` would raise `KeyError` only when the flag is **off**. In module mode (line 711), the value is always passed explicitly regardless of its value. Consistently passing `allow_unsupported_nodes` (even when False) across both paths — or at least a comment explaining the intentional absence — would make the contract safer to extend.

tests/unit/build/test_common.py
```diff
@@ -0,0 +1,113 @@
...
+                max_optim_iterations=1,
+            )
+
+    def test_warns_instead_of_raising_when_allowed(self, tmp_path: Path, caplog) -> None:
+        model = tmp_path / "in.onnx"
+        model.write_text("mock")
+        optimized = tmp_path / "out.onnx"
+        p_opt, p_analyze, p_copy = _patched_loop(tmp_path)
+
+        with p_opt, p_analyze, p_copy, caplog.at_level("WARNING"):
```
DingmaomaoBJTU: `caplog.at_level('WARNING')` without a `logger=` kwarg captures WARNING records from *all* loggers. The assertion `'Unsupported nodes persist' in caplog.text` is correct, but another module logging a WARNING during the test could cause a false-positive pass in theory. Prefer `caplog.at_level('WARNING', logger='winml.modelkit.build.common')` to pin capture to exactly the module under test.

src/winml/modelkit/inference/engine.py
xieofxie: Good call — added the note to the `load()` docstring clarifying the flag has no effect on the pre-built ONNX and cached build-dir paths (no analyze step runs there). Fixed in 5c4f5841.

src/winml/modelkit/commands/build.py
xieofxie: Agreed. Now setting `extra_kwargs['allow_unsupported_nodes']` unconditionally (even when False) so the downstream `.pop` default is no longer load-bearing and the single-model path matches module mode. Fixed in 5c4f5841.

tests/unit/build/test_common.py
xieofxie: Done — pinned `caplog.at_level('WARNING', logger='winml.modelkit.build.common')` so capture is scoped to the module under test. Fixed in 5c4f5841.

src/winml/modelkit/models/auto.py
```diff
@@ -431,6 +436,7 @@ def from_pretrained(
             cache_key=cache_key,
             ep=resolved_ep,
             device=device,
+            allow_unsupported_nodes=allow_unsupported_nodes,
```
DingmaomaoBJTU, line 445: `allow_unsupported_nodes` is forwarded for both the ONNX-file path (line 318) and the HF-build path (line 439), but the composite model shortcut at line 345 (`WinMLCompositeModel.from_pretrained`) is skipped entirely. Since `allow_unsupported_nodes` is a *named* parameter of `from_pretrained` it is not part of `**kwargs`, so it is silently dropped when the composite path is taken (T5, Qwen, etc.). Should add `allow_unsupported_nodes=allow_unsupported_nodes` to that call.

src/winml/modelkit/models/auto.py
```diff
@@ -311,6 +315,7 @@ def from_pretrained(
                 cache_dir=cache_dir,
                 use_cache=use_cache,
                 force_rebuild=force_rebuild,
+                allow_unsupported_nodes=allow_unsupported_nodes,
```
DingmaomaoBJTU, line 323: `from_onnx` doesn't declare `allow_unsupported_nodes` in its signature — it flows through `**kwargs` into `build_onnx_model` which pops it. This works, but callers who look at the signature of `from_onnx` (e.g., for type checking or IDE completion) won't know the parameter exists. Consider adding it as an explicit keyword-only param parallel to `skip_build` (already explicit on this method). At minimum, add it to the docstring's `**kwargs` section.

src/winml/modelkit/models/auto.py
xieofxie: Good catch — this was a real drop for composite models (T5/Qwen/etc.). `allow_unsupported_nodes` is now passed to `WinMLCompositeModel.from_pretrained`, which forwards it through to each sub-model's `WinMLAutoModel.from_pretrained`. Added a regression test for the composite dispatch path. Fixed in f156c7d0.

src/winml/modelkit/models/auto.py
xieofxie: Done — `from_onnx` now declares `allow_unsupported_nodes` as an explicit keyword-only param (parallel to `skip_build`), documents it, and forwards it explicitly to `build_onnx_model` rather than via `**kwargs`. Added a forwarding test. Fixed in f156c7d0.

DingmaomaoBJTU (approved):

Issue #824, opened 2026-06-08
bug: non exist code are imported


xieofxie:
See https://github.com/microsoft/winml-cli/blob/cf25bfd41f8a22a4d35b24936c1b90a25019674f/src/winml/modelkit/core/unified_optimizer.py#L267

apply_htp_optimizations

apply_usage_based_optimizations

xieofxie:
Also this file https://github.com/microsoft/winml-cli/blob/main/src/winml/modelkit/core/universal_hierarchy_exporter.py look like dead code

DingmaomaoBJTU:
**Triage summary**

- **Component**: Optimizer
- **Priority**: P2

`unified_optimizer.py` imports `apply_htp_optimizations` and `apply_usage_based_optimizations` which do not exist. These are dead imports that should be removed or the missing functions should be implemented.

DingmaomaoBJTU:
## Triage Summary
Priority: **P2**
Component: graph-optimizer
Owner: @DingmaomaoBJTU

**Analysis**: `unified_optimizer.py` imports `apply_htp_optimizations` and `apply_usage_based_optimizations` which no longer exist in the codebase. These are dead imports that will cause `ImportError` if the code paths are exercised.

**Action**: Remove or replace the stale imports. If the optimization functions were intentionally removed, clean up all references. If they were moved, update the import paths.

DingmaomaoBJTU:
## Triage Summary (2026-06-25)

**Priority**: P2 (confirmed — dead code import, not user-facing crash but code hygiene issue)
**Component**: graph-optimizer / unified_optimizer
**Root Cause**: `apply_htp_optimizations` and `apply_usage_based_optimizations` are imported but functions do not exist.
**Owner**: @xieofxie
**Action**: Remove dead imports or implement the missing functions if they were planned.

_Triaged by automated daily triage._

PR #825, opened 2026-06-08
chore(typecheck): clean config/core/data; enforce mypy in CI for cleaned folders

xieofxie:
## Summary
- Clear all mypy errors in `winml.modelkit.{config,core,data}` (90 → 0). Fixes follow the project's "narrow at the source, don't widen at the call site" preference: `QuantType` Literal added to `precision.py` and propagated; `policy.weight_type`/`activation_type` correlation made explicit; `loader_config.task`/`model_type` assertions after `resolve_loader_config`; `tag_utils.py` `any` (builtin) → `typing.Any` cleared 22 cascading errors; missing `__init__`/`__len__`/`__getitem__` annotations across the data datasets.
- Annotate `cli_utils.verbosity_options()` as `Callable[[F], F]` — one source fix clears 11 cascading `[untyped-decorator]` errors across `commands/`.
- Add a **required** mypy CI step over `analyze`, `build`, `cache`, `commands`, `compiler`, `config`, `core`, `data`. New errors in those folders now block PRs. The existing full-package mypy step stays **advisory** for the rest of the tree while the backlog is worked down.
- Switch CI dependency install from `uv sync --all-extras` → `uv sync --all-extras --all-groups` so type stubs in `[dependency-groups] dev` (`pandas-stubs`, `types-jsonschema`, etc.) are actually installed.

## Known follow-ups
Six `# type: ignore[import-not-found]` were added to silence imports of modules that don't exist in the repo: `..strategies.{htp,usage_based,fx}[.optimizations]` in `core/unified_optimizer.py` and `..constants` in `core/universal_hierarchy_exporter.py`. The first two `try/except ImportError`-guarded ones are graceful degradation; the rest are unguarded and will raise at runtime if reached. Docstrings hint at "iterations 17-18" experimental code — worth deciding whether to implement or delete in a follow-up.


src/winml/modelkit/core/universal_hierarchy_exporter.py
```diff
@@ -1,881 +1,881 @@
...
+
+def export_bert_tiny_with_validation() -> dict[str, Any]:
+    """Export BERT-tiny and validate against ground truth.
+
+    This demonstrates the complete workflow following all requirements.
+    """
+    from transformers import AutoModel, AutoTokenizer
+
+    # Load model (CARDINAL RULE: NO HARDCODED LOGIC - this works with any HF model)
+    from ..constants import DEFAULT_TEST_MODEL  # type: ignore[import-not-found]
```
xieofxie: tracked via https://github.com/microsoft/winml-cli/issues/824

src/winml/modelkit/core/unified_optimizer.py
```diff
@@ -1,424 +1,426 @@
...
+            profile.optimizations_applied.append("batch_processing")
+
+    def _apply_strategy_optimizations(
+        self, exporter: Any, strategy_name: str, profile: OptimizationProfile
+    ) -> None:
+        """Apply strategy-specific optimizations."""
+        if strategy_name == "htp":
+            # Apply HTP-specific optimizations from iteration 17
+            try:
+                from ..strategies.htp.optimizations import (  # type: ignore[import-not-found]
```
xieofxie: tracked via https://github.com/microsoft/winml-cli/issues/824

.github/workflows/lint.yml
```diff
@@ -34,11 +34,23 @@ jobs:
...
-      # Type-checking is currently advisory: src/ has many pre-existing
-      # mypy errors against the strict config in pyproject.toml. The job
-      # surfaces type issues in CI logs without blocking PRs while the
-      # backlog is worked down. Flip continue-on-error to false (or drop
-      # it) once src/ is clean.
-      - name: Type check (advisory)
+      # Required type check: these folders are clean against the strict
+      # config in pyproject.toml. Any new mypy error here blocks the PR.
+      # Expand the package list as more folders are cleaned up.
+      - name: Type check (required)
```
timenick, line 40: I ran `uv run mypy -p <pkg>` against HEAD `6982a63` for each of the eight packages in this required step. Only five pass today:

- `winml.modelkit.build` ✅
- `winml.modelkit.cache` ✅
- `winml.modelkit.compiler` ✅
- `winml.modelkit.config` ✅ (cleaned by this PR)
- `winml.modelkit.data` ✅ (cleaned by this PR)
- `winml.modelkit.analyze` ❌ many errors (missing pandas/tqdm stubs, real `[no-any-return]` errors in `runtime_checker_query.py`, etc.)
- `winml.modelkit.commands` ❌ 2 errors (`commands/optimize.py` needs `types-PyYAML`, `commands/_live_chart.py:136` skipping unanalyzed module)
- `winml.modelkit.core` ❌ 6 errors — all of them are `Unused "type: ignore"` on the very `# type: ignore[import-not-found]` lines this PR adds to `core/unified_optimizer.py` and `core/universal_hierarchy_exporter.py` (see other comment).

As written this required job will be red the moment it merges. Two safe paths:

1. Narrow the required list to just the packages that pass today (`build`, `cache`, `compiler`, `config`, `data`) and add `analyze`/`commands`/`core` in follow-up PRs once they're actually clean.
2. Clean up `analyze`/`commands` and the `core` unused-ignores in this PR before merging.

src/winml/modelkit/core/unified_optimizer.py
```diff
@@ -264,7 +264,9 @@ def _apply_strategy_optimizations(
         if strategy_name == "htp":
             # Apply HTP-specific optimizations from iteration 17
             try:
-                from ..strategies.htp.optimizations import apply_htp_optimizations
+                from ..strategies.htp.optimizations import (  # type: ignore[import-not-found]
```
timenick, line 267: All six `# type: ignore[import-not-found]` added in this file and in `core/universal_hierarchy_exporter.py:828` reference modules that don''t exist in the repo:

- `..strategies.htp.optimizations`, `..strategies.usage_based.optimizations`, `..strategies.usage_based`, `..strategies.htp`, `..strategies.fx` — no `src/winml/modelkit/strategies/` directory exists.
- `..constants` (used in `universal_hierarchy_exporter.py:828`) — no `constants.py` at the modelkit root.

And the surrounding code is dead — I greped `src/` and `tests/`: `create_optimized_exporter`, `_apply_strategy_optimizations` (only called from `optimize_exporter` which is itself uncalled), `compare_optimized_vs_original`, and `export_bert_tiny_with_validation` have zero callers anywhere.

Mypy correctly flags these ignores as `Unused "type: ignore"` because the missing modules are silently ignored (they hit the default `ignore_missing_imports` path), not raised as `[import-not-found]`. So adding `[import-not-found]` is a workaround that mypy itself rejects under this repo''s strict config.

Cleanest fix: delete the dead code paths (and shrink `core/__init__.py` accordingly). That also makes `core/` actually clean for the new required CI step, instead of needing a different ignore code.

src/winml/modelkit/config/build.py
```diff
@@ -217,9 +217,9 @@ def validate(self) -> None:
             errors.append("loader.task is required for full model builds")
         # export=None is valid for ONNX builds
 
-        # 2. optim config always required
+        # 2. optim config always required (runtime callers may pass None despite the type)
         if self.optim is None:
-            errors.append("optim config is required")
+            errors.append("optim config is required")  # type: ignore[unreachable]
```
timenick, line 222: The comment and the `# type: ignore[unreachable]` contradict each other:

```python
# 2. optim config always required (runtime callers may pass None despite the type)
if self.optim is None:
    errors.append("optim config is required")  # type: ignore[unreachable]
```

The comment says the check is a real safety net against runtime callers passing `None`. The `[unreachable]` ignore tells mypy the branch is dead code that we want to silence. One of them is wrong:

- If runtime callers really can pass `None`, the field annotation should be `WinMLOptimizationConfig | None` and the check is genuine — drop the ignore.
- If they truly can''t, delete the check (and the comment).

Keeping both is the worst of both worlds: a half-disabled type system plus a runtime guard the next reader can''t tell is alive or dead.

src/winml/modelkit/config/build.py
```diff
@@ -800,20 +804,24 @@ class name (HF path only).
...
+        "model_type": model_type,
+        "override": override,
+        "shape_config": shape_config,
+        "library_name": library_name,
+        "device": device,
+        "precision": precision,
+        "trust_remote_code": trust_remote_code,
+        "ep": ep,
+    }
+    if module is None:
```
timenick, line 822: Two notes on the branch split + `**common_kwargs: dict[str, Any]` workaround:

1. `**common_kwargs: dict[str, Any]` widens every keyword type to `Any` at the callsite. So `task`, `model_class`, `precision` etc. are no longer statically checked against `generate_hf_build_config`''s signature here — mypy will accept anything for those fields and silently let bugs through. The comment says this is intentional ("per-kwarg type checks happen at the callee"), but it''s only true _inside_ the callee body; the contract _at the callsite_ is now `Any → Any`.
2. The branching itself only exists to give mypy a literal `module=None` vs `module=<concrete>` at the call. If `generate_hf_build_config` really needs different signatures depending on `module`, the right fix is `@overload` decorators on _the callee_, which preserves call-site checking and avoids the 16-line dict construction. If it doesn''t actually have different signatures, the original single `return generate_hf_build_config(...)` was fine and mypy should be made happy a different way (e.g., a single `assert module is None or isinstance(module, str)`).

Not blocking, but worth flagging — this pattern will spread if it''s adopted as the canonical "how we handle non-overloaded callees" idiom.

timenick (commented):
Type-check cleanup PR. Four notes below — most importantly, the new `Type check (required)` step doesn''t actually pass against this PR''s HEAD on three of the eight packages it lists.

src/winml/modelkit/core/unified_optimizer.py
xieofxie: already tracked via https://github.com/microsoft/winml-cli/issues/824 could handle in a new pr

src/winml/modelkit/config/build.py
xieofxie: Currently they are kept because not all codes are checked against mypy now, so many redundant checks exist. 
Create an issue to track https://github.com/microsoft/winml-cli/issues/830

.github/workflows/lint.yml
xieofxie: you forget to install [dev] dependecies

timenick (approved):

PR #832, opened 2026-06-08
Fix rich Live thread crashes.

### LINKED ISSUE #795: winml build --compile: rich Live thread crashes with OSError [WinError 1] Incorrect function on Windows
## Command

```
winml build -m microsoft/resnet-50 -o ... --ep openvino --device gpu --compile
```

## Log

A `rich` Live display thread crashes while refreshing the console, raising `OSError: [WinError 1] Incorrect function`:

```
Exception in thread Thread-3:
Traceback (most recent call last):
  File "...\Lib\threading.py", line 1075, in _bootstrap_inner
    self.run()
  File "...\Lib\site-packages\rich\live.py", line 38, in run
    self.live.refresh()
  File "...\Lib\site-packages\rich\live.py", line 270, in refresh
    with self.console:
         ^^^^^^^^^^^^
  File "...\Lib\site-packages\rich\console.py", line 864, in __exit__
    self._exit_buffer()
  File "...\Lib\site-packages\rich\console.py", line 820, in _exit_buffer
    self._check_buffer()
  File "...\Lib\site-packages\rich\console.py", line 2055, in _check_buffer
    self._write_buffer()
  File "...\Lib\site-packages\rich\console.py", line 2103, in _write_buffer
    write(text)
OSError: [WinError 1] Incorrect function
```

## Notes

- Occurs on Windows.
- The exception is raised from `rich`'s background `Live` refresh thread when writing to the console buffer.

chinazhangchao:
Issue:
https://github.com/microsoft/winml-cli/issues/795

**Root cause:** Inside the Rich `Live` refresh thread, a stderr write fails with `OSError: [WinError 1]` ("Incorrect function"). On Windows this typically happens when a native EP dependency (here OpenVINO inside `compile_onnx`'s `session.run` validation) puts the console handle into a state where ANSI/control writes no longer work. Rich's daemon refresh thread doesn't catch this, dies, and Python dumps its traceback into the middle of build output. The build itself completes — the visible log lines after the traceback are proof of that.

**Fix:** console.py

- Added `_SafeLive`, a `rich.live.Live` subclass that catches `OSError` inside `refresh()`. After the first failure it disables further refreshes for that Live (logged at DEBUG), so the daemon thread can't die or spam tracebacks.
- Added a tiny `_safe_call` helper and wrapped `StageLive.__enter__` / `__exit__` / `_update` Live operations with it, so any console hiccup on `start`/`update`/`stop` is also swallowed.
- Net effect: the stage may stop animating its final frame after such an event, but the surrounding pipeline continues cleanly with no traceback noise.

### DIFF
--- a/src/winml/modelkit/utils/console.py
+++ b/src/winml/modelkit/utils/console.py
@@ -266,6 +266,48 @@ def get_onnx_total_size(onnx_path: Path) -> int:
 # ══════════════════════════════════════════════════════════════════════════
 
 
+class _SafeLive(Live):
+    """A :class:`rich.live.Live` that survives Windows console hiccups.
+
+    Some native dependencies (e.g. OpenVINO / ONNX Runtime EPs) can put
+    the Windows console handle into a state where ANSI/control writes
+    fail with ``OSError: [WinError 1]`` ("Incorrect function").  That
+    error is raised from Rich's daemon refresh thread, which then dies
+    and dumps an ugly traceback into the middle of build output even
+    though the build itself succeeds.
+
+    This subclass catches :class:`OSError` in :meth:`refresh` and, after
+    the first failure, disables further refreshes for the lifetime of
+    this Live instance.  The visible effect is that the final frame may
+    not animate further, but no traceback escapes and the surrounding
+    pipeline continues normally.
+    """
+
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
+        super().__init__(*args, **kwargs)
+        self._refresh_disabled: bool = False
+
+    def refresh(self) -> None:  # type: ignore[override]
+        if self._refresh_disabled:
+            return
+        try:
+            super().refresh()
+        except OSError:
+            # Console handle is unusable (typically VT/handle state damaged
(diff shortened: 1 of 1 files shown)


src/winml/modelkit/utils/console.py
```diff
@@ -266,6 +266,48 @@ def get_onnx_total_size(onnx_path: Path) -> int:
...
+            super().refresh()
+        except OSError:
+            # Console handle is unusable (typically VT/handle state damaged
+            # by a native library).  Disable further refreshes so the
+            # daemon thread does not spam tracebacks.
+            self._refresh_disabled = True
+            logger.debug("Disabling Live refresh after OSError", exc_info=True)
+
+
+def _safe_call(func: Any, *args: Any, **kwargs: Any) -> None:
```
vortex-captain: nit: maybe it would be slightly cleaner to define _safe_call as a decorator inside _SafeLive, and decorate `start, stop, update`, so caller of `Live` would use `_SafeLive` in the same way.

vortex-captain (approved):

PR #834, opened 2026-06-08
chore(typecheck): clean datasets/eval/export/inference; enforce mypy on 12 packages

xieofxie:
## Summary

Extend the strict-mypy required-CI gate from 8 → 12 packages by cleaning `datasets/`, `eval/`, `export/`, and `inference/` (189 errors → 0). Also restructures the dispatcher pattern in `config/build.py` to fix a regression timenick flagged on #825 (kwarg type-checking lost behind `**dict[str, Any]`).

### Dispatcher / overload cleanup (`config/build.py`, `commands/config.py`)
- `generate_hf_build_config`: added a third `@overload` accepting `module: str | None` and returning the union of the two narrow returns. Lets the `generate_build_config` dispatcher collapse to a single call (no kwarg duplication, no `Any` widening, no `cast`).
- `commands.config` command: ripped out the `_shared_kwargs: dict[str, Any] = {...}` workaround and replaced the `if module:` discriminator with `isinstance(result, list)` narrowing. Restores full call-site type checking against the callee's signature.

### Cleaning 4 new folders (189 → 0)
Notable source-level fixes that cascaded:
- **`core/time_utils.format_timestamp_iso`** — added `@overload`s so `float → str` and `None → None` propagate; cleared 3 downstream `[arg-type]` chains.
- **`export/htp/base_writer.step` decorator** — replaced `-> Any` factory with `Callable[[F], F]`, dropped no-op `@wraps` wrapper. One source fix cleared 18 cascading `[untyped-decorator]` errors across writers.
- **`export/htp/exporter`** — typed the `None`-init core fields with proper `Optional` annotations + single non-None `assert` in `_apply_hierarchy_tags`; cleared the union-attr chain.
- **`datasets/*`** — common pattern: class-level `_image_col: str = ""` default instead of runtime `None`-init in `_detect_columns`; eliminates `Any | None` property-return errors across image/segmentation/detection/depth datasets.
- **`eval/*`** — `BaseImageProcessor | None` guards across evaluators; `pipeline()` call-overload suppressions for runtime task strings that can't be statically matched against transformers' 60+ Literal overloads.

### Configuration
- `pyproject.toml` `ignore_missing_imports`: added `soundfile`, `sklearn.*`, `evaluate(.*)`.
- Dev deps: added `scipy-stubs`, `types-psutil`.
- `.github/workflows/lint.yml`: switched dependency install to `uv sync --all-extras --all-groups` so type stubs in `[dependency-groups] dev` are present in CI. Extended the required mypy loop from 8 to 12 packages.

## Tracked follow-ups

- **#859** — two unguarded `winml.modelkit.inference.onnx_config.*` imports in `export/htp/config_generator.py` reference a subpackage that doesn't exist in the repo. Silenced with `# type: ignore[import-not-found]` referencing the issue; #859 tracks delete-vs-restore decision.

### DIFF
--- a/src/winml/modelkit/commands/config.py
+++ b/src/winml/modelkit/commands/config.py
@@ -326,26 +326,25 @@ def config(
                 )
                 return
 
-            # Generate config(s) - module parameter selects overload:
-            # module=str → list[WinMLBuildConfig], module=None → WinMLBuildConfig.
-            # ``module`` is the only differing kwarg, so build a shared dict
-            # once and add it only on the list-returning branch. This keeps
-            # the overload dispatch but avoids repeating the other 10 kwargs.
-            _shared_kwargs: dict[str, Any] = {
-                "model_id": hf_model,
-                "task": task,
-                "model_class": model_class,
-                "model_type": model_type,
-                "override": override,
-                "shape_config": shape_config,
-                "library_name": library_name,
-                "device": device,
-                "precision": precision,
-                "trust_remote_code": trust_remote_code,
-                "ep": ep,
-            }
-            if module:
-                configs = generate_hf_build_config(module=module, **_shared_kwargs)
+            # Generate config(s). The ``module: str | None`` overload of
+            # generate_hf_build_config returns WinMLBuildConfig | list[...],
+            # which isinstance(result, list) narrows for the branches below.
+            result = generate_hf_build_config(
+                model_id=hf_model,
+                task=task,
+                model_class=model_class,
+                model_type=model_type,
+                module=module,
+                override=override,
+                shape_config=shape_config,
+                library_name=library_name,
+                device=device,
+                precision=precision,
+                trust_remote_code=trust_remote_code,
+                ep=ep,
+            )
+            if isinstance(result, list):
+                configs = result
                 for cfg in configs:
                     _apply_stage_overrides(cfg, no_quant=no_quant, no_compile=no_compile)
                 output_data = [cfg.to_dict() for cfg in configs]
@@ -353,7 +352,7 @@ def config(
                 # Use first config for display metadata
                 config_obj = configs[0] if configs else None
             else:
-                config_obj = generate_hf_build_config(**_shared_kwargs)
+                config_obj = result
                 configs = []
                 _apply_stage_overrides(config_obj, no_quant=no_quant, no_compile=no_compile)
                 output_data = config_obj.to_dict()
--- a/src/winml/modelkit/config/build.py
+++ b/src/winml/modelkit/config/build.py
@@ -489,6 +489,29 @@ def generate_hf_build_config(
 ) -> list[WinMLBuildConfig]: ...
 
 
+@overload
+def generate_hf_build_config(
+    model_id: str | None = None,
+    *,
+    task: str | None = None,
+    model_class: str | None = None,
+    model_type: str | None = None,
+    # Catch-all for callers that hold ``module`` as ``str | None`` (e.g. the
+    # ``generate_build_config`` dispatcher). Without this overload, mypy can't
+    # resolve the call against the two narrower overloads above and fails with
+    # "too many union combinations".
+    module: str | None,
+    override: WinMLBuildConfig | None = None,
+    shape_config: dict | None = None,
+    library_name: str = "transformers",
+    device: str = "auto",
+    precision: str = "auto",
+    trust_remote_code: bool = False,
+    ep: EPNameOrAlias | None = None,
+    no_compile: bool = False,
+) -> WinMLBuildConfig | list[WinMLBuildConfig]: ...
+
+
 def generate_hf_build_config(
     model_id: str | None = None,
     *,
@@ -804,24 +827,24 @@ def generate_build_config(
             ep=ep,
             override=override,
         )
-    # Split branches so mypy can pick the matching overload of generate_hf_build_config.
-    # Typed as dict[str, Any] so per-kwarg type checks happen at the callee, not on the
-    # widened Union mypy would otherwise infer from this heterogeneous literal.
-    common_kwargs: dict[str, Any] = {
-        "task": task,
-        "model_class": model_class,
-        "model_type": model_type,
-        "override": override,
-        "shape_config": shape_config,
-        "library_name": library_name,
-        "device": device,
-        "precision": precision,
-        "trust_remote_code": trust_remote_code,
-        "ep": ep,
-    }
-    if module is None:
-        return generate_hf_build_config(model_id, module=None, **common_kwargs)
-    return generate_hf_build_config(model_id, module=module, **common_kwargs)
+    # Single call resolves against generate_hf_build_config's `module: str | None`
+    # overload, which returns WinMLBuildConfig | list[WinMLBuildConfig] — matching
+    # this dispatcher's implementation return type. The dispatcher's own
+    # narrowing overloads above still tighten the return type for its callers.
+    return generate_hf_build_config(
+        model_id,
+        task=task,
+        model_class=model_class,
+        model_type=model_type,
+        module=module,
+        override=override,
+        shape_config=shape_config,
+        library_name=library_name,
+        device=device,
+        precision=precision,
+        trust_remote_code=trust_remote_code,
+        ep=ep,
+    )
 
 
 # =============================================================================
--- a/src/winml/modelkit/datasets/processor_utils.py
+++ b/src/winml/modelkit/datasets/processor_utils.py
@@ -13,7 +13,9 @@ from __future__ import annotations
 import logging
 from typing import Any
 
-from transformers.image_processing_utils import ImageProcessingMixin
+from transformers.image_processing_utils import (  # type: ignore[attr-defined]
+    ImageProcessingMixin,
+)
 
 
 logger = logging.getLogger(__name__)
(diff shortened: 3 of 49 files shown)


src/winml/modelkit/commands/config.py
```diff
@@ -326,34 +326,33 @@ def config(
...
-                "precision": precision,
-                "trust_remote_code": trust_remote_code,
-                "ep": ep,
-            }
-            if module:
-                configs = generate_hf_build_config(module=module, **_shared_kwargs)
+            # Generate config(s). The ``module: str | None`` overload of
+            # generate_hf_build_config returns WinMLBuildConfig | list[...],
+            # which isinstance(result, list) narrows for the branches below.
+            result = generate_hf_build_config(
```
xieofxie, line 333: revert of https://github.com/microsoft/winml-cli/pull/789/changes#diff-45b62ef5ac66eb35c9a46086706b3be0ecf8f6ab0f44c1eaad14df18858ab3a9

src/winml/modelkit/config/build.py
```diff
@@ -804,24 +827,24 @@ class name (HF path only).
...
-        "ep": ep,
-    }
-    if module is None:
-        return generate_hf_build_config(model_id, module=None, **common_kwargs)
-    return generate_hf_build_config(model_id, module=module, **common_kwargs)
+    # Single call resolves against generate_hf_build_config's `module: str | None`
+    # overload, which returns WinMLBuildConfig | list[WinMLBuildConfig] — matching
+    # this dispatcher's implementation return type. The dispatcher's own
+    # narrowing overloads above still tighten the return type for its callers.
+    return generate_hf_build_config(
```
xieofxie, line 834: revert as comment in https://github.com/microsoft/winml-cli/pull/825

src/winml/modelkit/datasets/processor_utils.py
```diff
@@ -13,7 +13,9 @@
 import logging
 from typing import Any
 
-from transformers.image_processing_utils import ImageProcessingMixin
+from transformers.image_processing_utils import (  # type: ignore[attr-defined]
```
xieofxie, line 16: the module DOES exist (transformers ships with py.typed and inline annotations) —
   mypy can find and read it. The complaint is one level deeper: ImageProcessingMixin is defined in some other submodule (e.g.
  transformers.image_processing_base) and image_processing_utils.py only does an internal from .image_processing_base import
  ImageProcessingMixin — without exposing it via __all__ or the explicit-re-export form import X as X.

DingmaomaoBJTU (commented):
## Type-checking PR review

**Overall:** Clean, well-scoped PR. The approach of progressively moving packages into the required mypy gate is sound. Found one structural correctness issue, two warnings about suppressions masking real problems, and a few low-priority notes.

### Bug - dead runtime guard in tensor_similarity_evaluator.py

mypy proves the `isinstance(model, WinMLCompositeModel)` branch is **unreachable** given the function's declared parameter type. The two `# type: ignore[unreachable]` suppressions silence the warning and make the guard invisible to callers. Either widen the parameter type to `WinMLPreTrainedModel | WinMLCompositeModel` so the check is live, or remove it and document that composite models are excluded at the call site.

### Warning - cast admits subtype incompatibility (evaluate.py)

`cast("str | dict[str, str | Path]", config.model_path)` with a comment admitting it is "subtype-incompatible-but-runtime-compatible" suppresses a genuine type mismatch. The right fix is to widen `WinMLEvaluationConfig.model_path`'s declared type to match what `WinMLAutoModel.from_onnx` actually accepts, rather than casting at each call site.

### Warning - dead try-branches in config_generator.py

Neither `inference.onnx_config.input_generator` nor `inference.onnx_config.patterns` exists at runtime. Both try-blocks always fall through to their fallback paths. The `# type: ignore[import-not-found]` suppresses mypy but the code inside the try-blocks is dead. Consider removing the unreachable import bodies entirely (leaving only the fallback logic) until issue #859 resolves the module's fate.

### Warning - duplicate if TYPE_CHECKING blocks (pipeline.py)

Two separate `if TYPE_CHECKING:` blocks appear at lines 26 and 30. They should be merged into one block. As-is it reads as if the second was added without noticing the first.

### Info - class-level empty-string sentinel (datasets/image.py)

Replacing `None` with `""` as the pre-detection sentinel is fine since all checks use `if not self._image_col:` consistently. Worth noting: a subclass that accidentally skips `_detect_columns()` will inherit the class-level `""` and fail with `KeyError: ''` rather than a more debuggable `AttributeError`. Not a blocker.

### Info - @step decorator TypeVar bound (base_writer.py)

`F = TypeVar("F", bound="Callable[..., int]")` is stricter than the old wrapper. Any `@step`-decorated handler that does not return `int` will now get a mypy error. Confirm all existing handlers annotate `-> int`; if any return `None`, they'll surface as new failures once dependent packages are gated.

zhenchaoni:
Heads-up on the `dict(processor(...))` / `dict(tokenizer(...))` wrappers added in `datasets/{depth_estimation,image,object_detection,text,image_segmentation}.py`:

These shallow-copy the returned `BatchEncoding` into a plain `dict`, which drops `BatchEncoding`-specific methods like `.to(device)` and the `.encodings` property. Should be fine here since the result is immediately consumed by `dataset.map` (which plain-dictifies anyway), but it's the one change in this PR with non-trivial runtime surface area. A single smoke run through the datasets path would confirm nothing downstream was relying on the `BatchEncoding` type.

xieofxie:
> ## Type-checking PR review
> **Overall:** Clean, well-scoped PR. The approach of progressively moving packages into the required mypy gate is sound. Found one structural correctness issue, two warnings about suppressions masking real problems, and a few low-priority notes.
> 
> ### Bug - dead runtime guard in tensor_similarity_evaluator.py
> mypy proves the `isinstance(model, WinMLCompositeModel)` branch is **unreachable** given the function's declared parameter type. The two `# type: ignore[unreachable]` suppressions silence the warning and make the guard invisible to callers. Either widen the parameter type to `WinMLPreTrainedModel | WinMLCompositeModel` so the check is live, or remove it and document that composite models are excluded at the call site.
> 
> ### Warning - cast admits subtype incompatibility (evaluate.py)
> `cast("str | dict[str, str | Path]", config.model_path)` with a comment admitting it is "subtype-incompatible-but-runtime-compatible" suppresses a genuine type mismatch. The right fix is to widen `WinMLEvaluationConfig.model_path`'s declared type to match what `WinMLAutoModel.from_onnx` actually accepts, rather than casting at each call site.
> 
> ### Warning - dead try-branches in config_generator.py
> Neither `inference.onnx_config.input_generator` nor `inference.onnx_config.patterns` exists at runtime. Both try-blocks always fall through to their fallback paths. The `# type: ignore[import-not-found]` suppresses mypy but the code inside the try-blocks is dead. Consider removing the unreachable import bodies entirely (leaving only the fallback logic) until issue #859 resolves the module's fate.
> 
> ### Warning - duplicate if TYPE_CHECKING blocks (pipeline.py)
> Two separate `if TYPE_CHECKING:` blocks appear at lines 26 and 30. They should be merged into one block. As-is it reads as if the second was added without noticing the first.
> 
> ### Info - class-level empty-string sentinel (datasets/image.py)
> Replacing `None` with `""` as the pre-detection sentinel is fine since all checks use `if not self._image_col:` consistently. Worth noting: a subclass that accidentally skips `_detect_columns()` will inherit the class-level `""` and fail with `KeyError: ''` rather than a more debuggable `AttributeError`. Not a blocker.
> 
> ### Info - @step decorator TypeVar bound (base_writer.py)
> `F = TypeVar("F", bound="Callable[..., int]")` is stricter than the old wrapper. Any `@step`-decorated handler that does not return `int` will now get a mypy error. Confirm all existing handlers annotate `-> int`; if any return `None`, they'll surface as new failures once dependent packages are gated.

Fixed and one issue created for you https://github.com/microsoft/winml-cli/issues/859

DingmaomaoBJTU (commented):
## Re-review after draft -> open promotion

Three of the four previously raised issues are resolved in commit 6caff1ff. One warning remains open.

### Resolved

**Bug (tensor_similarity_evaluator.py)** - Fixed. Parameter widened to `WinMLPreTrainedModel | WinMLCompositeModel`; the `isinstance` guard is now live and the `# type: ignore[unreachable]` suppression was correctly removed.

**Warning (evaluate.py:212)** - Fixed. The `cast` was dropped entirely because `WinMLAutoModel.from_onnx` and `WinMLCompositeModel.from_onnx` now both accept `Mapping[str, str | Path]` rather than `dict[str, str | Path]`. `dict[str, str]` is a `Mapping[str, str]` which satisfies `Mapping[str, str | Path]` due to covariance on the value type, so no cast is needed. Clean fix.

**Warning (pipeline.py duplicate TYPE_CHECKING)** - Fixed. The two blocks were merged into one.

### Still open

**Warning (config_generator.py:204)** - The two dead try-branches importing `inference.onnx_config.input_generator` and `inference.onnx_config.patterns` are unchanged. Both modules do not exist; the try-blocks always fall through; `# type: ignore[import-not-found]` silences mypy while leaving unreachable runtime code. Please remove the try-bodies (keeping only the fallback logic) and leave a comment pointing to #859 for restoration tracking.

### No new issues found

The changes in the latest commit are targeted and correct. No new type-annotation or logic issues.

DingmaomaoBJTU (approved):
LGTM! 3 of 4 prior findings resolved. The remaining dead try-branches in config_generator.py can be tracked in #859.

PR #836, opened 2026-06-08
Enable static quantization for Qwen3-0.6B decoder (transformer-only)

spalne:
Adds a transformer-only ONNX export path for Qwen3 that emits a fused (GQA) GroupQueryAttention op (with built-in rotary), LpNormalization RMSNorm, and 1×1 Conv projections, backed by an FP16 KV cache. The path is opt-in via install(), which hot-patches the build registries to produce two graphs (prefill seq=64, decode seq=1) without embeddings or lm_head. Quantization runs w8a16 static PTQ on these graphs using GSM8K calibration

Results

Produces two transformer-only ONNX files (prefill + decode) plus their w8a16-quantized variants.


src/winml/modelkit/models/hf/qwen_transformer_only.py
```diff
@@ -0,0 +1,354 @@
...
+    global _INSTALLED
+    if _INSTALLED:
+        return
+
+    # 1) Per-model build config + wrapper-class lookup live on the parent
+    #    ``models.hf`` package as module-level dicts; mutating them is the
+    #    documented hook for adding/overriding a model_type.
+    from .. import hf as _hf_pkg  # noqa: PLC0415
+
+    _hf_pkg.MODEL_BUILD_CONFIGS["qwen3"] = QWEN_TRANSFORMER_ONLY_CONFIG
```
DingmaomaoBJTU: The structural thing that doesn't match the rest of the repo: this `install()` hot-patches the global registries (`MODEL_BUILD_CONFIGS` / `MODEL_CLASS_MAPPING` / `COMPOSITE_MODEL_REGISTRY` + the optimum onnx overwrite) at runtime. Every other model under `models/hf` registers declaratively at import time - `qwen.py` uses `@register_onnx_overwrite` / `@register_composite_model` and gets merged through `hf/__init__.py`. This module isn't imported by `__init__.py` at all, so the path only exists if someone manually calls `install()`, and once called it permanently overrides qwen3 for the whole process with no way back (the original eager path is shadowed).

Can we make this a first-class variant the normal way instead of monkey-patching? e.g. a distinct task/model_type registered at import time, or a flag on the build config that selects the transformer-only export path. That also removes the "MUST call install() before importing the composite machinery" ordering trap.

test_qwen.py
```diff
@@ -0,0 +1,235 @@
+"""E2E test for the transformer-only Qwen3 export path.
```
DingmaomaoBJTU: This is a standalone runner at the repo root that drives the build via `subprocess` and judges success by "did a fresh artifact file appear". Repo convention (and `tests/CLAUDE.md`) is pytest under `tests/` with code-generated expectations - there are no other root-level `test_*.py` scripts. Could this move under `tests/e2e/` as a real pytest (marked `e2e`/`npu`/`qnn`), or under `examples/` if it's really a demo rather than a test? As-is it'll get picked up by name but isn't a pytest, and it lives outside the tree the suite runs from.

test_qwen.py
```diff
@@ -0,0 +1,235 @@
...
+        force_rebuild=True,
+        shape_config={"max_cache_len": MAX_CACHE, "seq_len": seq_len},
+    )
+    # The QNN/ORT teardown segfaults (0xC0000005) on interpreter shutdown
+    # AFTER the artifact is fully written. Skip the buggy cleanup with a hard
+    # exit so the parent sees a clean exit code 0.
+    print(f"BUILD COMPLETE: task={task} seq_len={seq_len}", flush=True)
+    sys.stdout.flush()
+    sys.stderr.flush()
+    os._exit(0)
```
DingmaomaoBJTU: `os._exit(0)` to skip interpreter teardown (because QNN/ORT segfaults on shutdown), combined with judging success purely by artifact mtime, hides a real native crash and makes the subprocess exit code meaningless. If the teardown crash is reproducible it's worth a tracked issue / fix at the EP layer rather than papering over it in the test harness. At minimum this deserves a code comment pointing at a tracking issue, otherwise a genuine build failure that still happens to touch the file would read as success.

qwen3_transformer_only_quantize.py
```diff
@@ -0,0 +1,230 @@
+"""Transformer-only w8a16 quantization for Qwen3.
```
DingmaomaoBJTU: Quantization in winml-cli is normally config-driven through `WinMLBuildConfig.quant` and runs as part of the build pipeline. This adds a parallel standalone quant entry point at the repo root that reaches into `sub_models[*]._onnx_path` directly and is "run via test_qwen.py". Could the transformer-only calibration reader be wired into the standard quant flow so it's reachable from `winml build` / the config instead of a bespoke script? Also minor: `Qwen3TransformerOnlyCalibReader` structurally satisfies `winml.modelkit.quant.config.CalibrationDataReader` but doesn't declare it - worth importing/typing against the protocol so it stays in sync.

qwen3_transformer_only_quantize.py
```diff
@@ -0,0 +1,230 @@
...
+            hf_model.config,
+            token_ids_list,
+            seq_len=seq_len,
+            max_cache_len=max_cache_len,
+        )
+        cfg = WinMLQuantizationConfig(
+            samples=num_samples,
+            weight_type=weight_type,  # type: ignore[arg-type]
+            activation_type=activation_type,  # type: ignore[arg-type]
+            calibration_method="minmax",
```
DingmaomaoBJTU: Accuracy concern worth resolving before this lands. I ran the produced w8a16 graphs against the FP `optimized` graphs on the same GSM8K-style input (ORT CPU EP): the FP export matches HF eager exactly (top-1 next token identical), but the w8a16 output flips the top-1 token on **both** prefill and decode - top-5 overlap 0-1/5, KL(FP||quant) 0.66 / 2.75, output_hidden_states cosine 0.64-0.72. The present-KV path is ~0.999, so the damage is concentrated in the residual stream.

Likely causes: `minmax` calibration over a residual stream with large outliers (+/-76), calibrating with an all-zero KV cache, and only 30 samples. Suggest trying `calibration_method="percentile"` (or entropy), feeding a realistic non-zero KV during calibration, and reporting an actual task metric (e.g. GSM8K logits/top-1 agreement) so we can see the quant is acceptable, not just that QDQ nodes were inserted.

DingmaomaoBJTU (commented):
**Summary - structurally sound export, but registration/test/quant integration don't match repo conventions, and w8a16 accuracy regresses.**

Nice work getting a fused GQA + LpNorm RMSNorm + 1x1-Conv transformer-only export running end-to-end on QNN, and the export itself is faithful - the FP `optimized` graph reproduces HF eager's next-token exactly. Three things to address before this is review-ready:

**1. Registration is non-standard (highest priority).** `qwen_transformer_only.install()` hot-patches the global registries at runtime and isn't imported by `models/hf/__init__.py`. Every other model registers declaratively at import time (`@register_onnx_overwrite` / `@register_composite_model`, merged in `__init__.py`). Please make this a first-class variant (distinct task/model_type or a build-config flag) instead of monkey-patching; it also removes the "must call install() before importing the composite machinery" ordering trap and the no-way-back override of the eager path.

**2. Test & quant entry points violate repo layout.** `test_qwen.py` and `qwen3_transformer_only_quantize.py` are standalone scripts at the repo root; `test_qwen.py` is a subprocess driver that judges success by artifact mtime and uses `os._exit(0)` to mask a native QNN/ORT teardown crash. Convention (`tests/CLAUDE.md`) is pytest under `tests/`. Move the runner to `tests/e2e/` (or `examples/`), and wire the calibration reader into the config-driven quant flow (`WinMLBuildConfig.quant`) rather than a bespoke quantizer.

**3. w8a16 accuracy is not yet acceptable.** Measured against the FP graph on the same GSM8K-style input, the quantized model flips the top-1 next token on both prefill and decode (top-5 overlap 0-1/5, KL 0.66/2.75; hidden-state cosine 0.64-0.72), while present-KV stays ~0.999 - i.e. the residual stream is the casualty. Likely `minmax` + all-zero KV calibration + only 30 samples. Please try percentile/entropy calibration with a realistic non-zero KV feed and report an actual task metric, not just QDQ node count.

Naming and the custom-op export pattern look good and match the codebase.

qwen3_transformer_only_quantize.py
```diff
@@ -0,0 +1,230 @@
...
+
+
+def quantize_built_model(
+    model: WinMLCompositeModel,
+    *,
+    model_id: str = DEFAULT_MODEL_ID,
+    max_cache_len: int = DEFAULT_MAX_CACHE,
+    prefill_seq: int = DEFAULT_PREFILL_SEQ,
+    num_samples: int = DEFAULT_NUM_SAMPLES,
+    weight_type: str = "uint8",
```
DingmaomaoBJTU: **Weights come out asymmetric uint8 instead of symmetric int8 — likely an accuracy regression.**

Comparing the model this path produces against the reference quantized model (`qwen3_gqa_fp16_ctx.onnx`), all 196 weight tensors differ in scheme:

| | zero_point dtype | non-zero zp | scheme |
|---|---|---|---|
| this PR | uint8, ≈127 mean | 196/196 | asymmetric |
| reference | int8, 0 | 0/196 | symmetric |

e.g. same weight in Netron: PR `model.model.layers.0.mlp.down_proj.weight_zero_point` = uint8 `124` vs ref int8 `0` (scales are ~equal, only zp differs).

`weight_type="uint8"` here + the quantizer's `symmetric=False` default produces asymmetric weights. For zero-centered Linear weights this shifts the grid and loses precision vs symmetric int8, and is also less friendly to NPU/HTP kernels that assume symmetric weights. Was uint8/asymmetric intended, or should this match the reference (`weight_type="int8"` + symmetric)?

qwen3_transformer_only_quantize.py
```diff
@@ -0,0 +1,230 @@
...
+    activation_type: str = "uint16",
+) -> dict[str, Path]:
+    """Quantize the transformer-only ONNX files in-place.
+
+    Returns ``{sub_model_name: quantized_path}``.
+    """
+    # Locate the un-compiled ONNX for each sub-model (no surgery — file is
+    # already transformer-only).
```
DingmaomaoBJTU: **Every GQA attention gets an extra QuantizeLinear that the reference model doesn't have.**

Because this runs the default quantizer with no GQA-aware surgery, each `GroupQueryAttention.output[0]` ends up quantized:

- this PR: `GroupQueryAttention.output[0]` → `QuantizeLinear` (×28 layers)
- reference: `GroupQueryAttention.output[0]` → `Cast` to fp32 (×28), **not** quantized

The reference keeps the attention output in float (its `add_qdq_after_gqa` surgery in `lm_to_onnx/onnx_utils/onnx_surgeries.py` restructures QDQ around GQA rather than quantizing the output). Quantizing it on all 28 layers adds extra error on the attention path that accumulates layer-by-layer through the residual stream — probably the larger contributor to the accuracy drop. Should this path exclude GQA outputs from quantization to match the reference?

src/winml/modelkit/models/hf/qwen3_modeling.py
```diff
@@ -0,0 +1,237 @@
...
+    def prepare_for_onnx_export(self) -> None:
+        # Pre-multiply the gain into the weight (LpNorm has unit gain).
+        n = self.weight.numel()
+        scale = torch.sqrt(
+            torch.tensor([n], device=self.weight.device, dtype=self.weight.dtype)
+        )
+        if torch.any(self.weight.data != torch.ones_like(self.weight)).item():
+            new_w = scale * self.weight
+        else:
+            new_w = scale
```
DingmaomaoBJTU: **Bug: RMSNorm weight shape mismatch when weights are all 1.0**

When `self.weight` is all ones (the default init), `new_w = scale` produces a `[1]`-shaped tensor, not `[hidden_size]`. The ONNX initializer exports with shape `[1]` instead of `[hidden_size]`, which broadcasts silently in PyTorch but may cause shape errors in downstream ONNX tooling.

The branch is also logically redundant (scale * ones == scale). Simplify to:
```python
self.weight = nn.Parameter(scale * self.weight)
```

test_qwen.py
```diff
@@ -0,0 +1,229 @@
...
+projections.
+
+Generation (``model.generate(...)``) is NOT supported by this build path —
+the inference feeds in ``WinMLDecoderOnlyModel`` still target the eager
+I/O signature. Use the eager ``WinMLQwen3Model`` build path for end-to-end
+generation.
+
+Run::
+
+    python test_qwen_transformer_only.py
```
DingmaomaoBJTU: **Bug: Wrong filename in docstring** — says `python test_qwen_transformer_only.py` but the file is `test_qwen.py`.

src/winml/modelkit/models/hf/qwen3_export_ops.py
```diff
@@ -0,0 +1,211 @@
...
+            "onnx::LpNormalization",
+            input,
+            axis_i=int(axis),
+            p_i=int(p),
+        )
+        return output.setType(output_type)
+
+    @staticmethod
+    def forward(ctx, input, axis, p):  # noqa: ARG004
+        return input  # placeholder — real compute happens in symbolic
```
DingmaomaoBJTU: **Warning: Eager-mode forward returns incorrect (un-normalized) results**

`LpNormOnnxExport.forward` returns `input` unchanged (identity). This is only correct during ONNX tracing where `symbolic` runs instead. Any eager execution (unit tests, calibration debug runs) silently gets un-normalized values. Consider computing the real norm for eager mode or raising `NotImplementedError` to make misuse obvious.

src/winml/modelkit/models/hf/qwen3_export_ops.py
```diff
@@ -0,0 +1,211 @@
...
+        past_value,
+        seqlens_k,
+        total_sequence_length,
+        cos_cache,
+        sin_cache,
+        do_rotary,
+        kv_num_heads,
+        num_heads,
+    ):  # noqa: ARG004
+        return query, past_key, past_value  # placeholder shapes
```
DingmaomaoBJTU: **Warning: Stale KV cache in eager mode**

`GroupQueryAttentionOnnxExport.forward` returns `(query, past_key, past_value)` — the present_keys/present_values are the old un-updated tensors. Eager execution silently produces a KV cache that never advances. A `NotImplementedError` here would be safer than a silently-wrong placeholder.

src/winml/modelkit/models/hf/qwen3_export_ops.py
```diff
@@ -0,0 +1,211 @@
...
+        WinMLQwen3RMSNorm,
+    )
+
+    def _bind(module: nn.Module, owner: type) -> None:
+        module.forward = owner.forward.__get__(module, type(module))
+
+    # Identify Qwen3 submodules by their (stock HF) class name so we don't
+    # depend on importing ``transformers.models.qwen3`` here.
+    def _is(module: nn.Module, name: str) -> bool:
+        return type(module).__name__ == name
```
DingmaomaoBJTU: **Warning: Fragile class-name string matching**

`type(module).__name__ == name` breaks silently if HuggingFace renames a Qwen3 module in a future release — the forward won't be bound and the export will be silently broken. Consider adding a post-patch assertion that the expected number of attention/MLP/RMSNorm modules were patched.

src/winml/modelkit/quant/calibration/qwen3_transformer_only.py
```diff
@@ -0,0 +1,262 @@
...
+            )
+        sub_paths[name] = final_path
+
+    for name, p in sub_paths.items():
+        print(f"  {name}: {p}")
+
+    print("\n=== Loading HF embed_tokens for calibration ===")
+    hf_model = AutoModelForCausalLM.from_pretrained(model_id, torch_dtype=torch.float32)
+    hf_model.eval()
+    embed_tokens = hf_model.get_input_embeddings()
```
DingmaomaoBJTU, line 393: **Warning: HF model not freed after calibration data is pre-built**

`hf_model` stays live for the entire quantization loop. Since `Qwen3TransformerOnlyCalibReader.__init__` materializes all samples in `self._samples`, the model weights are no longer needed after reader construction. Add `del hf_model; gc.collect()` before the quantization loop to free the model memory.

qwen3_transformer_only_quantize.py
```diff
@@ -0,0 +1,262 @@
...
+
+    quant_paths: dict[str, Path] = {}
+    for sub_name, fused_path in sub_paths.items():
+        if sub_name not in seq_by_sub:
+            print(f"\n--- Skipping unknown sub-model {sub_name!r} ---")
+            continue
+
+        seq_len = seq_by_sub[sub_name]
+        quant_path = fused_path.with_name(
+            fused_path.stem + f"_w{weight_type[-1]}a{activation_type[-2:]}.quant.onnx"
```
DingmaomaoBJTU: **Warning: Brittle string slicing for output filename**

`weight_type[-1]` and `activation_type[-2:]` work for int8/uint16 but produce wrong suffixes for other valid types (e.g. `activation_type='uint4'` -> `_t4`). Use an explicit dict mapping or strip the numeric suffix with a regex/lstrip instead.

test_qwen.py
```diff
@@ -0,0 +1,229 @@
...
+    #    Judge success by whether a FRESH npu_ctx artifact appeared, NOT by the
+    #    subprocess exit code: the native QNN/ORT layer segfaults (0xC0000005)
+    #    on teardown AFTER the artifact is fully written to disk.
+    import time as _time
+
+    for name, (task, seq_len, prefix) in SUB_MODELS.items():
+        print(f"\n########## BUILD {name} (task={task}, seq_len={seq_len}) ##########", flush=True)
+        before = _latest_ctx_mtime(prefix)
+        start = _time.time()
+        rc = subprocess.run(
```
DingmaomaoBJTU: **Warning: subprocess.run() has no timeout**

If the QNN/ORT build stalls, this blocks indefinitely. Add `timeout=1800` (or similar) and catch `subprocess.TimeoutExpired` to surface a clear CI failure.

test_qwen.py
```diff
@@ -0,0 +1,229 @@
...
+        force_rebuild=True,
+        shape_config={"max_cache_len": MAX_CACHE, "seq_len": seq_len},
+    )
+    # The QNN/ORT teardown segfaults (0xC0000005) on interpreter shutdown
+    # AFTER the artifact is fully written. Skip the buggy cleanup with a hard
+    # exit so the parent sees a clean exit code 0.
+    print(f"BUILD COMPLETE: task={task} seq_len={seq_len}", flush=True)
+    sys.stdout.flush()
+    sys.stderr.flush()
+    os._exit(0)
```
DingmaomaoBJTU: **Info: os._exit(0) as a crash workaround**

Documented, but worth tracking as an upstream ORT/QNN issue. Consider adding a TODO comment with a link to a filed issue so this workaround isn't forgotten.

test_qwen.py
```diff
@@ -0,0 +1,229 @@
+"""E2E test for the transformer-only Qwen3 export path.
```
DingmaomaoBJTU: **Info: Not a pytest test**

Despite the `test_` prefix, this file uses `__main__`, `sys.path` mutations at import time, subprocess orchestration, and `os._exit`. It lives in the repo root and won't be collected by `uv run pytest tests/`. Consider renaming to `scripts/run_qwen3_quant.py` to avoid accidental pytest collection, or convert to a proper pytest integration test with hardware skip markers.

DingmaomaoBJTU (commented):
## Code Review — PR #836 (Draft)

Well-structured PR. The transformer-only export topology (fused GQA, LpNorm RMSNorm, 1x1 Conv), GSM8K calibration pipeline, and model_type override mechanism are solid. A few correctness bugs and infrastructure concerns should be resolved before marking ready for merge.

**Not approving** since this is a draft PR.

src/winml/modelkit/quant/calibration/qwen3_transformer_only.py
```diff
@@ -0,0 +1,262 @@
...
+                # seq_len-token query, which makes the GQA prefill kernel read
+                # out of bounds → native access violation.)
+                "past_seq_len": np.array([[self.seq_len - 1]], dtype=np.int32),
+                "total_seq_len": np.array([self.max_cache_len], dtype=np.int32),
+            }
+            kv_shape = (1, self.num_kv_heads, self.max_cache_len, self.head_dim)
+            zeros = np.zeros(kv_shape, dtype=np.float16)
+            for i in range(self.num_layers):
+                feed[f"past_keys_{i}"] = zeros
+                feed[f"past_values_{i}"] = zeros
```
DingmaomaoBJTU, line 206: Follow-up from the accuracy verification: the symmetric-int8 + GQA-exclusion fix correctly aligns the **graph** with the reference, but the produced model still **degenerates in generation** — it emits a correct first token and then collapses into repeated garbage (e.g. `Paris` → `Parisammedammed...`).

I traced it to this calibration reader, specifically the **decode (`seq_len=1`, `decoder_gen`)** path. Stage-isolation (mixing FP/quant ctx & iter) shows the **prefill quant is fine; only the quantized iter breaks decode**, and since ctx/iter share identical int8 weights and activations are `uint16` (plenty of precision), the culprit is the **iter activation calibration ranges**, i.e. the data fed here:

1. `ids = ids[:, : self.seq_len]` with `seq_len=1` → after `apply_chat_template`, every sample is the **same first token** `<|im_start|>` (151644). I verified across prompts: the decode reader sees one identical token repeated, so MinMax effectively calibrates on a single activation value.
2. `past_seq_len = seq_len - 1 = 0` → declares an empty context.
3. `past_keys_/past_values_ = np.zeros(...)` → attention is calibrated with an **empty KV cache**.

So the iter activation ranges are derived from "one unrepresentative token + empty KV + length 0", which is far narrower than real decode. Real decode activations then saturate → collapse. (Prefill survives because `seq_len=64` gives ~1920 diverse real states.)

**Verified fix (single variable changed):** I re-quantized the FP iter changing **only** `calibration_data` — keeping `weight_type=int8`/`weight_symmetric=True`, `activation_type=uint16`, `calibration_method=minmax`, and the same `nodes_to_exclude=gqa_nodes` — and fed a **real decode trajectory** instead: run a short FP `prefill + N decode` loop and capture each step's true feed (current token embedding, accumulated KV, growing `past_seq_len`), 10 prompts × 16 steps = 160 samples. Result (e2e greedy, same ctx/emb/head):

| prompt | current iter calib | trajectory calib | fp16 reference |
|---|---|---|---|
| The capital of France is | `Parisammedammed...` | `Paris, and the capital of Italy is Rome...` | `Paris, and the capital of Italy is Rome...` (token-identical) |
| 2 + 2 = | garbage | `4, 2 + 2 = 4...` | `4, 3 + 3 = 6...` |
| The opposite of hot is | garbage | `cold, and the opposite of cold is hot...` | `cold, and the opposite of cold is hot...` |

QDQ count is unchanged (2279), so this is purely a calibration-data issue, not graph structure or bit-width.

**Suggested change:** for the `decoder_gen` (`seq_len=1`) sub-model only, replace the single-token + zeroed-KV feed with real decode-step states (a brief FP prefill+decode trajectory). The `decoder_prefill` (`seq_len=64`) path is fine as-is. Happy to share the prototype script that produces the trajectory calibration if useful.

qwen3_transformer_only_quantize.py
```diff
@@ -0,0 +1,262 @@
...
+    # Locate the un-compiled ONNX for each sub-model (no surgery — file is
+    # already transformer-only).
+    sub_paths: dict[str, Path] = {}
+    for name, sub in model.sub_models.items():
+        final_path = Path(sub._onnx_path)
+        if final_path.name.endswith("_model.onnx"):
+            stem = final_path.name[: -len("_model.onnx")]
+            optimized = final_path.with_name(f"{stem}_optimized.onnx")
+            if optimized.exists():
+                sub_paths[name] = optimized
```
DingmaomaoBJTU: Non-blocking / out of scope for the quant fix — just recording a graph-hygiene difference for later.

The `*_optimized.onnx` consumed here carries a bloated set of opset imports compared with the reference graph (measured on the produced ctx model):

- **This PR (MINE):** main domain **opset 17** plus 8 extra domains — `com.microsoft`, `com.microsoft.nchwc`, `ai.onnx.ml`, `ai.onnx.training`, `ai.onnx.preview.training`, `com.microsoft.experimental`, `org.pytorch.aten`, `com.microsoft.dml`.
- **Reference:** main domain **opset 18** + `com.microsoft` only (2 domains total).

The extra `nchwc`/`dml`/`training`/`aten` domains are artifacts of the ORT optimize pass that produces `_optimized.onnx`; they're not used by the actual nodes. This doesn't change quant numerics, but it does inflate the graph metadata and can confuse downstream EP partitioning/loaders. Worth cleaning up the optimize pass (or stripping unused opset imports) at some point — not required for this PR.

src/winml/modelkit/models/hf/qwen3/qwen_transformer_only.py
```diff
@@ -0,0 +1,351 @@
...
+    allow_new=True,
+)
+
+
+def _transformer_only_inputs(num_layers: int, kv_seq_axis: str = "max_seq_len") -> dict[str, dict[int, str]]:
+    """Input ordering: past KV pairs, then hidden states, then seq lens."""
+    result: dict[str, dict[int, str]] = {}
+    for i in range(num_layers):
+        result[f"past_keys_{i}"] = {2: kv_seq_axis}
+        result[f"past_values_{i}"] = {2: kv_seq_axis}
```
DingmaomaoBJTU, line 253: Non-blocking / out of scope for the quant fix — recording another export/pipeline difference vs the reference graph.

The KV time axis is declared symbolic here (`kv_seq_axis="max_seq_len"`, applied to `past_keys_{i}`/`past_values_{i}` axis 2), which matches the reference. **But the produced/optimized graph ends up with a static axis** (measured on the ctx model):

- **This PR (MINE):** `past_keys_0` axis2 = **256 (static)**.
- **Reference:** `past_keys_0` axis2 = **`max_seq_len` (symbolic)**.

So the symbolic dim declared here is being frozen to the concrete `max_cache_len` (256) somewhere downstream — most likely the same ORT optimize pass that produces `_optimized.onnx`. A static 256 cache hard-codes the max sequence length into the graph (less flexible for longer contexts / different cache sizes) whereas the reference stays parametric. Doesn't affect quant numerics; flagging so the symbolic axis is preserved through the optimize/export step if that flexibility is wanted.

src/winml/modelkit/quant/calibration/qwen3_transformer_only.py
DingmaomaoBJTU: Non-blocking follow-up / request for the reference **calibration recipe**.

I compared this path's w8a16 transformer initializers against the reference bundle (`qwen3_gqa_fp16_ctx.onnx` / `qwen3_gqa_fp16_iter.onnx`) by hashing every initializer's bytes and classifying each FLOAT scale as a *weight scale* vs an *activation scale* (by whether the (De)QuantizeLinear it feeds quantizes an INT8 weight init or a runtime tensor):

| | INT8 weights | weight scales | activation scales |
|---|---|---|---|
| prefill / ctx | **392/392 identical** | **196/196 identical** | 122/704 identical |
| decode / iter | **392/392 identical** | **196/196 identical** | 121/704 identical |

So the **quantization scheme + the deterministic RTN weights + the weight scales are already byte-identical to the reference** — the only thing that diverges is ~83% of the **activation scales**, which are purely calibration-derived. That's expected (activation ranges are data-dependent), and I don't think byte-matching them is the right goal — functional task-metric parity is. But it does mean the only free variable left between the two pipelines is the calibration.

To make the accuracy comparison apples-to-apples (and to rule out "we calibrated on GSM8K, the reference used something else" as a confound), could you share the reference **calibration recipe** — not just the dataset, but the whole thing:

1. **dataset** (and how many samples / selection order / tokenization),
2. **calibration method** (minmax / percentile / entropy, and any clipping),
3. **how the activations + KV are fed during calibration for the iter (decode, seq_len=1) sub-model** — i.e. the equivalent of the reference's `CalibrationDataReader`. This is the important one: as documented above, the decode collapse here comes from feeding a single token + zeroed KV, and I expect the reference feeds real decode-trajectory states. The dataset alone won't fix that — the feed mechanics will.

Note this isn't a blocker for landing: the decode fix (real-trajectory calibration) is independent of the reference data, and we can validate via a task metric. The recipe would just let us do a clean side-by-side on the same distribution and potentially close the activation-scale gap.

<sub>Evidence scripts: `temp/weight_overlap.py` (initializer byte-overlap by dtype) and `temp/scale_classify.py` (weight-scale vs activation-scale classification + overlap).</sub>

src/winml/modelkit/models/hf/qwen3_export_ops.py
DingmaomaoBJTU: Fixed in caada383. `LpNormOnnxExport.forward` now computes the real L2 normalization (`input / ||input||_p` along `axis`) instead of returning the input unchanged. This is export-invariant — the exported node still comes solely from `symbolic`, the eager body is never traced into the graph — and it makes any eager execution (unit tests, calibration debug runs, the exporter shape-tracing pass) correct. Verified: e2e export + decode parity still pass.

src/winml/modelkit/models/hf/qwen_transformer_only.py
```diff
@@ -0,0 +1,407 @@
...
+from ...export import register_onnx_overwrite
+from ...export.config import WinMLExportConfig
+from ..winml import register_specialization
+from ..winml.composite_model import register_composite_model
+from ..winml.decoder_only import WinMLDecoderOnlyModel
+from ..winml.kv_cache import WinMLSlidingWindowCache
+from .qwen3_modeling import apply_transformer_only_export_prep
+
+
+logger = logging.getLogger(__name__)
```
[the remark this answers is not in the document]
DingmaomaoBJTU: Fixed in caada383. The unused module-level `logger` (and its `import logging`) were removed.

src/winml/modelkit/models/hf/qwen3_export_ops.py
DingmaomaoBJTU: Kept as-is by design, with an expanded docstring in caada383 explaining why. `forward` cannot raise on eager execution: the HTP export pipeline runs a real eager `forward` pass to capture the module hierarchy (`export/htp/hierarchy.py::trace_model_execution`), and there both `torch.jit.is_tracing()` and `torch.onnx.is_in_onnx_export()` are False — indistinguishable from "misuse" — so raising there breaks the actual build (empirically confirmed: it failed e2e at `trace_model_execution`). A faithful eager attention is also not cheaply available since correct attention grows the sequence axis that static-shape export freezes. The real op is emitted by `symbolic`; this module is export-only and is never run for real inference (calibration loads a fresh real model).

qwen3_transformer_only_quantize.py
DingmaomaoBJTU: Confirmed out of scope for this quant PR. The extra opset imports are a downstream artifact of the ORT optimize pass on the exported graph, not something this change introduces; there is no clean fix at this layer. Leaving as a future graph-hygiene cleanup.

src/winml/modelkit/models/hf/qwen3/qwen_transformer_only.py
DingmaomaoBJTU: Confirmed out of scope here. The symbolic KV axis being frozen to a static dimension is a property of the static-shape HTP export path (the exporter does not apply dynamic_axes for this model_type), not of the quantization change. Tracked as a future export-pipeline improvement.

qwen3_transformer_only_quantize.py
DingmaomaoBJTU: Closing this out: the ``qwen3_transformer_only_quantize.py`` standalone script this note was anchored to has been removed in the config-driven refactor (quantization now runs through the standard build pipeline + ``winml_finalize_quant_config`` hook, covered by the e2e test). The opset-import bloat itself is a downstream ORT optimize-pass artifact on the produced graph, non-blocking and unchanged in scope. Resolving.

src/winml/modelkit/commands/build.py
```diff
@@ -1388,6 +1392,32 @@ def _name(base: str) -> str:
     config_path.write_text(json.dumps(config.to_dict(), indent=2))
 
     # ── Quantize stage ───────────────────────────────────────────
+    # Some model types finalize their quant config only once the exported ONNX
+    # exists (calibration feeds / nodes-to-exclude derived from the graph).
+    # Resolve the model-type-specific quant policy from the quant registry,
+    # keyed on the live ``model_type`` — mirrors build.hf.build_hf_model so the
+    # CLI and library pipelines apply the same scheme. Unregistered types return
+    # None → the quantizer uses its standard task-aware DatasetCalibrationReader.
+    if config.quant is not None:
```
zhenchaoni: I find this logic is repeated in build.py and hf.py, I suggest putting it inside quantize. The quantize method also takes a onnx model path, you can resolve the logic inside.

src/winml/modelkit/loader/config.py
```diff
@@ -220,6 +220,19 @@ def resolve_loader_config(
...
+    # variant (e.g. "qwen3_transformer_only") rather than the architecture's
+    # native type. The model_type-only path above (AutoConfig.for_model) is
+    # unaffected because it only runs when model_id is None.
+    if model_id is not None and model_type is not None and hf_config.model_type != model_type:
+        logger.info(
+            "Overriding resolved model_type '%s' -> '%s' (explicit request)",
+            hf_config.model_type,
+            model_type,
+        )
+        hf_config.model_type = model_type
```
zhenchaoni: It could be risky to override the hf_config's model type but keep other config value not changed. This could potentially break the consistency of the config object. I would suggest to keep using the model type from huggingface then further resolve it in the downstream logic.

src/winml/modelkit/loader/config.py
```diff
@@ -220,6 +220,19 @@ def resolve_loader_config(
             f"attribute. Cannot proceed with config generation."
         )
 
+    # Explicit model_type override alongside a model_id: honor the requested
+    # type so downstream class / build-config / export resolution selects the
+    # variant (e.g. "qwen3_transformer_only") rather than the architecture's
```
zhenchaoni: where do you get "qwen3_transformer_only"? If I just give you model id "Qwen/Qwen3-0.6B" or model task, which api will tell you the model type is qwen3_transformer_only?

src/winml/modelkit/quant/calibration/registry.py
```diff
@@ -0,0 +1,75 @@
...
+# ``model_type`` -> submodule that defines (and self-registers) its finalizer.
+# Looked up lazily so the heavy module loads only when needed. Keys must match
+# the live ``model_type`` string verbatim (no ``_`` -> ``-`` normalization),
+# since lookup is keyed on the exported model's ``config.model_type``.
+_KNOWN_FINALIZER_MODULES: dict[str, str] = {
+    "qwen3_transformer_only": ".qwen3_transformer_only",
+}
+
+
+def register_quant_finalizer(model_type: str) -> Callable[[type], type]:
```
zhenchaoni: No need to have a decorator or register method. It is good but a bit overkill. You can simply expose the dict. We use this code pattern in many places in the repo.

src/winml/modelkit/loader/config.py
DingmaomaoBJTU: Done - no longer mutating hf_config.model_type. The explicit type is threaded through as model_type_override into resolve_task, and only the loader config's model_type surfaces the variant. The HF config (and everything downstream that reads it - exporter, Optimum patcher) keeps the architecture's native type.

src/winml/modelkit/loader/config.py
DingmaomaoBJTU: It isn't auto-detected - it's the explicit --model-type qwen3_transformer_only the user passes for this build variant. There's no inference from the model id/task; it's a synthetic build-variant tag that selects the transformer-only export + w8a16 policy.

src/winml/modelkit/commands/build.py
DingmaomaoBJTU: Moved the dispatch into quantize_onnx: it resolves the model-type policy from config.model_type and applies it before the mode handler, so the build command and build_hf_model no longer duplicate it. The pipelines just carry model_type onto the quant config.

src/winml/modelkit/quant/calibration/registry.py
DingmaomaoBJTU: Done - dropped the decorator/registration indirection. QUANT_FINALIZERS is now a plain dict[str, (module, class)] and get_quant_finalizer just looks it up and lazily instantiates.

DingmaomaoBJTU (approved):

spalne:
> @spalne please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> @microsoft-github-policy-service agree [company="{your company}"]
> ```

@microsoft-github-policy-service agree

DingmaomaoBJTU (approved):

ssss141414 (approved):

Issue #829, opened 2026-06-08
winml compile --ep qnn --device gpu ignores device flag, compiles for HTP/NPU instead of GPU


DingmaomaoBJTU:
## Summary

`winml compile --ep qnn --device gpu` produces an EPContext binary that only works on NPU, not GPU. The `--device gpu` flag is effectively ignored when `--ep` is specified.

## Root Cause

In `_resolve_compile_provider()` ([compile.py:271](src/winml/modelkit/commands/compile.py#L271)):

```python
def _resolve_compile_provider(device: str, ep: str | None) -> str:
    if ep:
        return ep.lower()   # returns "qnn" immediately, ignores device
    ...
```

When `--ep qnn` is passed, it short-circuits and returns `"qnn"` without considering the `--device gpu` flag. QNN then compiles using the default HTP (NPU) backend, producing a binary that cannot be loaded on GPU.

## Reproduction

```bash
# Compile with --device gpu --ep qnn
winml compile -m model.onnx --ep qnn --device gpu -o compiled.onnx

# Fails on GPU (error 1002: Failed to create context from binary)
winml perf -m compiled.onnx --device gpu --ep qnn

# But works on NPU - proving it was compiled for NPU, not GPU
winml perf -m compiled.onnx --device npu --ep qnn
```

## Expected Behavior

`winml compile --ep qnn --device gpu` should compile using QNN's GPU backend so the resulting EPContext model can run on GPU. The `--device` flag should be passed through to the QNN backend configuration, not ignored.

Assign to @hualiang (need GitHub ID).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

DingmaomaoBJTU:
**Triage summary**

- **Component**: Compile
- **Priority**: P0

`_resolve_compile_provider()` silently ignores `--device gpu` when `--ep qnn` is specified, producing an NPU-only EPContext binary. This is silent wrong behavior — the user gets an unusable artifact with no warning. Suggested fix: pass the device flag through to QNN backend configuration so the correct target (HTP vs GPU) is selected.

xieofxie:
not reproduceable..

DingmaomaoBJTU:
## Triage Summary
Priority: **P1**
Component: compile
Owner: @xieofxie

[...]

PR #837, opened 2026-06-09
refactoring: simplify compile ep selection logic

xieofxie:
Found when investigating https://github.com/microsoft/winml-cli/issues/829

# Change 1
When `resolve_device`, the returned device is already checked against ep, so use it directly.

So the previous code check is useless, simplified it and change test to check against `compile` directly

# Change 2
in `resolve_device` check against the hard code list first instead of into real ort ep device map, see issue #860

# Change 3
in `register_ep`, add a dup check for provider name

### DIFF
--- a/src/winml/modelkit/commands/compile.py
+++ b/src/winml/modelkit/commands/compile.py
@@ -264,5 +264,12 @@ def _resolve_compile_provider(resolved_device: str, ep: EPNameOrAlias | None) ->
 
     ``ep`` overrides the device mapping. Returns
     the canonical EP name (e.g., ``"QNNExecutionProvider"``).
+
+    Device/EP policy compatibility is enforced upstream by ``resolve_device``;
+    this helper trusts its inputs and only normalizes.
     """
-    return normalize_ep_name(ep) if ep else resolve_eps(resolved_device)[0]
+    if ep is not None:
+        normalized = normalize_ep_name(ep)
+        if normalized is not None:
+            return normalized
+    return resolve_eps(resolved_device)[0]
--- a/src/winml/modelkit/commands/compile.py
+++ b/src/winml/modelkit/commands/compile.py
@@ -28,7 +28,7 @@ from rich.console import Console
 from ..onnx import is_compiled_onnx
 from ..sysinfo import resolve_device, resolve_eps
 from ..utils import cli as cli_utils
-from ..utils.constants import EP_SUPPORTED_DEVICES, normalize_ep_name
+from ..utils.constants import normalize_ep_name
 
 
 if TYPE_CHECKING:
@@ -265,21 +265,4 @@ def _resolve_compile_provider(resolved_device: str, ep: EPNameOrAlias | None) ->
     ``ep`` overrides the device mapping. Returns
     the canonical EP name (e.g., ``"QNNExecutionProvider"``).
     """
-    if ep:
-        canonical = normalize_ep_name(ep)
-        if canonical is None:
-            raise click.UsageError(f"Unknown EP: {ep}")
-        supported = EP_SUPPORTED_DEVICES[canonical]
-        if resolved_device.lower() not in supported:
-            raise click.UsageError(
-                f"--ep {ep} cannot run on --device {resolved_device}. "
-                f"{canonical} supports: {', '.join(supported)}."
-            )
-        # EP host-availability is enforced by ``resolve_device`` upstream,
-        # no need for an extra check here
-        return canonical
-
-    eps = resolve_eps(resolved_device)
-    if not eps:
-        return "CPUExecutionProvider"
-    return eps[0]
+    return normalize_ep_name(ep) if ep else resolve_eps(resolved_device)[0]
--- a/src/winml/modelkit/session/ep_registry.py
+++ b/src/winml/modelkit/session/ep_registry.py
@@ -269,8 +269,12 @@ class WinMLEPRegistry:
         """Load EP catalog from WinML."""
         from windowsml import EpCatalog
 
+        checked_eps: set[EPName] = set()
         with EpCatalog() as catalog:
             for provider in catalog.find_all_providers():
+                if provider.name in checked_eps:
+                    continue
+                checked_eps.add(provider.name)
                 try:
                     _ensure_provider_ready(provider)
                 except OSError as e:
--- a/src/winml/modelkit/session/ep_registry.py
+++ b/src/winml/modelkit/session/ep_registry.py
@@ -289,7 +289,7 @@ class WinMLEPRegistry:
                 if provider.library_path == "":
                     continue
                 self._ep_paths[cast("EPName", provider.name)] = provider.library_path
-                logger.debug("Found EP: %s at %s", provider.name, provider.library_path)
+                logger.info("Found EP: %s at %s", provider.name, provider.library_path)
 
     def register_to_ort(self) -> list[EPName]:
         """Register discovered EPs to ONNX Runtime.
--- a/src/winml/modelkit/sysinfo/device.py
+++ b/src/winml/modelkit/sysinfo/device.py
@@ -177,10 +177,12 @@ def resolve_device(
         ep_full = normalize_ep_name(ep)
         if ep_full not in EP_SUPPORTED_DEVICES:
             raise ValueError(f"Unknown EP '{ep}'. Expected one of: {sorted(EP_SUPPORTED_DEVICES)}")
+        # Static policy gate (see issue #860): reject EP/device combos the EP
+        # architecture cannot support, before consulting runtime ORT availability.
         if device != "auto" and device not in EP_SUPPORTED_DEVICES[ep_full]:
             raise ValueError(
                 f"EP '{ep}' does not support device '{device}'. "
-                f"Supported devices for {ep_full}: {', '.join(EP_SUPPORTED_DEVICES[ep_full])}."
+                f"Supported devices: {', '.join(EP_SUPPORTED_DEVICES[ep_full])}."
             )
         device_ep_map = {dev: (ep_full,) for dev, eps in device_ep_map.items() if ep_full in eps}
         if not device_ep_map:
--- a/src/winml/modelkit/sysinfo/device.py
+++ b/src/winml/modelkit/sysinfo/device.py
@@ -177,6 +177,11 @@ def resolve_device(
         ep_full = normalize_ep_name(ep)
         if ep_full not in EP_SUPPORTED_DEVICES:
             raise ValueError(f"Unknown EP '{ep}'. Expected one of: {sorted(EP_SUPPORTED_DEVICES)}")
+        if device != "auto" and device not in EP_SUPPORTED_DEVICES[ep_full]:
+            raise ValueError(
+                f"EP '{ep}' does not support device '{device}'. "
+                f"Supported devices for {ep_full}: {', '.join(EP_SUPPORTED_DEVICES[ep_full])}."
+            )
         device_ep_map = {dev: (ep_full,) for dev, eps in device_ep_map.items() if ep_full in eps}
         if not device_ep_map:
             raise ValueError(
--- a/tests/e2e/test_compile_e2e.py
+++ b/tests/e2e/test_compile_e2e.py
@@ -180,8 +180,7 @@ def assert_by_run_inference(
 
 
 def _find_qairt_sdk_root() -> Path | None:
-    """Locate an installed QAIRT SDK on this host, or None.
-    """
+    """Locate an installed QAIRT SDK on this host, or None."""
     for env_var in ("QNN_SDK_ROOT", "QAIRT_SDK_ROOT"):
         value = os.environ.get(env_var)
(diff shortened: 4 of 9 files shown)


DingmaomaoBJTU (commented):
## Code Review: PR #837 - Simplify compile EP selection logic

### Summary

The core idea - push device/EP policy enforcement into `resolve_device()` and let `_resolve_compile_provider` trust its inputs - is sound. The log-level adjustments are intentional and sensible. However, two concrete issues in the new `_resolve_compile_provider` need attention.

---

### Bug 1: Unguarded resolve_eps()[0] - potential IndexError

**File:** `src/winml/modelkit/commands/compile.py`, line 275

The old code guarded the empty-list case with `if not eps: return "CPUExecutionProvider"`.

`resolve_eps()` returns `[]` when the device has EP entries in `_get_device_ep_map_from_ort()` that don't appear in `_DEVICE_EP_MAP` (the priority table). Because `resolve_device()` checks against `_get_device_ep_map_from_ort()` while `resolve_eps()` cross-references `_DEVICE_EP_MAP`, the two can disagree - `resolve_device()` succeeds but `resolve_eps()` returns `[]`. The new code will then raise `IndexError`, which bubbles up as a confusing generic error.

Suggested fix:
```python
eps = resolve_eps(resolved_device)
if not eps:
    raise click.ClickException(f"No known EP available for device '{resolved_device}'.")
return eps[0]
```

---

### Warning: Unknown EP alias now silently falls through

**File:** `src/winml/modelkit/commands/compile.py`, lines 271-274

When `normalize_ep_name(ep)` returns `None` (unrecognized EP alias), the function silently ignores the user's `--ep` flag and uses the device default. The old code raised `click.UsageError("Unknown EP: {ep}")`. In the main execution path this is safe because `resolve_device()` catches the unknown EP first, but `_resolve_compile_provider` is also called from the `--list-compilers` branch (line 175) and is a semi-public helper. A defensive raise eliminates a subtle footgun.

---

### Info: Test reliability note

The updated tests write `b"fake"` to a `.onnx` file and invoke the CLI. They pass because `resolve_device()` raises the policy error before `is_compiled_onnx()` is called. This is correct today but fragile if call order in `compile()` changes. An explicit `_patch_device_ep_map` mock would make the tests independent of call order.

---

### What looks good

- The new policy check in `resolve_device()` fires before any ORT system calls - correct placement.
- `test_explicit_device_ep_policy_mismatch_raises` is a well-structured parametrized test.
- Log-level changes match the stated goal of making `-v` output more useful.
- The `test_explicit_ep_and_device_both_required` update from `(cpu, qnn)` to `(gpu, qnn)` is correct: `(cpu, qnn)` would now be rejected by the new policy gate before reaching the disambiguation logic.

src/winml/modelkit/commands/compile.py
```diff
@@ -264,22 +264,12 @@ def _resolve_compile_provider(resolved_device: str, ep: EPNameOrAlias | None) ->
...
-
-    eps = resolve_eps(resolved_device)
-    if not eps:
-        return "CPUExecutionProvider"
-    return eps[0]
+    if ep is not None:
+        normalized = normalize_ep_name(ep)
+        if normalized is not None:
+            return normalized
+    return resolve_eps(resolved_device)[0]
```
DingmaomaoBJTU, line 275: **Bug: unguarded index on potentially-empty list.** resolve_eps() returns [] when the device's runtime EPs (from _get_device_ep_map_from_ort()) do not intersect _DEVICE_EP_MAP; [0] then raises IndexError. The old code had 'if not eps: return CPUExecutionProvider' as a safety net. Suggest guarding: eps = resolve_eps(resolved_device); if not eps: raise click.ClickException(...); return eps[0]

src/winml/modelkit/commands/compile.py
```diff
@@ -264,22 +264,12 @@ def _resolve_compile_provider(resolved_device: str, ep: EPNameOrAlias | None) ->
...
-        # no need for an extra check here
-        return canonical
-
-    eps = resolve_eps(resolved_device)
-    if not eps:
-        return "CPUExecutionProvider"
-    return eps[0]
+    if ep is not None:
+        normalized = normalize_ep_name(ep)
+        if normalized is not None:
```
DingmaomaoBJTU, line 273: **Warning: unknown EP alias silently ignored.** When normalize_ep_name(ep) returns None (unrecognized alias), the new code falls through to resolve_eps(resolved_device)[0] instead of raising an error. The old code raised click.UsageError('Unknown EP: {ep}'). In the compile() main path resolve_device() catches this first, but _resolve_compile_provider is also called from the --list-compilers branch (line 175) and is a reusable helper. A defensive raise here keeps the function self-contained: if normalized is None: raise click.UsageError(f'Unknown EP: {ep!r}')

src/winml/modelkit/commands/compile.py
xieofxie: resolved_device already checked first

src/winml/modelkit/commands/compile.py
xieofxie: currently we only support known eps, click option also validate against that

src/winml/modelkit/sysinfo/device.py
```diff
@@ -177,6 +177,11 @@ def resolve_device(
         ep_full = normalize_ep_name(ep)
         if ep_full not in EP_SUPPORTED_DEVICES:
             raise ValueError(f"Unknown EP '{ep}'. Expected one of: {sorted(EP_SUPPORTED_DEVICES)}")
+        if device != "auto" and device not in EP_SUPPORTED_DEVICES[ep_full]:
```
DingmaomaoBJTU, line 182: Design note: static policy check fires before runtime ORT check. This hardcoded EP_SUPPORTED_DEVICES table gates the request before consulting what ORT actually reports on the system. If a future driver or version makes an EP genuinely run on a policy-forbidden device, this check would wrongly reject it. Suggest adding a comment: '# Static policy gate: rejects combos the EP architecture cannot support, independent of runtime.'

src/winml/modelkit/sysinfo/device.py
```diff
@@ -177,6 +177,11 @@ def resolve_device(
...
+            raise ValueError(
+                f"EP '{ep}' does not support device '{device}'. "
+                f"Supported devices for {ep_full}: {', '.join(EP_SUPPORTED_DEVICES[ep_full])}."
```
DingmaomaoBJTU: Nit: error message mixes alias and canonical name in the same sentence. First line uses the user-supplied alias (ep='qnn'), second line uses the internal full name (ep_full='QNNExecutionProvider'). Suggest keeping only the alias for consistency: remove '{ep_full}:' from the second line, e.g. 'Supported devices: npu, gpu.'

tests/unit/session/test_winml_session.py
```diff
@@ -681,12 +681,16 @@ def spy(self_sess, ep_devices, provider_opts):
         assert "C" in outputs
 
     def test_explicit_ep_and_device_both_required(self, simple_matmul_onnx: Path):
-        """Regression: ep=qnn + device=cpu must bind QNN-on-CPU, not QNN-on-NPU.
+        """Regression: ep=qnn + device=gpu must bind QNN-on-GPU, not QNN-on-NPU.
```
DingmaomaoBJTU, line 684: Behavioral change hidden in test update: the original regression test used (cpu, qnn) to verify disambiguation. It was changed to (gpu, qnn) because (cpu, qnn) is now rejected by the new static policy gate before reaching the disambiguation logic. This means (cpu, qnn) semantics changed from 'valid but ambiguous' to 'hard policy error' — that's a real behavior change, not just a refactor. Worth calling out explicitly in the PR description so reviewers know this is intentional.

src/winml/modelkit/sysinfo/device.py
xieofxie, line 182: Use this to hide one known issue https://github.com/microsoft/winml-cli/issues/860

xieofxie: added

src/winml/modelkit/sysinfo/device.py
xieofxie: updated

tests/unit/session/test_winml_session.py
xieofxie: added

tests/e2e/test_compile_e2e.py
```diff
@@ -768,15 +795,10 @@ def test_bad_input_device_ep_conflict(
     require_ep(ep)
     src_hash = _sha256(simple_matmul_onnx)
     result = _invoke("-m", str(simple_matmul_onnx), "--device", device, "--ep", ep)
-    # Two valid rejection paths depending on whether ORT enumerates ``ep``
-    # on ``device``:
-    #   * Yes (e.g. QNN on the cpu node) -> ``_resolve_compile_provider``
-    #     policy check raises ``--ep X cannot run on --device Y``.
-    #   * No (e.g. NvTensorRTRTX is GPU-only) -> ``resolve_device`` raises
-    #     ``Device 'X' requested but no compatible EP is available``.
+    # ``resolve_device`` raises `EP '{ep}' does not support device '{device}'``.
```
xieofxie, line 798: update one test case

src/winml/modelkit/session/ep_registry.py
```diff
@@ -269,8 +269,12 @@ def _load_ep_catalog(self) -> None:
         """Load EP catalog from WinML."""
         from windowsml import EpCatalog
 
+        checked_eps: set[EPName] = set()
         with EpCatalog() as catalog:
             for provider in catalog.find_all_providers():
+                if provider.name in checked_eps:
```
xieofxie, line 275: In my setup, it will check qnn twice and the second one failed. But it should not be the case, so add a set

src/winml/modelkit/commands/compile.py
DingmaomaoBJTU, line 275: Bug still unresolved: resolve_eps(resolved_device)[0] remains unguarded. resolve_eps() can return [] when runtime EPs from _get_device_ep_map_from_ort() do not intersect _DEVICE_EP_MAP, causing IndexError. The old 'if not eps: return CPUExecutionProvider' guard was removed. Please add: eps = resolve_eps(resolved_device); if not eps: raise click.ClickException(f'No known EP available for device {resolved_device!r}.'); return eps[0]

src/winml/modelkit/commands/compile.py
DingmaomaoBJTU, line 273: Warning still unresolved: when normalize_ep_name(ep) returns None (unrecognized EP alias), the code silently falls through to resolve_eps() rather than raising. In the normal compile() path resolve_device() catches this first, but _resolve_compile_provider is also called from the --list-compilers branch (line 175) before any validation. A one-liner fix: if normalized is None: raise click.UsageError(f'Unknown EP: {ep!r}')

xieofxie: _resolve_compile_provider is only reached after resolve_device(device, ep=ep) at line 167, which raises ValueError -> UsageError for any unrecognized EP. That runs before both the --list-compilers branch (175) and the main path (193), and --ep is also Click-constrained to the known set. So normalize_ep_name returning None here is unreachable; the docstring already states the helper trusts validated inputs. Resolving as won't-fix.

src/winml/modelkit/commands/compile.py
xieofxie: resolved_device is returned by resolve_device, which guarantees the device has at least one runtime EP. resolve_eps then filters the static _DEVICE_EP_MAP by that same ORT runtime map, so for a resolved device the result is non-empty (both derive from the same known-EP set). An empty list isn't reachable for a resolved device; the helper documents that it trusts validated inputs. Resolving as won't-fix.

tests/unit/commands/test_ep_device_pair.py
```diff
@@ -88,11 +88,18 @@ class TestCompileIncompatiblePair:
...
-        from winml.modelkit.commands.compile import _resolve_compile_provider
+    def test_incompatible_pair_rejected(self, device: str, ep: str, tmp_path) -> None:
+        from click.testing import CliRunner
+
+        from winml.modelkit.commands.compile import compile
+
+        model_file = tmp_path / "model.onnx"
+        model_file.write_bytes(b"fake")
+
+        result = CliRunner().invoke(compile, ["-m", str(model_file), "-d", device, "--ep", ep])
```
DingmaomaoBJTU, line 99: Info: test relies on call order inside compile() to trigger the error. The policy rejection fires before is_compiled_onnx() today, but if the call order changes the test will silently stop exercising the right code path. A _patch_device_ep_map mock (already used in tests/unit/sysinfo/test_device.py) would make this independent of compile() internals.

xieofxie: the mock _patch_device_ep_map is not related to order of is_compiled_onnx. I think we could revisit this when the order actually changed

DingmaomaoBJTU (approved):

Issue #838, opened 2026-06-09
detect_task is architecture-blind for encoder-decoder model types — all auto-detect as feature-extraction


timenick:
## Summary

`detect_task()` — the detector behind `winml inspect` (and the eval task-surfacing path) — returns `feature-extraction` for **every** encoder-decoder model type (`bart`, `t5`, `marian`, `mu2`, `qwen3`, `blip`, `vision-encoder-decoder`), regardless of the model's architecture head. A `BartForSequenceClassification` and a `BartForConditionalGeneration` both resolve to `feature-extraction`.

Note that `winml config` / `build` / `perf` auto-detect through a **different** path (`resolve_task_and_model_class` Case 1 → `_detect_task_and_class_from_config`), which **is** architecture-aware — so the two detectors already disagree on the same model today (see table below).

## Reproduction

```python
from transformers import AutoConfig
from winml.modelkit.loader.task import detect_task
for mid in ["facebook/bart-large-cnn", "facebook/bart-large-mnli", "facebook/bart-base"]:
    cfg = AutoConfig.from_pretrained(mid)
    print(mid, cfg.architectures[0], "->", detect_task(cfg)[0])
```

All three print `feature-extraction`, including `bart-large-mnli`, which is unambiguously sequence-classification.

## Already divergent across commands (today)

| model | architectures | `inspect` (`detect_task`) | `config` / `build` (`task=None`) |
|---|---|---|---|
| facebook/bart-large-cnn | BartForConditionalGeneration | feature-extraction | **fill-mask** |
| facebook/bart-large-mnli | BartForSequenceClassification | feature-extraction | **text-classification** |
| facebook/bart-base | BartModel | feature-extraction | feature-extraction |

`inspect` is head-blind. `config` / `build` are head-aware, but still mis-map the seq2seq generation head to `fill-mask` (should be `text2text-generation`).

## Blast radius

Every `model_type` with >1 entry in `MODEL_CLASS_MAPPING` is affected (detection picks the first key by insertion order):

```
bart / blip / marian / mu2 / qwen3 / t5 / vision-encoder-decoder
    = [feature-extraction, text2text-generation]      -> always picks feature-extraction
clip / siglip = [feature-extraction, image-feature-extraction] -> picks feature-extraction
sam / sam2 / sam2-video = [None, ...]                 -> None sentinel first -> falls through to arch-aware (correct)
```

## Root cause

Two unrelated concerns share one table:

[...]

PR #841, opened 2026-06-09
fix(task): make detect_task architecture-aware for multi-task model types

timenick:
## What

`detect_task` short-circuited to the first `(model_type, task)` key in `MODEL_CLASS_MAPPING`, so every encoder-decoder model type (`bart`, `t5`, `marian`, `mu2`, `qwen3`, `blip`, `vision-encoder-decoder`) auto-detected as `feature-extraction` regardless of the architecture head — and disagreed with the already-head-aware `config` / `build` path (which uses `_detect_task_and_class_from_config`).

## Change

Short-circuit step 1 only when a `model_type` maps to exactly one real (non-`None`) task. With multiple distinct tasks, fall through to architecture-aware detection. A `(model_type, None)` default-class sentinel no longer forces a fall-through, so single-real-task types like `sam` resolve to their real task (`mask-generation`).

## Result — `inspect` now agrees with `config` / `build`

| model | before (inspect) | after (inspect) | config / build |
|---|---|---|---|
| facebook/bart-large-mnli | feature-extraction | text-classification | text-classification |
| facebook/bart-large-cnn | feature-extraction | fill-mask\* | fill-mask\* |
| facebook/bart-base | feature-extraction | feature-extraction | feature-extraction |
| facebook/sam-vit-base | feature-extraction | mask-generation | mask-generation |
| clip / siglip | feature-extraction | feature-extraction | feature-extraction |

\* `fill-mask` for seq2seq generation models is still semantically wrong (optimum maps `BartForConditionalGeneration` to `fill-mask`). This PR only makes `inspect` consistent with `config` / `build`. Correcting `*ForConditionalGeneration -> text2text-generation` — which changes build artifacts and needs a `text2text-generation` export path — is **Step 2**, tracked in #838.

## Tests

- Unit: ambiguous multi-task type falls through to architecture-aware detection; a single-real-task type with a `None` sentinel (`sam`) resolves to its real task; a single-entry type still short-circuits.
- Integration (`test_task_consistency`): added `bart-large-mnli -> text-classification` and `sam-vit-base -> mask-generation`.

Addresses #838 (Step 1 of 2).

### DIFF
--- a/src/winml/modelkit/loader/task.py
+++ b/src/winml/modelkit/loader/task.py
@@ -345,7 +345,7 @@ def detect_task(config: PretrainedConfig) -> tuple[str, str]:
     #    to exactly one *real* (non-None) task. With multiple distinct tasks the
     #    architecture head is what disambiguates, so fall through to step 3 instead
     #    of guessing; the None sentinel alone never decides the task.
-    distinct_tasks = {
+    distinct_tasks: set[str] = {
         mapped
         for mt, mapped in HF_MODEL_CLASS_MAPPING
         if mt == model_type_normalized and mapped is not None
--- a/src/winml/modelkit/loader/task.py
+++ b/src/winml/modelkit/loader/task.py
@@ -337,11 +337,21 @@ def detect_task(config: PretrainedConfig) -> tuple[str, str]:
     task: str | None = None
     source = "none"
 
-    # 1. Explicit (model_type, task) mapping wins.
-    for mt, mapped_task in HF_MODEL_CLASS_MAPPING:
-        if mt == model_type_normalized:
-            task, source = mapped_task, "HF_MODEL_CLASS_MAPPING"
-            break
+    # 1. Explicit (model_type, task) mapping wins — but only when unambiguous.
+    #    The mapping is keyed by (model_type, task) and a model_type may appear
+    #    under several tasks: encoder-decoder types (bart/t5: feature-extraction +
+    #    text2text-generation) plus an optional (model_type, None) default-class
+    #    sentinel. Detection therefore short-circuits only when the model_type maps
+    #    to exactly one *real* (non-None) task. With multiple distinct tasks the
+    #    architecture head is what disambiguates, so fall through to step 3 instead
+    #    of guessing; the None sentinel alone never decides the task.
+    distinct_tasks = {
+        mapped
+        for mt, mapped in HF_MODEL_CLASS_MAPPING
+        if mt == model_type_normalized and mapped is not None
+    }
+    if len(distinct_tasks) == 1:
+        task, source = next(iter(distinct_tasks)), "HF_MODEL_CLASS_MAPPING"
 
     # 2. Wrapped-library model types (e.g. timm via "timm_wrapper") carry no
     #    `architectures`; resolve through their wrapped library instead of the
--- a/tests/integration/test_task_consistency.py
+++ b/tests/integration/test_task_consistency.py
@@ -20,6 +20,12 @@ from winml.modelkit.inspect import detect_task as inspect_detect_task
 from winml.modelkit.loader import detect_task
 
 
+# Every test here calls AutoConfig.from_pretrained, which hits the network unless
+# the HF config is already cached; mark the module so offline runs can skip via
+# ``-m 'not network'`` instead of failing with a confusing ConnectionError.
+pytestmark = pytest.mark.network
+
+
 # (model_id, expected WinMLTask) — exercises the D2 modality upgrade (DINOv2),
 # a top-level-nested vision config that must NOT upgrade (CLIP), and a text control.
 CASES = [
--- a/tests/integration/test_task_consistency.py
+++ b/tests/integration/test_task_consistency.py
@@ -26,6 +26,14 @@ CASES = [
     ("facebook/dinov2-small", "image-feature-extraction"),
     ("openai/clip-vit-base-patch32", "feature-extraction"),  # image_size nested -> no D2
     ("bert-base-uncased", "fill-mask"),
+    # bart maps to >1 task (feature-extraction + text2text-generation) in
+    # MODEL_CLASS_MAPPING; detect_task must consult the architecture head instead
+    # of short-circuiting to the first key, so a sequence-classification BART
+    # resolves to text-classification rather than the lossy feature-extraction.
+    ("facebook/bart-large-mnli", "text-classification"),
+    # sam has a (sam, None) default-class sentinel plus a single real task; the
+    # sentinel must not make detection fall through — it resolves to mask-generation.
+    ("facebook/sam-vit-base", "mask-generation"),
 ]
 
 
--- a/tests/unit/loader/test_detect_task.py
+++ b/tests/unit/loader/test_detect_task.py
@@ -83,6 +83,40 @@ def test_detect_task_falls_back_to_hf_task_defaults() -> None:
     assert source == "HF_TASK_DEFAULTS"
 
 
+def test_detect_task_does_not_short_circuit_for_ambiguous_model_type() -> None:
+    """A model_type with >1 distinct task in MODEL_CLASS_MAPPING (e.g. bart:
+    feature-extraction + text2text-generation) cannot be disambiguated by
+    model_type alone, so detect_task must fall through to architecture-aware
+    detection instead of short-circuiting to the first key (feature-extraction)."""
+    cfg = _FakeConfig("bart", architectures=["BartForSequenceClassification"])
+    with patch(_DETECT, return_value="text-classification") as m:
+        task, source = detect_task(cfg)
+    assert task == "text-classification"
+    assert source == "TasksManager"
+    m.assert_called_once()
+
+
+def test_detect_task_uses_single_real_task_despite_none_sentinel() -> None:
+    """A model_type with a None default-class sentinel plus exactly one real task
+    (sam: (sam, None) + (sam, mask-generation)) short-circuits to that real task
+    rather than falling through on the None sentinel."""
+    cfg = _FakeConfig("sam")
+    with patch(_DETECT) as m:
+        task, source = detect_task(cfg)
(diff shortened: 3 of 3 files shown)


src/winml/modelkit/loader/task.py
```diff
@@ -337,11 +337,21 @@ def detect_task(config: PretrainedConfig) -> tuple[str, str]:
...
+    # 1. Explicit (model_type, task) mapping wins — but only when unambiguous.
+    #    The mapping is keyed by (model_type, task) and a model_type may appear
+    #    under several tasks: encoder-decoder types (bart/t5: feature-extraction +
+    #    text2text-generation) plus an optional (model_type, None) default-class
+    #    sentinel. Detection therefore short-circuits only when the model_type maps
+    #    to exactly one *real* (non-None) task. With multiple distinct tasks the
+    #    architecture head is what disambiguates, so fall through to step 3 instead
+    #    of guessing; the None sentinel alone never decides the task.
+    distinct_tasks = {
+        mapped
```
DingmaomaoBJTU, line 349: **Type safety: `distinct_tasks` is inferred as `set[str | None]` by mypy/pyright.**

The `and mapped is not None` guard in the comprehension filter narrows the *runtime* type to `str`, but mypy/pyright cannot narrow types inside comprehension conditions the same way they do `if` blocks — they will infer `set[str | None]` for `distinct_tasks`. This causes `next(iter(distinct_tasks))` on line 354 to type-check as `str | None` rather than `str`.

Fix with an explicit annotation:
```python
distinct_tasks: set[str] = {
    mapped
    for mt, mapped in HF_MODEL_CLASS_MAPPING
    if mt == model_type_normalized and mapped is not None
}
```

src/winml/modelkit/loader/task.py
```diff
@@ -337,11 +337,21 @@ def detect_task(config: PretrainedConfig) -> tuple[str, str]:
...
+        for mt, mapped in HF_MODEL_CLASS_MAPPING
+        if mt == model_type_normalized and mapped is not None
+    }
+    if len(distinct_tasks) == 1:
+        task, source = next(iter(distinct_tasks)), "HF_MODEL_CLASS_MAPPING"
```
DingmaomaoBJTU, line 355: **Performance: O(n) scan over `HF_MODEL_CLASS_MAPPING` on every `detect_task` call.**

`HF_MODEL_CLASS_MAPPING` is a module-level constant that never changes at runtime, yet `distinct_tasks` is recomputed from a full key scan on every invocation. As the mapping grows with new model families, this becomes an unbounded linear scan per call.

Consider precomputing a `dict[str, frozenset[str]]` index once at import time (inside the lazy-import block or lazily-cached on first access) so the lookup is O(1).

tests/unit/loader/test_detect_task.py
```diff
@@ -83,6 +83,40 @@ def test_detect_task_falls_back_to_hf_task_defaults() -> None:
...
+def test_detect_task_uses_single_real_task_despite_none_sentinel() -> None:
+    """A model_type with a None default-class sentinel plus exactly one real task
+    (sam: (sam, None) + (sam, mask-generation)) short-circuits to that real task
+    rather than falling through on the None sentinel."""
+    cfg = _FakeConfig("sam")
+    with patch(_DETECT) as m:
+        task, source = detect_task(cfg)
+    assert (task, source) == ("mask-generation", "HF_MODEL_CLASS_MAPPING")
+    m.assert_not_called()
+
```
DingmaomaoBJTU, line 108: **Missing coverage for `sam2` / `sam2-video` — the multi-task fallthrough case.**

`sam2` has four real entries in `MODEL_CLASS_MAPPING` (`image-segmentation`, `feature-extraction`, `image-feature-extraction`, `mask-generation`), so `len(distinct_tasks) == 4` and the code falls through to TasksManager, which returns `feature-extraction`. The `(sam2, None)` sentinel has zero effect on `detect_task` for multi-task model types (this matches the pre-PR behaviour, but is undocumented).

No test currently pins this, so a future mapping change could silently regress `sam2` detection. Suggested addition:
```python
def test_detect_task_falls_through_for_multi_task_model_type_sam2() -> None:
    """sam2 has 4 real tasks in MODEL_CLASS_MAPPING; must fall through to TasksManager."""
    cfg = _FakeConfig("sam2")
    with patch(_DETECT, return_value="feature-extraction") as m:
        task, source = detect_task(cfg)
    assert source == "TasksManager"
    m.assert_called_once()
```

tests/integration/test_task_consistency.py
```diff
@@ -26,6 +26,14 @@
...
     ("bert-base-uncased", "fill-mask"),
+    # bart maps to >1 task (feature-extraction + text2text-generation) in
+    # MODEL_CLASS_MAPPING; detect_task must consult the architecture head instead
+    # of short-circuiting to the first key, so a sequence-classification BART
+    # resolves to text-classification rather than the lossy feature-extraction.
+    ("facebook/bart-large-mnli", "text-classification"),
+    # sam has a (sam, None) default-class sentinel plus a single real task; the
+    # sentinel must not make detection fall through — it resolves to mask-generation.
+    ("facebook/sam-vit-base", "mask-generation"),
 ]
```
DingmaomaoBJTU, line 43: **`AutoConfig.from_pretrained` makes a network call unless configs are pre-cached.**

The new `facebook/bart-large-mnli` and `facebook/sam-vit-base` entries require either a pre-populated HuggingFace cache or live internet access. In offline CI this yields a `ConnectionError` rather than a test assertion failure, making the failure mode confusing.

Consider adding a `@pytest.mark.network` (or equivalent) marker so the parametrize group can be explicitly skipped in offline runs with `-m 'not network'`.

DingmaomaoBJTU (commented):
Overall this is a correct and clean fix. The core logic is sound: collect only the non-None task entries for a given model_type, and short-circuit only when there is exactly one — otherwise fall through to the architecture-aware TasksManager path. Three issues worth addressing below.

src/winml/modelkit/loader/task.py
timenick: Fixed in `72794fd7` — added `distinct_tasks: set[str]`. With the explicit target type mypy narrows the `mapped is not None` filter, so `next(iter(distinct_tasks))` types as `str` (verified `reveal_type` → `str`).

Generated with Claude Code

src/winml/modelkit/loader/task.py
timenick: I'd prefer to leave this as-is. `detect_task` is a cold path (once per model in inspect/config/eval, not a hot loop), and the scan is over a small constant mapping (~25 entries) whose cost is dwarfed by what the call already does — `from ..models import HF_MODEL_CLASS_MAPPING` triggers model registrations, and the fall-through `_detect_task_from_config` imports optimum and runs TasksManager. Precomputing a module-level index would also force that `HF_MODEL_CLASS_MAPPING` import eagerly at module load, whereas it's deliberately a lazy import inside the function to avoid import-time cost/cycles. Happy to revisit if the mapping grows substantially.

Generated with Claude Code

tests/unit/loader/test_detect_task.py
timenick: Added in `72794fd7` — `test_detect_task_falls_through_for_multi_task_model_type_sam2` asserts sam2 (4 real tasks) falls through to TasksManager (`source == "TasksManager"`, `_detect_task_from_config` called once).

Generated with Claude Code

tests/integration/test_task_consistency.py
timenick: Added in `72794fd7` — module-level `pytestmark = pytest.mark.network` (the marker is already registered in `pyproject.toml`), so the whole file, including the pre-existing dinov2/clip/bert cases, can be skipped offline with `-m 'not network'`.

Generated with Claude Code

src/winml/modelkit/loader/task.py
```diff
@@ -337,11 +337,21 @@ def detect_task(config: PretrainedConfig) -> tuple[str, str]:
...
+    #    to exactly one *real* (non-None) task. With multiple distinct tasks the
+    #    architecture head is what disambiguates, so fall through to step 3 instead
+    #    of guessing; the None sentinel alone never decides the task.
+    distinct_tasks: set[str] = {
+        mapped
+        for mt, mapped in HF_MODEL_CLASS_MAPPING
+        if mt == model_type_normalized and mapped is not None
+    }
+    if len(distinct_tasks) == 1:
+        task, source = next(iter(distinct_tasks)), "HF_MODEL_CLASS_MAPPING"
```
DingmaomaoBJTU, line 355: **Still open: O(n) scan on every `detect_task` call (performance suggestion from previous review).**

`HF_MODEL_CLASS_MAPPING` is a module-level constant, yet this set comprehension rescans all its keys on every invocation. For the current mapping size this is negligible, but as new model families land the cost grows linearly.

Not a blocker, but consider precomputing a `dict[str, frozenset[str]]` index at import time:
```python
# Built once inside the lazy-import block, e.g.:
_TASKS_BY_MODEL_TYPE: dict[str, frozenset[str]] = {}
for _mt, _task in HF_MODEL_CLASS_MAPPING:
    if _task is not None:
        _TASKS_BY_MODEL_TYPE.setdefault(_mt, set()).add(_task)
_TASKS_BY_MODEL_TYPE = {k: frozenset(v) for k, v in _TASKS_BY_MODEL_TYPE.items()}

# Then the hot path becomes O(1):
distinct_tasks = _TASKS_BY_MODEL_TYPE.get(model_type_normalized, frozenset())
```

DingmaomaoBJTU (approved):
3 of 4 previously raised issues are resolved. One remains open (see inline comment).

- **Type safety** (`distinct_tasks: set[str]`) ✅ Fixed.
- **`sam2` unit test** (`test_detect_task_falls_through_for_multi_task_model_type_sam2`) ✅ Added.
- **Network marker** (`pytestmark = pytest.mark.network`) ✅ Applied module-wide with a helpful comment.
- **Performance (O(n) scan)** — Still open; see inline comment. Not a blocker given the current mapping size, but worth tracking.

PR #844, opened 2026-06-09
chore(cli): convert boolean is_flag options to --flag/--no-flag pairs

### LINKED ISSUE #843: chore(cli): convert boolean is_flag options to --flag/--no-flag pairs for consistency
## Summary

Several boolean CLI flags use bare `is_flag=True` (single-direction, e.g. `--trust-remote-code`) instead of Click's `--flag/--no-flag` pair pattern. This is inconsistent with newer options like `--skip-build/--no-skip-build` (added in #659) and makes it harder for users to:

- Explicitly set a flag to `False` in scripts/automation
- Override a config-file default from the CLI
- Understand the current state from `--help` output

## Flags to convert

The following `is_flag=True` options should become `--xxx/--no-xxx` boolean pairs:

| Command | Current flag | Proposed pair |
|---------|-------------|---------------|
| `build` | `--use-cache` | `--use-cache/--no-use-cache` |
| `build` | `--rebuild` | `--rebuild/--no-rebuild` |
| `build` | `--no-quant` | `--quant/--no-quant` |
| `build` | `--no-analyze` | `--analyze/--no-analyze` |
| `build` | `--no-optimize` | `--optimize/--no-optimize` |
| `compile` | `--embed` | `--embed/--no-embed` |
| `eval` | `--streaming` | `--streaming/--no-streaming` |
| `export` | `--with-report` | `--with-report/--no-with-report` |
| `export` | `--clean-onnx` | `--clean-onnx/--no-clean-onnx` |
| `export` | `--dynamo` | `--dynamo/--no-dynamo` |
| `export` | `--no-hierarchy` | `--hierarchy/--no-hierarchy` |
| `inspect` | `--hierarchy` | `--hierarchy/--no-hierarchy` |
| `perf` | `--no-quantize` | `--quantize/--no-quantize` |
| `perf` | `--rebuild` | `--rebuild/--no-rebuild` |
| `perf` | `--ignore-cache` | `--ignore-cache/--no-ignore-cache` |
| `perf` | `--monitor` | `--monitor/--no-monitor` |
| `quantize` | `--per-channel` | `--per-channel/--no-per-channel` |
| `quantize` | `--symmetric` | `--symmetric/--no-symmetric` |
| `serve` | `--multi` | `--multi/--no-multi` |
| `serve` | `--auto-reload` | `--auto-reload/--no-auto-reload` |
| `cli.py` | `--trust-remote-code` | `--trust-remote-code/--no-trust-remote-code` |
| `cli.py` | `--allow-unsupported-nodes` | `--allow-unsupported-nodes/--no-allow-unsupported-nodes` |

**Out of scope** (action triggers, not toggles): `--list`, `--schema`, `--list-tasks`, `--list-capabilities`, `--list-rewrites`, `--list-device`, `--list-ep`, `--connect`, `--quiet`

## Acceptance Criteria

- [ ] All listed flags converted to `--flag/--no-flag` Click boolean pairs
- [ ] Existing tests updated to reflect new flag names where needed
- [ ] `--no-xxx` named flags (e.g. `--no-quant`, `--no-analyze`) renamed to positive form with negation pair
- [ ] `--help` output shows both forms and default value
- [ ] No breaking change for users already using the existing `--flag` form (it remains valid)

DingmaomaoBJTU:
## Summary

Converts all `is_flag=True` Click boolean options to `--flag/--no-flag` pairs for consistency with the existing `--skip-build/--no-skip-build` pattern, as specified in #843.

## Changes

**Positive flags** (`--use-cache`, `--rebuild`, `--embed`, `--streaming`, `--with-report`, `--dynamo`, `--per-channel`, `--symmetric`, `--multi`, `--auto-reload`, `--ignore-cache`, `--monitor`, `--trust-remote-code`, `--allow-unsupported-nodes`) now accept an explicit `--no-xxx` counterpart.

**Negative flags** (`--no-quant`, `--no-analyze`, `--no-optimize`, `--no-quantize`, `--no-hierarchy`) are renamed to positive form (`--quant/--no-quant`, etc.) with `default=True`, so `--no-xxx` still works without breaking existing scripts.

**Export `--clean-onnx`** is preserved as a hidden deprecated flag that maps to `--no-hierarchy` behavior.

All `--help` output now shows both forms and the default value via `show_default=True`.

## Backward Compatibility

- All existing flag forms remain valid (e.g., `--no-quant` still works as the negative side of the pair)
- `--clean-onnx` preserved as hidden deprecated alias for `--no-hierarchy`
- No changes to internal helper function signatures or dataclass fields

Closes #843

### DIFF
--- a/src/winml/modelkit/commands/export.py
+++ b/src/winml/modelkit/commands/export.py
@@ -71,24 +71,31 @@ def _delete_onnx_with_external_data(onnx_path: Path) -> None:
 )
 @cli_utils.output_option("Output ONNX file path (e.g., model.onnx)", required=True)
 @click.option(
-    "--with-report",
-    is_flag=True,
+    "--with-report/--no-with-report",
     default=False,
+    show_default=True,
     help="Generate full export reports (markdown, JSON, console tree)",
 )
 @click.option(
-    "--clean-onnx",
-    "--no-hierarchy",
-    "no_hierarchy",
-    is_flag=True,
-    default=False,
-    help="Skip embedding hierarchy_tag metadata in ONNX (clean ONNX output)",
+    "--hierarchy/--no-hierarchy",
+    "hierarchy",
+    default=True,
+    show_default=True,
+    help="Embed hierarchy_tag metadata in ONNX output",
 )
 @click.option(
-    "--dynamo",
-    "dynamo",
+    "--clean-onnx",
+    "clean_onnx",
     is_flag=True,
     default=False,
+    hidden=True,
+    help="Deprecated alias for --no-hierarchy.",
+)
+@click.option(
+    "--dynamo/--no-dynamo",
+    "dynamo",
+    default=False,
+    show_default=True,
     help="Enable PyTorch 2.9+ dynamo export for rich node metadata",
 )
 @click.option(
@@ -132,7 +139,8 @@ def export(
     verbose: int,
     quiet: bool,
     with_report: bool,
-    no_hierarchy: bool,
+    hierarchy: bool,
+    clean_onnx: bool,
     dynamo: bool,
     torch_module: str | None,
     task: str | None,
@@ -195,14 +203,21 @@ def export(
         _build_export_dict = ec
         if not cli_utils.is_cli_provided(ctx, "task") and "task" in lc:
             task = lc["task"]
-        if not cli_utils.is_cli_provided(ctx, "no_hierarchy") and "enable_hierarchy_tags" in ec:
-            no_hierarchy = not ec["enable_hierarchy_tags"]
+        if (
+            not cli_utils.is_cli_provided(ctx, "hierarchy")
+            and not cli_utils.is_cli_provided(ctx, "clean_onnx")
+            and "enable_hierarchy_tags" in ec
+        ):
+            hierarchy = ec["enable_hierarchy_tags"]
         if not cli_utils.is_cli_provided(ctx, "dynamo") and "dynamo" in ec:
             dynamo = ec["dynamo"]
 
     from ..export import InputTensorSpec, OutputTensorSpec, WinMLExportConfig
     from ..export import export_pytorch as export_onnx
     from ..loader import load_hf_model
+
+    if clean_onnx and not cli_utils.is_cli_provided(ctx, "hierarchy"):
+        hierarchy = False
 
     # Configure logging — stderr only, shared format with the rest of the CLI.
     configure_logging(verbosity=verbose, quiet=quiet)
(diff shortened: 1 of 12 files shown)


tests/unit/commands/test_boolean_flag_pairs.py
```diff
@@ -0,0 +1,241 @@
...
+            if param.name == param_name:
+                found = True
+                # For boolean flag pairs, check that the flag string appears
+                # in either the primary or secondary opts
+                all_opts = list(param.opts) + list(param.secondary_opts)
+                # The flag or its bare form should be recognized
+                flag_bare = flag.lstrip("-")
+                assert (
+                    any(flag_bare in opt.lstrip("-") for opt in all_opts)
+                    or param.name == param_name
```
timenick: This test as written doesn’t actually verify what its name and docstring promise. A few specifics:

1. The `expected_value` parameter on the test signature is never asserted against.
2. The fixture `_capture_params` defined above (~L651) is also never used.
3. The final assertion is a tautology:

```python
for param in command.params:
    if param.name == param_name:
        found = True
        ...
        assert (
            any(flag_bare in opt.lstrip("-") for opt in all_opts)
            or param.name == param_name  # always True here
        )
        break
```

Because we only enter the `if` branch when `param.name == param_name`, the `or` clause is always `True`, so the `any(...)` check is effectively dead.

Net effect: this test only confirms (a) Click accepts the flag with `--help` and (b) a param by that name exists on the command. A typo making `--quant` set `quant=False` (or `--no-quant` set `quant=True`) would still pass.

Suggested fix — actually run the command with the flag and capture `ctx.params`. Something like:

```python
@pytest.fixture
def captured():
    return {}

def test_flag_sets_expected_value(command, flag, param_name, expected_value, captured):
    @command.callback.__wrapped__ if False else ...  # or use ctx.invoke wrapper
    runner = CliRunner()
    # Use standalone_mode=False so we can intercept ctx.params before the body runs.
    ...
    assert captured[param_name] is expected_value
```

Or, simpler: parse via `command.make_context(...)` directly and check `ctx.params[param_name]` — no need to invoke the callback at all.

src/winml/modelkit/commands/export.py
```diff
@@ -195,15 +203,22 @@ def export(
...
+        ):
+            hierarchy = ec["enable_hierarchy_tags"]
         if not cli_utils.is_cli_provided(ctx, "dynamo") and "dynamo" in ec:
             dynamo = ec["dynamo"]
 
     from ..export import InputTensorSpec, OutputTensorSpec, WinMLExportConfig
     from ..export import export_pytorch as export_onnx
     from ..loader import load_hf_model
 
+    if clean_onnx and not cli_utils.is_cli_provided(ctx, "hierarchy"):
```
timenick: `--clean-onnx` is preserved as a hidden alias but the runtime path doesn’t emit a deprecation warning. Existing scripts using `--clean-onnx` will keep working silently, with no signal that they should migrate to `--no-hierarchy`. Since the alias is hidden from `--help`, this is the only realistic channel to surface the rename.

Consider a one-liner when `clean_onnx` is truthy, e.g.:

```python
if clean_onnx:
    click.echo(
        "warning: --clean-onnx is deprecated; use --no-hierarchy instead.",
        err=True,
    )
    if not cli_utils.is_cli_provided(ctx, "hierarchy"):
        hierarchy = False
```

timenick (commented):
Two observations — the first is a real test gap, the second is a UX nit.

timenick (approved):
Both previous comments are addressed correctly. Verified locally — all 41 parametrized cases of test_flag_sets_expected_value pass, and test_export_clean_onnx_disables_hierarchy_tags confirms the deprecation warning. LGTM.

Issue #850, opened 2026-06-09
winml config/build for encoder-decoder models without --task emits a decoder-only, non-runnable model


timenick:
## Summary

For encoder-decoder model types (t5, marian, mu2, bart, vision-encoder-decoder, blip), running `winml config` / `build` **without `--task`** produces a **decoder-only** export — it takes `encoder_hidden_states` as an *input* but contains no encoder to produce it, so the result can't run standalone.

## Reproduction

```
winml config -m t5-small      # no --task
```

```
loader.task  : text2text-generation
loader.class : T5DecoderWrapper
inputs       : decoder_input_ids, encoder_hidden_states, past_0_key/value … past_5_*
outputs      : logits, present_0..5_key/value
```

The export is the decoder half only; `encoder_hidden_states` has no producer in the build.

## Root cause

The full encoder+decoder export is built via the **composite path**, which is keyed on concrete tasks (`(t5, summarization)`, `(t5, translation)`, `(bart, summarization)`, …) and only fires for an explicit `--task` — `_resolve_composite_model_components` returns early when `task is None`. With no `--task`, auto-detection resolves `text2text-generation`, which (a) is not a composite key and (b) maps to the single `*DecoderWrapper` in `MODEL_CLASS_MAPPING`, so only the decoder is exported.

## Scope

All encoder-decoder model types: t5, marian, mu2, vision-encoder-decoder, blip. (bart currently auto-detects `fill-mask` so it hits a *different* failure today; once the detection fix in #838 lands, bart joins the others in this decoder-only path.)

## Proposed fixes (pick one)

- **Route no-`--task` seq2seq through the full composite** — when auto-detection yields an encoder-decoder generation task for an `is_encoder_decoder` model, build the encoder+decoder composite instead of the decoder alone.
- **Fail loudly** — reject a no-`--task` build for encoder-decoder models with a clear message to pass `--task summarization|translation|…`, instead of silently emitting a decoder-only model.

DingmaomaoBJTU:
## Triage Summary

**Priority:** P1
**Component:** config / build / loader
**Assigned:** @timenick

**Rationale:** Encoder-decoder models (t5, marian, bart, blip, etc.) without explicit --task produce decoder-only exports that cannot run standalone. This silently produces a broken artifact — users won't discover the issue until inference time. Impacts all encoder-decoder model families.

**Next steps:** When task is not specified for encoder-decoder architectures, either auto-detect the composite task or emit a clear error/warning that --task is required for these model types.

PR #851, opened 2026-06-09
fix(task): detect encoder-decoder fill-mask heads as text2text-generation

### LINKED ISSUE #838: detect_task is architecture-blind for encoder-decoder model types — all auto-detect as feature-extraction

timenick:
## What

optimum's `TasksManager` maps some encoder-decoder `*ForConditionalGeneration` classes (e.g. `BartForConditionalGeneration`) to `fill-mask`, so `bart` auto-detected as `fill-mask` in `inspect` and `config` / `build`, while `t5` / `marian` correctly resolved to `text2text-generation`.

## Change

A universal correction (architecture-agnostic — keyed on the `is_encoder_decoder` flag, not model names): a config reported as `fill-mask` that is `is_encoder_decoder=True` is actually a seq2seq generator → `text2text-generation`. Applied at both detection sites:

- `_detect_task_from_config` → fixes `inspect` / `detect_task`.
- `_detect_task_and_class_from_config` → fixes `config` / `build` and makes the class resolve to `BartDecoderWrapper` (consistent with how t5 resolves to `T5DecoderWrapper`).

The flag must be explicitly `True` (HF configs set a real bool), so partial/duck-typed configs are never silently upgraded.

## Result

| model | before (inspect / config) | after |
|---|---|---|
| facebook/bart-large-cnn | fill-mask / AutoModelForMaskedLM | text2text-generation / BartDecoderWrapper |
| bert / roberta (real fill-mask) | fill-mask | unchanged |
| t5 / marian | text2text-generation | unchanged |

## Tests

- Unit: `_upgrade_fill_mask_for_seq2seq` — upgrades encoder-decoder fill-mask; leaves encoder-only fill-mask; only touches `fill-mask`.
- Integration: `bart-large-cnn → text2text-generation` (detect_task) and `→ (text2text-generation, BartDecoderWrapper)` (config/build); `bert → fill-mask` guard.

Fixes #838 (Step 2 of 2). The separate decoder-only no-`--task` seq2seq build issue is tracked in #850.

### DIFF
--- a/src/winml/modelkit/loader/task.py
+++ b/src/winml/modelkit/loader/task.py
@@ -498,6 +498,11 @@ def _detect_task_and_class_from_config(config: PretrainedConfig) -> tuple[str, t
                 # default. If one ever exposes multiple, supported[0] is an
                 # arbitrary pick -- warn (listing the tasks) but still proceed;
                 # pass --task to choose a different one.
+                # No _upgrade_fill_mask_for_seq2seq here: this task comes from the
+                # wrapped library's ONNX export list (get_supported_tasks), not from
+                # class->task inference, so the optimum fill-mask mislabel cannot
+                # arise on this path; rewriting could also yield a task outside that
+                # list. The correction is intentionally scoped to the class->task paths.
                 task = supported[0]
                 if len(supported) > 1:
                     logger.warning(
--- a/src/winml/modelkit/loader/task.py
+++ b/src/winml/modelkit/loader/task.py
@@ -271,7 +271,24 @@ def _detect_task_from_config(config: PretrainedConfig) -> str:
     """
     model_class = _resolve_model_class_from_config(config)
     task = _detect_task_from_model_class(model_class)
+    task = _upgrade_fill_mask_for_seq2seq(task, config)
     logger.info("Detected task: %s (from %s)", task, model_class.__name__)
+    return task
+
+
+def _upgrade_fill_mask_for_seq2seq(task: str, config: PretrainedConfig) -> str:
+    """Correct Optimum's ``fill-mask`` mislabel for encoder-decoder generation heads.
+
+    ``TasksManager`` maps some encoder-decoder ``*ForConditionalGeneration`` classes
+    (e.g. ``BartForConditionalGeneration``) to ``fill-mask``. A real masked-LM is
+    encoder-only, so a config that is ``is_encoder_decoder`` yet reported as
+    ``fill-mask`` is actually a seq2seq generator -> ``text2text-generation``.
+    Architecture-agnostic: keyed on the ``is_encoder_decoder`` flag, not model names.
+    Requires the flag to be explicitly ``True`` (HF configs set a real bool) so a
+    partial/duck-typed config without the field is never silently upgraded.
+    """
+    if task == "fill-mask" and getattr(config, "is_encoder_decoder", False) is True:
+        return "text2text-generation"
     return task
 
 
@@ -508,6 +525,7 @@ def _detect_task_and_class_from_config(config: PretrainedConfig) -> tuple[str, t
 
     # [2] Infer task from model class
     task = _detect_task_from_model_class(arch_model_class)
+    task = _upgrade_fill_mask_for_seq2seq(task, config)
     logger.info("Detected task: %s (from %s)", task, arch_name)
 
     # [3] Get model_type - REQUIRED for specialization lookup
--- a/tests/integration/loader/test_detect_task_and_class.py
+++ b/tests/integration/loader/test_detect_task_and_class.py
@@ -114,6 +114,9 @@ class TestSeq2SeqFillMaskCorrection:
         task, resolved_class = _detect_task_and_class_from_config(config)
 
         assert task == "text2text-generation"
+        # BartDecoderWrapper is WinML's encoder-decoder decoder sub-component class
+        # registered for (bart, text2text-generation) — i.e. the seq2seq route,
+        # not the AutoModelForMaskedLM that the fill-mask mislabel would produce.
         assert resolved_class.__name__ == "BartDecoderWrapper"
 
     def test_bert_masked_lm_stays_fill_mask(self):
--- a/tests/integration/loader/test_detect_task_and_class.py
+++ b/tests/integration/loader/test_detect_task_and_class.py
@@ -100,3 +100,26 @@ class TestSupportedModelsIntegration:
 
         assert task == "image-classification"
         assert "ImageClassification" in resolved_class.__name__
+
+
+@pytest.mark.slow
+class TestSeq2SeqFillMaskCorrection:
+    """Encoder-decoder generation heads that Optimum mislabels as fill-mask must
+    resolve to text2text-generation, not the encoder-only masked-LM task."""
+
+    def test_bart_conditional_generation_is_text2text(self):
+        """BartForConditionalGeneration -> text2text-generation / BartDecoderWrapper."""
+        config = AutoConfig.from_pretrained("facebook/bart-large-cnn")
+
+        task, resolved_class = _detect_task_and_class_from_config(config)
+
+        assert task == "text2text-generation"
+        assert resolved_class.__name__ == "BartDecoderWrapper"
+
+    def test_bert_masked_lm_stays_fill_mask(self):
+        """Encoder-only masked-LM is untouched (is_encoder_decoder is False)."""
+        config = AutoConfig.from_pretrained("bert-base-uncased")
+
+        task, _ = _detect_task_and_class_from_config(config)
+
+        assert task == "fill-mask"
--- a/tests/integration/test_task_consistency.py
+++ b/tests/integration/test_task_consistency.py
@@ -40,6 +40,9 @@ CASES = [
     # sam has a (sam, None) default-class sentinel plus a single real task; the
     # sentinel must not make detection fall through — it resolves to mask-generation.
     ("facebook/sam-vit-base", "mask-generation"),
+    # optimum mislabels BartForConditionalGeneration as fill-mask; an encoder-decoder
+    # reported as fill-mask is a seq2seq generator -> text2text-generation.
+    ("facebook/bart-large-cnn", "text2text-generation"),
 ]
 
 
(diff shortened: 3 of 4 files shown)


src/winml/modelkit/loader/task.py
```diff
@@ -271,10 +271,27 @@ def _detect_task_from_config(config: PretrainedConfig) -> str:
...
+    ``TasksManager`` maps some encoder-decoder ``*ForConditionalGeneration`` classes
+    (e.g. ``BartForConditionalGeneration``) to ``fill-mask``. A real masked-LM is
+    encoder-only, so a config that is ``is_encoder_decoder`` yet reported as
+    ``fill-mask`` is actually a seq2seq generator -> ``text2text-generation``.
+    Architecture-agnostic: keyed on the ``is_encoder_decoder`` flag, not model names.
+    Requires the flag to be explicitly ``True`` (HF configs set a real bool) so a
+    partial/duck-typed config without the field is never silently upgraded.
+    """
+    if task == "fill-mask" and getattr(config, "is_encoder_decoder", False) is True:
+        return "text2text-generation"
```
vortex-captain, line 291: does `"text2text-generation"` for t5-small config+build?

timenick: Yes — t5-small resolves to `text2text-generation` in both inspect and config/build, but it gets there **directly via optimum**, not through this change. On `main` (no helper):

```
t5-small        optimum_native=text2text-generation  config=(text2text-generation, T5DecoderWrapper)
bart-large-cnn  optimum_native=fill-mask             config=(fill-mask, AutoModelForMaskedLM)
```

`TasksManager.infer_task_from_model(T5ForConditionalGeneration)` is already `text2text-generation`; only `BartForConditionalGeneration` maps to `fill-mask`. The helper's `task == "fill-mask"` guard means it fires only for that mislabel, so this PR changes **only bart** (`fill-mask → text2text-generation` / `BartDecoderWrapper`). t5/marian were already correct and are unaffected.

Generated with Claude Code

DingmaomaoBJTU (commented):
## Review: fix(task): detect encoder-decoder fill-mask heads as text2text-generation

**Overall**: The core logic is sound and well-documented. The is True identity check (not ==) is a nice touch that prevents truthy-coercion from triggering the upgrade on non-bool values. Architecture-agnostic keying on is_encoder_decoder is preferable to model-name matching.

**One structural gap**: The _upgrade_fill_mask_for_seq2seq call is missing from the wrapped-library early-return path (~line 520) inside _detect_task_and_class_from_config. Currently harmless (no wrapped library produces fill-mask today), but logically inconsistent with the rest of the function. See inline comment.

**Test coverage**: Good unit tests for the three boolean cases (True/False/other-task). Missing a test for the *absent-attribute* case that the docstring explicitly promises (see inline). Integration tests using real model IDs are appropriate for @pytest.mark.slow.

**Minor**: The 
esolved_class.__name__ == 'BartDecoderWrapper' assertion in the integration test couples to an internal implementation detail. Low risk, but worth a comment explaining the expected class.

timenick:
Thanks for the review! Addressed in `063b704c`:

**1. Missing absent-attribute test** ✅ — Added `test_upgrade_fill_mask_for_seq2seq_ignores_config_without_flag`: a config lacking `is_encoder_decoder` is not upgraded (the `getattr(..., False)` default), matching the docstring's promise.

**2. `BartDecoderWrapper` assertion coupling** ✅ — Added a comment explaining `BartDecoderWrapper` is WinML's encoder-decoder decoder sub-component for `(bart, text2text-generation)` — the seq2seq route, not the `AutoModelForMaskedLM` the fill-mask mislabel would produce. (Kept the class-name assertion, since proving the class *changed away from* `AutoModelForMaskedLM` is the point of the test.)

**3. Wrapped-library early-return path (~line 520)** — I'd keep the correction scoped *out* of that path, and documented why in-code. That path derives its task from the wrapped library's ONNX export list (`get_supported_tasks`), not class→task inference — so optimum's `fill-mask` mislabel (which originates in `infer_task_from_model`) can't arise there. Applying it would be a no-op today (the only wrapped type, timm, is `is_encoder_decoder=False` / `image-classification`) and could, in a hypothetical future, rewrite to a `text2text-generation` that isn't in that library's supported list — breaking the path's invariant that the task is always export-supported. Happy to add the call if you'd prefer uniformity, but the scoping is intentional.

Generated with Claude Code

DingmaomaoBJTU (approved):
All three previous review findings have been addressed:

1. **Warning resolved** — The wrapped-library early-return path now has a clear explanatory comment (lines 501-505) documenting *why* _upgrade_fill_mask_for_seq2seq is intentionally omitted there: the task comes from get_supported_tasks (the library's export list), not from class->task inference, so the Optimum mislabel cannot arise on that path. This is a better fix than blindly adding the call — it's a deliberate scoping decision, now documented.

2. **Info resolved** — 	est_upgrade_fill_mask_for_seq2seq_ignores_config_without_flag has been added, using a minimal _NoFlagConfig class with no is_encoder_decoder attribute. This directly tests the getattr(..., False) default behavior promised by the docstring.

3. **Info resolved** — The 
esolved_class.__name__ == 'BartDecoderWrapper' assertion now has an inline comment explaining that BartDecoderWrapper is the WinML encoder-decoder decoder sub-component class registered for (bart, text2text-generation), distinguishing it from the AutoModelForMaskedLM that the fill-mask mislabel would have produced.

No new issues found. LGTM. ✅

PR #856, opened 2026-06-10
fix: use resolved path and cache_key in WinMLAutoModel.from_onnx

### LINKED ISSUE #827: bug: WinMLAutoModel.from_onnx uses incorrect cache dir/file name
After https://github.com/microsoft/winml-cli/pull/659, verified by `--no-skip-build`

- WinMLAutoModel uses `output_dir = get_model_dir(onnx_path.stem, cache_dir=cache_dir_path)` instead of hash of model file / path
- build_onnx_model uses `stem` instead of hash of config `winml_build_config`

chinazhangchao:
Fixes #827

## Problem

Two cache-naming bugs in `WinMLAutoModel.from_onnx` introduced after PR #659:

1. **Cache directory collision** – `from_onnx` called `get_model_dir(onnx_path.stem, ...)`, so two ONNX files with the same filename (e.g. `model.onnx`) from different directories would map to the same cache directory and overwrite each other's build outputs.

2. **Config change not reflected in cache** – `build_onnx_model` had no `cache_key` support (unlike `build_hf_model`), so different build configs for the same ONNX file would collide on the same `model.onnx` artifact path.

## Fix

**`src/winml/modelkit/models/auto.py` — `from_onnx`**
- Replace `get_model_dir(onnx_path.stem, ...)` with `get_model_dir(str(onnx_path.resolve()), ...)` so the cache dir is unique per absolute file path.
- Compute `cache_key = get_cache_key(task_abbrev, config.generate_cache_key())` (mirrors `from_pretrained`) and pass it to `build_onnx_model`.

**`src/winml/modelkit/build/onnx.py` — `build_onnx_model`**
- Add `cache_key: str | None = None` parameter.
- Add `_name()` helper (same pattern as `build_hf_model`) so all artifact paths are optionally prefixed — enabling multiple config variants to coexist in one directory.
- Rebuild cleanup uses the cache_key-scoped glob pattern to avoid removing unrelated artifacts.

## Tests

- 5 new tests in `test_auto_onnx.py`: resolved path used for model dir, same-stem/different-path collision prevention, cache_key passed through.
- 5 new tests in `test_onnx.py`: `cache_key=None` backward compat, prefixed final artifact, prefixed config path, reuse with prefixed path, unrelated artifact preservation on rebuild.


src/winml/modelkit/models/auto.py
```diff
@@ -199,10 +199,12 @@ def from_onnx(
...
             )
 
-        # Resolve output directory
+        # Resolve output directory and cache key
+        task_abbrev = get_task_abbrev(resolved_task) if resolved_task else "onnx"
+        cache_key = get_cache_key(task_abbrev, config.generate_cache_key())
         if use_cache:
             cache_dir_path = get_cache_dir(override=cache_dir)
-            output_dir = get_model_dir(onnx_path.stem, cache_dir=cache_dir_path)
+            output_dir = get_model_dir(str(onnx_path.resolve()), cache_dir=cache_dir_path)
```
xieofxie: should we use path or hash of onnx model? If use path, if user replace the model content, it will be wrong

src/winml/modelkit/build/onnx.py
```diff
@@ -101,15 +105,19 @@ def build_onnx_model(
...
     # Define output paths
-    stem = onnx_path.stem
-    optimized_path = output_dir / f"{stem}_optimized.onnx"
-    quantized_path = output_dir / f"{stem}_quantized.onnx"
-    compiled_path = output_dir / f"{stem}_compiled.onnx"
-    final_path = output_dir / "model.onnx"
-    config_path = output_dir / "winml_build_config.json"
-    manifest_path = output_dir / "build_manifest.json"
-    analyze_result_path = output_dir / "analyze_result.json"
+    optimized_path = output_dir / _name("optimized.onnx")
```
xieofxie, line 114: is the naming convention be same as perf a hf id ?

src/winml/modelkit/models/auto.py
chinazhangchao: Good point. I accepted this and changed the ONNX cache model directory to use a content hash of the ONNX file, including referenced external data, so replacing a model at the same path no longer reuses stale artifacts.

src/winml/modelkit/build/onnx.py
chinazhangchao: Yes, this is intentional: the ONNX build artifacts now use the same cache-key-prefixed convention as the HF path, for example `{cache_key}_optimized.onnx` and `{cache_key}_model.onnx`. I kept the unprefixed names only for direct build_onnx_model calls without a cache_key to preserve backward compatibility.

tests/unit/models/auto/test_auto_onnx.py
```diff
@@ -235,10 +236,127 @@
...
+
+        onnx_path.write_bytes(b"second-content")
+        model_dir_b = get_model_dir(f"onnx-{get_onnx_model_hash(onnx_path)}", cache_dir=cache)
+
+        assert model_dir_a != model_dir_b
+
+    def test_onnx_model_hash_includes_external_data(self, tmp_path: Path):
+        """Changing external data changes the ONNX model content hash."""
+        import numpy as np
+        import onnx
```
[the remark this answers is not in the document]
chinazhangchao: Accepted. I removed the mixed `import onnx` and `from onnx import ...` usage in this test and now access helper and TensorProto through the `onnx` module, keeping behavior unchanged while satisfying the CodeQL import-style check.

src/winml/modelkit/onnx/external_data.py
```diff
@@ -154,6 +155,37 @@ def has_external_data(model_path: str | Path) -> bool:
...
+    """Stream *path* into an existing hash object."""
+    with path.open("rb") as f:
+        for chunk in iter(lambda: f.read(1024 * 1024), b""):
+            hash_obj.update(chunk)
+
+
+def get_onnx_model_hash(model_path: str | Path) -> str:
+    """Compute a content hash for an ONNX model and referenced external data."""
+    model_path = Path(model_path).resolve()
+    hash_obj = hashlib.sha256()
```
xieofxie, line 172: how about use hash path + size + mtime? weight could be big

src/winml/modelkit/onnx/external_data.py
```diff
@@ -154,6 +155,41 @@ def has_external_data(model_path: str | Path) -> bool:
     return len(get_external_data_files(model_path)) > 0
 
 
+def _update_hash_from_path_metadata(hash_obj: Any, path: Path) -> None:
+    """Update *hash_obj* with a file's resolved path, size, and mtime."""
+    path = path.resolve()
+    stat = path.stat()
```
xieofxie, line 161: **Possible crash: `get_onnx_model_hash` doesn't guard against missing external-data files**

`get_external_data_files()` (called at L176) returns the `location` strings declared in the model proto regardless of whether those files exist on disk. The per-file loop then calls `_update_hash_from_path_metadata()` → `path.stat()` here with no existence check, so if a sidecar isn't resolvable at the referenced path (model copied/moved without its `.data`, embedded absolute path, mismatched relative name), this raises an uncaught `FileNotFoundError`.

The `try/except` at L175–179 only covers `get_external_data_files()`, not this stat loop — so the exception propagates out of `WinMLAutoModel.from_onnx` (default `use_cache=True`) as a crash during cache-dir resolution. That's a new failure mode vs. the previous `onnx_path.stem` code, which never touched the sidecars. External-data (>2GB) models are an explicit target of this tool, so it's worth handling.

`try_load_external_initializer_array` in this same file already guards exactly this with `if not data_path.exists(): return None`. Suggest the same here: skip missing sidecars (or fold a sentinel into the hash) instead of stat-ing blindly.

xieofxie (approved):

PR #861, opened 2026-06-10
feat: add --memory flag to measure RAM and VRAM per phase

DingmaomaoBJTU:
## Summary

Add `--memory` / `--no-memory` CLI flag to `winml perf` that measures process RAM (RSS) and VRAM (GPU/NPU dedicated + shared) at phase boundaries: baseline → compile → inference.

## Output

```
Memory:
  RAM:  670.7 MB -> model load: +77.3 MB | inference: +1.4 MB | total: +78.7 MB
  VRAM: 17.2/17.6 MB (local/shared) -> model load: +17.2/+17.6 MB | inference: +0.0/+0.0 MB | total: +17.2/+17.6 MB
```

## Design

- **RAM**: psutil RSS snapshots at phase boundaries (baseline after `_load_model()`, after `compile()`, after benchmark loop)
- **VRAM**: PDH `\GPU Process Memory` counters queried at same phase boundaries, reported as local/shared separately
- **Default**: enabled (`--no-memory` to disable). Snapshots are taken between phases and add zero overhead to latency measurements
- **VRAM line only shown** when device memory > 0 (hidden for CPU-only runs)

## Display changes

- Hardware section: `Mem` → `RAM`, `Device Mem` → `VRAM`
- Memory section: single line per metric with `->` showing deltas

## Files changed

- `src/winml/modelkit/commands/perf.py` — memory integration in `_run_single()`, console display, `_resolve_adapter_luid()`
- `src/winml/modelkit/session/monitor/memory_tracker.py` — `get_rss_mb()` + `get_vram_mb()` (~50 lines)
- `src/winml/modelkit/commands/_live_chart.py` — rename labels
- `src/winml/modelkit/session/monitor/live_display.py` — rename labels
- `tests/unit/session/monitor/test_memory_tracker.py` — unit tests


xieofxie:
I think we could just enabled it, it doesn't take much time

xieofxie:
Also consider adding this to --monitor

src/winml/modelkit/commands/perf.py
```diff
@@ -309,14 +327,31 @@ def run(self) -> BenchmarkResult:
...
             "Running benchmark: %d iterations + %d warmup",
             self.config.iterations,
             self.config.warmup,
         )
         stats = self._run_benchmark()
 
+        if self.config.memory:
+            rss_after_inference = get_rss_mb()
+            adapter_luid = self._resolve_adapter_luid()
+            device_mb = get_device_memory_mb(adapter_luid)
```
xieofxie: why only get it in the end?

DingmaomaoBJTU: Device memory only grows during a session (allocations are never freed until the session is destroyed), so querying once at the end captures the peak. Added a comment in code to clarify.

DingmaomaoBJTU: Removed device memory from this PR — the WDDM PDH counters only work reliably for GPU (DML). For NPU/QNN they report tiny numbers that don't reflect actual device usage. We now only measure process RSS at phase boundaries (baseline → compile → inference).

src/winml/modelkit/commands/perf.py
```diff
@@ -852,7 +886,6 @@ def display_console_report(result: BenchmarkResult, console: Console) -> None:
         console.print("[bold]Hardware (during benchmark)[/bold]")
         cpu = result.hw_monitor.get("cpu", {})
         ram = result.hw_monitor.get("ram", {})
-        dev_mem = result.hw_monitor.get("device_memory", {})
```
xieofxie, line 954: should we delete this? it is set by --monitor

DingmaomaoBJTU: This line is now gone — we no longer read device_memory from hw_monitor in the console summary. Instead, VRAM is measured independently at phase boundaries (via PDH point-in-time queries) and reported as local/shared in the Memory section. The hw_monitor's to_dict() still emits device_memory for the JSON output.

xieofxie (approved):

PR #862, opened 2026-06-10
fix(config): build full encoder-decoder composite for no-task seq2seq (#850)

timenick:
## What

`winml config`/`build` for an encoder-decoder model **without `--task`** exported a decoder-only graph: it took `encoder_hidden_states` as an input with no producer, so the artifact could not run standalone (#850). The full encoder+decoder export only fired for an explicit `--task`.

## Change

The no-`--task` path now detects the task with the **same `detect_task` that `inspect` uses**, then expands to the full encoder+decoder composite only when the detected task is one a seq2seq composite serves (`text2text-generation` / `image-to-text`).

- Seq2seq generators (t5, marian, mu2, blip, vision-encoder-decoder) → encoder + decoder composite.
- A non-generation checkpoint of a seq2seq-capable architecture stays a single model, **consistent with `inspect`**:
  - `facebook/bart-large-mnli` (`BartForSequenceClassification`) → `text-classification`
  - a T5 encoder → `feature-extraction`
- Gated on the `WinMLEncoderDecoderModel` base class — not the raw `is_encoder_decoder` flag, which BLIP reports `False`. Decoder-only (qwen3) and dual-encoder (clip/siglip) composites are excluded from auto-expansion; the explicit-`--task` path serves every composite kind, unchanged.

## Relationship to #851

`bart` currently auto-detects `fill-mask` (a `TasksManager` mislabel), so it stays single until #851 flips the detection to `text2text-generation`, at which point it expands automatically — no further change needed here. The two PRs compose; this one is based on `main`.

## Result (no `--task`)

| model | before | after |
|---|---|---|
| t5 / marian / mu2 / blip / vision-encoder-decoder | decoder-only (non-runnable) | encoder + decoder composite |
| bart-large-mnli | decoder-only / wrong | text-classification (single) |
| bart-large-cnn | decoder-only | single until #851, then composite |

Addresses #850 (bart completes once #851 lands).

### DIFF
--- a/src/winml/modelkit/commands/config.py
+++ b/src/winml/modelkit/commands/config.py
@@ -497,11 +497,12 @@ def _resolve_composite_model_components(
     *component* task (``text2text-generation``) and export the decoder alone -- a
     non-runnable half whose ``encoder_hidden_states`` input has no producer. So the
     no-task path detects the task with the *same* ``detect_task`` that ``inspect``
-    uses, then expands to the encoder-decoder composite only when the detected task
-    is one a seq2seq composite serves. A non-generation checkpoint of a
-    seq2seq-capable architecture (a sequence-classification BART ->
-    ``text-classification``, a T5 encoder -> ``feature-extraction``) therefore stays
-    a single model, consistent with ``inspect``.
+    uses, then expands to the registered composite when the detected task is one that
+    composite serves -- so no-task routing matches explicit ``--task`` routing for the
+    same model (e.g. qwen3 -> its decoder prefill+gen composite). A checkpoint whose
+    detected task is not a composite task (a sequence-classification BART ->
+    ``text-classification``, a T5 encoder -> ``feature-extraction``) stays a single
+    model, consistent with ``inspect``.
     """
     import winml.modelkit.models.hf  # noqa: F401  # trigger pipeline registrations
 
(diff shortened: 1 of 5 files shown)


src/winml/modelkit/commands/config.py
```diff
@@ -485,31 +485,114 @@ def _resolve_composite_model_components(
...
+    from transformers import AutoConfig
+
+    try:
+        if hf_model is not None:
+            config = AutoConfig.from_pretrained(hf_model, trust_remote_code=trust_remote_code)
+        elif model_type is not None:
+            config = AutoConfig.for_model(model_type)
+        else:
+            return None
+    except Exception:
```
vortex-captain: would `Exception` be too broad, hiding other potential root causes?

src/winml/modelkit/commands/config.py
```diff
@@ -485,31 +485,114 @@ def _resolve_composite_model_components(
...
+    from ..models.winml.composite_model import COMPOSITE_MODEL_REGISTRY
+
+    # The export-side generation task every encoder-decoder composite's decoder
+    # produces; detect_task yields this for a seq2seq generator. Universal HF/Optimum
+    # task taxonomy, not a model name.
+    seq2seq_generation_task = "text2text-generation"
+
+    distinct: dict[tuple, type[WinMLEncoderDecoderModel]] = {}
+    for (m_type, reg_task), cls in COMPOSITE_MODEL_REGISTRY.items():
+        if m_type != model_type or not issubclass(cls, WinMLEncoderDecoderModel):
```
vortex-captain: qwen is `WinMLDecoderOnlyModel`, and also has text2text-generation component:
```python
    _SUB_MODEL_CONFIG: ClassVar[dict[str, str]] = {
        "decoder_prefill": "feature-extraction",
        "decoder_gen": "text2text-generation",
    }
```
Shall we check `WinMLDecoderOnlyModel` as well? Alternatively, we could consider checking for all `WinMLCompositeModel` subclasses

src/winml/modelkit/commands/config.py
timenick: Ended up removing the `try/except` entirely rather than narrowing it. The probe and the main build path (`generate_hf_build_config`) load the model config the same way, so the probe can't fail where the build would succeed — the catch gave no real-usage robustness and only swallowed load errors for tests that pass placeholder model ids. Now the probe just loads the config (a bad id errors clearly, same as before, via the command's outer handler), and the few tests that used fake ids mock `AutoConfig` at the boundary, like they already mock the loader.

Generated with Claude Code

src/winml/modelkit/commands/config.py
timenick: Good call — widened the gate from `WinMLEncoderDecoderModel` to `WinMLCompositeModel`, so any registered composite participates (decoder-only and dual-encoder included). Motivation: explicit `--task text-generation` on qwen3 already routes to its prefill+gen composite, so no-task should match — now it does. Task-mismatched checkpoints still stay single (a no-task CLIP detects `feature-extraction`, not its `zero-shot-image-classification` composite task, so it doesn't over-route). Renamed the helper to `_composite_components_for_task` since it's no longer encoder-decoder-specific, and added tests for qwen3 (expands) and CLIP (stays single).

Generated with Claude Code

vortex-captain (approved):

PR #866, opened 2026-06-10
fix(perf): support composite (dual-encoder) models in winml perf

xieofxie:
## Problem

`winml perf` crashed on composite (dual-encoder) models such as SigLIP/CLIP:

```
winml perf --ep openvino --device cpu -m google/siglip-base-patch16-224 \
  --task zero-shot-image-classification
...
AttributeError: 'WinMLModelForZeroShotImageClassification' object has no attribute 'io_config'
```

`PerfBenchmark` assumed every model exposes a single `io_config` / `_session`. Composite
models have neither — they orchestrate multiple sub-models (e.g. an image encoder and a
text encoder), each with its own ONNX session. The failure is device-independent: the
`(model_type, task)` registry routes SigLIP to the composite class regardless of `--device`.

## Fix

Benchmark each sub-model **individually** — the same way `--module` benchmarks each
submodule — instead of timing the aggregate `forward()` pass:

- **`run()`** returns a single `BenchmarkResult` for single-session models, or a
  `{sub_model_name: BenchmarkResult}` mapping for composites.
- **`_run_sub_models()`** iterates `sub_models` and benchmarks each through the standard
  single-session pipeline by spawning a child `PerfBenchmark` with the already-loaded
  sub-model. Every sub-model reuses the existing compile / input-generation / simple +
  monitored loops / result-collection code, so per-sub-model measurement semantics match
  single-model runs exactly (pure-ORT time inside `session.perf()`).
- Each sub-model reports **its own** I/O, task, device, and EP — no aggregation.
- **`report_composite_results()`** renders a per-sub-model summary table and writes one
  combined JSON nesting each sub-model's full `BenchmarkResult.to_dict()` under
  `components` (`model_id`, `task`, `component_count`, `components`).

This removes the previous aggregate machinery (`_aggregate_io_config`,
`_describe_outputs`, `_probe_composite_outputs`, `_composite_run_iteration`) and the
composite branches inside the benchmark loops.

`_is_composite` stays duck-typed on `sub_models` (rather than
`isinstance(..., WinMLCompositeModel)`) on purpose: an `isinstance` check needs a runtime
import of `composite_model`, which imports torch and would blow the `winml perf --help`
import budget (`tests/cli/test_import_time.py`). Keeping `WinMLCompositeModel` a
TYPE_CHECKING-only import also lets the unit tests use lightweight duck-typed fakes.

## Result

```
Sub-model: image-encoder

Device:      cpu / OpenVINOExecutionProvider
Task:        image-feature-extraction
Inputs:      pixel_values         [1, 3, 224, 224]       float32
Outputs:     last_hidden_state    [1, 196, 768]
             pooler_output        [1, 768]


Sub-model: text-encoder

Device:      cpu / OpenVINOExecutionProvider
Task:        feature-extraction
Inputs:      input_ids            [1, 64]                int32
Outputs:     last_hidden_state    [1, 64, 768]
             pooler_output        [1, 768]


                                                    Per-Sub-Model Perf
┏━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━┳━━━━━━━━━━┳━━━━━━━━━━┳━━━━━━━━━━┓
┃ Sub-Model     ┃ Task                     ┃ Device                          ┃ Mean (ms) ┃ P90 (ms) ┃ Min (ms) ┃ Max (ms) ┃
┡━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━╇━━━━━━━━━━╇━━━━━━━━━━╇━━━━━━━━━━┩
│ image-encoder │ image-feature-extraction │ cpu / OpenVINOExecutionProvider │     81.14 │   104.97 │    63.49 │   124.89 │
│ text-encoder  │ feature-extraction       │ cpu / OpenVINOExecutionProvider │     39.19 │    51.89 │    24.99 │   109.08 │
└───────────────┴──────────────────────────┴─────────────────────────────────┴───────────┴──────────┴──────────┴──────────┘
```

## Tests

`tests/unit/commands/test_perf_composite.py` is rewritten for the per-sub-model behavior:
each sub-model produces its own `BenchmarkResult` with its own I/O / task / device / EP,
every sub-session is compiled and run `warmup + iterations` times, and
`report_composite_results` emits the combined per-component JSON (and JSON-mode stdout).
Existing `test_perf_cli.py` and `test_perf_optracing_cli.py` still pass — no regression.


DingmaomaoBJTU (commented):
Overall this PR is well-structured and the approach of duck-typing on `sub_models` (to avoid torch import at load time) is a reasonable trade-off with a clear explanation. A few issues worth addressing before merge.

src/winml/modelkit/commands/perf.py
```diff
@@ -290,34 +291,101 @@ class PerfBenchmark:
...
+    def _run_sub_models(self) -> dict[str, BenchmarkResult]:
+        """Benchmark each sub-model of a composite individually.
+
+        Each sub-model is itself a single-session ``WinMLAutoModel``, so it is
+        benchmarked through the standard single-model pipeline by spawning a
+        child ``PerfBenchmark`` with the already-loaded sub-model. Results are
+        keyed by sub-model name for per-component reporting.
+        """
+        results: dict[str, BenchmarkResult] = {}
+        for name, sub in self._sub_models().items():
```
DingmaomaoBJTU: **No per-sub-model error isolation.** If `child._run_single()` raises on the second sub-model, the first sub-model's already-computed result is silently discarded and the caller gets an undecorated exception with no indication of which sub-model failed. Consider wrapping each iteration in a `try/except` that logs the failing sub-model name before re-raising.

src/winml/modelkit/commands/perf.py
```diff
@@ -290,34 +291,101 @@ class PerfBenchmark:
...
+        on purpose: an ``isinstance`` check needs a runtime import of
+        ``composite_model``, which imports torch and would blow the
+        ``winml perf --help`` import budget (see tests/cli/test_import_time.py) by
+        pulling torch in at module load. Keeping ``WinMLCompositeModel`` a
+        TYPE_CHECKING-only import also lets the unit tests use lightweight
+        duck-typed fakes instead of constructing real torch-backed composites.
+        ``sub_models`` is the defining member of the composite base, so it is a
+        reliable marker.
+        """
+        return hasattr(self._model, "sub_models")
```
DingmaomaoBJTU: **Duck-type check is fragile.** `hasattr(self._model, 'sub_models')` will silently misidentify any future single-session model that happens to carry a `sub_models` attribute for another purpose. The torch-import avoidance rationale is well documented, but tightening to `hasattr(..., 'sub_models') and isinstance(self._model.sub_models, dict)` would narrow the false-positive surface without any import cost.

src/winml/modelkit/commands/perf.py
```diff
@@ -290,34 +291,101 @@ class PerfBenchmark:
...
+
+    def _sub_models(self) -> dict[str, WinMLPreTrainedModel]:
```
DingmaomaoBJTU, line 315: **Inconsistent accessor style.** `_sub_models()` is a regular method while `_is_composite` and `_single` are `@property` accessors. Callers must remember to add `()` for one but not the others. Promote this to a property to match the other two.

src/winml/modelkit/commands/perf.py
```diff
@@ -1323,6 +1460,21 @@ def perf(
...
+            report_composite_results(
+                result,
+                console=console,
+                json_mode=json_mode,
+                output_path=output,
+                model_id=hf_model,
+                task=task,
+            )
+            console.print(f"[green]Results saved to:[/green] {output}")
+            return
```
DingmaomaoBJTU, line 1490: **`--op-tracing` is silently ignored for composite models.** This early return means any `--op-tracing` flag the user passed is swallowed with no feedback. Print a yellow warning before returning if `op_tracing` is truthy: `console.print('[yellow]Warning:[/yellow] --op-tracing is not supported for composite models and will be skipped.')`

tests/unit/commands/test_perf_composite.py
```diff
@@ -0,0 +1,264 @@
...
+
+class TestPerfBenchmarkComposite:
+    """PerfBenchmark benchmarks each sub-model of a composite individually."""
+
+    def test_detects_composite(self) -> None:
+        bench, _ = _composite_benchmark()
+        assert bench._is_composite is True
+
+    def test_run_returns_result_per_sub_model(self) -> None:
+        bench, _ = _composite_benchmark()
```
DingmaomaoBJTU, line 179: **Missing edge-case: empty `sub_models` dict.** When a composite model has zero sub-models `_run_sub_models()` returns `{}`, the `isinstance(result, dict)` branch fires, and `report_composite_results` writes a `component_count: 0` JSON file with no sub-model rows. A short test asserting this doesn't crash and produces an empty result dict would guard the boundary.

src/winml/modelkit/commands/perf.py
```diff
@@ -33,7 +34,10 @@
 
 
 if TYPE_CHECKING:
+    from collections.abc import Callable, Iterable
+
     from ..models.winml.base import WinMLPreTrainedModel
+    from ..models.winml.composite_model import WinMLCompositeModel
```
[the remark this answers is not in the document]
xieofxie: Resolved in 629ca816c. The import is needed so the type checker can resolve the `cast("WinMLCompositeModel", ...)` forward reference, but CodeQL only saw it inside a string. `_sub_models` now references the name via a bare local annotation (`composite: WinMLCompositeModel = cast(...)`), which is a real `Name` reference for CodeQL yet never evaluated at runtime (local annotations aren't), so the TYPE_CHECKING-only import stays and the import-time budget is unaffected.

src/winml/modelkit/commands/perf.py
xieofxie: Resolved in 629ca816c. `_run_sub_models` now wraps each `child._run_single()` in `try/except Exception` that logs `"Benchmarking sub-model '%s' failed"` with the sub-model name before re-raising, so a failure points at the offending sub-model.

src/winml/modelkit/commands/perf.py
xieofxie: Resolved in 629ca816c. `_is_composite` is now `hasattr(self._model, "sub_models") and isinstance(getattr(self._model, "sub_models", None), dict)`, narrowing the false-positive surface at no import cost. Docstring updated to note the guard.

src/winml/modelkit/commands/perf.py
xieofxie: Resolved in 629ca816c. `_sub_models` is now an `@property`, matching `_is_composite` and `_single`; the sole caller was updated to drop the `()`.

src/winml/modelkit/commands/perf.py
xieofxie: Resolved in 629ca816c. Before the composite early return, if `op_tracing` is truthy we now print `[yellow]Warning:[/yellow] --op-tracing is not supported for composite models and will be skipped.`

tests/unit/commands/test_perf_composite.py
xieofxie: Resolved in 629ca816c. Added `test_empty_sub_models_returns_empty_dict` (asserts `_run_sub_models()` returns `{}` without crashing) and `test_empty_components_writes_zero_count` (asserts `report_composite_results({}, ...)` writes `component_count: 0` with empty `components`).

src/winml/modelkit/commands/perf.py
xieofxie: Updated in 5d23bfebd to use the concrete type as you suggested: `_is_composite` is now `isinstance(self._model, WinMLCompositeModel)` with a **function-local** import of `composite_model`. The import runs only when the property is read (i.e. after a model is loaded, when torch is already in memory), never at module load, so the `winml perf --help` import budget (tests/cli/test_import_time.py — 68 passed) is unaffected. `_sub_models` narrows the same way via `assert isinstance(...)`, and the test fake now subclasses `WinMLCompositeModel` so the check matches.

src/winml/modelkit/commands/perf.py
```diff
@@ -290,34 +291,110 @@ class PerfBenchmark:
...
+            logger.info("Benchmarking sub-model '%s'", name)
+            Console(stderr=True).print(f"\n[bold]Sub-model:[/bold] {name}")
+            child = PerfBenchmark(self.config)
+            child._model = sub
+            try:
+                results[name] = child._run_single()
+            except Exception:
+                logger.error("Benchmarking sub-model '%s' failed", name)
+                raise
+        return results
```
DingmaomaoBJTU, line 374: **Finding 1 — partially addressed.** `logger.error` now names the failing sub-model, which is a good improvement. However, the bare `raise` re-propagates the original exception unchanged, so the message the *user* sees at the CLI carries no sub-model context unless they inspect logs. Consider `raise RuntimeError(f"Sub-model '{name}' failed: {exc}") from exc` (capture `except Exception as exc`) so the error is self-documenting without needing a log viewer.

DingmaomaoBJTU (commented):
Follow-up review: 4 of 5 previous findings are fully resolved. One remaining concern on error isolation (see inline comments), and one new minor issue introduced via the merge.

DingmaomaoBJTU (commented):
**New info-level issue (not inlineable — modified line from merge context):** perf.py:719 — _resolution variable is named with an underscore prefix, signalling ''intentionally unused'', yet occupies a named binding. Use plain _ as the fourth discard: parent_loader_cfg, _, _, _ = resolve_loader_config(...). If the TaskResolution value is needed later, rename it at that point without the leading underscore.

src/winml/modelkit/commands/perf.py
xieofxie: Resolved in caed1e492. Changed to capture the exception and re-raise with context: `except Exception as exc: ... raise RuntimeError(f"Sub-model '{name}' failed: {exc}") from exc`. The CLI-visible error now names the failing sub-model directly, and `from exc` preserves the original traceback/cause. Added `test_sub_model_failure_names_component` asserting the `RuntimeError` matches the component name and that `__cause__` is the original exception.

DingmaomaoBJTU (approved):
4/5 previous findings are fully resolved. The two remaining items are non-blocking:

- **perf.py:374** (warning, partial): \logger.error\ now names the failing sub-model, which is good. The bare \
aise\ still surfaces the original exception without sub-model context at the CLI, but this is a UX improvement rather than a correctness issue — the log output is sufficient for diagnosis.
- **perf.py:719** (info): \_resolution\ could be plain \_\ since it's unused, but this is cosmetic.

Core logic is correct: sub-models are benchmarked independently, results are keyed per component, the duck-type check is properly tightened, the \--op-tracing\ gap is closed, and the test coverage is solid. Approving.

src/winml/modelkit/commands/perf.py
xieofxie: applied

PR #871, opened 2026-06-11
compile: multi-model shared EP context with selectable backend

vortex-captain:
## Summary

Enables compiling **multiple ONNX models into a single shared EP context** (weight sharing) from `winml compile`, and adds a **selectable compile backend** via a new `--compiler` choice.

## New compilation enablements
- **Multiple models, shared EP context** — `winml compile -m A.onnx -m B.onnx ...` compiles each model sharing one weights binary (`ep.share_ep_contexts` / `ep.stop_share_ep_contexts`); every compiled model references the single shared `.bin`.
- **Selectable backend via `--compiler`**:
  - `ort` (default) — `ort.ModelCompiler`
  - `ort_session` — `ort.InferenceSession` (`ep.context_enable`), which produces loadable EPContext models for graphs the ModelCompiler path over-fragments
  - `qairt` — unchanged

  (`winml compile --list` shows the available compilers for the selected EP.)
- **EP provider options via `--config`** — the `compile.provider_options` block (e.g. QNN `htp_arch` / `soc_model` / `vtcm_mb`) is forwarded to the compile session.
- **Per-model reporting** — every model's result is logged (not just the first failure); a missing artifact on an otherwise-successful compile is reported as a warning, not an error.

## CLI / behavior
- `-m/--model` is now repeatable.
- `--compiler` gains the `ort_session` choice (this replaces the earlier `--use-inference-session` flag).
- Output:
  - **single model**: `-o`/`--output` (a file) or `--output-dir` (a directory).
  - **multiple models**: `--output-dir` is **required** (a single `-o`/`--output` file is rejected), since outputs are written by filename into the directory; same-named inputs are de-duplicated with an integer suffix (`model_ctx.onnx`, then `model_1_ctx.onnx`) and a warning.
- The single-model default path is unchanged.

## Implementation
- `compile_multiple_onnx(model_paths, output_path, config)` drives the per-model loop. `output_path` may be a file (single model only) or a directory; it asserts a directory when compiling multiple models.
- The backend is taken from `config.ep_config.compiler` and surfaced via the `CompileContext.use_inference_session` property (`compiler == "ort_session"`).
- `Compiler` carries `n_total_models` / `n_compiled_models` and the reused shared `SessionOptions` (`shared_session_options`).
- `CompileStage` selects the backend, reuses the shared options, and toggles the share/stop-share session entries; the default single-model `ModelCompiler` path is untouched.

## Tests
- **e2e**: shared-weight multi-model test parametrized over both backends. The `ort_session` output is loaded, run, and checked against a CPU reference with `np.allclose`; the `ort` (ModelCompiler) output is a file-level smoke check.
- **unit**: output rules (file vs directory; multiple models require a directory), same-name de-duplication, backend dispatch, and `--list` including the new compiler.

### DIFF
--- a/src/winml/modelkit/commands/compile.py
+++ b/src/winml/modelkit/commands/compile.py
@@ -232,6 +232,9 @@ def compile(
 
     config.validate = validate
     config.verbose = bool(verbose)
+    # CLI flag (merged with any config-file value above) overrides the config; the
+    # actual compilation reads use_inference_session from the config.
+    config.use_inference_session = use_inference_session
 
     # Set compiler options
     config.ep_config.compiler = compiler
@@ -272,9 +275,7 @@ def compile(
             # Multi-model (shared EP context) and/or inference-session backend.
             # compile_multiple_onnx writes each model as <stem>_ctx.onnx into a folder;
             # multiple models require --output-dir (enforced above).
-            results = compile_multiple_onnx(
-                models, output_dir, config, use_inference_session=use_inference_session
-            )
+            results = compile_multiple_onnx(models, output_dir, config)
 
         # Report every model's result (not just the first failure).
         multi = len(results) > 1
--- a/src/winml/modelkit/commands/compile.py
+++ b/src/winml/modelkit/commands/compile.py
@@ -45,8 +45,10 @@ console = Console()
     "--model",
     "-m",
     required=False,
+    multiple=True,
     type=click.Path(exists=True, path_type=Path),
-    help="Input ONNX model file (required unless --list)",
+    help="Input ONNX model file. Repeat -m to compile multiple models with a shared "
+    "EP context (weight sharing). Required unless --list.",
 )
 @cli_utils.output_option("Output file path (e.g., model_compiled.onnx)")
 @click.option(
@@ -91,6 +93,13 @@ console = Console()
     help="Embed EP context in ONNX file (default: external .bin file)",
 )
 @click.option(
+    "--use-inference-session",
+    is_flag=True,
+    default=False,
+    help="Compile via ort.InferenceSession (ep.context_enable) instead of the default "
+    "ort.ModelCompiler backend.",
+)
+@click.option(
     "--list",
     "list_compilers_flag",
     is_flag=True,
@@ -102,7 +111,7 @@ console = Console()
 @click.pass_context
 def compile(
     ctx: click.Context,
-    model: Path | None,
+    model: tuple[Path, ...],
     output: Path | None,
     output_dir: Path | None,
     device: str,
@@ -113,6 +122,7 @@ def compile(
     compiler: str,
     qnn_sdk_root: Path | None,
     embed: bool,
+    use_inference_session: bool,
     list_compilers_flag: bool,
     config_file: Path | None,
 ) -> None:
@@ -140,15 +150,24 @@ def compile(
 
     # Apply build config defaults (CLI explicit options take precedence).
     # Read raw JSON so missing keys are distinguishable from dataclass defaults.
+    config_provider_options: dict[str, str] = {}
     if config_file is not None:
         _, raw_cfg = cli_utils.load_build_config(config_file)
         cc = raw_cfg.get("compile") or {}
+        # EP provider options (e.g. QNN htp_arch/soc_model/vtcm_mb) for the compile session.
+        if "provider_options" in cc:
+            config_provider_options = dict(cc["provider_options"])
         if not cli_utils.is_cli_provided(ctx, "ep") and "execution_provider" in cc:
             ep = cc["execution_provider"]
         if not cli_utils.is_cli_provided(ctx, "compiler") and "compiler" in cc:
             compiler = cc["compiler"]
         if not cli_utils.is_cli_provided(ctx, "embed") and "embed_context" in cc:
             embed = cc["embed_context"]
+        if (
+            not cli_utils.is_cli_provided(ctx, "use_inference_session")
+            and "use_inference_session" in cc
+        ):
+            use_inference_session = cc["use_inference_session"]
         if not cli_utils.is_cli_provided(ctx, "validate") and "validate" in cc:
             validate = cc["validate"]
         # Config-file verbosity fallback. CLI flags always win: only honor the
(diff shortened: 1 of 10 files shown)


src/winml/modelkit/commands/compile.py
```diff
@@ -90,6 +92,13 @@
     show_default=True,
     help="Embed EP context in ONNX file (default: external .bin file)",
 )
+@click.option(
```
xieofxie: we have `--compiler`, please use that?

src/winml/modelkit/commands/compile.py
```diff
@@ -90,6 +92,13 @@
...
+    "--use-inference-session",
+    is_flag=True,
```
timenick: Convention nit: `--use-inference-session` is added as `is_flag=True`, but the same command already uses `--embed/--no-embed` (line 90) and `--validate/--no-validate` (line 73) pairs. PR #844 converted boolean options across the CLI to the `--flag/--no-flag` form; new boolean options should follow that convention. Suggest `--use-inference-session/--no-use-inference-session`.

src/winml/modelkit/compiler/compiler.py
```diff
@@ -199,3 +238,55 @@ def compile_onnx(
     """
     compiler = Compiler()
     return compiler.compile(model_path=model_path, output_path=output_path, config=config)
+
+
+def compile_multiple_onnx(
```
timenick, line 240: `compile_multiple_onnx(model_paths, output_dir=None, ...)` accepts `output_dir=None` even when `model_paths` has more than one entry. The CLI guards against this in `commands/compile.py` (raises when `--output-dir` is omitted with multiple `-m`), but the public API doesn’t. Called programmatically with `output_dir=None` and multiple models, each model resolves its own `<stem>_ctx.onnx` next to the input, which silently breaks the “shared EP context = one `.bin` in one directory” invariant the docstring relies on. Suggest raising `ValueError` when `len(model_paths) > 1 and output_dir is None`, or documenting the constraint explicitly.

src/winml/modelkit/compiler/compiler.py
```diff
@@ -53,6 +59,28 @@ class Compiler:
...
+                same shared ``SessionOptions`` is reused across every ``compile``.
+            use_inference_session: Select the ``ort.InferenceSession``
+                (``ep.context_enable``) backend instead of the default
+                ``ort.ModelCompiler``.
+        """
+        self.n_total_models = n_total_models
+        self.use_inference_session = use_inference_session
+        # The shared SessionOptions: created by CompileStage on the first model and
+        # reused for the rest (kept here so it survives between compile() calls).
+        self.inference_session: object | None = None
```
timenick: `Compiler.inference_session` (here) and `CompileContext.inference_session` (`context.py:57`, annotated `ort.SessionOptions | None`) both hold an `ort.SessionOptions`, not an `ort.InferenceSession` — `stages/compile.py:639` confirms: `sess_options = context.inference_session`. The field name is going to mislead the next reader (and the `object | None` annotation here hides the real type). Suggest renaming both fields to e.g. `shared_session_options` (or `shared_sess_options`) and tightening the annotation.

src/winml/modelkit/commands/compile.py
vortex-captain: added new choice to `--compiler`: `ort_inference_session`, removed `use_inference_session`

src/winml/modelkit/commands/compile.py
vortex-captain: `--use-inference-session` has been removed

src/winml/modelkit/compiler/compiler.py
vortex-captain: fixed

src/winml/modelkit/compiler/compiler.py
vortex-captain: fixed

src/winml/modelkit/compiler/compiler.py
```diff
@@ -199,3 +235,74 @@ def compile_onnx(
...
+        One :class:`CompileResult` per input model, in order.
+    """
+    paths = [Path(mp) for mp in model_paths]
+    out = Path(output_path) if output_path is not None else None
+    # A path with a suffix (e.g. ".onnx") is a file; otherwise it's a directory.
+    out_is_file = out is not None and bool(out.suffix)
+
+    if len(paths) > 1:
+        out_is_dir = out is not None and not out_is_file
+        assert out_is_dir, (
```
timenick: Nit but worth flagging: using `assert` for input validation here is risky — `assert` statements are stripped under `python -O` / `PYTHONOPTIMIZE=1`, so this guard silently disappears in optimized builds and the multi-model `output_path` invariant the docstring relies on is no longer enforced. Suggest `raise ValueError("output_path must be a directory ...")` instead.

The new unit tests in `tests/unit/compiler/test_compile_multiple.py` (`test_multiple_models_require_directory`, `test_multiple_models_reject_none_output`) currently `pytest.raises(AssertionError, ...)` — they’ll need to follow the exception type along.

src/winml/modelkit/commands/compile.py
```diff
@@ -74,9 +76,10 @@
 )
 @click.option(
     "--compiler",
-    type=click.Choice(["ort", "qairt"]),
+    type=click.Choice(["ort", "ort_jit", "qairt"]),
```
xieofxie: or ort_session? ort_jit is a technique, not a object

src/winml/modelkit/commands/compile.py
xieofxie: if possible, change this to type

src/winml/modelkit/compiler/compiler.py
vortex-captain: fixed

timenick (approved):

src/winml/modelkit/commands/compile.py
vortex-captain: fixed

src/winml/modelkit/commands/compile.py
vortex-captain: renamed

xieofxie (approved):

PR #872, opened 2026-06-11
feat: precision-driven quantization (FP16, RTN int4, static QDQ) via --precision flag

### LINKED ISSUE #867: feat: add --enable-fp16-conversion to winml optimize and --precision to winml build/export
## Problem

There is no official `winml` CLI path to produce an FP16 ONNX model. This blocks the `autoconfig` skill from testing FP16 as an optimization hypothesis for QNN GPU and DML targets via a reproducible CLI path.

### Unified `--precision` flag on `winml build` and `winml export`

`--precision` already exists on `winml perf`, `winml eval`, `winml config` — `build` and `export` should be consistent:

```bash
winml build  -m facebook/convnext-tiny-224 --device gpu --precision fp16 -o out/
winml export -m facebook/convnext-tiny-224 -o model.onnx --precision fp16
```

`winml build --precision fp16` should choose the implementation path based on model source:
- **HuggingFace source** → export-time (`torch_dtype=float16` when loading PyTorch model) — cleanest, native FP16
- **Pre-exported ONNX** → optimize-stage post-export cast (see below)

### `--enable-fp16-conversion` on `winml optimize`

Consistent with the existing `--enable-X / --disable-X` pattern in `winml optimize`. Declarative, not implementation-specific:

```bash
# Full FP16 — inputs + weights + activations (QNN GPU / DML use case)
winml optimize -m model.onnx -o model_fp16.onnx --enable-fp16-conversion

# Keep model I/O as FP32 (CPU-safe fallback)
winml optimize -m model.onnx -o model_fp16.onnx --enable-fp16-conversion --fp16-keep-io-types

# Keep precision-sensitive ops (e.g. LayerNorm, Softmax) in FP32
winml optimize -m model.onnx -o model_fp16.onnx --enable-fp16-conversion --fp16-op-block-list LayerNorm,Softmax
```

Internally backed by `onnxruntime.transformers.float16.convert_float_to_float16` (ORT built-in, no new dependencies).

## Implementation priority

| Phase | Work | Effort |
|---|---|---|
| **P0** | `winml optimize --enable-fp16-conversion` (post-export cast via ORT transformers) | Low |
| **P1** | `winml build/export --precision fp16` wired to P0 for ONNX inputs | Low (after P0) |
| **P2** | `winml build --precision fp16` using export-time `torch_dtype=float16` for HF sources | Medium |

## Empirical motivation

Tested on ConvNext (facebook/convnext-tiny-224), QNN GPU (Adreno X1-85):

| Version | p50 | p90 | std |
|---|---|---|---|
| FP32 baseline | 17.7ms | 19.7ms | 1.0ms |
| FP16 (post-export cast) | **8.8ms** | 32ms | 9ms |

FP16 p50 is 2× faster, making it the primary optimization lever for Adreno GPU. Without this CLI path, `autoconfig` must mark the FP16 hypothesis as `SKIPPED — CLI gap`.

Note: high p90 variance (32ms) is a separate issue likely related to DVFS on Adreno GPU — tracked in #865 (`--ep-option htp_performance_mode`).

## Acceptance criteria (P0)
- `winml optimize -m model.onnx --enable-fp16-conversion` produces a valid FP16 ONNX
- `--fp16-keep-io-types` flag preserves FP32 I/O
- Model structure is preserved (only dtypes change); node count difference ≤ 2 (Cast nodes at I/O boundary)
- Operation is recorded in optimize output metadata for reproducibility

DingmaomaoBJTU:
## Summary

Adds a unified `--precision` flag to `winml quantize` and `winml build` that auto-selects the quantization algorithm based on the target precision. This replaces the need to manually configure `--weight-type`, `--activation-type`, and algorithm-specific flags.

Closes #867

## Precision → Algorithm Mapping

| `--precision` | Algorithm | Description |
|---|---|---|
| `fp16` | FP16 conversion | Weights + activations → FP16 (I/O stays FP32) |
| `int4` | RTN (weight-only) | 4-bit weight via MatMulNBits, activation stays FP32 |
| `int8` | Static QDQ | Calibrated QDQ (uint8 weight + uint8 activation) |
| `int16` / `w16a16` | Static QDQ | Calibrated QDQ (int16 weight + uint16 activation) |
| `w8a16` | Static QDQ | Calibrated QDQ (uint8 weight + uint16 activation) |
| `w8a8` | Static QDQ | Calibrated QDQ (uint8 weight + uint8 activation) |

## Architecture

This PR implements **single-pass** quantization only. Each `--precision` value maps to exactly one quantization operation:

- **FP16**: `_quantize_fp16()` — float16 conversion via onnxconverter-common
- **RTN**: `_quantize_rtn()` — weight-only int4 via MatMulNBits
- **QDQ**: `_quantize_qdq()` — calibrated static quantization via onnxruntime

Dispatch is handled via a `_mode_handlers` dict in `quantize_onnx()`. The command layer builds a `WinMLQuantizationConfig` with `mode` set to `"fp16"`, `"rtn"`, `"static"`, or `"dynamic"`, then calls `quantize_onnx(config=...)`.

### Key Design Decisions

- `WinMLQuantizationConfig.mode` is the unified field: `Literal["static", "dynamic", "rtn", "fp16"]`
- RTN and FP16 paths skip calibration entirely — warnings shown if calibration flags are provided
- QDQ precisions (int8, int16, w8a16, etc.) still require calibration data
- `from_dict()` provides backward compat: reads `"algorithm"` key and maps to `mode`

## Supported Commands

| Command | `--precision` support | Notes |
|---|---|---|
| `winml build` | ✅ | Full pipeline: export → optimize → quantize → compile |
| `winml quantize` | ✅ | Standalone quantization on existing ONNX |
| `winml config` | ✅ | Config generation respects precision |

## E2E Test Results (convnext-tiny-224)

| Command | Result | Notes |
|---|---|---|
| `winml quantize --precision fp16` | ✅ | 109→54.6MB, 4.7s |
| `winml quantize --precision int4` | ✅ | 109→23.7MB, 3.7s (RTN 4-bit) |
| `winml quantize --precision int8` | ✅ | 109→28.0MB, 46s (QDQ) |
| `winml build --precision fp16` | ✅ | Full pipeline, 87s |
| `winml build --precision int4` | ✅ | Full pipeline with RTN |

## Follow-up PRs

- #964: Refactor into `Quantizer` class with `BaseQuantPass` pipeline — enables multi-pass (e.g., `w4a16` = int4 + fp16)
- #963: Share `fp16_op_block_list` / `fp16_keep_io_types` across QDQ and FP16 paths


src/winml/modelkit/commands/build.py
```diff
@@ -1378,6 +1423,14 @@ def _name(base: str) -> str:
 
     stage_timings.append(("Export", _export_elapsed))
 
+    # ── FP16 conversion (when --precision fp16) ──────────────────
+    _precision = extra_kwargs.pop("precision", None)
+    if _precision == "fp16":
+        current_path = _run_fp16_stage(
+            model_path=current_path,
+            stage_timings=stage_timings,
+        )
```
timenick: **Pipeline order is inconsistent between the pytorch and ONNX build paths.**

In `_build_pytorch_pipeline` (here) FP16 runs **between Export and Optimize**, but elsewhere FP16 runs **after** Optimize:
- `_build_onnx_pipeline` (line ~1547): `Optimize → FP16 → Quantize`
- Standalone `winml optimize --precision fp16` (`optimize.py:435`): converts to FP16 **after** `optimizer.optimize(...)` returns
- The PR description's mermaid diagram: `Export → Optimize → FP16 → Quantize → Compile`

Practical impact: `winml build -m hf-model --precision fp16` will hand an FP16 graph to the optimizer, while `winml build -c cfg.json -m model.onnx --precision fp16` runs optimize on FP32 first. Two different graphs out for the same logical request, depending on input format.

Suggest moving the FP16 stage to run after `_run_optimize_stage` so both pipelines match the documented order.

src/winml/modelkit/commands/build.py
```diff
@@ -1119,6 +1125,45 @@ def _on_reoptimize(autoconf_dict: dict) -> None:
...
+        t0 = time.monotonic()
+
+        model = load_onnx(model_path)
+        model = convert_to_fp16(model, keep_io_types=True)
+        save_onnx(model, model_path)
+
+        elapsed = time.monotonic() - t0
+        sl.set_done(elapsed)
+        sl.detail("[dim]I/O types preserved as FP32[/dim]")
+        sl.artifact(str(model_path), 0)
```
timenick: `sl.artifact(str(model_path), 0)` hardcodes the size to `0`. Every other stage in this file passes `_safe_size(...)` (see lines 1121, 1249, 1322, 1421), so the FP16 stage will render as `0 B` in the stage summary. Should be:

```python
sl.artifact(str(model_path), _safe_size(model_path))
```

tests/unit/optim/test_fp16.py
```diff
@@ -0,0 +1,130 @@
...
+- CARDINAL RULE #3: Tests must run and pass
+"""
+
+from __future__ import annotations
+
+import numpy as np
+import onnx
+from onnx import TensorProto, numpy_helper
+
+from winml.modelkit.optim.fp16 import convert_to_fp16
```
timenick: Per `CLAUDE.md`, test code must use absolute imports at the package level, not reach into internal submodules for non-`_`-prefixed symbols. This line imports from the deep submodule `winml.modelkit.optim.fp16`, but `convert_to_fp16` is a new public function used by three commands (`build`, `optimize`, `export`).

Suggest exporting it from `src/winml/modelkit/optim/__init__.py` (add to both the imports and `__all__`) and changing this import to:

```python
from winml.modelkit.optim import convert_to_fp16
```

timenick (commented):
Three findings on PR #872.

src/winml/modelkit/quant/config.py
```diff
@@ -78,10 +89,23 @@ class WinMLQuantizationConfig:
...
+    # RTN-specific settings (only used when algorithm="rtn")
+    rtn_bits: int = 4
+    rtn_block_size: int = 128
+    rtn_symmetric: bool = True
+    rtn_accuracy_level: int = 0
+
+    # FP16 post-processing (can combine with rtn or static algorithm)
+    # When True, remaining FP32 tensors/ops are converted to FP16 after
+    # quantization. This produces mixed-precision models (e.g. RTN int4 + FP16).
+    fp16_postprocess: bool = False
```
xieofxie: a little weird.. we could also rtn + QDQ, correct? it more looks like quantize should also be applied multiple times in the build process

xieofxie: we should keep quantize a one operation tool - just one at one time and you could run it multiple times

Actually for QDQ, there could be exclude / includes nodes like Olive, so we could even run multiple passes to use different precisions

src/winml/modelkit/quant/config.py
```diff
@@ -47,11 +47,22 @@ class WinMLQuantizationConfig:
...
     """
 
+    # Quantization algorithm
+    algorithm: Literal["static", "dynamic", "rtn", "fp16"] = "static"
+    # "static"  — Calibrated QDQ quantization (requires calibration data)
+    # "dynamic" — Dynamic quantization (no calibration) [planned, not yet wired]
+    # "rtn"     — Round-To-Nearest weight-only (no calibration, block-wise)
+    # "fp16"    — Pure FP16 conversion only (no quantization)
+
     mode: Literal["qdq", "static", "dynamic"] = "qdq"
```
xieofxie: we already have mode here, do we need algorithm?
I think we just keep one?

src/winml/modelkit/commands/quantize.py
```diff
@@ -174,6 +187,103 @@ def quantize(
     # Import quantizer (late import to speed up CLI)
     from ..quant import WinMLQuantizationConfig, quantize_onnx
 
+    # ── FP16 fast path ───────────────────────────────────────────
```
zhenchaoni: There are many code duplicated across the 3 quantization strategy (fp16, RTN and QDQ). I would suggest you to use a strategy based pattern for the 3 algorithms. Otherwise, if you want to add another quantization algorithm, you will end up adding another path here and duplicate many logic. Furthermore, the config file is not honored in the 2 new added algorithm.

I would suggest you to always "translate" the CLI option to a WinMLQuantizationConfig for all the 3 quantization algorithms then hand over to quantize_onnx. `quantize_onnx` should be responsible to pick an algorithm based on the config.

All the console print, output folder logic, error message, etc should be reused.

src/winml/modelkit/commands/quantize.py
```diff
@@ -253,7 +363,14 @@ def _resolve_quant_types(
         Tuple of (weight_type, activation_type).
     """
     from ..config import is_quantized_precision, resolve_quant_types
+    from ..config.precision import is_weight_only_precision
 
+    if precision and is_weight_only_precision(precision.lower()):
```
zhenchaoni: if precision is weight only, then you can simply make activation_type to be f32 or NONE. So this method is extensible to all the 3 algorithms. The code will be less coupled.

src/winml/modelkit/quant/quantizer.py
```diff
@@ -22,28 +22,170 @@ def quantize_onnx(
...
+
+    model_path = Path(model_path)
+    config = config or WinMLQuantizationConfig()
+
+    if output_path is not None:
+        output_path = Path(output_path)
+    else:
+        output_path = model_path.parent / f"{model_path.stem}_qdq.onnx"
+
+    # If precision is provided, expand and run multi-pass
```
zhenchaoni: It is hard to get the reason why you run multi-pass if the precision is provided. I would suggest you to directly add an pass field in the WinMLQuantizationConfig. Then you can decide by that field. In the pass field, you can explicitly define the quantization algorithm in each pass. So the logic is configurable.

src/winml/modelkit/quant/quantizer.py
```diff
@@ -84,16 +223,98 @@ def quantize_onnx(
...
-    if output_path is not None:
-        output_path = Path(output_path)
-    else:
-        output_path = model_path.parent / f"{model_path.stem}_qdq.onnx"
-
     errors: list[str] = []
     warnings: list[str] = []
 
     try:
+        # ── Pure FP16 fast path (no quantization, only FP16 conversion) ──
```
zhenchaoni: I suggest you to have 3 methods, fp16, qdq and RTN for the 3 algorithms. You can put a registry dict and resolve by config.algorithm

src/winml/modelkit/quant/quantizer.py
```diff
@@ -233,6 +451,9 @@ def quantize_onnx(
             1 for node in quantized_model.graph.node if node.op_type in QDQ_OP_TYPES
         )
 
+        # Step 8: Save the final model
+        save_onnx(quantized_model, output_path, use_external_data=use_external_data)
```
zhenchaoni: Any reason why you move the save method call below?

zhenchaoni (changes requested):

src/winml/modelkit/quant/quantizer.py
```diff
@@ -22,28 +22,170 @@ def quantize_onnx(
     model_path: str | Path,
     output_path: str | Path | None = None,
     config: WinMLQuantizationConfig | None = None,
+    *,
+    precision: str | None = None,
```
xieofxie: a little weird.. how about the function still focus on one config just as WinMLQuantizationConfig , but in the command, we loop the multiple quantizes? i.e. when we use --precision w4a16, generate two WinMLQuantizationConfig  one for rtn and one for fp16 and then run them in a helper function. so eval, perf, build could also use it. But it will only work for command

src/winml/modelkit/commands/build.py
```diff
@@ -1133,67 +1161,97 @@ def _run_quantize_stage(
...
-        )
+            raise RuntimeError(f"{stage_name} failed: {errors}")
+
+        elapsed = time.monotonic() - t0
+        sl.set_done(elapsed)
+
+        # Show algorithm-specific result details
+        if is_fp16_only:
+            sl.detail("[dim]I/O types preserved as FP32[/dim]")
+        elif config.quant.algorithm == "rtn":
```
zhenchaoni: This comment is not just for this PR. I think `_run_quantize_stage` needs to be refactored. We cannot put quantization internal logic to build. The logic is very coupled here. The `build` command should just know it needs to run quantize by the config it has. It should not know internal logic inside quantization, otherwise the code encapsulation is broken.

src/winml/modelkit/commands/quantize.py
```diff
@@ -174,6 +187,103 @@ def quantize(
...
+
+        return
+
+    # ── Weight-only (RTN) path ───────────────────────────────────
+    from ..config.precision import (
+        extract_weight_bits,
+        is_weight_only_precision,
+    )
+
+    is_rtn = precision and is_weight_only_precision(precision.lower())
```
xieofxie: technically rtn is not weight_only, weight_only could have other algorithms? it is a little strange but no harm.

for the different log in 3 cases, maybe create 3 help functions and orchestrate in the main function to make logic clear like
```
precision_list = resolve_precision(precision)
for precision in precision_list:
    fp16 -> fp16();
    rtn -> rtn();
    qdq -> qdq();

```

src/winml/modelkit/quant/config.py
```diff
@@ -78,10 +87,20 @@ class WinMLQuantizationConfig:
...
 
+    # RTN-specific settings (only used when mode="rtn")
+    rtn_bits: int = 4
+    rtn_block_size: int = 128
+    rtn_symmetric: bool = True
+    rtn_accuracy_level: int = 0
+
+    # FP16 conversion settings (only used when mode="fp16")
+    fp16_keep_io_types: bool = True
+    fp16_op_block_list: list[str] | None = None
```
xieofxie, line 102: consider also apply to qdq and vice versa

a trick is get all types from input model to get the reverse list

src/winml/modelkit/quant/quantizer.py
DingmaomaoBJTU: Thanks for the suggestion! We considered moving multi-pass orchestration to the command layer, but decided to keep it inside `quantize_onnx(precision=...)` because:

1. **Reusability** — `build`, `eval`, and `perf` commands all call `quantize_onnx` under the hood. If multi-pass lives in the cmd layer, each caller would need to duplicate the expand-loop-cleanup logic.
2. **Encapsulation** — the precision expansion (e.g. w4a16 -> [int4, fp16]) is an implementation detail of quantization. Callers should only need to say "I want w4a16" and get the final model back.
3. **Intermediate file management** — the multi-pass path creates temp files between passes and cleans them up; keeping this in one place avoids leaking temp file concerns to every caller.

The single-config path (`quantize_onnx(config=...)`) still works exactly as before for one-shot use.

src/winml/modelkit/quant/quantizer.py
DingmaomaoBJTU: The multi-pass logic is triggered only for composite precisions like `w4a16` (which expands to `[int4, fp16]` — two sequential passes). Single precisions like `int4`, `fp16`, `int8` go directly to `_quantize_single_pass` with no loop. Added a clarifying comment in the docstring.

src/winml/modelkit/quant/config.py
DingmaomaoBJTU: Good idea — will address in a follow-up PR.

DingmaomaoBJTU: Tracking in #963.

src/winml/modelkit/quant/quantizer.py
zhenchaoni: I also feel a bit weird to add another precision parameter to the method. Can we just read the precision inside the config?

src/winml/modelkit/quant/config.py
xieofxie: if we want to add it later, we should not name it like this in first place

src/winml/modelkit/quant/quantizer.py
```diff
@@ -10,40 +10,347 @@
...
 
 
+if TYPE_CHECKING:
+    from collections.abc import Callable
+
+
 logger = logging.getLogger(__name__)
 
 
 def quantize_onnx(
```
xieofxie, line 25: Another option, expand this to 
```suggestion
@overload
def quantize_onnx(
    model_path: str | Path,
    output_path: str | Path | None = None,
    config: WinMLQuantizationConfig | None = None,
   ):
  @overload
 def quantize_onnx(
    model_path: str | Path,
    output_path: str | Path | None = None,
    configs: list[WinMLQuantizationConfig] | None = None,
   ):
    
   
```

src/winml/modelkit/quant/quantizer.py
DingmaomaoBJTU: This PR now focuses on single-pass only — `quantize_onnx` takes one `WinMLQuantizationConfig` and runs one operation. Multi-pass orchestration (e.g., w4a16 = int4 + fp16) will be implemented in a follow-up PR via a Quantizer class with BaseQuantPass pipeline (#964).

src/winml/modelkit/quant/quantizer.py
DingmaomaoBJTU: Good suggestion! This aligns with the Quantizer pipeline design in #964. For this PR we keep it single-pass (`config: WinMLQuantizationConfig`). The `configs: list[...]` overload will come with the multi-pass pipeline implementation.

src/winml/modelkit/quant/config.py
DingmaomaoBJTU: Tracked in #963 — will be part of the multi-pass pipeline work.

src/winml/modelkit/commands/build.py
```diff
@@ -1137,35 +1162,47 @@
...
+            try:
+                import datasets
 
-            datasets.disable_progress_bars()
-            _datasets_available = True
-        except ImportError:
-            pass  # datasets package not installed; progress bar suppression not needed
+                datasets.disable_progress_bars()
+                _datasets_available = True
+            except ImportError:
```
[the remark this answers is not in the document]
DingmaomaoBJTU: Fixed — added explanatory comment: `# datasets package is optional; calibration falls back to random data`

src/winml/modelkit/quant/config.py
```diff
@@ -128,8 +155,15 @@ def from_dict(cls, data: dict) -> WinMLQuantizationConfig:
         Returns:
             WinMLQuantizationConfig instance.
         """
+        # Backward compat: prefer "algorithm" (authoritative in old configs)
+        # over deprecated "mode" (which defaulted to "qdq").
+        # Map legacy "qdq" value to "static".
+        raw_mode = data["algorithm"] if "algorithm" in data else data.get("mode", "static")
```
xieofxie: we don't add algorithm

src/winml/modelkit/quant/config.py
```diff
@@ -47,11 +47,20 @@ class WinMLQuantizationConfig:
         # Custom config
         config = WinMLQuantizationConfig(samples=100, weight_type="int8")
         result = quantize_onnx("model.onnx", config)
+
+        # FP16 conversion (pure FP16, no quantization)
+        config = WinMLQuantizationConfig(mode="fp16")
+        result = quantize_onnx("model.onnx", config)
     """
 
-    mode: Literal["qdq", "static", "dynamic"] = "qdq"
```
xieofxie, line 52: do we use qdq before? what does it mean?

DingmaomaoBJTU: Yes — `"qdq"` was the original default in all existing recipe configs (~50+ JSON files in `examples/recipes/`). QDQ stands for "Quantize-Dequantize" which is the same thing as static quantization. This shim maps `"qdq"` to `"static"` so those configs keep working. Tracked full removal in #971.

src/winml/modelkit/quant/config.py
DingmaomaoBJTU: You are right — `"algorithm"` was never used in any config file. Removed it. Now the code just reads `data.get("mode", "static")` with the `qdq → static` compat shim.

zhenchaoni (approved):

xieofxie:
1. commands/quantize.py:287 — --precision w4a16 --weight-type uint8 silently ignores the documented weight-type override.
The dispatch checks elif precision_lower and _is_weight_only(precision_lower) (true for w4a16/w4a8) before the QDQ branch that honors explicit --weight-type/--activation-type. The precision_option help advertises that explicit type flags override precision, but for a weight-only precision the RTN path runs unconditionally. Failure: user runs winml quantize --precision w4a16 --weight-type uint8 expecting uint8 QDQ, gets RTN int4 weight-only instead (a yellow warning is printed, but it contradicts the documented precedence).

2. config/precision.py:737 / quantizer.py — --precision w4a16 (and w4a8) is accepted but runs int4-only, silently dropping the activation pass.
_is_valid_precision/is_weight_only_precision accept w4a16, and quantize_onnx runs a single RTN int4 pass. expand_precision("w4a16") == ["int4", "fp16"] documents that the a16 (FP16 activation) pass should follow, but multi-pass is deferred to #964 and never invoked. Failure: --precision w4a16 produces an int4/w4a32-equivalent model that doesn't match its name, with no error. Consider rejecting w4a16/w4a8 at the CLI until #964 lands, rather than producing a mislabeled artifact.

Let's use another item to fix this?

One option is also add weight / activation type flags to build and add a common resolve_helper to consider all 3

src/winml/modelkit/commands/build.py
```diff
@@ -1137,35 +1162,47 @@ def _run_quantize_stage(
...
-        sl.kv(
-            "Calibration:",
-            f"[cyan]{config.quant.samples}[/cyan] samples"
-            f"  [dim]({config.quant.calibration_method})[/dim]",
-        )
-        # Suppress tqdm/datasets progress bars during quantize
-        # to keep Live display clean
+    # Determine stage label from precision/algorithm
+    precision = config.precision
+    is_fp16_only = precision and precision.lower() == "fp16"
```
xieofxie: FP16 stage display is keyed off `config.precision`, but the actual algorithm is driven by `config.quant.mode`. `_patch_device` only sets `cfg.precision` when the CLI `--precision` arg is truthy, while it always sets `cfg.quant.mode`.

So `winml build -c config.json --device cpu` (or `gpu`) with no `--precision` resolves `mode="fp16"` but leaves `config.precision=None`. `is_fp16_only` is then falsy and the stage takes the QDQ branch — "Quantizing (uint8)…", Dataset/Calibration rows, the `datasets` progress-bar handling — and records the timing as "Quantize", even though `quantize_onnx` actually performs a pure FP16 conversion.

The RTN branch already keys its display off `config.quant.mode`; FP16 detection should do the same, e.g. `is_fp16_only = config.quant.mode == "fp16"`.

src/winml/modelkit/commands/quantize.py
```diff
@@ -174,69 +182,101 @@ def quantize(
...
     except Exception as e:
-        console.print(f"\n[bold red]Quantization failed:[/bold red] {e}")
-        logger.exception("Quantization failed")
-        raise click.ClickException(f"Quantization failed: {e}") from e
+        console.print(f"\n[bold red]{label} failed:[/bold red] {e}")
+        logger.exception("%s failed", label)
+        raise click.ClickException(f"{label} failed: {e}") from e
+
+
+def _is_weight_only(precision: str) -> bool:
```
xieofxie: `_is_weight_only` is a one-line wrapper around `is_weight_only_precision`, which is already exported from `config/__init__.py` and imported directly elsewhere (e.g. `build.py`). The wrapper adds an extra name + a duplicated late import for the same behavior.

Its second use also feeds an unreachable guard in `_resolve_quant_types` (`# Should not reach here — RTN path returns early above.`) — that branch is dead because the RTN path returns before `_resolve_quant_types` is ever called.

Relatedly, `WinMLQuantizationConfig.fp16_op_block_list` is serialized and forwarded to `convert_to_fp16`, but no CLI/policy entry point ever sets it to a non-`None` value, so it is currently inert config surface. Suggest dropping the wrapper + dead branch and deferring `fp16_op_block_list` until there's an entry point that populates it.

src/winml/modelkit/commands/build.py
DingmaomaoBJTU: Fixed — now uses `config.quant.mode == "fp16"` directly instead of `config.precision`. This correctly detects FP16 regardless of whether `--precision` was explicitly passed or resolved from config.

src/winml/modelkit/commands/quantize.py
DingmaomaoBJTU: Fixed — removed `_is_weight_only` wrapper (now imports `is_weight_only_precision` directly at module level), and removed the dead RTN guard in `_resolve_quant_types`. Re: `fp16_op_block_list` — it is wired to `convert_to_fp16()` in quantizer.py but currently has no CLI entry point to populate it. Tracked in #963 for when the CLI surface is added.

xieofxie (approved):

PR #875, opened 2026-06-11
feat: add running model path / model precision to perf output

### LINKED ISSUE #81: Feat: need more perf results
When I run wmk perf --hf-model microsoft/resnet-50, it results: 
 "benchmark_info": {
    "model_id": "microsoft/resnet-50",
    "task": "auto-detected",
    "device": "cpu",
    "precision": "auto",
    "iterations": 100,
    "warmup": 10,
    "batch_size": 1,
    "timestamp": "2026-03-03T09:14:06.464939+00:00"
  },

My questions:
1. Where is the output model? Seems no parameter I can set for the model path.
2. Anyway to know the task and precision?
@zhengte_microsoft

xieofxie:
## Summary

Closes #81.

The `perf` JSON report did not record which ONNX model was actually run, so users had no way to locate the built/compiled model on disk (issue #81, question 1). This adds the **real running model path** — the ONNX file ORT actually loads, which can differ from the input `model_id`/`onnx_path` when an EPContext (compiled) model is produced or a cached one is reused.

`benchmark_info` now includes `running_model_path`:

```json
"benchmark_info": {
    "model_id": "microsoft/resnet-50",
    "running_model_path": "C:\Users\...\.cache\winml\...\model_cpu_ctx.onnx",
    "task": "auto-detected",
    "device": "cpu",
    ...
}
```

## Changes

- **`session/session.py`**: track the resolved model path ORT loads via a new `running_model_path` property (set in `compile()`; falls back to the input path before compilation).
- **`session/qairt/qairt_session.py`**: record the EPContext path the QAIRT session loads.
- **`models/winml/base.py`**: expose `running_model_path` on the model, delegating to the session.
- **`commands/perf.py`**: emit `running_model_path` in `benchmark_info`, and add it to each per-module instance entry in `--module` mode.

## Tests

- Extended `tests/e2e/test_perf_e2e.py::test_benchmark_cpu` to assert `running_model_path` is an existing `.onnx` file.
- Verified locally (CPU, `-m e2e`): `test_benchmark_cpu` and `test_module_benchmark_cpu` pass; `tests/unit/commands/test_perf_cli.py` (28) passes.

@

### DIFF
--- a/src/winml/modelkit/commands/perf.py
+++ b/src/winml/modelkit/commands/perf.py
@@ -101,6 +101,10 @@ class BenchmarkResult:
     output_names: list[str] = field(default_factory=list)
     output_shapes: list[list[int]] = field(default_factory=list)
 
+    # Resolved model precision from io_config (None if the model does not
+    # expose one). Distinct from the requested config.precision policy.
+    model_precision: str | None = None
+
     # Latency stats (milliseconds)
     mean_ms: float = 0.0
     min_ms: float = 0.0
@@ -172,6 +176,8 @@ class BenchmarkResult:
             },
             "raw_samples_ms": [round(s, 3) for s in self.raw_samples_ms],
         }
+        if self.model_precision:
+            result["model_info"]["precision"] = self.model_precision
         if self.hw_monitor:
             result["hw_monitor"] = self.hw_monitor
         return result
@@ -502,6 +508,7 @@ class PerfBenchmark:
             input_types=[str(t) for t in io_config["input_types"]],
             output_names=io_config["output_names"],
             output_shapes=[list(s) if s else [] for s in io_config["output_shapes"]],
+            model_precision=io_config.get("precision"),
             # Latency stats
             mean_ms=stats.mean_ms,
             min_ms=stats.min_ms,
--- a/src/winml/modelkit/commands/perf.py
+++ b/src/winml/modelkit/commands/perf.py
@@ -126,6 +126,10 @@ class BenchmarkResult:
     actual_task: str = ""
     actual_ep: EPName | None = None
 
+    # ONNX model ORT actually loaded (may be an EPContext model, differing
+    # from the input model_id when compiled or a cached one is reused)
+    running_model_path: str = ""
+
     # Hardware monitor metrics (from HWMonitor.to_dict())
     hw_monitor: dict[str, Any] | None = None
 
@@ -134,6 +138,7 @@ class BenchmarkResult:
         result = {
             "benchmark_info": {
                 "model_id": self.config.model_id,
+                "running_model_path": self.running_model_path,
                 "task": self.actual_task,
                 "device": self.actual_device,
                 "ep": self.actual_ep,
@@ -515,6 +520,7 @@ class PerfBenchmark:
             actual_device=self._model.device,
             actual_task=self._model.task or self.config.task or "auto-detected",
             actual_ep=self._model.ep_name,
+            running_model_path=str(self._model.running_model_path),
             # Hardware monitor metrics (only present when --monitor is used)
             hw_monitor=getattr(self, "_hw_metrics", None),
         )
@@ -707,6 +713,7 @@ def _perf_modules(
                 mod_stats = stats
                 result_entry: dict[str, Any] = {
                     "module_path": module_path,
+                    "running_model_path": str(session.running_model_path),
                     "mean_ms": round(mod_stats.mean_ms, 3),
                     "p50_ms": round(mod_stats.p50_ms, 3),
                     "p90_ms": round(mod_stats.p90_ms, 3),
--- a/src/winml/modelkit/commands/perf.py
+++ b/src/winml/modelkit/commands/perf.py
@@ -139,7 +139,7 @@ class BenchmarkResult:
 
     def to_dict(self) -> dict[str, Any]:
         """Convert to dictionary for JSON serialization."""
-        result = {
+        result: dict[str, Any] = {
             "benchmark_info": {
                 "model_id": self.config.model_id,
                 "running_model_path": self.running_model_path,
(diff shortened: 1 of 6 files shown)


DingmaomaoBJTU (commented):
## Code Review: feat: add running model path / model precision to perf output

Overall this is clean and well-documented. A few issues to address, from most to least critical.

### Summary of findings

1. `if self.model_precision:` should be `is not None` - falsy guard on a string creates an unstable JSON schema (see inline comment on perf.py:179).
2. Test assertion will KeyError when precision is undetectable - minfo["precision"] raises rather than failing cleanly when the key is absent (see inline on test_perf_e2e.py:231).
3. _running_model_path set before session creation in QAIRT - the path is recorded even if ort.InferenceSession throws, leaving a misleading stale value (see inline on qairt_session.py:236).
4. _perf_modules new field has no test coverage - the running_model_path key added to result_entry is not verified in any test.

src/winml/modelkit/commands/perf.py
```diff
@@ -167,6 +176,8 @@ def to_dict(self) -> dict[str, Any]:
             },
             "raw_samples_ms": [round(s, 3) for s in self.raw_samples_ms],
         }
+        if self.model_precision:
```
DingmaomaoBJTU: Using a falsy guard `if self.model_precision:` means the `precision` key is conditionally absent from `model_info`. When `_get_precision()` returns `None` (no detectable precision signal), the key is silently omitted, and any consumer doing `data["model_info"]["precision"]` gets a `KeyError` instead of a handled null.

Prefer always emitting the key with a null for unknown:
```python
result["model_info"]["precision"] = self.model_precision
```
This matches how `running_model_path` is always emitted in `benchmark_info` (even as empty string) and makes the schema stable and predictable.

tests/e2e/test_perf_e2e.py
```diff
@@ -226,6 +228,7 @@ def test_benchmark_cpu(self, tmp_path: Path, model_arg: str):
         assert len(minfo["input_names"]) >= 1
         assert isinstance(minfo["output_names"], list)
         assert len(minfo["output_names"]) >= 1
+        assert minfo["precision"] == "fp32"
```
DingmaomaoBJTU, line 231: If `_get_precision()` returns `None` for the test model (e.g., a stub or minimal ONNX with no weight initializers), `model_info` will not have a `precision` key (see `perf.py` where it is conditionally emitted). This line then raises `KeyError` rather than an `AssertionError`, producing a confusing failure message.

Safer pattern:
```python
assert minfo.get("precision") == "fp32"
```
Or, if you want to assert the key is always present for fp32 models:
```python
assert "precision" in minfo, "model_info.precision missing - _get_precision returned None"
assert minfo["precision"] == "fp32"
```

src/winml/modelkit/session/qairt/qairt_session.py
```diff
@@ -233,6 +233,7 @@ def _create_inference_session(self) -> None:
         import onnxruntime as ort
 
         sess_options, _, _ = self._build_session_options(self._device)
+        self._running_model_path = self._ctx_path
```
DingmaomaoBJTU: `_running_model_path` is assigned before `ort.InferenceSession` is created (next line). If the constructor throws, `_running_model_path` will be set to `_ctx_path` even though no session was ever successfully loaded. A caller accessing `running_model_path` after a failed `_create_inference_session` will get a stale, misleading path.

The same pattern exists in the base `session.py` (line 253 vs 262), so this is at least consistent — but consider moving the assignment after a successful session creation in both places, or resetting it in the error path.

src/winml/modelkit/commands/perf.py
```diff
@@ -707,6 +720,7 @@ def _perf_modules(
                 mod_stats = stats
                 result_entry: dict[str, Any] = {
                     "module_path": module_path,
+                    "running_model_path": str(session.running_model_path),
```
DingmaomaoBJTU, line 722: The new `running_model_path` field in the module-perf result entry has no test coverage. `TestPerfModuleParameterForwarding` mocks `WinMLSession` with `perf.side_effect = RuntimeError`, so the benchmark loop never completes and `result_entry` is never populated.

Consider adding an assertion in an existing or new test that verifies `"running_model_path"` appears in the per-module result dict — even if only by checking the mock return value of `session.running_model_path`.

tests/e2e/test_perf_e2e.py
xieofxie: it exists for this case

src/winml/modelkit/session/qairt/qairt_session.py
xieofxie: updated both

src/winml/modelkit/commands/perf.py
xieofxie: test added

src/winml/modelkit/commands/perf.py
xieofxie: added

DingmaomaoBJTU (commented):
Follow-up review: all 4 previously flagged issues are now addressed.

- **perf.py: falsy precision guard** — fixed. `"precision": self.model_precision` is now always emitted inline in `model_info`, making the schema stable (null when precision is unknown). ✅
- **test_perf_e2e.py: KeyError risk** — resolved as a consequence of the above fix; the key is always present now. ✅
- **qairt_session.py + session.py: stale path on failed session** — fixed in both files. `_running_model_path` is now set only after `ort.InferenceSession` constructs successfully. ✅
- **perf.py:723 / _perf_modules: no test coverage** — fixed. `test_running_model_path_in_module_result` covers the full round-trip (successful benchmark → JSON report → `running_model_path` value matches mock). ✅

One non-blocking observation: the **error-path dict** in `_perf_modules` (~line 743) omits `running_model_path`, while successful entries always include it. Consumers iterating `instances` must handle both shapes. Not a regression — the `"error"` key already distinguishes them — but worth noting if a formal schema is ever added.

DingmaomaoBJTU (approved):
LGTM! All 4 prior findings resolved. One non-blocking observation: the error-path dict in _perf_modules omits running_model_path while success entries always include it — worth aligning if a formal output schema is added later.

Issue #877, opened 2026-06-11
Refactor: unify task detection into one core; derive modality from model class, not config-field heuristics


timenick:
## Summary

`detect_task` (used by `inspect`/`eval`) and `_detect_task_and_class_from_config`
(used by `config`/`build`) are two implementations of the same task decision over
the same `MODEL_CLASS_MAPPING` data, and a **third** entry point
(`resolve_loader_config` step 2, used by `--model-type`) resolves the task a fourth
way via `get_supported_tasks(type)[0]`. They have drifted, and the
modality-disambiguation step (D2) reconstructs modality from config field names —
a heuristic that is provably weaker than information the pipeline already holds.

Proposal: extract a single task-detection core + a single task-override lookup that
**all three entry points** consult, and derive modality from the resolved model
class's `main_input_name`.

Internal refactor only — public `detect_task` / `resolve_task_and_model_class`
signatures stay unchanged.

## Motivation 1: the D2 false positive (concrete bug)

D2 upgrades `feature-extraction → image-feature-extraction` when the config has a
top-level `image_size`/`patch_size` (OR semantics). `patch_size` is **not exclusive
to vision** — spectrogram transformers patchify their mel-spectrogram. Verified via
the real detection path:

| declared architecture | `inspect` task today | correct |
|---|---|---|
| `Wav2Vec2Model` | `feature-extraction` → text dataset → fails | audio |
| `WhisperModel` | `feature-extraction` → text dataset → fails | audio |
| `ASTModel` | `image-feature-extraction` → image dataset → fails | audio |

## Motivation 2: sam2 resolves three different ways (entry-point drift)

`sam2` is a multi-task type (`MODEL_CLASS_MAPPING` registers feature-extraction,
image-feature-extraction, image-segmentation, mask-generation) **plus** a
`(sam2, None)` sentinel whose canonical export target is the mask-generation decoder.
The three entry points disagree (verified):

| entry point | path | sam2 result |
|---|---|---|
| `inspect` (`detect_task`) | short-circuit fires only for a single real task → falls through to TasksManager on `Sam2Model` (headless) | `feature-extraction` |
| `build -m <ckpt>` (`_detect_task_and_class_from_config`) | applies the `(sam2, None)` sentinel | **`mask-generation`** (decoder) |
| `config/build --model-type sam2` (`resolve_loader_config` step 2) | `get_supported_tasks("sam2")[0]` — bypasses the sentinel entirely | `feature-extraction` (encoder) |

[...]

PR #878, opened 2026-06-12
refactor(task): unify detection into resolve_task/TaskResolution; fix inspect source

timenick:
## What

Unifies WinML's task detection into a single resolver. `detect_task` (inspect/eval), `_detect_task_and_class_from_config` (build via `-m`), and `resolve_loader_config` step 2 (config/build via `--model-type`) were three implementations of the same decision over the same `MODEL_CLASS_MAPPING` data, and had drifted. Modality was reconstructed from config field names — a heuristic that misroutes models whose fields aren't modality-exclusive.

All three are replaced by a single `resolve_task(config, *, task=None, model_class=None) -> TaskResolution` in a new `loader/resolution.py`, consumed by every entry point. `loader/task.py` is reduced to data tables + boundary utils. Modality derives from the model class's `main_input_name`. Addresses #877.

**API change:** legacy `detect_task` / `resolve_task_and_model_class` / `_detect_task_and_class_from_config` are removed; callers use `resolve_task` + `TaskResolution` fields. `resolve_loader_config` returns the `TaskResolution` as a 4th tuple element.

## Why (concrete bugs fixed)

- **Inspect reported a misleading task source.** `winml inspect` recomputed provenance post-hoc by scanning `MODEL_CLASS_MAPPING`, so a model detected via Optimum (e.g. TAPEX/BART) was mislabeled `HF_MODEL_CLASS_MAPPING`. It now reports the resolver's real `source` (`tasks-manager`).
- **AST misroute:** `ASTModel` carries `patch_size`, so the old config-field heuristic upgraded it to `image-feature-extraction` → image dataset → fail. `main_input_name` (`input_values`) keeps it `feature-extraction`.
- **sam2/sam2-video disagreement:** `feature-extraction` via inspect/`--model-type` but `mask-generation` via `-m` (different artifacts for the same model). Now `mask-generation` everywhere.
- **`load_hf_model` returned an inconsistent task** (Optimum-canonical in the `user_script` branch, modality-aware otherwise). Both now return the modality-aware task.
- **Calibration crash:** a backbone that stays `feature-extraction` but exports non-text inputs routed to a text dataset and crashed; calibration falls back to `RandomDataset` on a modality mismatch.

## How

- **`resolve_task` (5 stages):** user override → detection (override / no-architectures / TasksManager + fill-mask→seq2seq correction / default) → model class → modality upgrade (detection path only) → composite tag. Returns `TaskResolution(task, optimum_task, model_class, source, composite)` with the invariant `optimum_task == to_optimum_task(task)`.
- **`TaskSource` enum** replaces stringly-typed provenance with semantic values (`tasks-manager`, `sentinel-default`, `model-id-default`, `wrapped-library`, `hf-task-default`, `user-task`, `user-class`) — no internal data-structure names leak into `inspect` output.
- **`resolve_composite`** + the seq2seq bridge move from `commands/config.py` into the resolver; `TaskResolution.composite` carries sub-components so `winml config` no-task routing matches explicit `--task`.
- **`_resolve_task_override`** (model-id default → `(model_type, None)` sentinel reverse-lookup) is the single override source; a default-task override is declared **only** by an explicit sentinel, not by a lone `(model_type, task)` class-fix entry (e.g. segformer image-segmentation) — the architecture head decides otherwise.
- **Modality** derives from the architecture class's `main_input_name` (`pixel_values` → `image-feature-extraction`; text/audio stay `feature-extraction`).
- **`--model-type` fallback** (first ONNX-exportable task) and the timm wrapped-library stage merge into one no-architectures stage in `resolve_task`.
- Detection helpers relocate from `task.py` into `resolution.py`. Calibration degrades to `RandomDataset` (real input specs from the ONNX) on a modality mismatch.

## Behavior changes

| case | before | after |
|---|---|---|
| inspect task source (seq2seq, e.g. TAPEX) | mislabeled `HF_MODEL_CLASS_MAPPING` | `tasks-manager` (real source) |
| AST (`ASTModel`) | `image-feature-extraction` | `feature-extraction` |
| sam2 / sam2-video | inspect/`--model-type`: `feature-extraction`; `-m`: `mask-generation` | `mask-generation` everywhere |
| segformer classification ckpt (build) | `image-segmentation` | `image-classification` |
| `prajjwal1/bert-tiny` (inspect) | head-detected | `feature-extraction` (model-id default) |
| `load_hf_model` (`user_script` branch) | Optimum-canonical task | modality-aware task |
| undetectable config | `detect_task` fell back; loader path raised | unified to `hf-task-default` fallback |
| task `source` strings | `TasksManager` / `HF_MODEL_CLASS_MAPPING` / … | cleaned kebab-case `TaskSource` values |

## Out of scope (deferred)

- Inspect rendering the full composite structure (encoder+decoder + available pipeline tasks) in the formatter.
- A real `audio-feature-extraction` evaluator + dataset (and the task name itself).
- Video task support — optimum has no ONNX export config for videomae/timesformer/vivit.


DingmaomaoBJTU (commented):
## Code Review

### Summary
This PR correctly unifies task detection by replacing scattered, divergent paths with a single `resolve_task()` 5-stage resolver. The sentinel-based override mechanism for SAM/SAM2, the modality upgrade via `main_input_name`, and the composite component tagging are all well-designed. Below are the substantive issues I found.

---

### Bug: USER_CLASS path skips modality upgrade (`resolution.py` line 367–369)

When `model_class` is provided without `--task`, the USER_CLASS stage returns `opt_task` directly as `TaskResolution.task` — no call to `_resolve_task_modality`. Example:

`resolve_task(vit_config, model_class="ViTModel")`
→ `TaskResolution(task="feature-extraction", ...)` ← should be `"image-feature-extraction"`

The class docstring says `task` is always WinML modality-aware. The detection path (Stage 3) applies `_resolve_task_modality`; the USER_CLASS path does not. This causes `TASK_DATASET_MAPPING` to route vision models to `TextDataset` for calibration.

Fix:
`python
# resolution.py, after opt_task is set in the USER_CLASS branch:
surfaced = _resolve_task_modality(config, opt_task)
return TaskResolution(surfaced, to_optimum_task(surfaced), resolved, TaskSource.USER_CLASS, None)
`

No existing test covers this combination; it should be added to `test_resolve_task.py`.

---

### Warning: Stage 1b `KeyError` propagates uncaught (`resolution.py` line 417)

`python
resolved = TasksManager.get_model_class_for_task(opt_task, framework="pt")
`

This line has no `try/except`. Stage 2's error handling does **not** protect it — `resolved` is set here, before Stage 2 checks `if resolved is None`. If the first supported task from `get_supported_tasks` cannot be looked up under `framework="pt"` (e.g. a timm_wrapper whose classes are under the timm library, not transformers), a raw `KeyError` escapes. Stage 2's `except Exception` pattern should be applied here too.

---

### Warning: `assert` eliminated by `-O` (`resolution.py` line 447)

`python
assert source is not None
`

Python's `-O` flag strips `assert` statements. If an unexpected code path were to leave `source = None`, this guard disappears in optimised builds. Replace with:

`python
if source is None:
    raise RuntimeError("resolve_task: internal invariant violated — source was not set")
`

---

### Warning: `image-feature-extraction` not pinned in `TASK_SYNONYM_EXTENSIONS` (`task.py` line 222)

`to_optimum_task("image-feature-extraction")` relies on Optimum's `TasksManager.map_from_synonym` to collapse it to `"feature-extraction"`. The docstring documents the collapse; the tests verify it works today. But unlike `"mask-generation"` (which **is** pinned in `TASK_SYNONYM_EXTENSIONS` to prevent Optimum from mis-normalising it), `"image-feature-extraction"` has no fallback. If a future Optimum update changes the mapping, all downstream Optimum API calls will silently receive the unknown task string and fail.

Suggested addition:

`python
TASK_SYNONYM_EXTENSIONS: dict[str, str] = {
    "image-feature-extraction": "feature-extraction",   # add this
    "next-sentence-prediction": "text-classification",
    "mask-generation": "mask-generation",
}
`

---

### Info: audio feature-extraction lands in `TextDataset` (`datasets/__init__.py` line 44)

`_FEATURE_MODALITY_BY_MAIN_INPUT` only upgrades `pixel_values` → `image-feature-extraction`. Audio models (`input_values`, `input_features`) stay as `"feature-extraction"` and hit `TextDataset`. The new RandomDataset fallback handles the mismatch gracefully, but it would help to add a `# TODO: add audio-feature-extraction once AudioDataset exists` comment at line 44 so this gap is visible.

---

### Info: missing test for USER_CLASS with vision feature-extraction

`test_resolve_task.py` has no case for `resolve_task(vit_config, model_class="ViTModel")`. Even if the modality gap is intentional (user is being explicit), the expected behaviour should be pinned by a test so future refactors cannot silently change it.

src/winml/modelkit/loader/resolution.py
```diff
@@ -0,0 +1,448 @@
...
+            resolved = TasksManager.get_model_class_for_task(
+                opt_task, framework="pt", model_class_name=model_class
+            )
+        except (KeyError, AttributeError) as e:
+            raise ValueError(
+                f"Model class '{model_class}' not found for task '{opt_task}'. "
+                f"Check that the class name is correct and available in transformers."
+            ) from e
+        return TaskResolution(
+            opt_task, to_optimum_task(opt_task), resolved, TaskSource.USER_CLASS, None
```
DingmaomaoBJTU: **🐛 Bug: USER_CLASS path skips modality upgrade.**

`opt_task` (e.g. `"feature-extraction"` for `ViTModel`) is returned as `TaskResolution.task` without calling `_resolve_task_modality`. This contradicts the dataclass docstring ("task is WinML modality-aware") and causes downstream consumers — calibration dataset selection, eval routing — to receive `"feature-extraction"` instead of `"image-feature-extraction"`, routing vision models to `TextDataset`.

The detection path (Stage 3, line 442) applies `_resolve_task_modality`; this branch should match:

```python
# replace lines 367-369 with:
surfaced = _resolve_task_modality(config, opt_task)
return TaskResolution(surfaced, to_optimum_task(surfaced), resolved, TaskSource.USER_CLASS, None)
```

A test for `resolve_task(vit_config, model_class="ViTModel")` is also needed to pin the expected value.

src/winml/modelkit/loader/resolution.py
```diff
@@ -0,0 +1,448 @@
...
+    if opt_task is None and not getattr(config, "architectures", None) and model_type:
+        # Populate Optimum's ONNX export-config registry before querying it;
+        # get_supported_tasks returns [] if this hasn't been imported.
+        import optimum.exporters.onnx.model_configs  # noqa: F401
+
+        supported = get_supported_tasks(model_type, resolve_optimum_library(model_type))
+        if supported:
+            opt_task = supported[0]
+            source = TaskSource.WRAPPED_LIBRARY
+            resolved = TasksManager.get_model_class_for_task(opt_task, framework="pt")
```
DingmaomaoBJTU: **⚠️ Warning: uncaught `KeyError` can escape from Stage 1b.**

`TasksManager.get_model_class_for_task(opt_task, framework="pt")` is called here without a `try/except`. Stage 2's `except Exception` block (line 437) does **not** protect this call — it only runs when `resolved is None`, but `resolved` is being assigned here. For wrapped-library models (e.g. `timm_wrapper`, whose classes live under the timm framework, not `"pt"`), a `KeyError` escapes with no actionable message.

Suggested fix — consistent with Stage 2:
```python
try:
    resolved = TasksManager.get_model_class_for_task(opt_task, framework="pt")
except Exception:
    resolved = None  # Stage 2 fallback will handle it
```

DingmaomaoBJTU (commented):
## Code Review: fix(task): unify task detection across entry points

The unified `resolve_task()` 5-stage resolver is a sound design — sentinel-based overrides, the `TaskResolution`/`TaskSource` dataclasses, and the composite tagging are all well-structured. The cross-entry-point tests in `test_task_entry_point_consistency.py` are a good addition.

Five issues below (one bug, three warnings, one info).

---

### ⚠️ `task.py` line 222 — `TASK_SYNONYM_EXTENSIONS` missing `image-feature-extraction` entry

`to_optimum_task("image-feature-extraction")` relies on Optimum's `map_from_synonym` to collapse this to `"feature-extraction"`. Comments in `clip.py` and `blip.py` confirm Optimum knows the synonym today, but unlike `"mask-generation"` (which is explicitly pinned here to prevent Optimum mis-normalisation), there is no fallback. If Optimum changes the mapping, all downstream Optimum API calls silently receive the unknown string `"image-feature-extraction"` and fail.

Suggested fix:
```python
TASK_SYNONYM_EXTENSIONS: dict[str, str] = {
    "image-feature-extraction": "feature-extraction",  # add
    "next-sentence-prediction": "text-classification",
    "mask-generation": "mask-generation",
}
```

src/winml/modelkit/loader/resolution.py
```diff
@@ -0,0 +1,448 @@
...
+            except Exception:
+                resolved = _resolve_model_class_from_config(config)  # arch fallback
+
+    # --- Stage 3: modality upgrade (surfaced task only) -------------------
+    surfaced = _resolve_task_modality(config, opt_task)
+
+    # --- Stage 4: composite tag (detection path) --------------------------
+    composite = _composite_components_for_task(model_type, opt_task) if model_type else None
+
+    assert source is not None
```
DingmaomaoBJTU: **⚠️ Warning: `assert` is stripped by Python `-O`.**

Python's `-O` (optimize) flag eliminates all `assert` statements. This guard disappears in optimised builds, converting what should be a hard error into a `None` being silently passed into `TaskResolution`.

Replace with an explicit raise:
```python
if source is None:
    raise RuntimeError(
        "resolve_task: internal invariant violated — source was never assigned. "
        "This is a bug in the stage-1 detection logic."
    )
```

tests/unit/loader/test_resolve_task.py
```diff
@@ -0,0 +1,128 @@
...
+    cfg._name_or_path = "prajjwal1/bert-tiny"
+    r = resolve_task(cfg)
+    assert r.task == "feature-extraction"
+    assert r.source == TaskSource.MODEL_ID_DEFAULT
+
+
+def test_user_class_unknown_raises_friendly_error():
+    cfg = _cfg("bert", ["BertModel"])
+    with pytest.raises(ValueError, match="not found for task"):
+        resolve_task(cfg, model_class="NotARealClass")
```
DingmaomaoBJTU, line 77: **ℹ️ Info: missing test for USER_CLASS path with a vision feature-extraction model.**

There is no test case for `resolve_task(vit_config, model_class="ViTModel")` (or any vision-class USER_CLASS call without an explicit task). Whether the expected `task` is `"feature-extraction"` (intentional opt-out of upgrade) or `"image-feature-extraction"` (consistent with detection path) should be pinned here so future refactors cannot silently change the contract.

Example case to add:
```python
def test_user_class_feature_extraction_vision_returns_expected_task():
    cfg = _cfg("vit", ["ViTModel"])
    r = resolve_task(cfg, model_class="ViTModel")
    assert r.source == TaskSource.USER_CLASS
    # Pin whichever value is intended ("feature-extraction" or "image-feature-extraction"):
    assert r.task == "image-feature-extraction"  # or "feature-extraction" if intentional
```

timenick:
Thanks for the thorough review — all six addressed in `5d47ce9e`.

**1. USER_CLASS skips modality (bug)** — Fixed, with one refinement to your suggested patch. Applying `_resolve_task_modality` *unconditionally* would also upgrade a user-given `--task` (e.g. `--model-class ViTModel --task feature-extraction`), which should stay verbatim like the USER_TASK path. So I apply modality only when the task is **inferred** (no `--task`):

```python
if task is not None:
    opt_task = normalize_task(task); surfaced = opt_task          # preserve user intent
else:
    opt_task = _infer_task_from_architecture(config)
    surfaced = _resolve_task_modality(config, opt_task)           # inferred -> modality-aware
```

`resolve_task(vit_cfg, model_class="ViTModel")` now returns `image-feature-extraction`. Pinned by two new tests.

**2. Stage 1b KeyError uncaught** — Fixed by **removing** the eager `get_model_class_for_task` from 1b and letting Stage 2 resolve the class uniformly (it already has the `try/except` + arch fallback) — DRY, and the lookup can no longer escape as a raw `KeyError`. timm routing tests still green. (Legacy had the same un-wrapped call, so not a regression, but worth hardening.)

**3. `assert` stripped by `-O`** — Replaced with `if source is None: raise RuntimeError(...)`.

**4. `image-feature-extraction` not pinned** — Agreed; confirmed it's absent from Optimum's `get_all_tasks()`, so it's a WinML-owned name. Pinned in `TASK_SYNONYM_EXTENSIONS` → `feature-extraction`.

**5. audio → TextDataset (info)** — Added a `# TODO` at `_FEATURE_MODALITY_BY_MAIN_INPUT` for `input_values`/`input_features` → `audio-feature-extraction` once an AudioDataset exists.

**6. missing USER_CLASS test** — Added `test_user_class_inferred_task_is_modality_aware` + `test_user_class_with_explicit_task_preserved_verbatim`.

Generated with Claude Code

src/winml/modelkit/loader/resolution.py
timenick: Fixed in `5d47ce9e`, with a refinement: modality is applied only when the task is *inferred* (no `--task`), so a user-given `--model-class ... --task feature-extraction` stays verbatim like USER_TASK. `resolve_task(vit_cfg, model_class="ViTModel")` now returns `image-feature-extraction`; pinned by two new tests.

Generated with Claude Code

src/winml/modelkit/loader/resolution.py
timenick: Fixed in `5d47ce9e`: removed the eager lookup from Stage 1b entirely and let Stage 2 resolve the class uniformly under its `try/except` (+ arch fallback), so no raw `KeyError` can escape. timm routing tests stay green.

Generated with Claude Code

src/winml/modelkit/loader/resolution.py
timenick: Fixed in `5d47ce9e`: replaced with `if source is None: raise RuntimeError(...)`.

Generated with Claude Code

tests/unit/loader/test_resolve_task.py
timenick: Added in `5d47ce9e`, pinned to `image-feature-extraction` (modality-aware, consistent with the detection path), plus `test_user_class_with_explicit_task_preserved_verbatim` for the explicit-`--task` case.

Generated with Claude Code

DingmaomaoBJTU (commented):
## Follow-up review after 5d47ce9e

All five issues from the previous review have been addressed:
- ✅ **Bug**: USER_CLASS modality upgrade — fixed (inferred-task branch now calls `_resolve_task_modality`).
- ✅ **Warning**: Stage 1b raw `KeyError` — fixed (call removed; Stage 2 handles class resolution).
- ✅ **Warning**: `assert source is not None` stripped by `-O` — fixed (`if/raise RuntimeError`).
- ✅ **Warning**: `image-feature-extraction` not pinned in `TASK_SYNONYM_EXTENSIONS` — fixed.
- ✅ **Info**: Missing test for USER_CLASS + vision feature-extraction — fixed (two new tests added).

One new issue introduced by the fix:

### ⚠️ USER_CLASS explicit-task branch normalizes away the user's original task string (inconsistent with USER_TASK)

In the explicit-task sub-branch (line 358–361), `surfaced = normalize_task(task)`. The comment says *"preserve it verbatim (like USER_TASK)"*, but the USER_TASK path (line 381–398) stores `original = task` and returns that, **not** the normalized form. So the two paths are inconsistent:

```python
# USER_TASK: resolve_task(cfg, task="causal-lm") -> TaskResolution(task="causal-lm", ...)
# USER_CLASS: resolve_task(cfg, task="causal-lm", model_class="X") -> TaskResolution(task="text-generation", ...)
```

The new test `test_user_class_with_explicit_task_preserved_verbatim` only checks `task="feature-extraction"` (already canonical, no normalization happens) so it does not catch this divergence.

Fix: use `surfaced = task` (the original user string) instead of `surfaced = opt_task` (normalized), keeping `opt_task = normalize_task(task)` for the Optimum API call only — identical to what USER_TASK does with `original` vs `normalized`.

src/winml/modelkit/loader/resolution.py
```diff
@@ -0,0 +1,461 @@
...
+    model_type = getattr(config, "model_type", None)
+    model_type_norm = model_type.lower().replace("_", "-") if model_type else ""
+    model_id = getattr(config, "_name_or_path", "") or None
+
+    # --- Stage 0: user override (short-circuits detection) ----------------
+    if model_class is not None:
+        if task is not None:
+            # User gave an explicit task: preserve it verbatim (like USER_TASK).
+            opt_task = normalize_task(task)
+            surfaced = opt_task
```
DingmaomaoBJTU: **⚠️ Warning: `surfaced = opt_task` loses the user's original task string, inconsistent with USER_TASK.**

The comment above says *"preserve it verbatim (like USER_TASK)"* but `opt_task = normalize_task(task)` collapses aliases. Setting `surfaced = opt_task` means the returned `TaskResolution.task` is the *normalized* form, not the original user input.

The USER_TASK path (two branches down) does this correctly:
```python
original = task
normalized = normalize_task(task)
# ... class lookup uses `normalized` ...
return TaskResolution(original, to_optimum_task(original), ...)
```

This branch should match:
```python
opt_task = normalize_task(task)  # for Optimum API call only
surfaced = task                  # preserve verbatim, like USER_TASK
```

Practical impact: `resolve_task(cfg, task="causal-lm", model_class="GPT2LMHeadModel")` returns `task="text-generation"` while `resolve_task(cfg, task="causal-lm")` returns `task="causal-lm"`. Callers that downstream-key on the task string (e.g. dataset routing) see different values for the same user input depending on whether `--model-class` was also given.

tests/unit/loader/test_resolve_task.py
```diff
@@ -0,0 +1,147 @@
...
+    r = resolve_task(_cfg("vit", ["ViTModel"]), model_class="ViTModel")
+    assert r.source == TaskSource.USER_CLASS
+    assert r.task == "image-feature-extraction"
+    assert r.optimum_task == "feature-extraction"
+
+
+def test_user_class_with_explicit_task_preserved_verbatim():
+    """USER_CLASS WITH an explicit --task preserves the user's task (no modality upgrade)."""
+    r = resolve_task(
+        _cfg("vit", ["ViTModel"]), model_class="ViTModel", task="feature-extraction"
```
DingmaomaoBJTU, line 148: **ℹ️ Info: test uses a canonical task name — alias normalization is not exercised.**

`task="feature-extraction"` is already canonical; `normalize_task("feature-extraction")` returns `"feature-extraction"` unchanged, so the `surfaced = opt_task` vs `surfaced = task` discrepancy is invisible here.

To pin the verbatim-preservation contract described in the docstring, add a case with an alias:
```python
def test_user_class_with_explicit_task_alias_preserved_verbatim():
    """USER_CLASS with an alias task preserves the user's original string."""
    r = resolve_task(
        _cfg("gpt2", ["GPT2LMHeadModel"]),
        model_class="GPT2LMHeadModel",
        task="causal-lm",  # alias for text-generation
    )
    assert r.task == "causal-lm"  # verbatim, not "text-generation"
    assert r.source == TaskSource.USER_CLASS
```

timenick:
**Correction on item 4** (`image-feature-extraction` synonym pin): reverted in `fddd4254`. `TASK_SYNONYM_EXTENSIONS` is also consulted by `commands.build._validate_task_supported_for_model` — adding the entry moved `image-feature-extraction` from the Optimum-synonym branch (which emits the cross-modality visibility warning) to the WinML-extension branch (silent), breaking `test_silently_accepts_cross_modality_feature_extraction` and silencing a deliberately-added safety warning.

The forward-looking robustness concern is already guarded: `test_task_boundary.py` asserts `to_optimum_task("image-feature-extraction") == "feature-extraction"`, so a future Optimum change would fail CI. Left a `NOTE` in `TASK_SYNONYM_EXTENSIONS` so it isn't re-added. Items 1/2/3/5/6 stand.

Generated with Claude Code

src/winml/modelkit/loader/resolution.py
timenick: Good catch — fixed in `c6e6e8dd`. The explicit-task branch now sets `surfaced = task` (original) and uses `opt_task = normalize_task(task)` only for the Optimum class lookup, matching USER_TASK. So `resolve_task(cfg, task="causal-lm", model_class="X")` now returns `task="causal-lm"` (was `text-generation`).

Generated with Claude Code

tests/unit/loader/test_resolve_task.py
timenick: Added `test_user_class_with_explicit_task_alias_preserved_verbatim` in `c6e6e8dd` — `task="causal-lm"` -> `r.task == "causal-lm"` and `r.optimum_task == "text-generation"`, exercising the alias path the existing test did not.

Generated with Claude Code

src/winml/modelkit/loader/resolution.py
timenick: Correction: reverted the verbatim change (`c6e6e8dd`) in `2260bda7`. On closer look, USER_CLASS normalizing the task (`masked-lm` -> `fill-mask`) is an intentional, tested contract (`test_task_normalized_with_model_class`), distinct from USER_TASK, which preserves the original string to keep WinML modality-aware names (`image-feature-extraction`) un-collapsed. So the USER_TASK / USER_CLASS divergence is by design: keeping the normalize behavior rather than break that test.

Generated with Claude Code

src/winml/modelkit/loader/resolution.py
```diff
@@ -0,0 +1,461 @@
...
+
+    # --- Stage 0: user override (short-circuits detection) ----------------
+    if model_class is not None:
+        if task is not None:
+            # USER_CLASS normalizes the surfaced task (e.g. masked-lm -> fill-mask),
+            # pinned by test_task_normalized_with_model_class. This INTENTIONALLY differs
+            # from USER_TASK, which preserves the original string to keep WinML
+            # modality-aware names (image-feature-extraction) un-collapsed.
+            opt_task = normalize_task(task)
+            surfaced = opt_task
```
DingmaomaoBJTU: **⚠️ Warning: `normalize_task` collapses WinML modality-aware names, breaking downstream routing for `--task image-feature-extraction --model-class X`.**

The comment says this normalization is "intentional", justified by `test_task_normalized_with_model_class` (which pins `masked-lm → fill-mask`). Alias normalization (`masked-lm → fill-mask`) is reasonable — same semantic task, different spelling. But `image-feature-extraction → feature-extraction` is a **semantic downgrade**:

```python
# normalize_task("image-feature-extraction") → "feature-extraction"  (via Optimum's map_from_synonym)
# surfaced = "feature-extraction"
```

Consequence: `resolve_task(vit_cfg, task="image-feature-extraction", model_class="ViTModel")` returns `TaskResolution(task="feature-extraction", ...)`. Downstream — `TASK_DATASET_MAPPING`, `TASK_TO_EVALUATOR`, etc. — receives `"feature-extraction"` and routes to `TextDataset`, not `ImageDataset`. A user who explicitly passes `--task image-feature-extraction` to select the image modality loses that intent when they also pass `--model-class`.

This is distinct from the `masked-lm` case (where normalization doesn't change routing).

**Two options:**
1. Special-case WinML modality-aware names: use `task` verbatim (like USER_TASK) instead of `normalize_task(task)` as `surfaced`, keeping `opt_task = normalize_task(task)` only for the Optimum API call.
2. Accept the collapse as intentional and add an explicit test that documents it:
   ```python
   def test_user_class_image_feature_extraction_collapses_to_canonical():
       r = resolve_task(_cfg("vit", ["ViTModel"]), model_class="ViTModel", task="image-feature-extraction")
       assert r.task == "feature-extraction"  # intentional: normalization collapses modality
   ```
   Without this test, the behavior is undocumented for modality-aware names specifically.

src/winml/modelkit/loader/task.py
```diff
@@ -644,6 +220,10 @@ def normalize_task(task: str) -> str:
 # WinML task-synonym extensions — extend Optimum's ``TasksManager.map_from_synonym``
 # for tasks it does not recognize or mis-maps. Entries here take priority over Optimum.
 TASK_SYNONYM_EXTENSIONS: dict[str, str] = {
+    # NOTE: do NOT add "image-feature-extraction" here. This set is also consulted by
+    # commands.build._validate_task_supported_for_model (its "WinML extension" branch),
+    # so adding it would silence the cross-modality visibility warning. Its Optimum-synonym
+    # collapse to feature-extraction is guarded by tests/unit/loader/test_task_boundary.py.
```
DingmaomaoBJTU, line 226: **ℹ️ Info: `TASK_SYNONYM_EXTENSIONS` has two coupled but semantically distinct roles.**

This comment correctly explains why `image-feature-extraction` must NOT be added here: the dict is also consulted by `commands.build._validate_task_supported_for_model` (line 326) as a bypass-validation allowlist. But this means any future maintainer adding a translation entry (for Optimum API calls) must also reason about whether that entry should bypass the build validator — a non-obvious coupling.

The dual purpose is:
1. **Optimum translation** — `to_optimum_task()` uses it to collapse WinML-specific names.
2. **Build validator bypass** — `build._validate_task_supported_for_model` uses it to skip the supported-task check.

These have different semantics: `next-sentence-prediction` and `mask-generation` legitimately belong in the bypass list (Optimum doesn't know them). `image-feature-extraction` should belong in the translation list (collapse for Optimum calls) but NOT in the bypass list.

Low-priority, but worth a follow-up issue: separate these two concerns — a dedicated `_VALIDATION_BYPASS_TASKS` set in `build.py` vs. `TASK_SYNONYM_EXTENSIONS` purely for translation — so entries in one don't accidentally affect the other.

DingmaomaoBJTU (commented):
## Follow-up review after 275b0221

**Status of previous open items:**
- ✅ **Warning (resolution.py:361)** — Addressed as an intentional design decision, documented in code. The `test_task_normalized_with_model_class` test pins the normalization behavior.
- ✅ **Info (test_resolve_task.py:144)** — Test renamed to `test_user_class_explicit_task_is_not_modality_upgraded`, more accurately describing the behavior.

**One new warning** found in the current state — see inline comment.

src/winml/modelkit/loader/resolution.py
timenick: Fixed in `e65968da` — and I need to walk back my earlier "intentional, tested contract" reply on this thread: that framing was wrong, the collapse was a real bug.

The fix, in the USER_CLASS explicit-task branch:

```python
opt_task = normalize_task(task)
surfaced = _resolve_task_modality(config, opt_task)
```

Canonicalize the alias for the Optimum class lookup, then re-apply modality so a WinML modality-aware name survives. So `--task image-feature-extraction --model-class ViTModel` now surfaces `image-feature-extraction` (→ ImageDataset/evaluator) instead of collapsing to `feature-extraction` (→ TextDataset). `optimum_task` is unchanged — `to_optimum_task(surfaced)` still collapses to `feature-extraction` — so the class lookup and the `optimum_task == to_optimum_task(task)` invariant hold.

This separates the two things `normalize_task` was conflating:
- **(a) alias canonicalization** (`masked-lm → fill-mask`) — kept. `test_task_normalized_with_model_class` stays green because `_resolve_task_modality` is a no-op for non-`feature-extraction` tasks.
- **(b) modality collapse** (`image-feature-extraction → feature-extraction`) — that was the bug; gone now.

Took neither Option 1 (verbatim) nor Option 2 (document the collapse): verbatim re-introduces the masked-lm regression reverted in `2260bda7`, and documenting would ratify the bug. The branch is now consistent with the USER_CLASS *inferred* branch and USER_TASK — adding `--model-class` no longer changes the modality.

Tests: updated `test_user_class_explicit_task_is_not_modality_upgraded` → `..._is_modality_aware` (it was pinning the collapse) and added `test_user_class_explicit_image_feature_extraction_preserved` as the regression pin.

Generated with Claude Code

src/winml/modelkit/loader/task.py
timenick: Agreed — the dual role is a real, non-obvious coupling: `to_optimum_task()` uses this table for Optimum synonym translation, while `build._validate_task_supported_for_model` uses membership as a validate-gate bypass allowlist. An entry added for one purpose silently affects the other — which is exactly what bit the reverted `image-feature-extraction` pin (`fddd4254`). Splitting a dedicated `_VALIDATION_BYPASS_TASKS` set in `build.py` from `TASK_SYNONYM_EXTENSIONS` is the right shape.

Keeping it out of this PR though — it's orthogonal to the modality fix and touches a currently-passing path, so folding the refactor in here would widen the review surface. Tracking it as a follow-up.

Generated with Claude Code

src/winml/modelkit/loader/resolution.py
```diff
@@ -0,0 +1,464 @@
...
+
+    # --- Stage 0: user override (short-circuits detection) ----------------
+    if model_class is not None:
+        if task is not None:
+            # USER_CLASS with an explicit task does two separable things to the surfaced
+            # task: (a) canonicalize the alias for the Optimum class lookup
+            # (masked-lm -> fill-mask) and (b) re-apply modality so a WinML modality-aware
+            # name survives (feature-extraction -> image-feature-extraction for a
+            # pixel_values arch). (b) is a no-op for non-feature-extraction tasks, so (a)
+            # is preserved. Consistent with the inferred branch below and USER_TASK —
```
DingmaomaoBJTU, line 361: **ℹ️ Info: comment overstates consistency with USER_TASK.**

This line says *"Consistent with the inferred branch below and USER_TASK"*, but the two explicit-task paths diverge for plain eature-extraction on a vision model:

`
resolve_task(vit_cfg, task="feature-extraction")                      # USER_TASK  → r.task = "feature-extraction"
resolve_task(vit_cfg, task="feature-extraction", model_class="ViT")   # USER_CLASS → r.task = "image-feature-extraction"
`

Adding --model-class silently upgrades the surfaced task, while USER_TASK keeps the verbatim value. 	est_user_task_preserved_verbatim_no_modality_upgrade (line 52) explicitly pins this divergence.

The comment is accurate about image-feature-extraction not being collapsed (both paths agree there), but the broad "consistent with USER_TASK" claim is misleading for the eature-extraction upgrade case. A more precise phrasing:

> *"(b) is consistent with the inferred branch and with the USER_TASK path for image-feature-extraction (preserves modality-aware names). Note: unlike USER_TASK, USER_CLASS additionally upgrades plain eature-extraction to image-feature-extraction for vision architectures — this is intentional so that adding --model-class improves rather than degrades routing accuracy."*

Low priority since the behavior is already pinned by tests, but the current wording is likely to confuse future readers.

DingmaomaoBJTU (commented):
## Fourth-pass review — commit e65968da

**Previous warning RESOLVED:** 
esolution.py:361 — surfaced = _resolve_task_modality(config, opt_task) is now applied in the USER_CLASS explicit-task branch. The image-feature-extraction modality is preserved when combined with --model-class, backed by two new tests (	est_user_class_explicit_feature_extraction_is_modality_aware, 	est_user_class_explicit_image_feature_extraction_preserved). ✅

**One info finding** remains — see inline comment on 
esolution.py:361.

**Open item from previous pass:**
- 🔵 **Info** (	ask.py:226) — TASK_SYNONYM_EXTENSIONS dual role (Optimum translation + build-validator bypass) is a structural design smell. Low priority; recommend a follow-up issue to separate the concerns.

DingmaomaoBJTU (approved):
LGTM! All bugs and warnings resolved across 4 review rounds. Two remaining info items (misleading comment on resolution.py:361 and TASK_SYNONYM_EXTENSIONS dual-role coupling) are low-priority and can be tracked as follow-up issues.

PR #896, opened 2026-06-16
type: clean mypy errors in inspect, loader, onnx, optim and enforce in CI

xieofxie:
## Summary

Brings four `winml.modelkit` packages to **zero mypy errors** under the strict `[tool.mypy]` config, and wires them into the required CI type-check gate so they cannot regress.

- **inspect, loader, onnx, optim** — now 0 errors under `strict = true`.
- **`.github/workflows/lint.yml`** — adds the four packages to the `Type check (required)` step (12 → 16 packages). New errors in them now block the PR.

Fixes follow the repos typing playbook: narrow/fix at the source first; `cast()` only at genuine untyped third-party boundaries; `Any` only for true dynamic pass-throughs.

## Notable fixes

- **optim** — `BasePipe` is now generic over its config (`BasePipe[ConfigT]`, `ConfigT` bound to `PipeConfig`). This resolves all 6 Liskov `[override]` errors structurally and lets the pre-existing `# type: ignore[override]` in `rewrite.py` be removed. Test-driven `__post_init__` type validators in `registry.py` kept behind a file-level `# mypy: disable-error-code="unreachable"` pragma.
- **inspect** — corrected a stale `process_module` return annotation (cleared 3 `[unreachable]` + a `[return-value]` together); `assert export is not None` matching the existing codebase convention; hf-hub exceptions imported from their defining module `huggingface_hub.errors`.
- **onnx** — declared `SupportedONNXType`s dynamically-built lookup maps as typed `ClassVar`s (fixes the attr-defined + no-any-return cluster); `list[int | None]` for dynamic tensor dims; targeted `[call-arg]` ignores for onnxs pybind11-enum stubs.
- **loader** — boundary `cast()`s at untyped optimum/transformers returns; a typed local for the Case-2 model-class resolution instead of reusing a `type`-typed name.

## Housekeeping

- Normalized line endings back to **LF** on a handful of files that had been flipped to CRLF on Windows (pure line-ending change, no content difference).

## Out of scope

- `models` is left untouched (still has pre-existing errors) for a follow-up pass; it is intentionally **not** added to the required gate.


src/winml/modelkit/optim/pipes/surgery.py
```diff
@@ -314,9 +314,9 @@ def _remove_isnan_in_attention_mask(
             for i, inp in enumerate(node.input):
                 if inp in rewire_map:
                     node.input[i] = rewire_map[inp]
-        for out in model.graph.output:
-            if out.name in rewire_map:
-                out.name = rewire_map[out.name]
+        for graph_out in model.graph.output:
```
DingmaomaoBJTU, line 317: **Missing type parameter (inconsistency with this PR's own refactor).** `BasePipe` was made `Generic[ConfigT]` in this PR, and `ORTFusionPipe`, `ORTGraphPipe`, and `RewritePipe` were all updated to `BasePipe[ConcreteConfig]`. `SurgeryPipe` at line 79 still reads `class SurgeryPipe(BasePipe):` — it should be `class SurgeryPipe(BasePipe[SurgeryPipeConfig]):` for consistency and to get the narrowed types in `build_config`/`process`/`should_process`.

src/winml/modelkit/loader/resolution.py
```diff
@@ -402,7 +408,7 @@ def resolve_task(
         )
 
     # --- Stage 1: detection -----------------------------------------------
-    opt_task: str | None = None
+    opt_task = None
```
DingmaomaoBJTU: **Redundant reassignment.** `opt_task` was just declared `str | None = None` before Stage 0 (line 357), and Stage 0 always returns early when entered, so `opt_task` is guaranteed `None` here. The second `opt_task = None` adds noise. The hoisted comment above explains the intent well — the Stage-1 reset can just be removed.

src/winml/modelkit/inspect/resolver.py
```diff
@@ -269,7 +269,10 @@ def resolve_exporter(
     # Check MODEL_BUILD_CONFIGS for predefined config
     if model_type_normalized in MODEL_BUILD_CONFIGS:
         config: WinMLBuildConfig = MODEL_BUILD_CONFIGS[model_type_normalized]
+        # MODEL_BUILD_CONFIGS entries are HF export configs; export is None only on
+        # the direct-ONNX build path, which never reaches this registry lookup.
         export_config = config.export
+        assert export_config is not None, "MODEL_BUILD_CONFIGS entries require export"
```
DingmaomaoBJTU: **Assert used for type narrowing — fragile in production.** The comment correctly explains that `export_config` is never `None` here by construction, but Python's `-O` flag disables asserts. Prefer an explicit guard: `if export_config is None: raise ValueError('MODEL_BUILD_CONFIGS entries require export config')`. This gives a clear error message in optimized builds and avoids relying on a runtime assert for an invariant of the registry.

src/winml/modelkit/optim/registry.py
```diff
@@ -16,6 +16,13 @@
 - Configuration validation
 """
 
+# The isinstance() guards in each capability's __post_init__ are unreachable per
+# the (statically-typed) field annotations, but are kept as runtime validation:
+# dataclass construction does not enforce field types, and tests in
+# test_registry_cli.py exercise these TypeError paths. Silence [unreachable] here
```
DingmaomaoBJTU: **Module-level mypy directive is broader than needed.** `# mypy: disable-error-code='unreachable'` silences `[unreachable]` for the *entire* `registry.py`. If a genuine unreachable branch is introduced elsewhere in this file in the future, mypy won't catch it. Prefer per-line `# type: ignore[unreachable]` on each specific `isinstance()` guard that triggers the warning, so the suppressor is narrowly scoped.

DingmaomaoBJTU (commented):
Reviewed 24 changed files. Type annotations are generally correct and well-motivated. Found one consistency gap (SurgeryPipe missing new type parameter), one redundant reset, one fragile assert-for-narrowing, and one overly-broad module-level mypy suppressor.

src/winml/modelkit/optim/pipes/surgery.py
xieofxie: Good catch — fixed in c2feb8a. `SurgeryPipe` is now `BasePipe[SurgeryPipeConfig]`. (The original edit existed but a `main` merge had reverted it, along with a loop-var rename in the same file.)

src/winml/modelkit/loader/resolution.py
xieofxie: Fixed in c2feb8a — removed the redundant Stage-1 `opt_task = None`; it now relies on the hoisted declaration.

src/winml/modelkit/inspect/resolver.py
xieofxie: Agreed on the `-O` concern — fixed in c2feb8a. Replaced the assert with an explicit `if export_config is None: raise ValueError(...)` carrying a clear message.

src/winml/modelkit/optim/registry.py
xieofxie: Fixed in c2feb8a — replaced the file-level pragma with per-line `# type: ignore[unreachable]` on the three validator lines (98/125/157), each with a short rationale comment.

src/winml/modelkit/optim/pipes/surgery.py
```diff
@@ -76,7 +76,7 @@ class SurgeryPipeConfig(PipeConfig):
 # =============================================================================
 
 
-class SurgeryPipe(BasePipe):
+class SurgeryPipe(BasePipe[SurgeryPipeConfig]):
```
DingmaomaoBJTU, line 79: **[RESOLVED]** Finding #1 (warning). `SurgeryPipe` now correctly inherits `BasePipe[SurgeryPipeConfig]`, consistent with `ORTFusionPipe`, `ORTGraphPipe`, and `RewritePipe`.

src/winml/modelkit/loader/resolution.py
```diff
@@ -402,7 +408,7 @@ def resolve_task(
         )
 
     # --- Stage 1: detection -----------------------------------------------
-    opt_task: str | None = None
+    # opt_task stays at its hoisted None until a detection sub-stage sets it.
```
DingmaomaoBJTU, line 411: **[RESOLVED]** Finding #2 (info). The redundant `opt_task = None` reset is replaced with a clarifying comment. Clean.

src/winml/modelkit/inspect/resolver.py
```diff
@@ -269,7 +269,14 @@ def resolve_exporter(
     # Check MODEL_BUILD_CONFIGS for predefined config
     if model_type_normalized in MODEL_BUILD_CONFIGS:
         config: WinMLBuildConfig = MODEL_BUILD_CONFIGS[model_type_normalized]
+        # MODEL_BUILD_CONFIGS entries are HF export configs; export is None only on
+        # the direct-ONNX build path, which never reaches this registry lookup.
         export_config = config.export
```
DingmaomaoBJTU, line 274: **[RESOLVED]** Finding #3 (info). `assert export_config is not None` replaced with an explicit `if export_config is None: raise ValueError(...)` that includes the model_type in the message and survives `python -O`.

src/winml/modelkit/optim/registry.py
```diff
@@ -95,7 +95,9 @@ class BoolCapability(CapabilityDef):
     def __post_init__(self) -> None:
         """Validate boolean default."""
         if not isinstance(self.default, bool):
-            msg = f"BoolCapability '{self.name}' must have bool default, got {type(self.default)}"
+            # Unreachable per the typed field, but kept as runtime validation:
```
DingmaomaoBJTU, line 98: **[RESOLVED]** Finding #4 (info). The file-level `# mypy: disable-error-code='unreachable'` pragma is removed. Each of the three `isinstance()` runtime guards in `BoolCapability`, `IntCapability`, and `ChoiceCapability` now carries its own `# type: ignore[unreachable]` with an explanatory comment. Perfectly scoped.

DingmaomaoBJTU (commented):
Follow-up review: all four findings from the previous pass are fully addressed in commit c2feb8a. No new issues found.

DingmaomaoBJTU (approved):
LGTM! All 4 findings resolved in c2feb8a: SurgeryPipe correctly parameterized, redundant Stage-1 reset replaced with comment, assert-for-narrowing replaced with explicit if/raise ValueError, and module-level mypy pragma replaced with three narrowly-scoped per-line ignores.

PR #901, opened 2026-06-16
Remove dead reference to inference.onnx_config.* subpackage in HTP config generator

### LINKED ISSUE #859: Remove dead reference to inference.onnx_config.* subpackage in HTP config generator
## Context

\`src/winml/modelkit/export/htp/config_generator.py\` references a \`winml.modelkit.inference.onnx_config\` subpackage that does not exist in the repo:

- \`config_generator.py:206\` — \`from ...inference.onnx_config.input_generator import InputSpecGenerator\`
- \`config_generator.py:268\` — \`from ...inference.onnx_config.patterns import TASK_TO_OUTPUTS\`

Both imports are inside \`try/except Exception\` blocks that fall through to empty specs / generic defaults, so the missing modules are masked at runtime — the broken branches are silently skipped.

Tracked from PR review (typecheck cleanup PR) — silenced for the type-check pass with \`# type: ignore[import-not-found]\` while we decide whether to:

1. Reintroduce \`inference.onnx_config.{input_generator,patterns}\` (if the integration was meant to land in a follow-up that never did), or
2. Delete the dead try-blocks and let the function unconditionally return its fallback (\`{}\` for input specs, generic outputs).

## Suggested action

Option 2 is the pragmatic cleanup unless there's a known plan to ship the \`inference.onnx_config\` subpackage. Option 1 needs design context this review doesn't have.

## Reference

- Skill: \`modelkit-typechecking\` — Pattern "Surface every \`# type: ignore[import-not-found]\`"
- Related: #824 (similar dead-module pattern in \`core/unified_optimizer.py\` and \`core/universal_hierarchy_exporter.py\`)

DingmaomaoBJTU:
## Summary

Removes dead `try`-blocks in `ExportConfigGenerator._generate_input_specs` and `_get_output_names` that referenced the non-existent `winml.modelkit.inference.onnx_config` subpackage.

## Changes

- `_generate_input_specs` — removed the unreachable import of `InputSpecGenerator` and all associated logic; the function now unconditionally returns `{}` (callers infer inputs from the ONNX graph directly).
- `_get_output_names` — removed the unreachable import of `TASK_TO_OUTPUTS`; the function now directly applies the task-based heuristic that was previously the fallback path.

Both imports were inside `try/except Exception` blocks that silently swallowed `ImportError`, so runtime behavior is unchanged.

Fixes #859


xieofxie:
Should we just remove the whole file? https://github.com/search?q=repo%3Amicrosoft%2Fwinml-cli%20config_generator&type=code it is not referenced any where and no one will maintain it

xieofxie (approved):

PR #582, opened 2026-05-12
Feat/sam3 onnx hub support (issue #324)

AChenreddy24:
Adds SAM 3 support and introduces a generic Hub-hosted ONNX input form (`<org>/<repo>/<path>.onnx`) that downloads pre-exported ONNX files from HuggingFace via `huggingface_hub`. SAM 3 is the first consumer.

The standard SAM 2-style export route is blocked: `optimum-onnx` pins `transformers<4.58`, but `facebook/sam3` requires `transformers>=5`. This PR uses the pre-exported `onnx-community/sam3-tracker-ONNX` artifacts via the existing Scenario D pipeline.

## Changes

- New Hub-ONNX URI resolver (`loader/onnx_hub.py`)
- Wired into `wmk config`, `wmk build`, `wmk inspect`, `wmk run`, `wmk serve`, `wmk perf`, `wmk eval`
- SAM 3 catalog entries (encoder + decoder) in the new schema
- Fixed `is_quantized_onnx` to detect QOperator format (`ConvInteger`, `MatMulInteger`, `QLinear*`) — was QDQ-only
- Fixed `is_pre_quantized` build branch to truly skip the optimize stage (previously crashed on QOperator models)

## Tests

- 4649 unit tests pass (+12 new regression tests)
- 6 integration tests pass (live HF download)
- Built decoder is byte-identical to input (verified numerically)
- Ruff lint + format clean

## Limitations

- SAM 3 vision encoder requires NPU EP (no CPU `ConvInteger` kernel in `onnxruntime-windowsml`); decoder runs on CPU + NPU.

PR #902, opened 2026-06-16
feat(e2e-eval): support pre-exported ONNX models

xieofxie:
## Summary

Adds support for evaluating models that ship a **pre-exported ONNX** in the e2e_eval harness, including HF refs that point at a specific `.onnx` file under a repo. This supports models such as the PaddleOCR `*_onnx` repos, where the native HF export path is blocked by the pinned `transformers` / `optimum-onnx` dependency stack.

When a registry entry points at an ONNX file, the harness downloads that file from the HF repo and feeds the local ONNX path to `winml config`, `winml build`, and `winml perf`, bypassing HF-to-ONNX export.

## Changes

- `run_eval.py`: resolves registry ONNX files before build/perf, preserves ONNX metadata through `--hf-model` task overrides, and resolves direct-ONNX build artifacts deterministically.
- Analyzer metadata: removes the analyzer-only opset floor so ONNX models are recorded with their actual opset instead of being rejected before downstream build/optimize handling.
- Registry data: updates PaddleOCR curated/all entries to use the ONNX file refs and explicit task metadata.
- `build_registry.py`: preserves ONNX metadata during registry rebuilds, adds `--recheck-downloads` to refresh download counts before ordering, preserves existing download counts if a recheck lookup fails, and normalizes nested `.onnx` HF refs to `org/model` only for HF API calls while keeping the full registry `hf_id`.
- Tests: adds/updates coverage for ONNX eval wiring, task-required cache validation, analyzer opset metadata recording, registry ONNX API normalization, and download rechecks.

## Validation

- `uv run pytest tests/unit/eval/test_run_eval_script.py`
- `uv run pytest tests/unit/commands/test_build.py -q`
- `uv run pytest tests/unit/analyze/models/test_onnx_model.py -q`
- `uv run pytest tests/unit/test_build_registry.py -q`
- End-to-end PaddleOCR ONNX perf run on CPU completed successfully for the detector model.


scripts/e2e_eval/run_eval.py
```diff
@@ -477,6 +478,73 @@
...
+def _ensure_min_opset(onnx_path: str, model_dir: Path) -> str:
+    """Upgrade an ONNX file to winml's minimum opset if it ships below it.
+
+    Some pre-exported ONNX models (e.g. PaddleOCR's ``inference.onnx``) ship at
+    opset 11. onnxruntime can run them, but winml's optimize stage rejects
+    opset < 12. Convert to :data:`_OPSET_UPGRADE_TARGET` and write the result
+    into ``model_dir`` so config/build consume the upgraded graph. Models already
+    at/above the minimum are returned unchanged (no needless rewrite).
+    """
+    import onnx
```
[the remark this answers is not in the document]
xieofxie: Fixed in 82f3bd483 — now uses a single consistent `import onnx` with attribute access (`onnx.version_converter` / `onnx.helper` / `onnx.TensorProto`), no longer mixing `import` and `from import`.

tests/unit/eval/test_run_eval_script.py
```diff
@@ -170,3 +172,119 @@
     def test_other_ep_omits_no_quant(self, run_eval, tmp_path):
         calls = self._invoke(run_eval, "dml", tmp_path)
         assert all("--no-quant" not in args for args in calls)
+
+
+def _write_onnx(path: Path, opset: int) -> None:
+    """Write a minimal single-node ONNX model at the given opset."""
+    import onnx
```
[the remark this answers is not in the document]
xieofxie: Fixed in 82f3bd483 — now uses a single consistent `import onnx` with attribute access (`onnx.version_converter` / `onnx.helper` / `onnx.TensorProto`), no longer mixing `import` and `from import`.

tests/unit/eval/test_run_eval_script.py
```diff
@@ -170,3 +172,119 @@
...
+    graph = helper.make_graph([helper.make_node("Identity", ["x"], ["y"])], "g", [x], [y])
+    model = helper.make_model(graph, opset_imports=[helper.make_opsetid("", opset)])
+    onnx.save(model, str(path))
+
+
+class TestEnsureMinOpset:
+    """``_ensure_min_opset`` upgrades sub-minimum ONNX to winml's opset floor."""
+
+    def test_below_minimum_is_upgraded(self, run_eval, tmp_path):
+        import onnx
```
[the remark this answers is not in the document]
xieofxie: Fixed in 82f3bd483 — now uses a single consistent `import onnx` with attribute access (`onnx.version_converter` / `onnx.helper` / `onnx.TensorProto`), no longer mixing `import` and `from import`.

xieofxie:
need to update this in favor of https://github.com/microsoft/winml-cli/pull/582

src/winml/modelkit/analyze/models/onnx_model.py
```diff
@@ -81,9 +81,9 @@ class ONNXModel(BaseModel):
     @field_validator("opset_version")
     @classmethod
     def validate_opset_version(cls, v: int) -> int:
-        """Validate opset version is >= 12 per assumption."""
-        if v < 12:
-            raise ValueError(f"Opset version {v} < 12 (minimum supported version)")
+        """Validate opset version is >= 11 per assumption."""
+        if v < 11:
```
xieofxie: for the paddle onnx

xieofxie: in winml-cli\src\winml\modelkit\export\config.py, we also guard against < 11

scripts/e2e_eval/build_registry.py
```diff
@@ -429,7 +440,22 @@ def build_registry(
...
+    if recheck_downloads:
+        safe_print("\n  Rechecking downloads for all entries...")
+        download_cache: dict[str, int] = {}
+        for e in all_entries:
+            hf_id = e.get("hf_id")
+            if not hf_id:
+                continue
+            if hf_id not in download_cache:
+                download_cache[hf_id] = get_model_metadata(hf_id).get("downloads", 0)
+            e["downloads"] = download_cache[hf_id]
```
DingmaomaoBJTU: `get_model_metadata()` returns `downloads: 0` on *any* failure — ImportError, network error, rate limit, private/removed repo (the `except Exception` fallback). Because `--recheck-downloads` overwrites `e["downloads"]` unconditionally here, a single transient HF API hiccup silently zeroes a model's real download count and drops it to the bottom of its task ranking, and the previously-good value is gone. Running this across all ~440 entries makes hitting at least one transient failure fairly likely, so a rerun can quietly reshuffle order.

Suggest only overwriting on a successful fetch — e.g. have `get_model_metadata` distinguish a failure (return `None`) from a genuine `0`, and keep the existing `e["downloads"]` when the lookup fails.

scripts/e2e_eval/build_registry.py
```diff
@@ -429,7 +440,22 @@ def build_registry(
...
+        safe_print(f"  Rechecked downloads for {len(download_cache)} model IDs")
+
+    # Phase 4: Rank within each task by downloads (descending).
```
DingmaomaoBJTU, line 477: Minor: bumping Rank to Phase 4 orphaned the comment at line 370 — `preserved_entry.pop("order", None)  # Phase 3 will reassign` now points at the new download-recheck phase, not the ranking that actually reassigns `order`. Update it to "Phase 4 will reassign".

src/winml/modelkit/analyze/models/onnx_model.py
```diff
@@ -81,9 +81,9 @@ class ONNXModel(BaseModel):
...
+            raise ValueError(f"Opset version {v} < 11 (minimum supported version)")
```
DingmaomaoBJTU: What's the basis for lowering the floor to 11? The PR description says sub-minimum opsets get "upgraded before WinML optimize," but there's no upgrade logic in this PR — this just drops the validator floor by one so the opset-11 PaddleOCR models pass. Could you note why 11 is safe downstream where 12 was previously required? The `>= 11 per assumption` docstring doesn't really justify it, and it'd be good to confirm this isn't masking a real minimum somewhere in the optimize/build path.

DingmaomaoBJTU (commented):
Reviewed the ONNX-eval enablement. The code changes themselves are small and mostly reasonable, but the PR description is significantly out of sync with what actually landed — worth reconciling before merge.

**Description vs. diff mismatch**
The "Changes" section describes functionality that isn't in this diff:
- `run_eval.py` is said to "resolve registry ONNX files before build/perf, preserve ONNX metadata through `--hf-model`, upgrade sub-minimum opsets, and resolve direct-ONNX build artifacts." The actual `run_eval.py` diff is formatting-only — there is no `onnx_file` handling anywhere in the file.
- The registry is said to carry "ONNX file metadata" / `onnx_file`, but there is no `onnx_file` key. The real mechanism encodes the ONNX path directly in `hf_id` (e.g. `PaddlePaddle/PP-OCRv6_medium_det_onnx/inference.onnx`) and leans on winml's existing `hub_onnx` resolution of `org/repo/path/file.onnx` refs. That's a cleaner design, but the description (and the earlier `load_curated_entries` passthrough commit) describe the reverted approach.

It looks like the "revert and update" commit changed the approach. Please update the description so reviewers know the ONNX support rides on winml's existing nested-onnx resolution plus the registry/opset/cache tweaks here — as written it's misleading about where the behavior lives and what's actually tested.

**models_all.json churn (FYI)**
The ~1.6k-line diff there is almost entirely `downloads`/`order` refresh from `--recheck-downloads` (440/443 entries changed downloads, 315 reordered). Only the two PaddleOCR entries actually change identity (v5 → v6 `_onnx`). Verified: no entries added/dropped and no duplicate keys.

Inline notes below.

src/winml/modelkit/analyze/models/onnx_model.py
xieofxie: I don't know why 12 is chosen

src/winml/modelkit/analyze/models/onnx_model.py
xieofxie: All versions of ONNX Runtime support ONNX opsets from ONNX v1.2.1+ (opset version 7 and higher).

We may also lower to 7

src/winml/modelkit/analyze/models/onnx_model.py
xieofxie: set when we use rules files as a whole zip of onnx version. now we use separate rule file for each op, so it is no longer valid now

scripts/e2e_eval/build_registry.py
xieofxie: Fixed in 92909f93. `--recheck-downloads` now uses a metadata fetch that can return failure separately from a real zero download count; on failure it leaves the existing `downloads` value unchanged and logs a warning instead of overwriting with 0.

scripts/e2e_eval/build_registry.py
xieofxie: Fixed in 92909f93. Updated the preserved-entry comment to point at Phase 4 ranking after the download recheck phase.

src/winml/modelkit/analyze/models/onnx_model.py
DingmaomaoBJTU: Lowering the floor from 12 to 11 is what lets the opset-11 PaddleOCR ONNX models pass analyze, but can we capture the *why* here? The docstring still just says "per assumption".

1. Is 11 the actual floor the optimize/QDQ pipeline supports, or just the lowest we happen to need right now?
2. The PR description says it "upgrades sub-minimum opsets before WinML optimize", but I don't see that logic in this diff, so opset-11 models now flow through un-upgraded. Have these been validated end-to-end through `optimize`, not just this validator?

src/winml/modelkit/commands/build.py
```diff
@@ -1054,13 +1054,9 @@ def _patch_device(cfg: WinMLBuildConfig) -> None:
                     model or "random-init",
                     cache_dir=get_cache_dir(),
                 )
-                if not task:
-                    raise click.UsageError(
-                        "--use-cache requires loader.task in config. "
-                        "The cache key is prefixed by the task abbreviation."
-                    )
+                task_abbrev = get_task_abbrev(task) if task else "unknown-task"
```
DingmaomaoBJTU: This removes the hard error for taskless `--use-cache` for *every* config, not just ONNX inputs. Previously a config missing `loader.task` errored out; now any such config silently builds under the `unknown-task_` prefix.

For pre-exported ONNX that's exactly right (there's no task). But is the relaxation meant to apply universally? A transformers config that simply forgot to set `task` would now proceed silently instead of failing fast. If it's intended, a one-line comment noting the `unknown-task` sentinel is deliberate, and safe because the model dir + `config.generate_cache_key()` already disambiguate, would help.

scripts/e2e_eval/testsets/models_curated.json
```diff
@@ -30,6 +30,6 @@
   {"hf_id": "timm/repghostnet_200.in1k", "group": "ISV", "priority": "P1"},
   {"hf_id": "openai/clip-vit-large-patch14-336", "group": "ISV", "priority": "P1"},
   {"hf_id": "timm/mobilenetv3_small_100.lamb_in1k", "group": "ISV", "priority": "P1"},
-  {"hf_id": "PaddlePaddle/PP-OCRv5_server_det", "group": "ISV", "priority": "P1"},
-  {"hf_id": "PaddlePaddle/PP-OCRv5_server_rec", "group": "ISV", "priority": "P1"}
+  {"hf_id": "PaddlePaddle/PP-OCRv6_medium_det_onnx/inference.onnx", "group": "ISV", "priority": "P1"},
+  {"hf_id": "PaddlePaddle/PP-OCRv6_medium_rec_onnx/inference.onnx", "group": "ISV", "priority": "P1"}
```
DingmaomaoBJTU: Worth flagging that this isn't only an "export format" swap, it changes the model itself: PP-OCRv5 **server** to PP-OCRv6 **medium** for both det and rec. So the ISV coverage shifts both a version (v5 to v6) and a size tier (server to medium).

Is that intended, or did we just want the pre-exported ONNX of the same v5 server models? If a v5-server `_onnx` repo exists, using it would keep this eval comparable to what was there before.

xieofxie: yes, intentional to upgrade

src/winml/modelkit/commands/build.py
xieofxie: good point! as a new thought, we should revert this change and add task to models_curated.json

src/winml/modelkit/analyze/models/onnx_model.py
xieofxie: we could totally remove this check

xieofxie: Fixed in 7527b4ca. Removed the analyzer opset floor entirely, so the analyzer records the model opset instead of enforcing an arbitrary 11/12 minimum. I also updated the PR body to remove the stale “upgrades sub-minimum opsets” wording.

src/winml/modelkit/commands/build.py
xieofxie: Fixed in 7527b4ca. Reverted the taskless `--use-cache` fallback so missing `loader.task` fails fast again, and added explicit `task` metadata to the PaddleOCR ONNX entries in `models_curated.json`.

DingmaomaoBJTU (approved):
Approving. The follow-up commit (`7527b4c`) cleanly addresses all three review threads:

- **onnx_model.py** — the analyzer opset floor is removed entirely, so the model records its opset without gating (a reasonable separation of concerns: analyze records, build enforces). Stale "upgrades sub-minimum opsets" wording dropped from the PR body.
- **build.py** — reverted to match `main` exactly, restoring the fail-fast on missing `loader.task`. Safe for the ONNX eval because `run_eval.py` passes `--task` from the registry entry, and the PaddleOCR entries now carry explicit `task` metadata.
- **models_curated.json** — explicit `task: image-to-text` added, keeping curated source and the built registry consistent across rebuilds. The v5-server → v6-medium swap is confirmed intentional.

Verified locally against the PR head: `test_onnx_model.py` + `test_build.py` = 127 passed, and `ruff check` is clean on the changed files.

Minor non-blocking note: removing the opset floor outright (vs. lowering it) means genuinely ancient opsets now surface only later in optimize rather than at analysis time — acceptable given the layering.

PR #903, opened 2026-06-16
feat(inspect): render composite pipeline structure (pipeline-led)

timenick:
## Summary

`winml inspect` now surfaces the **composite pipeline structure** for composite models (encoder-decoder seq2seq, etc.). Previously a composite was shown as its single exported component — e.g. a bart decoder rendered as `Task: text2text-generation` with a dangling `encoder_hidden_states` input ("non-runnable half"), with no hint that the model serves higher-level pipelines.

**Before**
```
Task            text2text-generation (via tasks-manager)
```

**After** (pipeline-led)
```
Task            summarization · table-question-answering [composite]
Export          encoder: feature-extraction, decoder: text2text-generation (via tasks-manager)
```
plus a dedicated **Composite Pipeline** panel listing the components and how to build a specific pipeline.

## Behavior

- **Scoped to auto-detected composites.** Gated on the resolver's detected composite (`TaskResolution.composite`), so only models whose *resolved* task bridges to a composite show the view. A CLIP inspected for `feature-extraction` (a runnable single encoder) and all non-composites are unchanged.
- **Architecture-agnostic.** The pipeline-task list is derived from the live `COMPOSITE_MODEL_REGISTRY` (new `loader.composite_pipeline_tasks`) — no hardcoded model/task names. bart → `summarization · table-question-answering`, marian → `translation`, qwen3 → `text-generation`, etc.
- **JSON contract preserved.** `task` stays the granular machine task (`text2text-generation`); `pipeline_tasks` + `composite` are additive.

## API

- `loader.composite_pipeline_tasks(model_type) -> list[str]`
- `inspect.CompositeInfo`, `inspect.resolve_composite_info(...)`

## Notes

- Explicit `--task` / `--model-class` yields no composite view (pins `composite=None`) — the pipeline-led view is intentionally scoped to auto-detection.
- `blip` (infers `image-text-to-text`, registers under `image-to-text`) is not reached by the seq2seq detection bridge today; out of scope here.


src/winml/modelkit/inspect/resolver.py
```diff
@@ -581,6 +582,31 @@ def resolve_cache(model_id: str) -> CacheInfo:
...
+    Returns ``None`` unless the model's resolved task bridges to a composite — i.e.
+    ``detected_components`` (from :attr:`TaskResolution.composite`) is set. This scopes
+    the composite view to genuine multi-component / non-runnable-half exports (a seq2seq
+    decoder, etc.) and avoids flagging every model_type that merely *could* serve a
+    composite pipeline (e.g. a CLIP inspected for plain feature-extraction).
+
+    ``pipeline_tasks`` (the higher-level pipelines the model_type serves) come from the
+    live composite registry; ``components`` is the resolver's detected breakdown.
+    """
+    if not detected_components:
```
DingmaomaoBJTU: **Bug: falsy check conflates `None` with empty dict, and does not guard against `pipeline_tasks=[]`.**

`if not detected_components:` is `True` for both `None` *and* `{}`. The intent is to guard on `None` (not populated by auto-detect); use `if detected_components is None:` instead.

More critically, even when `detected_components` is non-empty, `composite_pipeline_tasks(model_type)` may return `[]` if the `issubclass(cls, WinMLCompositeModel)` filter in `resolution.py` excludes the registered class. A `CompositeInfo(pipeline_tasks=[], components=...)` then reaches the formatter, which renders a broken Task row (` [composite]` with no pipeline name) and an empty *Serves pipelines:* line. Guard here:

```python
pipeline_tasks = composite_pipeline_tasks(model_type)
if not pipeline_tasks:
    return None  # registry divergence; don't surface a broken panel
return CompositeInfo(pipeline_tasks=pipeline_tasks, components=dict(detected_components))
```

src/winml/modelkit/loader/resolution.py
```diff
@@ -286,6 +286,28 @@ def resolve_composite(model_type: str, task: str) -> CompositeComponents | None:
...
+    """
+    import winml.modelkit.models.hf  # noqa: F401  # trigger composite registrations
+
+    from ..models.winml import WinMLCompositeModel
+    from ..models.winml.composite_model import COMPOSITE_MODEL_REGISTRY
+
+    return sorted(
+        task
+        for (mt, task), cls in COMPOSITE_MODEL_REGISTRY.items()
+        if mt == model_type and issubclass(cls, WinMLCompositeModel)
```
DingmaomaoBJTU: **Bug: `issubclass(WinMLCompositeModel)` filter is inconsistent with `resolve_composite()` and can silently produce `pipeline_tasks=[]`.**

`resolve_composite()` uses the same `COMPOSITE_MODEL_REGISTRY` without any subclass filter. If a composite is registered under a class not subclassing `WinMLCompositeModel`, `resolve_composite()` finds it (setting `TaskResolution.composite`) but `composite_pipeline_tasks()` silently excludes it and returns `[]`. That empty list propagates to `CompositeInfo.pipeline_tasks` and causes the formatter to render a Task row of ` [composite]` with no pipeline name (formatter.py line 267).

Either remove the filter here (trust the registry), or add the same filter to `resolve_composite()` so both functions agree on what constitutes a composite.

src/winml/modelkit/inspect/formatter.py
```diff
@@ -221,13 +260,28 @@ def output_table(console: Console, result: InspectResult, verbose: bool = False)
     model_table.add_column("Value")
 
     model_table.add_row("Model Type", result.model_type)
-    model_table.add_row("Task", f"{result.task} [dim](via {result.task_source})[/dim]")
+    if result.composite:
+        # Variant 1 (Pipeline-led): the Task row surfaces the pipeline tasks the
+        # model_type serves, and an Export row shows the granular component tasks.
+        pipelines = " · ".join(result.composite.pipeline_tasks)
+        model_table.add_row("Task", f"{pipelines} [magenta]\\[composite][/magenta]")
```
DingmaomaoBJTU, line 270: **Bug: no guard against `pipeline_tasks` being empty — renders a broken Task row.**

If `result.composite.pipeline_tasks` is `[]`, `pipelines` is `''` and the row becomes ` [composite]` with no pipeline name — visually confusing. The root fix belongs in `resolve_composite_info` (guard on empty `pipeline_tasks` before constructing `CompositeInfo`), but a local guard here makes the formatter robust in isolation:

```python
if result.composite.pipeline_tasks:
    pipelines = " · ".join(result.composite.pipeline_tasks)
    model_table.add_row("Task", f"{pipelines} [magenta]\\[composite][/magenta]")
    ...
else:
    model_table.add_row("Task", f"{result.task} [dim](via {result.task_source})[/dim]")
```

src/winml/modelkit/loader/resolution.py
```diff
@@ -286,6 +286,28 @@ def resolve_composite(model_type: str, task: str) -> CompositeComponents | None:
...
+    """Pipeline (composite) tasks a model_type can serve, sorted; ``[]`` for non-composites.
+
+    Registry-driven and architecture-agnostic (e.g. ``bart`` ->
+    ``["summarization", "table-question-answering"]``, ``marian`` -> ``["translation"]``,
+    ``qwen3`` -> ``["text-generation"]``). Surfaced by ``winml inspect`` to show which
+    higher-level pipelines a composite serves. The per-checkpoint pipeline is
+    config-indistinguishable, so the list is sorted (deterministic, model-id-independent) —
+    the order must not imply inspect knows which pipeline a given checkpoint is.
+    """
+    import winml.modelkit.models.hf  # noqa: F401  # trigger composite registrations
```
DingmaomaoBJTU: **Warning: fragile side-effect import as the sole registry population mechanism.**

`import winml.modelkit.models.hf` is the only thing ensuring the registry is populated before iteration. If the registrations ever move to a different module, this silently returns `[]` for every model_type with no error or diagnostic. Consider asserting the registry is non-empty after the import, or explicitly documenting in `COMPOSITE_MODEL_REGISTRY` that all callers must import `winml.modelkit.models.hf` first.

tests/unit/inspect/test_composite_rendering.py
```diff
@@ -0,0 +1,163 @@
...
+    """Gate contract: an explicit --task pins composite=None (USER_TASK), so no composite
+    view — even for a genuine seq2seq. Pins the design the inline comment documents, so a
+    revert to a registry-fallback (which would show it) can't slip through silently.
+    """
+    resolution = resolve_task(_bart_config(), task="text2text-generation")
+    assert resolution.composite is None
+    assert resolve_composite_info("bart", resolution.composite) is None
+    out = _render(_make_result(composite=None, task="text2text-generation"))
+    assert "Composite Pipeline" not in out
+    assert "[composite]" not in out
```
DingmaomaoBJTU, line 194: **Missing test: `pipeline_tasks=[]` edge case (registry divergence) is not covered.**

There is no test for the scenario where `detected_components` is non-empty but `composite_pipeline_tasks()` returns `[]`. This is the exact path that causes the broken formatter output described above. A test pinning the guard:

```python
def test_resolve_composite_info_returns_none_when_pipeline_tasks_empty(monkeypatch):
    monkeypatch.setattr(
        "winml.modelkit.inspect.resolver.composite_pipeline_tasks",
        lambda model_type: [],
    )
    assert resolve_composite_info("bart", {"encoder": "feature-extraction"}) is None
```

DingmaomaoBJTU (commented):
Good overall architecture — the pipeline-led composite view is clearly scoped (gated on resolver detection, not registry presence alone) and the additive JSON design is clean. I found a few correctness issues worth fixing before merge.

src/winml/modelkit/inspect/resolver.py
timenick: Fixed in `97c6c531` — added the registry-divergence guard:

```python
pipeline_tasks = composite_pipeline_tasks(model_type)
if not pipeline_tasks:
    return None  # don't surface a broken "[composite]" row
```

On `if not detected_components` vs `is None`: I kept `not` deliberately — it returns `None` for both `None` *and* the (never-produced) empty dict, which is the more defensive read of "no usable composite". `_composite_components_for_task` only ever yields a non-empty dict or `None`, so behaviour is identical; `not` just also covers the impossible `{}`.

Generated with Claude Code

src/winml/modelkit/loader/resolution.py
timenick: Looked into this — the divergence isn't reachable on the gate path. `resolve_composite_info` gates on `detected_components` = `TaskResolution.composite`, which is set by `_composite_components_for_task` — and that applies the **same** `issubclass(cls, WinMLCompositeModel)` filter as `composite_pipeline_tasks`. So whenever the gate is set, `composite_pipeline_tasks` (same filter) returns non-empty; the scenario where `resolve_composite` (no filter) finds it but `composite_pipeline_tasks` doesn't isn't on this path.

I left `resolve_composite` as-is on purpose: removing the filter from `composite_pipeline_tasks` would wrongly include generic, non-composite registrations (e.g. qwen3's `WinMLModelForGenericTask`), and adding the filter to `resolve_composite` would change behaviour for its other (explicit-task) callers. The new empty-`pipeline_tasks` guard makes this robust regardless.

Generated with Claude Code

src/winml/modelkit/inspect/formatter.py
timenick: Fixed at the root in `resolve_composite_info` (per your note that the root fix belongs there) — with the guard, `result.composite` is only ever non-`None` when `pipeline_tasks` is non-empty, so a formatter-side guard would be redundant. Happy to add a defensive one too if you'd prefer belt-and-suspenders.

Generated with Claude Code

src/winml/modelkit/loader/resolution.py
timenick: This mirrors the existing pattern — `resolve_composite` and `_composite_components_for_task` use the same `import winml.modelkit.models.hf` trigger — so a one-off assert here would be inconsistent with them. And the new empty-`pipeline_tasks` guard now degrades gracefully (returns `None`, no broken panel) if the registry ever comes up empty, which covers the silent-`[]` failure mode you flagged.

Generated with Claude Code

tests/unit/inspect/test_composite_rendering.py
timenick: Added `test_resolve_composite_info_none_when_pipeline_tasks_empty` (monkeypatches `composite_pipeline_tasks` to return `[]`, asserts `resolve_composite_info` returns `None`).

Generated with Claude Code

src/winml/modelkit/inspect/resolver.py
```diff
@@ -581,6 +582,39 @@ def resolve_cache(model_id: str) -> CacheInfo:
...
```
DingmaomaoBJTU: **Bug (partially resolved, semantic issue remains): `if not detected_components:` still conflates `None` with `{}`.**

The `pipeline_tasks=[]` downstream guard was correctly added below (lines 604–610) — thank you. However, the first guard on line 599 is unchanged: `if not detected_components:` is truthy for both `None` (not set) **and** `{}` (set to an empty dict). If the resolver ever sets `TaskResolution.composite = {}` for a composite model (e.g. a future composite with no sub-model breakdown), this silently suppresses the composite view even though detection succeeded.

The intent here is clearly *"was the composite path not taken?"*, which is `None`. Please change to:
```python
if detected_components is None:
    return None
```

src/winml/modelkit/loader/resolution.py
DingmaomaoBJTU: **Bug (unresolved): `issubclass(cls, WinMLCompositeModel)` filter is still present and still inconsistent with `resolve_composite()`.**

This is the root cause of Finding #2 from the previous review. `resolve_composite()` (defined ~10 lines above) does `COMPOSITE_MODEL_REGISTRY.get((model_type, task))` with **no** subclass filter. If a composite is registered under a class that does not subclass `WinMLCompositeModel`, `resolve_composite()` returns its config (setting `TaskResolution.composite` non-None) but `composite_pipeline_tasks()` silently excludes it here, returning `[]`.

The new guard in `resolve_composite_info` (line 605) now prevents a broken panel from rendering, but the two functions still disagree on what constitutes a composite — a `TaskResolution.composite` is set but `composite_pipeline_tasks` returns `[]`. This means the composite view is silently suppressed for valid composites that don't happen to subclass `WinMLCompositeModel`, with no diagnostic.

Fix: remove the `issubclass` filter here (trust the registry, same as `resolve_composite()`), or add the same filter to `resolve_composite()` so both agree.

src/winml/modelkit/inspect/formatter.py
DingmaomaoBJTU, line 270: **Bug (unresolved): formatter still has no guard against `pipeline_tasks=[]`.**

The upstream guard in `resolve_composite_info` now prevents `CompositeInfo(pipeline_tasks=[])` from being produced via the normal path. However, `output_table` is a public function that accepts any `InspectResult`. Any caller that constructs `InspectResult(composite=CompositeInfo(pipeline_tasks=[], ...))` directly (e.g. in tests, or a future code path) will still get a broken Task row of `' [composite]'` with no pipeline name.

The formatter should be independently robust:
```python
if result.composite and result.composite.pipeline_tasks:
    pipelines = " · ".join(result.composite.pipeline_tasks)
    model_table.add_row("Task", f"{pipelines} [magenta]\\[composite][/magenta]")
    ...
else:
    model_table.add_row("Task", f"{result.task} [dim](via {result.task_source})[/dim]")
```
Same fix applies to the `_output_composite_panel` call below (line 282) — it should also check `result.composite.pipeline_tasks`.

src/winml/modelkit/loader/resolution.py
DingmaomaoBJTU: **Warning (unresolved): bare side-effect import is still the sole registry population mechanism.**

This is unchanged from Finding #4. `import winml.modelkit.models.hf` is the only thing ensuring `COMPOSITE_MODEL_REGISTRY` is populated before the comprehension on lines 304–308 runs. If composite registrations ever move to a different module path, every call to `composite_pipeline_tasks()` silently returns `[]` for all model types — no error, no log, just an empty list that (now) suppresses the composite panel.

A minimal improvement: assert the registry is non-empty after the import, or add a comment that explicitly names the import as a required trigger so it is not accidentally removed during refactoring:
```python
import winml.modelkit.models.hf  # noqa: F401  # REQUIRED: populates COMPOSITE_MODEL_REGISTRY
assert COMPOSITE_MODEL_REGISTRY, "COMPOSITE_MODEL_REGISTRY is empty after hf import"
```

DingmaomaoBJTU (commented):
Follow-up review after commit 90f24b8a.

**Resolved:** Finding #1 (downstream): the `pipeline_tasks=[]` guard was added in `resolve_composite_info` — `CompositeInfo` with empty `pipeline_tasks` no longer reaches the formatter. Finding #5: the registry-divergence monkeypatch test was added.

**Still open:** Three issues remain — the falsy `None`/`{}` conflation in the same guard, the `issubclass` registry inconsistency that is the root cause of finding #2, and the formatter's own lack of a guard (finding #3). The side-effect import warning (#4) is also unaddressed.

src/winml/modelkit/inspect/resolver.py
timenick: You're right — changed to `if detected_components is None:` in `47b27a92`. My "defensive `not`" reasoning doesn't hold: `{}` means the composite path *was* taken (detection succeeded) but produced no breakdown — semantically distinct from `None` (path not taken), and shouldn't be suppressed the same way. The downstream `pipeline_tasks` guard still governs rendering; `test_resolve_composite_info_empty_components_not_suppressed` pins that an empty-components composite now surfaces instead of being silently dropped.

Generated with Claude Code

src/winml/modelkit/loader/resolution.py
timenick: You're right the two functions were written inconsistently — and I need to correct my earlier reply: I claimed removing the filter would wrongly include a "qwen3 generic non-composite registration," which was **factually wrong**. Every `COMPOSITE_MODEL_REGISTRY` entry already subclasses `WinMLCompositeModel` (qwen3 → `WinMLQwen3Model` → `WinMLDecoderOnlyModel(WinMLCompositeModel, …)`), so the `issubclass` filter was a **no-op** and the divergence couldn't actually fire — but the inconsistency was real as written.

Fixed at the root in `47b27a92`: `register_composite_model` now **enforces** the invariant (raises `TypeError` if a registered class isn't a `WinMLCompositeModel`), so the registry is homogeneous at its single write point. With that guarantee I removed the redundant `issubclass` filters from **both** `composite_pipeline_tasks` and `_composite_components_for_task`, so every reader (incl. `resolve_composite`) uniformly trusts the registry and they can no longer disagree. `test_composite_model.py` pins the enforcement and that all live registrations satisfy it.

Generated with Claude Code

src/winml/modelkit/inspect/formatter.py
timenick: Done in `47b27a92` — both composite branches now guard on `pipeline_tasks`: the Task row uses `if result.composite and result.composite.pipeline_tasks:` (plain-task fallback otherwise), and the `_output_composite_panel` call gets the same check. A directly-constructed `InspectResult(composite=CompositeInfo(pipeline_tasks=[], ...))` through the public `output_table` now renders the plain Task row and skips the panel instead of a bare ` [composite]`. `test_composite_with_empty_pipeline_tasks_renders_plain_task` pins it.

Generated with Claude Code

src/winml/modelkit/loader/resolution.py
timenick: Fair — "returns None gracefully" is still *silent*, which is exactly the failure mode you're flagging. Fixed in `fcc86983` by centralizing the trigger instead of scattering it: a new `_composite_registry()` accessor is the single load point all three readers (`resolve_composite` / `composite_pipeline_tasks` / `_composite_components_for_task`) now share, and it **raises `RuntimeError` if the registry is empty** after the import — so a moved/renamed registration fails loudly instead of silently disabling composites. That also resolves the consistency concern from my earlier reply (one trigger, not three bare imports that can drift). `test_registry_accessor_raises_loudly_when_empty` pins it.

Generated with Claude Code

DingmaomaoBJTU (approved):
All four findings from the previous reviews are now addressed:

1. **resolver.py** — if detected_components is None: (was falsy 
ot). ✅
2. **resolution.py** — issubclass(WinMLCompositeModel) filter removed from composite_pipeline_tasks(); both it and 
esolve_composite() now go through the shared _composite_registry() helper, so they are consistent. ✅
3. **formatter.py** — Guard changed to if result.composite and result.composite.pipeline_tasks: on both the Task row and the panel call. ✅
4. **resolution.py** — _composite_registry() centralises the side-effect import and adds a RuntimeError if the registry is empty after import — fail-loud instead of silently returning []. All three registry readers now go through it. ✅

The implementation is clean: the shared _composite_registry() accessor elegantly solves findings 2 and 4 together. LGTM.

PR #904, opened 2026-06-16
fix(cli): quiet third-party optimum logger noise in normal output

timenick:
## Summary

`optimum` emits informational WARNINGs that leak into normal CLI output and interleave with the rich progress spinner. The most visible:

```
No model type passed for the task image-to-text, that may be mapped to several loading classes (('AutoModelForVision2Seq', 'AutoModel')). Defaulting to AutoModelForVision2Seq to load the model.
```

It surfaces when inspecting `image-to-text` / vision-encoder-decoder models (e.g. `microsoft/trocr-base-printed`), mangling the live `⠧ Resolving …` spinner line.

## Fix

`configure_logging` now floors noisy third-party loggers (`optimum`) at `ERROR` in normal output; with `-v`/`-vv` they follow the CLI verbosity level, so the detail stays available on demand.

This is a logging-hygiene fix at the configuration boundary — it does **not** match message text and does **not** change how `optimum` is called.

## Notes

- **Pre-existing**, not from the recent task-detection work: verified at a pre-#878 commit with period-correct deps (`uv run`, not `--no-sync`) — still reproduces. `optimum>=2` has been pinned since the initial commit.
- Before: normal `winml inspect -m microsoft/trocr-base-printed` leaks the warning onto the spinner line.
- After: normal output is clean; `-v` still shows the notice.


src/winml/modelkit/utils/logging.py
```diff
@@ -82,6 +88,13 @@ def configure_logging(
     # the prior ``logging.basicConfig`` behavior, which never set a handler level.
     root.setLevel(log_level)
 
+    # Keep noisy third-party library chatter out of normal output. Their loggers float
+    # up to the CLI level only when the user opts into verbosity (-v/-vv); otherwise they
+    # are pinned at ERROR so library notices never leak into / interleave with output.
+    library_level = log_level if (verbosity > 0 or verbose) else logging.ERROR
```
DingmaomaoBJTU: **Suppression floor is `ERROR`, not `WARNING` — silences optimum warnings in default mode.**

The code pins noisy library loggers at `logging.ERROR` in default mode, which means `WARNING`-level messages from `optimum` are also swallowed in normal CLI use. The motivation in the PR description is INFO chatter (e.g. *"No model type passed for the task …"*), but `ERROR` goes further. A genuine optimum `WARNING` (e.g. a deprecated API, unsupported quantisation combination) would silently disappear for the user.

Consider using `logging.WARNING` as the default floor so only INFO and below are suppressed:

```python
library_level = log_level if (verbosity > 0 or verbose) else logging.WARNING
```

DingmaomaoBJTU: **`or verbose` in the condition is dead code after the backward-compat transform.**

By line 94, lines 66-67 have already set `verbosity = int(verbose)` when `verbose=True and verbosity==0`, so `verbosity > 0` already covers every `verbose=True` case. The `or verbose` branch can never fire on its own at this point — not a bug, but it misleads future readers into thinking `verbose` still has independent effect here. Consider simplifying to just `verbosity > 0`.

tests/unit/utils/test_logging.py
```diff
@@ -0,0 +1,39 @@
...
+from winml.modelkit.utils.logging import configure_logging
+
+
+@pytest.fixture(autouse=True)
+def _restore_optimum_level():
+    """configure_logging mutates global logger state; restore it after each test."""
+    logger = logging.getLogger("optimum")
+    before = logger.level
+    yield
+    logger.setLevel(before)
```
DingmaomaoBJTU: **Fixture only restores the `optimum` logger — root logger level leaks between tests.**

`configure_logging` mutates `logging.getLogger().level` (the root logger) on every call, but `_restore_optimum_level` only saves/restores the `optimum` child logger. A test that calls `configure_logging(verbosity=2)` leaves the root logger at DEBUG, which can affect subsequent tests that rely on root-logger default state.

```python
@pytest.fixture(autouse=True)
def _restore_optimum_level():
    root = logging.getLogger()
    root_before = root.level
    logger = logging.getLogger("optimum")
    before = logger.level
    yield
    root.setLevel(root_before)
    logger.setLevel(before)
```

DingmaomaoBJTU (commented):
Overall the change is clean and well-motivated. Three findings worth addressing before merge:

src/winml/modelkit/utils/logging.py
timenick: Good catch that `ERROR` is broader than `WARNING` — but I chose it deliberately, and a `WARNING` floor wouldn't actually fix the reported noise:

1. **The target message *is* a `WARNING`.** `"No model type passed for the task …"` is emitted via `logger.warning(...)` on `optimum.exporters.tasks`. A floor of `WARNING` still lets `WARNING`-level records through (`30 >= 30`), so the noise would keep printing in default mode. I verified this empirically — only a floor *above* WARNING (i.e. `ERROR`) suppresses it. So `else logging.WARNING` would be a no-op for the exact case this PR targets.
2. **Genuine optimum warnings aren't lost — they're gated behind verbosity.** The floor is verbosity-conditional: at `-v`/`-vv` it drops to the CLI level, so every optimum `WARNING` (deprecations, quant combos, …) reappears. The intent is that `optimum` is internal plumbing whose warnings are noise to a CLI user by default; anything that genuinely matters should be surfaced through WinML's own logger. `-v` is the escape hatch — nothing is permanently swallowed.

Happy to revisit if you think a specific optimum `WARNING` should be promoted to default visibility.

Generated with Claude Code

timenick: Fixed in `1b3f0e64` — simplified to just `verbosity > 0`. `verbose=True` is already folded into `verbosity` at lines 66–67, so the `or verbose` branch could never fire on its own. Thanks.

Generated with Claude Code

tests/unit/utils/test_logging.py
timenick: Fixed in `1b3f0e64` — the fixture now saves and restores the root logger level too, so a `configure_logging(verbosity=2)` call no longer leaks DEBUG into later tests:

```python
@pytest.fixture(autouse=True)
def _restore_logger_levels():
    root = logging.getLogger()
    optimum = logging.getLogger("optimum")
    root_before, optimum_before = root.level, optimum.level
    yield
    root.setLevel(root_before)
    optimum.setLevel(optimum_before)
```

Thanks for catching this.

Generated with Claude Code

src/winml/modelkit/utils/logging.py
```diff
@@ -82,6 +88,14 @@ def configure_logging(
     # the prior ``logging.basicConfig`` behavior, which never set a handler level.
     root.setLevel(log_level)
 
+    # Keep noisy third-party library chatter out of normal output. Their loggers float
+    # up to the CLI level only when the user opts into verbosity (-v/-vv); otherwise they
+    # are pinned at ERROR so library notices never leak into / interleave with output.
+    # (verbose=True is folded into verbosity above, so verbosity > 0 covers it.)
+    library_level = log_level if verbosity > 0 else logging.ERROR
```
DingmaomaoBJTU, line 95: **[Previous finding, still unresolved] Suppression floor is `ERROR`, not `WARNING` -- silences optimum WARNING-level messages in default mode.**

Findings 2 and 3 from the prior review were fixed in commit 1b3f0e64 -- thank you. This one was not.

The floor is still `logging.ERROR` (level 40), which swallows any WARNING (level 30) from `optimum` in default mode -- e.g. deprecation notices, unsupported quantisation flags, or model-type mismatches that the library classifies as warnings. The stated motivation is INFO chatter (level 20); pinning to ERROR goes two levels further and discards warnings too.

Suggested fix:
`python
# Floor at WARNING: suppress INFO noise, but let genuine library warnings through.
library_level = log_level if verbosity > 0 else logging.WARNING
`

The `_NOISY_LIBRARY_LOGGERS` module-level comment (line 40) also says 'floored at ERROR' and would need updating if the floor changes.

DingmaomaoBJTU (commented):
Follow-up review: two of the three previous findings are resolved. One warning-severity finding remains open.

DingmaomaoBJTU (approved):

Issue #284, opened 2026-04-09
[Task] keypoint-detection model support


DingmaomaoBJTU:
## Overview

Keypoint detection (a.k.a. human pose estimation) models predict the locations of anatomical landmarks — joints, facial landmarks, hand keypoints — on people in images or video frames. This enables downstream reasoning about body posture, gesture, and movement.

## Agent Scenarios

- **Fitness / sports coaching agent**: analyze exercise form in real time, flag incorrect posture or track rep counts from a webcam feed
- **Accessibility agent**: interpret sign language or body gestures as input to a conversational agent
- **AR / avatar agent**: drive avatar animations by mapping detected pose to a virtual skeleton
- **Safety monitoring agent**: detect falls or abnormal postures in elderly care or industrial environments

## ModelKit Integration

Models must pass the full wmk pipeline on all EPs:
```
wmk config → wmk build (ONNX export) → wmk perf → wmk eval
```

## Acceptance Criteria

- [ ] usyd-community/vitpose-plus-base
- [ ] usyd-community/vitpose-base-simple
- [ ] usyd-community/vitpose-plus-small
- [ ] usyd-community/vitpose-plus-large
- [ ] usyd-community/vitpose-plus-huge
- [ ] stanfordmimi/synthpose-vitpose-huge-hf

PR #905, opened 2026-06-16
feat(keypoint-detection): enable ViTPose config/build/perf

jeon185:
Enables the 6 ViTPose keypoint-detection models from #284 to pass `wmk config -> build -> perf` (CPU and OpenVINO). Eval isn't included here - I'll do the accuracy side in a follow-up since it needs a couple of design decisions first.

Two things were blocking all 6 models:

1. `wmk config` failed with "Task 'keypoint-detection' not supported by TasksManager". Optimum has the ViTPose ONNX export config but no task->class entry for keypoint-detection, and `AutoModelForKeypointDetection` only covers SuperPoint. Added the `(vitpose, keypoint-detection) -> VitPoseForPoseEstimation` mapping, same way we already do it for CLIP/SAM.

2. The plus checkpoints (MoE backbone) crashed during export with "dataset_index must be provided when using multiple experts". Optimum's VitPoseModelPatcher injects a constant dataset_index, but `patch_model_for_export` defaults model_kwargs to None so it crashed on init. Passing an explicit `model_kwargs={}` fixes that. The trace step (Step 3) was also running the model outside the patcher context, so I wrapped it the same way the export step already is.

The exporter change isn't ViTPose-specific - it helps any MoE model whose patcher injects forward args.

Verified config/build/perf on all 6: vitpose-base-simple, vitpose-plus-small/base/large/huge, and synthpose-vitpose-huge-hf. Added unit tests for the mapping and the patcher model_kwargs handling.

One note: you still need to pass `--task keypoint-detection` explicitly for now - the task isn't auto-detected from the config yet. I left auto-detection out of this PR to keep it small; can add it here or as a follow-up if you'd prefer.

Refs #284.

### DIFF
--- a/src/winml/modelkit/export/htp/exporter.py
+++ b/src/winml/modelkit/export/htp/exporter.py
@@ -229,7 +229,13 @@ class HTPExporter:
             monitor.update(ExportStep.INPUT_GEN, **input_gen_data)
 
             # Step 3: Hierarchy Building
-            self._trace_model_hierarchy(model, inputs)
+            # Trace under the Optimum patcher so models that inject constant
+            # forward arguments at export time (e.g. ViTPose MoE's dataset_index)
+            # are traced with the same inputs they are exported with. The export
+            # in Step 4 re-enters the patcher; the contexts are sequential, not
+            # nested.
+            with self._get_optimum_patcher(model, task):
+                self._trace_model_hierarchy(model, inputs)
 
             execution_steps = (
                 self._hierarchy_builder.get_execution_summary().get("execution_steps", 0)
@@ -487,7 +493,11 @@ class HTPExporter:
                 task=to_optimum_task(task),
                 library_name="transformers",
             )
-            return cfg_cls(model_config).patch_model_for_export(model)
+            # Pass an explicit empty model_kwargs so patchers that inject extra
+            # forward arguments can populate it. Some patchers (e.g. ViTPose MoE,
+            # which sets a constant dataset_index) assume a mutable dict and crash
+            # on the None default from patch_model_for_export.
+            return cfg_cls(model_config).patch_model_for_export(model, model_kwargs={})
         except KeyError:
             logger.debug(
                 "Model type '%s' (task='%s') not in Optimum registry; "
--- a/src/winml/modelkit/models/hf/vitpose.py
+++ b/src/winml/modelkit/models/hf/vitpose.py
@@ -1,0 +1,33 @@
+# -------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+# --------------------------------------------------------------------------
+"""ViTPose HuggingFace Model Configuration.
+
+ViTPose is a top-down human pose (keypoint-detection) model: a plain ViT
+backbone with a lightweight decoder that regresses keypoint heatmaps inside a
+given person box.
+
+This module provides:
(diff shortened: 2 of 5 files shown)


src/winml/modelkit/export/htp/exporter.py
```diff
@@ -229,7 +229,13 @@ def export(
             monitor.update(ExportStep.INPUT_GEN, **input_gen_data)
 
             # Step 3: Hierarchy Building
-            self._trace_model_hierarchy(model, inputs)
+            # Trace under the Optimum patcher so models that inject constant
+            # forward arguments at export time (e.g. ViTPose MoE's dataset_index)
+            # are traced with the same inputs they are exported with. The export
+            # in Step 4 re-enters the patcher; the contexts are sequential, not
+            # nested.
+            with self._get_optimum_patcher(model, task):
```
zhenchaoni, line 237: Change looks good and low-risk overall. The `model_kwargs={}` fix is effectively a no-op for non-MoE patchers (Optimum's `ModelPatcher.__init__` already coerces `None → {}`), so that one's safe.

My one concern is wrapping Step 3 (`_trace_model_hierarchy`) in the patcher. Models that already resolve a real Optimum patcher today — e.g. CLIP (`clip_text_model` / `clip_vision_model`), SAM, T5, SigLIP, whisper, VED — will now have their **hierarchy trace** run through patched `forward` for the first time. The ONNX graph is unaffected (export already ran patched), but the traced module path can shift, which could change the hierarchy / tag coverage on those models.

You verified the 6 ViTPose models, but those weren't being traced-under-patch before. Could you also run a before/after on at least one already-patched non-ViTPose model (CLIP is a good pick) and confirm the tag coverage / hierarchy stats are unchanged?

src/winml/modelkit/models/hf/vitpose.py
```diff
@@ -0,0 +1,33 @@
...
+"""
+
+from __future__ import annotations
+
+from transformers import VitPoseForPoseEstimation
+
+
+# (model_type, task) -> HuggingFace model class
+MODEL_CLASS_MAPPING: dict[tuple[str, str], type] = {
+    ("vitpose", "keypoint-detection"): VitPoseForPoseEstimation,
```
vortex-captain, line 46:
```suggestion
    ("vitpose", "keypoint-detection"): VitPoseForPoseEstimation,
    ("vitpose", None): VitPoseForPoseEstimation,
```
Could you help try test to see this line makes command with task omitted work?

src/winml/modelkit/export/htp/exporter.py
jeon185: I ran CLIP before and after the change — it goes through a real Optimum patcher (CLIPTextModelWithProjection), so it's a fair test. The hierarchy and tagging come out the same both ways: 52 traced modules, 153 hierarchy modules, 606 nodes, 606 tagged, 100% coverage. Only the build time differs, so tracing Step 3 under the patcher doesn't change tag coverage for the already-patched models.

src/winml/modelkit/models/hf/vitpose.py
jeon185: Added the ("vitpose", None) sentinel and it works. It makes the resolver reverse-look-up keypoint-detection from the matching class entry, so config and build pick the task up without --task now. I ran `winml config -m usyd-community/vitpose-base-simple` with no task flag and it resolves keypoint-detection on its own, and added unit tests for it. Same approach SAM uses for mask-generation.

zhenchaoni (approved):

vortex-captain (approved):

PR #906, opened 2026-06-17
analyze: optimize case_index encoding, persist model bytes for fast re-run, add --debug locator, and fix Pad/Gemm rule conflicts

fangyangci:
## Summary

This PR groups four related improvements to the `analyze` runtime-checker pipeline: a more compact and efficient `case_index` encoding, durable persistence of model bytes for fast re-runs, a new `--debug` locator that maps models/nodes/rules back to `case_index` values, and fixes for the Pad and MatMul+Add (Gemm) rule conflicts. It also pins `pyarrow`/`pandas` to keep generated parquet rule artifacts deterministic.

## 1. `case_index` encoding optimization

Introduces a compact 4-character key prefix so that lookups into / writes from `case_index` are faster and self-describing.

- New codec `analyze/utils/avalizble_ep_device_ops/case_index_key_codec.py`:
  - **char 1** -> EP/device (1 char)
  - **chars 2-3** -> op / pattern + opset version (2 chars)
  - **char 4** -> is-QDQ flag (`0`/`1`)
- Backed by two mapping files: `avaliable_providers.json` (EP/device -> 1-char key) and `available_ops_all.json` (op/pattern + version -> 2-char key).
- `encode_file_name_to_4char_key()` and `decode_4char_key_to_folder_and_file_name()` provide a stable round-trip between rule file names and the 4-char prefix.
- Mapping indexes are built once and cached globally, with validation for the allowed alphabet, key length, and duplicate-key detection so collisions fail fast.

## 2. Persist model bytes in `op_check_result` for fast re-run

Every persisted case now carries its serialized model so it can be replayed without regeneration.

- `pattern/op_input_gen/op_input_gen.py`: serialize the model and store `model_bytes_b64` in `final_tags` by default for every case (previously only populated in `dry_run`). This makes any stored `op_check_result` directly re-runnable.
- `dry_run` placeholder result now reports `success=False` / `not_run` so dry-run rows are not mistaken for real passes.
- `normalize_constraint_dict` keeps the constructed numpy array instead of converting to a list, preserving dtype/shape fidelity.
- Pinned `pyarrow==23.0.1` and `pandas==2.3.3` (and `result_processor` now sorts condition columns) so generated parquet artifacts are byte-stable across runs/machines and avoid spurious diffs.

## 3. `analyze --debug`: directional `case_index` locator

Adds a `--debug` mode that uses the debug parquet artifacts in `rules_debug` (which carry `case_indices`) to locate `case_index` values from the model / node / rule direction.

- New `--debug` flag on the `analyze` command. It requires `WINMLCLI_RULES_DIR_FOR_DEBUG` to point at a `rules_debug` directory containing `*/*.parquet`; clear diagnostics are emitted when it is missing or misconfigured.
- Produces a runtime debug summary grouped by support level (`supported` / `partial` / `unsupported`) and `node_stable_key`, each entry carrying `case_indices`, `table_path`, and `table_file`.
- Writes `<model>.analyze.<EP>.<DEVICE>.debug.json` next to the model file.
- New typed payloads: `RuntimeDebugDetails` / `RuntimeDebugStep` (in `models/runtime_checks.py`) and `RuntimeDebugSummaryEntry` plus the `runtime_debug_details_summary` field on `EPSupport` (in `models/output.py`).
- `rule_loader.py`: new `WINMLCLI_RULES_DIR_FOR_DEBUG` env var and `get_runtime_rules_debug_search_dirs()`; `resolve_rule_parquet_path()` now resolves from `<EP>_<DEVICE>/` subdirs, supports `for_debug`, and returns `None` when not found instead of a guessed path.
- The offline rule generator (`result_processor` CLI) gains `--debug` / `--rules-debug-dir` / `--rules-debug-json-dir` and aggregates `case_indices` into the debug artifacts.

## 4. Fix Pad and MatMul+Add (Gemm) rule conflicts

Both fixes add finite, derived properties so previously-conflicting rule rows become distinguishable.

- `pattern/op_input_gen/pad_input_generator.py`: re-enable the `pads_is_symmetric` finite property and add an asymmetric-padding sample case for branch coverage -- resolves the Pad rule conflict.
- `pattern/gemm_patterns.py`: add the `C_is_value_constraint` finite property to distinguish a fixed-value bias constraint from a shape-only constraint -- resolves the MatMul+Add (Gemm) pattern conflict.

## Testing

Extensive unit-test additions and updates across the analyze pipeline, including `test_analyzer.py`, `test_check_ops.py`, `test_rule_loader.py`, `test_static_analyzer_cli.py`, `test_runtime_checker_query_parquet.py`, `test_result_processor_per_op.py`, and `test_has_rule_data.py`.

### DIFF
--- a/src/winml/modelkit/analyze/runtime_checker/result_processor.py
+++ b/src/winml/modelkit/analyze/runtime_checker/result_processor.py
@@ -682,6 +682,7 @@ def _deduplicate_rule_rows(
     output_cols: list[str],
     compare_output_cols: list[str] | None = None,
     case_index_col: str = "case_index",
+    aggregate_case_indices: bool = False,
 ) -> tuple[pd.DataFrame, pd.DataFrame | None]:
     """Deduplicate rows by condition columns and detect conflicts.
 
@@ -723,6 +724,18 @@ def _deduplicate_rule_rows(
             continue
 
         row = group_df.iloc[0].copy()
+        if aggregate_case_indices and case_index_col in group_df.columns:
+            case_indices: list[Any] = []
+            for case_index_value in group_df[case_index_col].tolist():
+                if case_index_value is None:
+                    continue
+                if isinstance(case_index_value, float) and pd.isna(case_index_value):
+                    continue
+                case_indices.append(case_index_value)
+
+            # Keep deterministic order so generated artifacts are stable.
+            row["case_indices"] = tuple(sorted(set(case_indices), key=lambda value: str(value)))
+
         row["rule_row_count"] = len(group_df)
         dedup_rows.append(row)
 
@@ -783,6 +796,26 @@ if __name__ == "__main__":
         ),
     )
     parser.add_argument(
+        "--debug",
+        action="store_true",
+        help=(
+            "Generate debug rule artifacts that include case_indices, "
+            "written to --rules-debug-dir and --rules-debug-json-dir."
+        ),
+    )
+    parser.add_argument(
+        "--rules-debug-dir",
+        type=str,
+        default=None,
+        help="Directory where debug parquet rule files are written.",
+    )
+    parser.add_argument(
+        "--rules-debug-json-dir",
+        type=str,
+        default=None,
+        help="Directory where debug JSON rule files are written.",
+    )
+    parser.add_argument(
         "--domains",
         type=str,
         default="ai.onnx,com.microsoft",
@@ -817,6 +850,21 @@ if __name__ == "__main__":
     parquet_dir = Path(args.rules_dir) if args.rules_dir else get_runtime_rules_search_dirs()[0]
     parquet_dir.mkdir(parents=True, exist_ok=True)
 
+    debug_parquet_dir: Path | None = None
+    debug_output_dir: Path | None = None
+    if args.debug:
+        if not args.rules_debug_dir:
+            print("--rules-debug-dir is required when --debug is enabled")
+            sys.exit(1)
+        if not args.rules_debug_json_dir:
+            print("--rules-debug-json-dir is required when --debug is enabled")
+            sys.exit(1)
+
+        debug_parquet_dir = Path(args.rules_debug_dir)
+        debug_output_dir = Path(args.rules_debug_json_dir)
+        debug_parquet_dir.mkdir(parents=True, exist_ok=True)
+        debug_output_dir.mkdir(parents=True, exist_ok=True)
+
     qdq_generator = None
     processed = 0
     skipped = 0
@@ -824,6 +872,7 @@ if __name__ == "__main__":
 
     output_cols = ["compile_run_success"]
     compare_output_cols = ["compile_run_success"]
+    debug_output_cols = ["compile_run_success", "case_indices"]
     internal_cols = ["has_not_run_placeholder_reason", "compile_reason", "run_reason"]
 
     for json_file in sorted(json_files):
@@ -896,12 +945,16 @@ if __name__ == "__main__":
                 for c in rule_df.columns
                 if c not in output_cols and c not in infinite_properties and c != "case_index"
             ]
+            # Keep artifact schema stable across runs/machines regardless of
+            # intermediate dict/set insertion ordering.
+            condition_cols = sorted(condition_cols)
 
             dedup_df, conflict_df = _deduplicate_rule_rows(
                 rule_df,
                 condition_cols,
                 output_cols,
                 compare_output_cols=compare_output_cols,
+                aggregate_case_indices=args.debug,
             )
 
             if conflict_df is not None and not conflict_df.empty:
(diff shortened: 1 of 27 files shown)


src/winml/modelkit/commands/analyze.py
```diff
@@ -628,6 +611,93 @@ def _ep_name_device_display_name(ep_name: str, device_name: str) -> str:
...
+        result_device = str(ep_result.get("device_type") or "").upper()
+        if result_device and result_device != target_device:
+            continue
+
+        return _normalize_runtime_debug_summary_payload(
+            ep_result.get("runtime_debug_details_summary")
+        )
+
+    # Fallback: first available runtime_debug_details_summary in this run.
+    for ep_result in raw_results:
```
DingmaomaoBJTU, line 697: Bug: fallback returns debug data from wrong EP in multi-EP runs. The fallback loop (lines 684-690) has no EP name filter, so it returns the first result dict that has a runtime_debug_details_summary key regardless of which EP it belongs to. If the primary exact-match loop finds no match (e.g. because device_type is empty in the result), the debug JSON written for ep_name/target_device could silently contain data from a different EP. Add an EP name guard: if ep_result.get('ep_type') != ep_name: continue.

src/winml/modelkit/analyze/utils/op_utils.py
```diff
@@ -134,9 +135,21 @@ def _is_empty_top_level(value: Any) -> bool:
...
-    """Return a stable hash value for a case signature."""
-    return hashlib.sha256(signature.encode("utf-8")).hexdigest()
+def _hash_case_signature(signature: str) -> str:
+    """Return a stable 32-char hash value for case content."""
+    return hashlib.md5(signature.encode("utf-8"), usedforsecurity=False).hexdigest()
+
+
+def _compute_case_index_with_namespace_key(case: dict, *, namespace_key: str) -> str:
+    """Compute unified 36-char case_index using 4-char namespace key + 32-char md5."""
+    signature = compute_case_signature(case, namespace="")
```
DingmaomaoBJTU, line 145: Warning: silent backward-compatibility break for existing delta/filter files. The case_index format changed from 32-char SHA-256(namespace+content) to 4-char-codec-key+32-char-MD5(content, namespace=''). Stored filter files with old 32-char SHA-256 case_index values will never match the new 36-char keys, so every previously-filtered case will silently re-run with no warning. Consider documenting this as a breaking change, emitting a warning when all filter indices appear to be in the old format, or bumping a format version in the result JSON.

src/winml/modelkit/analyze/runtime_checker/result_processor.py
```diff
@@ -964,6 +1041,11 @@ def _encode_condition_columns_for_parquet(
         f"output_dir={output_dir}"
     )
     print(f"Parquet rule files written to: {parquet_dir}")
+    if args.debug:
+        assert debug_output_dir is not None
```
DingmaomaoBJTU: Warning: assert used as a runtime guard, disabled by python -O. assert debug_output_dir is not None is elided when Python runs optimized, after which debug_output_dir would be None and a subsequent write would crash with an AttributeError. Since this code is only reached when args.debug is True and the vars are already guaranteed non-None by the earlier validation block, the simplest fix is to remove both assertions. If you want an explicit invariant, use if debug_output_dir is None: raise RuntimeError(...) instead.

src/winml/modelkit/pattern/op_input_gen/op_input_gen.py
```diff
@@ -236,7 +236,7 @@ def normalize_constraint_dict(c: dict) -> dict:
     if "same_value" in c and "same_value_shape" in c:
         normalized = {k: v for k, v in c.items() if k not in ("same_value", "same_value_shape")}
         dtype = np.dtype(c["dtype"]) if "dtype" in c else None
-        normalized["value"] = np.full(c["same_value_shape"], c["same_value"], dtype=dtype).tolist()
+        normalized["value"] = np.full(c["same_value_shape"], c["same_value"], dtype=dtype)
```
DingmaomaoBJTU, line 242: Warning: drops .tolist() so return type changes from list to np.ndarray. The two call sites in result_processor.py are safe because encode_rule_condition_value_for_parquet and _safe_dump both handle ndarrays. However, the public contract of normalize_constraint_dict is now inconsistent (callers get different types depending on which keys are present), and any future caller that passes the dict to json.dumps without a custom handler will get a TypeError. Restore .tolist() to preserve the original list contract, or document that value may be ndarray.

src/winml/modelkit/analyze/utils/avalizble_ep_device_ops/case_index_key_codec.py
```diff
@@ -0,0 +1,273 @@
...
+    The mapping files are fixed and shipped alongside this module; the result is
+    parsed once per process and reused on every call.
+    """
+    providers_path, ops_path = _default_mapping_paths()
+    return _compute_indexes(str(providers_path), str(ops_path))
+
+
+def encode_file_name_to_4char_key(file_name: str) -> str:
+    """Encode a rule file name into its 4-char case_index key prefix."""
+    (
```
DingmaomaoBJTU, line 200: Info: the new codec module has no unit tests. encode_file_name_to_4char_key and decode_4char_key_to_folder_and_file_name are now critical primitives baked into every case_index written to disk, but there are no tests for encode/decode roundtrip correctness, KeyError on unknown EP/device, ValueError on malformed file names, or empty mappings in the JSON data files. A regression here would silently invalidate all case_index values and break delta/rerun mode without any obvious failure.

src/winml/modelkit/pattern/op_input_gen/pad_input_generator.py
```diff
@@ -138,10 +144,8 @@ def derive_properties(self, properties: dict[str, Any]) -> dict[str, Any]:
 
         # need for NVidia TRT RTX execution provider tests
         item["pads_all_zeros"] = bool(np.all(pads_array == 0))
-        # item["pads_is_symmetric"] = bool(
-        #     np.array_equal(pads_array[: len(pads_array) // 2],
-        #                    pads_array[len(pads_array) // 2 :])
-        # )
+        half = len(pads_array) // 2
+        item["pads_is_symmetric"] = bool(np.array_equal(pads_array[:half], pads_array[half:]))
```
DingmaomaoBJTU, line 148: Info: pads_is_symmetric was previously commented out with no explanation; re-enabling it changes the set of condition columns in existing parquet files. Any parquet rule files generated while the property was absent will not have a pads_is_symmetric column; lookups that now emit it as a condition may silently miss all rows. Consider regenerating affected parquet files and documenting this in the PR description.

src/winml/modelkit/commands/analyze.py
fangyangci: Deferring this intentionally. The EP/device selection and the debug-summary EP matching will be reworked together in a follow-up that unifies EP/device handling, and the explicit EP-name guard in the fallback loop will be added as part of that change. In the current code each `analyze()` call yields exactly one EP result, so the fallback does not misattribute debug data in practice today. Tracking the guard in the unified EP/device change.

src/winml/modelkit/analyze/utils/op_utils.py
fangyangci: By design — we intentionally do not support the old case_index format. All rule and result artifacts have been regenerated against the new format, and the legacy `case_index_ignore_ep_device` field is actively stripped on write (`_set_case_index_signatures` in op_utils.py, asserted gone by tests in test_check_ops.py). Old-format filter files are never mixed with the new artifacts, so we are deliberately not adding backward-compat detection or a format version. Old-version content is removed rather than supported.

src/winml/modelkit/analyze/runtime_checker/result_processor.py
fangyangci: Fixed. Removed both asserts: the values are already guaranteed non-None by the `args.debug` validation block above, and as you noted they would be elided under `python -O`. Replaced with a comment documenting the invariant.

src/winml/modelkit/pattern/op_input_gen/op_input_gen.py
fangyangci: We intentionally keep the ndarray and must NOT restore `.tolist()`. `.tolist()` coerces numpy scalars to Python types (e.g. numpy `bool_` -> `bool`), which loses the original dtype and breaks faithful restoration of the input value — restoring it would reintroduce that bug. The current call sites handle ndarrays correctly, and we have added an inline comment explaining why `.tolist()` must not be used (and updated the docstring wording from "list" to "array").

src/winml/modelkit/analyze/utils/avalizble_ep_device_ops/case_index_key_codec.py
fangyangci: Coverage exists. encode/decode round-trip over the real rule corpus is tested in test_rule_loader.py (asserts folder/file reconstruction), and the 36-char case_index format (4-char key + first-char-differs-by-EP/device, remaining 35 equal) is tested in test_check_ops.py. We additionally validated encode/decode against all 1782 real op_check_result file names with zero mismatches. The critical round-trip path is covered; dedicated error-path cases can be added later.

src/winml/modelkit/pattern/op_input_gen/pad_input_generator.py
fangyangci: All affected parquet rule files have been regenerated with the `pads_is_symmetric` condition column present, so there is no stale-column / missed-row scenario. The property is now live (the previously commented-out code is replaced) and is covered by the per-op and QDQ tests.

DingmaomaoBJTU (approved):
All review comments have been addressed — the assert fix lands in this PR, the intentional ndarray/format decisions are well-reasoned, the deferred EP-fallback guard is tracked for a follow-up, and parquet artifacts have been regenerated. LGTM.

Issue #64, opened 2026-03-31
feat: smart precision with device-aware auto defaults


DingmaomaoBJTU:
## Summary

Support \`--device auto --precision auto\` where auto precision intelligently selects the optimal quantization/conversion strategy based on the target device.

## Device-Aware Auto Precision

| Device | Auto Precision | Pipeline Behavior |
|--------|---------------|-------------------|
| **NPU** | int8 (quantized) | Enable QDQ quantization (uint8/int8) |
| **GPU** | fp16 (non-LLM) / w4a16 (LLM) | FP16 conversion for standard models; weight-int4 + activation-fp16 for LLMs |
| **CPU** | fp16 | FP16 conversion, no int quantization |

## Explicit Precision Overrides

| \`--precision\` | Behavior |
|----------------|----------|
| \`auto\` | Device-dependent (see table above) |
| \`fp32\` | No conversion — optimize + compile only |
| \`fp16\` | FP16 conversion, skip int quantization entirely |
| \`int8\` | QDQ int8 quantization (current default behavior) |
| \`int16\` | QDQ int16 quantization (#229) |

### Pipeline adaptation
- **fp16 path**: Skip quantize stage entirely, add fp16 conversion step (cast ops or mixed precision)
- **int8/int16 path**: Run quantize stage as today
- **fp32 path**: Skip quantize, standard optimize + compile

### LLM detection for GPU auto
- Heuristic: decoder-only architecture, large vocab size, model size > threshold
- Or explicit flag: \`--model-type llm\` to force w4a16 path
- Avoid hardcoding model names — use structural properties

### CLI integration
- \`wmk build -m model_id -o output/ --device npu --precision auto\`
- \`wmk build -m model_id -o output/ --device cpu --precision fp16\`
- Default: \`--device auto --precision auto\`

## Acceptance Criteria

- [ ] \`--precision auto\` resolves based on \`--device\` value
- [ ] NPU auto → int8 quantization
- [ ] GPU auto → fp16 (non-LLM) or w4a16 (LLM)
- [ ] CPU auto → fp16
- [ ] \`--precision fp16\` skips quantization stage
- [ ] \`--precision fp32\` skips quantization and conversion
- [ ] LLM detection uses structural properties, not hardcoded model names
- [ ] Backward compatible — existing behavior unchanged when precision not specified

DingmaomaoBJTU:
**Triage Summary** (auto-generated 2026-06-17)

| Field | Value |
|-------|-------|
[...]

PR #914, opened 2026-06-18
feat(cli): add --precision to build; share precision_option helper

xieofxie:
## Summary

Wires `--precision` into `winml build` and removes the duplicated `--precision`
option across commands by extracting a shared `cli_utils.precision_option()`
helper. Part of #64 (device-aware auto precision).

### `winml build` gains `--precision`
`winml build` previously had no `--precision` flag despite issue #64's
documented CLI (`wmk build … --precision auto`). It now accepts `-p/--precision`
and threads it through:
- the auto-generated config path (`generate_build_config(..., precision=...)`), and
- the device-patch path, which now also fires when only `--precision` is passed
  (not just `--device`), so `fp16`/`fp32` clear the quant config and
  `int8`/`w8a16` populate it.

The precision **engine** (`config/precision.py`, `resolve_precision`) already
existed and was fully tested — this is CLI wiring only. No new precision
semantics, no fp16-conversion stage (deferred).

### Shared `precision_option` helper
`--precision` was copy-pasted across `build`, `config`, `eval`, `perf`, and
`quantize` with drifting help text. Extracted into `cli_utils.precision_option()`:
- `type=str` (not `click.Choice`) so the `w{x}a{y}` mixed-precision format
  parses; invalid values are still rejected downstream (`resolve_precision`
  for build-path commands, `_resolve_quant_types` for `quantize`).
- consistent base help, with an optional command-specific note appended.
- registers `-p` by default (`eval`/`perf` gain the alias for consistency).
- `default` accepts `None` and a `help_text` override, so `quantize` — whose
  accepted set excludes `fp16`/`fp32` and whose default is `None` — shares the
  same helper without a misleading help string.

All five commands now use the helper; zero inline `--precision` definitions remain.

### Misc
- Fixed a stale `npu->int8` docstring in `models/auto.py` to `npu->w8a16`
  (matches the actual engine default).

## Tests
- `tests/unit/commands/test_build.py`: 3 new cases — `--precision int8` sets
  quant, `--precision fp16` clears quant on NPU, `--precision` alone (no
  `--device`) still patches quant.
- `tests/unit/utils/test_cli.py`: new `TestPrecisionOption` (default, `-p`,
  `--precision`, `include_short=False`, custom default, `default=None`,
  `help_text` override, help-text append, arbitrary-string acceptance).
- 389 affected unit tests pass; ruff clean.

### DIFF
--- a/src/winml/modelkit/commands/build.py
+++ b/src/winml/modelkit/commands/build.py
@@ -470,6 +470,9 @@ def _validate_loader_tasks_for_model(
     include_auto=True,
     optional_message="Default: auto-detect.",
 )
+@cli_utils.precision_option(
+    optional_message="With -c, applied only when --device or --precision is passed.",
+)
 @click.option(
     "--analyze/--no-analyze",
     "analyze",
@@ -509,6 +512,7 @@ def build(
     optimize: bool,
     ep: EPNameOrAlias | None,
     device: str,
+    precision: str,
     analyze: bool,
     max_optim_iterations: int | None,
     allow_unsupported_nodes: bool,
@@ -531,6 +535,12 @@ def build(
 
         # Full pipeline with explicit config
         winml build -c config.json -m microsoft/resnet-50 -o output/
+
+        # Device-aware auto precision (npu->w8a16, gpu/cpu->fp16)
+        winml build -m microsoft/resnet-50 -o output/ --device npu --precision auto
+
+        # Force fp16 (skips quantization)
+        winml build -m bert-base-uncased -o output/ --device gpu --precision fp16
 
         # Build from pre-exported ONNX file
         winml build -c config.json -m model.onnx -o output/
@@ -592,6 +602,7 @@ def build(
                 model_id,
                 trust_remote_code=trust_remote_code,
                 device=device,
+                precision=precision,
             )
             if not quant:
                 config_or_configs.quant = None
@@ -602,15 +613,19 @@ def build(
             if no_compile:
                 config_or_configs.compile = None
 
-        # If --device was explicitly provided, patch compile config and clear
-        # quant for CPU/GPU (neither device uses quantization by default).
-        if cli_utils.is_cli_provided(ctx, "device") and device:
+        # If --device or --precision was explicitly provided, patch quant/compile
+        # to honor the requested policy. fp16/fp32 clear quant; npu/int8 etc set it.
+        if (
+            cli_utils.is_cli_provided(ctx, "device") or cli_utils.is_cli_provided(ctx, "precision")
+        ) and device:
             from ..compiler.configs import WinMLCompileConfig
 
             def _patch_device(cfg: WinMLBuildConfig) -> None:
                 from ..config import resolve_quant_compile_config
 
-                resolved_quant, _ = resolve_quant_compile_config(device=device, ep=ep)
+                resolved_quant, _ = resolve_quant_compile_config(
+                    device=device, precision=precision, ep=ep
+                )
                 if not quant or resolved_quant is None:
                     cfg.quant = None
                 elif cfg.quant is None:
--- a/src/winml/modelkit/commands/build.py
+++ b/src/winml/modelkit/commands/build.py
@@ -615,9 +615,7 @@ def build(
 
         # If --device or --precision was explicitly provided, patch quant/compile
         # to honor the requested policy. fp16/fp32 clear quant; npu/int8 etc set it.
-        if (
-            cli_utils.is_cli_provided(ctx, "device") or cli_utils.is_cli_provided(ctx, "precision")
-        ) and device:
+        if cli_utils.is_cli_provided(ctx, "device") or cli_utils.is_cli_provided(ctx, "precision"):
             from ..compiler.configs import WinMLCompileConfig
 
             def _patch_device(cfg: WinMLBuildConfig) -> None:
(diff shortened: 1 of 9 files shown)


src/winml/modelkit/commands/eval.py
```diff
@@ -82,14 +82,8 @@
...
 @cli_utils.ep_option(required=False)
-@click.option(
-    "--precision",
-    type=str,
-    default="auto",
-    show_default=True,
-    help="Precision: auto, fp32, fp16, int8, int16, or w{x}a{y} (e.g., w8a16). "
-    "Applied during model build (fp16/fp32 skip quantization). "
-    "Ignored for pre-built ONNX inputs (precision is already baked in).",
+@cli_utils.precision_option(
```
DingmaomaoBJTU, line 85: **New `-p` short alias silently added to `eval`.** `eval` did not previously expose `-p` (only `--precision`). `precision_option()` defaults to `include_short=True`, so this refactor quietly adds a new CLI alias. If the intent is just consolidation (not expanding the surface), use `include_short=False` here to preserve the existing interface.

src/winml/modelkit/commands/eval.py
```diff
@@ -82,14 +82,8 @@
...
+    optional_message="Ignored for pre-built ONNX inputs (precision is already baked in).",
```
DingmaomaoBJTU: **Help text lost the 'Applied during model build' context.** The old help said: `'Applied during model build (fp16/fp32 skip quantization). Ignored for pre-built ONNX inputs...'`. The `optional_message` now only appends the second half. For `eval` users who wonder why `--precision` seems to have no effect on a pre-built ONNX model, this context matters. Consider: `optional_message='Applied during model build. Ignored for pre-built ONNX inputs (precision is already baked in).'`

src/winml/modelkit/commands/perf.py
```diff
@@ -1305,13 +1305,7 @@ def _run_simple_loop(
...
 )
 @cli_utils.device_option(required=False, default="auto", include_auto=True)
-@click.option(
-    "--precision",
-    type=str,
-    default="auto",
-    show_default=True,
-    help="Precision mode: auto, fp32, fp16, int8, int16, or w{x}a{y} (e.g., w8a16).",
-)
+@cli_utils.precision_option()
```
DingmaomaoBJTU, line 1308: **New `-p` short alias silently added to `perf`.** Same concern as `eval`: the old `perf --precision` had no `-p` alias. Using `precision_option()` with default `include_short=True` introduces it silently. If backward-compat of the CLI surface is a concern, pass `include_short=False`.

src/winml/modelkit/commands/build.py
```diff
@@ -602,15 +613,19 @@ def build(
...
                 config_or_configs.compile = None
 
-        # If --device was explicitly provided, patch compile config and clear
-        # quant for CPU/GPU (neither device uses quantization by default).
-        if cli_utils.is_cli_provided(ctx, "device") and device:
+        # If --device or --precision was explicitly provided, patch quant/compile
+        # to honor the requested policy. fp16/fp32 clear quant; npu/int8 etc set it.
+        if (
+            cli_utils.is_cli_provided(ctx, "device") or cli_utils.is_cli_provided(ctx, "precision")
+        ) and device:
```
DingmaomaoBJTU: **`and device` guard is truthy for `device='auto'` too.** When `--ep` is passed explicitly (so the resolution block at lines 576-586 is skipped) but `--device` is not set, `device` stays `'auto'` -- which is truthy. The guard intends to skip patching when no device is known, but `'auto'` passes it. In practice `resolve_quant_compile_config(device='auto', ...)` seems to handle this, but the intent should be made explicit (e.g. `and device != 'auto'`, or a comment explaining why `'auto'` is acceptable here).

src/winml/modelkit/utils/cli.py
```diff
@@ -294,6 +294,48 @@ def device_option(
...
+    param_decls = ["--precision", "precision"]
+    if include_short:
+        param_decls.insert(0, "-p")
+    return click.option(
+        *param_decls,
+        type=str,
+        default=default,
+        show_default=True,
+        help=help_text,
+    )
```
DingmaomaoBJTU, line 344: **`show_default=True` is new for `config` (minor help-text shift).** The old `config.py` option did not have `show_default=True`; it instead said `'Default: auto (based on device when device is specified).'` inline. The new shared helper drops that clarification and relies on Click's auto-rendered `[default: auto]`. Not a bug, but the note explaining *why* auto defaults to the device's precision is now missing from `config --help`.

src/winml/modelkit/commands/eval.py
xieofxie: all commands should align. So accepted. 
Another p is 
```
@click.option(
    "-P",
    "--param",
```

src/winml/modelkit/utils/cli.py
xieofxie: new help covered that

src/winml/modelkit/commands/eval.py
xieofxie: updated

src/winml/modelkit/commands/build.py
xieofxie: device is never None and resolve_quant_compile_config supports auto, so just remove it

DingmaomaoBJTU (approved):
LGTM! 3/5 findings resolved. The remaining two items (-p short alias on eval/perf) are confirmed intentional for CLI consistency. One new cosmetic info (quantize --help shows [default: None]) is acceptable.

Issue #913, opened 2026-06-18
Clean up unused inject_hub_metadata in utils/hub_utils.py


xieofxie:
While type-checking the `utils` package, `inject_hub_metadata` (in `src/winml/modelkit/utils/hub_utils.py`) appears to be **dead code**: it is exported in `utils/__init__.py` (`__all__`) but never called anywhere in `src/` or `tests/`.

It also contained a latent bug — `from ..version import __version__` referenced a non-existent `winml.modelkit.version` module (guarded by try/except, so `export_version` silently stayed `"unknown"`). The type-check PR fixes that import to `from .. import __version__`, but if the function is genuinely unused it should be removed entirely along with its export.

**Proposed cleanup:**
- Remove `inject_hub_metadata` from `hub_utils.py`
- Remove it from `utils/__init__.py` import + `__all__`
- Check for any other now-unused helpers it was the sole caller of

Found during the incremental mypy cleanup (utils folder).

xieofxie:
Also other functions in it like `is_hub_model`

DingmaomaoBJTU:
**Triage (2026-06-22)**

| Field | Value |
|-------|-------|
| Type | Refactor |
| Priority | P2 |
| Component | refactor |
| Owner | @xieofxie |

**Rationale:** Dead code removal — \inject_hub_metadata\ is unused and contains a latent import bug. Clean removal as part of ongoing mypy cleanup. Low risk, good first issue candidate.

PR #916, opened 2026-06-18
type: clean mypy in optracing, quant, serve, session, sysinfo, telemetry, utils and enforce in CI

xieofxie:
@
## Summary

Brings seven more `winml.modelkit` packages to **zero mypy errors** under the strict `[tool.mypy]` config and wires them into the required CI type-check gate (16 → 23 packages). Continues the incremental cleanup from #896.

Cleaned: **optracing, quant, serve, session, sysinfo, telemetry, utils** (sysinfo was already clean). Fixes follow the repos typing playbook — fix at the source first; `cast()` only at genuine untyped third-party boundaries; `Any` only for true dynamic pass-throughs.

## Real bugs surfaced (not just typing)

- **`utils/hub_utils.py` version import** — `from ..version import __version__` referenced a non-existent module; guarded by `try/except`, so the export version was silently always `"unknown"`. Fixed to `from .. import __version__`. (The containing helper `inject_hub_metadata` is dead code — tracked in #913.)
- **`utils/hub_utils.py`** — `model_info.modelId` → `model_info.id` (huggingface_hub renamed the attribute).

## Notable fixes

- **session** — bound a non-None local after auto-compile so narrowing survives into the lambda/comprehension; explicit guards for `resolved_ep is None` and single-mode `model_path`; widened PDH counter containers to `float`; class-level `_initialized` annotation; distinct ctypes struct locals.
- **telemetry** — annotated the `LoggerProvider`/`Logger` instance attributes (were inferred as `None`, cascading into unreachable/attr-defined); OTel/JSON boundary casts; typed the `@contextmanager`/`@asynccontextmanager` generators.
- **serve** — `cast()` at `json.loads`/`app.state` boundaries; `import binascii` to use `binascii.Error` (typeshed doesnt expose `base64.binascii`); explicit single-mode `model_path` guard.
- **utils** — removed stale `# type: ignore`s; `DataclassInstance` cast for generic-`TypeVar` dataclass reflection; `EPAlias` key cast for the alias lookup; loop-variable rename to fix `InputTensorSpec`/`OutputTensorSpec` shadowing.

## Config changes (`pyproject.toml`)

- **`qairt.*`** → `ignore_missing_imports` (external Qualcomm AI Runtime SDK; imported only inside `compile_qairt_bin.py`, which runs in a separate `venv-winml` subprocess — absent from the CI env).
- **`windowsml`** → `follow_untyped_imports = true` (installed but ships no `py.typed`; analyzing its source is clean and keeps its inline annotations instead of collapsing to `Any`).

## Verification

- Combined required-gate invocation (all 23 packages): `Success: no issues found in 300 source files`.
- ruff clean; unit tests green for the touched packages (2 pre-existing OpenVINO-EP-unavailable failures in this env, confirmed unrelated).
- Branch-vs-main diff is content-only (line endings normalized to LF to match main).
@


src/winml/modelkit/telemetry/library/exporter.py
```diff
@@ -226,7 +227,8 @@ def shutdown(self) -> None:
 
     def _to_envelope(self, ld: ReadableLogRecord) -> dict:
         record = ld.log_record
-        timestamp = _ns_to_datetime(record.timestamp)
+        # OTel types timestamp as Optional; fall back to now if a record omits it.
+        timestamp = _ns_to_datetime(record.timestamp or time.time_ns())
```
DingmaomaoBJTU: **Bug: falsy-timestamp misfire**

record.timestamp or time.time_ns() short-circuits on any falsy value — including 0 (Unix epoch), which is a valid OTel timestamp. If a record carries timestamp=0 the fallback fires silently.

Use an explicit None-check instead:
`record.timestamp if record.timestamp is not None else time.time_ns()`

src/winml/modelkit/utils/data_utils.py
```diff
@@ -51,7 +51,10 @@ def pad_inputs(
                 # Dynamic ONNX dims may be None or a string symbol; emit a
                 # (0, 0) pair so later pairs stay aligned with their dim index.
                 if not isinstance(exp, int):
-                    pad.extend([0, 0])
+                    # Forward-looking: expected is typed list[int] today, but ONNX
+                    # dynamic dims (None / str symbol) are a planned input (see TODO
+                    # above). Keep the guard until the signature widens for them.
+                    pad.extend([0, 0])  # type: ignore[unreachable]
```
DingmaomaoBJTU: **Warning: type-ignore[unreachable] papers over a real type mismatch**

exp is typed int (from expected_shape: list[int]), so 
ot isinstance(exp, int) is always False. Mypy's [unreachable] is correct. Suppressing it with a type: ignore keeps dead code alive and hides the inconsistency. Either widen the signature to list[int | str | None], or delete the guard and add a TODO comment.

src/winml/modelkit/session/monitor/_pdh.py
```diff
@@ -403,16 +405,17 @@ def __init__(
...
         self._cpu_samples: list[float] = []
-        self._ram_used_bytes: list[int] = []
+        self._ram_used_bytes: list[float] = []
         # Per-engtype snapshots of the monotonic Running Time counter. Stored
         # as dicts because an adapter exposes multiple engines (e.g. several
         # Compute_* on an NPU; 3D + Compute_* on a GPU) and the total adapter
         # time is the sum of per-engine deltas.
-        self._running_time_start_ns: dict[str, int] = {}
-        self._running_time_end_ns: dict[str, int] = {}
+        self._running_time_start_ns: dict[str, float] = {}
```
DingmaomaoBJTU: **Warning: overly broad annotation for ns-precision running-time counters**

Running-time counters use _PDH_FMT_LARGE (64-bit integer ns). Storing them as float can lose sub-100ns precision for large absolute timestamps (float64 mantissa is 53 bits; ns-since-epoch ~1.7e18 exceeds exact representation). Keeping dict[str, int] with an explicit int() conversion at the assignment site would be both more accurate and precision-safe.

src/winml/modelkit/session/session.py
```diff
@@ -327,6 +327,16 @@ def run(
         if self._session is None:
             self.compile()
 
+        # compile() populates self._session or raises; bind a non-None local so
+        # the narrowing survives into the lambda / comprehension below (mypy drops
+        # self-attribute narrowing inside nested scopes).
+        session = self._session
+        if session is None:
```
DingmaomaoBJTU, line 334: **Info: new InferenceError path has empty context**

The narrowing guard is correct and the comment explains why it exists. Two minor issues:
1. context={} gives no diagnostic information -- at minimum add the model path so a developer can identify which session failed.
2. This is a new exception type for a condition previously surfaced as an implicit AttributeError. Callers catching AttributeError would now miss it.

src/winml/modelkit/serve/cli_api.py
```diff
@@ -212,7 +212,10 @@ def _extract_json_from_stdout(stdout: str) -> dict[str, Any] | list[Any] | None:
                 if start == -1:
                     break
                 try:
-                    return json.loads(stdout[start : end + 1])
+                    return cast(
+                        "dict[str, Any] | list[Any] | None",
```
DingmaomaoBJTU: **Info: cast includes None but json.loads() cannot produce it here**

The code only reaches this point after finding matching bracket pairs, so the result is always a dict or list -- never null. The cast type dict[str, Any] | list[Any] | None misleads callers into thinking None is a possible return, making defensive None-checks on the call site dead code. Cast to dict[str, Any] | list[Any] to match the real contract.

src/winml/modelkit/telemetry/library/exporter.py
xieofxie: Good catch — fixed in f2091a2e. Switched to `record.timestamp if record.timestamp is not None else time.time_ns()` so a valid `timestamp=0` is kept.

src/winml/modelkit/utils/data_utils.py
xieofxie: Fixed in f2091a2e — widened `expected` to `Mapping[str, Sequence[int | str | None]]` and removed the ignore. Used the covariant `Mapping`/`Sequence` (rather than `dict`/`list`) so existing `dict[str, list[int]]` callers still fit without an invariance cascade, and the dynamic-dim guard is now genuinely reachable.

src/winml/modelkit/session/monitor/_pdh.py
xieofxie: Agreed — fixed in f2091a2e. Reverted the running-time dicts to `dict[str, int]` and convert with `int(v)` at the assignment site, so ns-since-epoch keeps full precision.

src/winml/modelkit/session/session.py
xieofxie: Fixed (1) in f2091a2e — context now carries `onnx_path` and `device`. On (2): this branch is outside the try/except, but the prior None-deref AttributeError was caught by that same block and re-raised as `InferenceError` anyway, so external callers already saw `InferenceError`, not `AttributeError` — the explicit raise just makes it a clear, typed message instead of a cryptic one.

src/winml/modelkit/serve/cli_api.py
xieofxie: Fixed in f2091a2e — cast is now `dict[str, Any] | list[Any]` (the `| None` of the function's return type comes from other paths, not this bracket-matched parse).

DingmaomaoBJTU (approved):
All five findings from the previous round are addressed in f2091a2e:

- **Bug fixed**: xporter.py timestamp now uses explicit is not None guard — epoch=0 is handled correctly.
- **Warning fixed**: data_utils.py signature widened to Mapping[str, Sequence[int | str | None]]; # type: ignore[unreachable] removed.
- **Warning fixed**: _pdh.py running-time counters reverted to dict[str, int] with explicit int() at assignment; delta computation is now clean.
- **Info addressed**: session.py InferenceError context now includes onnx_path and device.
- **Info addressed**: cli_api.py cast no longer includes None in the union type.

LGTM.

PR #919, opened 2026-06-18
analyze: resolve --ep/--device auto to a single best target

fangyangci:
## What

`winml analyze` previously treated `--ep auto` / `--device auto` as "fan out over **every** locally-available EP/device pair", which was inconsistent with `build` / `run` / `perf` (they resolve a single best target). This PR makes `analyze`'s `auto` resolution **single-target**, reusing the shared `sysinfo` helpers (`resolve_device` / `resolve_eps`).

`--ep all` / `--device all` keep their full fan-out semantics — **unchanged**.

## Resolution rules (per axis)

| value | resolves to |
|---|---|
| `auto` | a **single** best target via shared `resolve_device()` / `resolve_eps()` — local-availability based (device priority NPU > GPU > CPU, EP priority IHV-first) |
| `all` | the full rule-data-backed set (fan-out) — unchanged |
| concrete (`qnn`, `gpu`, …) | itself |

The cartesian product + `EP_SUPPORTED_DEVICES` filtering is unchanged, so mixed combos like `--ep all --device auto` work naturally (all EPs × the one resolved device). The old local-pairs intersection / half-auto warning block is removed — single-target `auto` is local by construction.

## Behavior changes

| command | before | after |
|---|---|---|
| `analyze` (default `auto` / `auto`) | all local pairs | **1 pair** (best device + its best EP) |
| `--ep openvino` (device auto) | all OpenVINO devices + warning | **1 pair** |
| `--ep auto --device gpu` | all GPU EPs + warning | **1 pair** |
| `--ep all --device all` | full matrix | **unchanged** |

### Breaking: `auto` now requires local hardware

Because `auto` resolves from local availability (like `build` / `run`), pinning a **non-local** EP with `--device auto` now exits `2` instead of warning-and-running:

```
$ winml analyze -m model.onnx --ep qnn      # on a machine without QNN
ERROR  Could not auto-select a device: Requested EP 'qnn' is not available on this system.
```

To statically analyze an EP you don't physically have, specify the device explicitly (`--ep qnn --device npu`) or use `--device all`.

## Verified on real hardware

On a box with Intel OpenVINO (NPU + CPU) and NVIDIA TensorRT-RTX (GPU):

| command | execution target(s) |
|---|---|
| `analyze` (default) | `OpenVINOExecutionProvider (NPU)` |
| `--device gpu` | `NvTensorRTRTXExecutionProvider (GPU)` |
| `--ep openvino --device auto` | `OpenVINOExecutionProvider (NPU)` |
| `--ep openvino --device all` | `OpenVINO × NPU / GPU / CPU` |
| `--ep qnn` | exit 2 (QNN not available) |

### DIFF
--- a/src/winml/modelkit/commands/analyze.py
+++ b/src/winml/modelkit/commands/analyze.py
@@ -725,7 +725,8 @@ def _build_runtime_debug_output_path(model_path: Path, ep_name: str, device_name
     type=click.Choice([*ALL_EP_NAMES, "all", "auto"], case_sensitive=False),
     help=(
         "Target execution provider. Supports canonical names, aliases, and all/auto. "
-        "all = evaluate all rule-data-backed EPs; auto = infer from local availability"
+        "all = evaluate all rule-data-backed EPs; "
+        "auto = infer a single best target from local availability"
     ),
 )
 @click.option(
@@ -736,7 +737,8 @@ def _build_runtime_debug_output_path(model_path: Path, ep_name: str, device_name
     type=click.Choice([*SUPPORTED_DEVICES, "all", "auto"], case_sensitive=False),
     help=(
         "Target device type. Supports CPU/GPU/NPU and all/auto. "
-        "all = all rule-data-backed devices; auto = infer from local availability"
+        "all = all rule-data-backed devices; "
+        "auto = infer a single best target from local availability"
     ),
 )
 @cli_utils.verbosity_options()
@@ -897,13 +899,25 @@ def analyze(
             logger.error("Searched directories: %s", searched)
             sys.exit(2)
 
-        devices: list[str]
-        if device == "auto":
-            from ..sysinfo.device import _get_available_devices
+        # Resolve the EP/device selection. `all` keeps the full rule-data-backed
+        # set (fan-out, unchanged). `auto` resolves to a single best target from
+        # local availability via the shared sysinfo helpers — the same path
+        # build/run/perf use. A concrete value is used as-is.
+        from ..sysinfo import resolve_device, resolve_eps
 
-            devices = list(_get_available_devices())
-        elif device == "all":
+        # Only a pinned (concrete) EP can constrain device auto-resolution.
+        ep_hint = ep if ep not in ("auto", "all") else None
+
+        devices: list[str]
+        if device == "all":
             devices = list(SUPPORTED_DEVICES)
+        elif device == "auto":
+            try:
+                resolved_device, _ = resolve_device(device="auto", ep=ep_hint)
+            except (ValueError, RuntimeError) as e:
+                logger.error("Could not auto-select a device: %s", e)
+                sys.exit(2)
+            devices = [resolved_device]
         elif device is not None:
             devices = [device]
         else:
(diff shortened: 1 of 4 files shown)


chinazhangchao (approved):

DingmaomaoBJTU (commented):
Overall this is a clean, well-motivated change — aligning analyze auto-resolution with build/run/perf via shared sysinfo helpers. A few issues worth addressing.

src/winml/modelkit/commands/analyze.py
```diff
@@ -897,26 +899,55 @@ def analyze(
...
+            try:
+                resolved_device, _ = resolve_device(device="auto", ep=ep_hint)
+            except (ValueError, RuntimeError) as e:
+                logger.error("Could not auto-select a device: %s", e)
+                sys.exit(2)
+            devices = [resolved_device]
         elif device is not None:
             devices = [device]
         else:
             devices = []
```
DingmaomaoBJTU, line 928: **Bug (latent): `else: devices = []` + downstream `devices[0]` = unguarded IndexError**

Line 945 (`ref_device = devices[0]`) is only reached when `ep == "auto"` and `device` is not `"all"`. But this `else` branch leaves `devices = []`. If `device` ever arrives as `None` (direct programmatic call, future `--device` config-file key, etc.) the subsequent `devices[0]` raises `IndexError` with no error message and no clean exit code — bypassing the `except (ValueError, RuntimeError)` guard used everywhere else.

Suggested fix before `devices[0]`:
```python
if not devices:
    logger.error("No device context available for EP auto-resolution.")
    sys.exit(2)
ref_device = devices[0]
```

src/winml/modelkit/commands/analyze.py
```diff
@@ -897,26 +899,55 @@ def analyze(
...
+                except (ValueError, RuntimeError) as e:
+                    logger.error("Could not auto-select an execution provider: %s", e)
+                    sys.exit(2)
+            else:
+                ref_device = devices[0]
+            compatible_eps = resolve_eps(ref_device)
+            if not compatible_eps:
+                logger.error("No execution provider is available for device '%s'.", ref_device)
+                sys.exit(2)
+            eps = [compatible_eps[0]]
```
DingmaomaoBJTU, line 950: **Warning: `ep="auto"` + `device="all"` EP is chosen by ref_device heuristic, then fanned out across ALL supported devices**

Flow when both flags combine:
1. `ref_device` = best locally available device (e.g. NPU, line 940).
2. `compatible_eps[0]` = best EP for that specific device (e.g. `OpenVINOExecutionProvider`).
3. Pair-building (lines 961–968) then iterates over **all** `SUPPORTED_DEVICES` → `(OpenVINO, GPU)` and `(OpenVINO, CPU)` are included even though the EP selection was driven purely by NPU local availability.

Result: analysis runs for EP×device combos the machine can't execute, without any warning. Before this PR, the old fan-out for `auto` was at least symmetric. Consider either documenting this in the help text, or when `device=="all"` picking the EP per-device rather than by a single ref_device.

tests/e2e/test_analyze_e2e.py
```diff
@@ -321,50 +321,46 @@ def test_optim_config_writes_valid_json(
...
+        With a synthetic rule present the run must complete cleanly; the auto
+        axis resolves from the real ORT device map (CPU EP is always available
+        as a fallback), so only documented exit codes are asserted."""
         _write_supported_rule(rules_dir, "QNNExecutionProvider", "NPU")
         result = _invoke(["-m", str(onnx_model_path), "--quiet"])
-        # Aggregate result depends on whether every probed EP is fully
+        # Aggregate result depends on whether the resolved EP is fully
         # supported; only assert documented exit codes.
-        assert result.exit_code in {0, 1}
+        assert result.exit_code in {0, 1, 2}
```
DingmaomaoBJTU, line 363: **Warning: adding exit code `2` makes this test a no-op when EP enumeration completely fails**

Previously `{0, 1}`. `2` here means the test passes silently on any CI runner where `_get_device_ep_map_from_ort()` throws (no Windows ML ORT build, driver missing, etc.) — exactly the hidden failure mode users are most likely to hit. Consider monkeypatching `_get_device_ep_map_from_ort` to always return at least `{"cpu": ("CPUExecutionProvider",)}` so exit 2 is never an expected outcome for this happy-path test.

PR #920, opened 2026-06-18
fix(perf): render live HW chart for --monitor in --module mode (#654)

### LINKED ISSUE #654: perf --monitor is not working for --module

xieofxie:
## Summary

Fixes #654 — `perf --monitor` did nothing visible in `--module` mode.

In single-model mode, `--monitor` drives a live hardware-utilization chart via `_run_monitored_loop()` → `LiveMonitorDisplay`, which is exactly what the flag's help promises (*"Show live hardware utilization chart"*).

In `--module` mode, the monitored branch created an `HWMonitor` and collected metrics into the JSON report, but **never instantiated the live chart** — it just spun the raw inference loop. So from the user's perspective `--monitor` appeared to do nothing for `--module`.

## Fix

Route the per-module monitored loop through the same shared `_run_monitored_loop()` helper that single-model mode already uses. Each module instance now renders the live chart, while `hw_monitor` metrics are still persisted to the JSON report (unchanged).

## Testing

- Added `TestPerfModuleMonitor::test_monitor_drives_live_chart_per_module`, which asserts `_run_monitored_loop` is invoked once per module instance under `--module --monitor` and that the HW metrics still land in the JSON report. Verified it **fails** without the source fix and **passes** with it.
- `uv run pytest tests/unit/commands/test_perf_module.py` — 12 passed.

### DIFF
--- a/src/winml/modelkit/commands/perf.py
+++ b/src/winml/modelkit/commands/perf.py
@@ -9,6 +9,7 @@ Benchmarks model inference performance using WinMLAutoModel and WinMLSession.
 Usage:
     winml perf -m microsoft/resnet-50
     winml perf -m microsoft/resnet-50 --device npu --iterations 100
+    winml perf -m microsoft/resnet-50 --module ResNetConvLayer
     winml perf -m bert-base-uncased --task text-classification
 """
 
--- a/src/winml/modelkit/commands/perf.py
+++ b/src/winml/modelkit/commands/perf.py
@@ -890,6 +890,9 @@ def _perf_modules(
                             model_id=label,
                             device=resolved_device,
                         )
+                        # Collect inside the `with` block: hw_ctx.__exit__
+                        # stops the monitor, so to_dict() must read while it's
+                        # still live (mirrors the single-model path).
                         hw_metrics = hw.to_dict()
                 else:
                     with session.perf(warmup=warmup) as stats:
--- a/src/winml/modelkit/commands/perf.py
+++ b/src/winml/modelkit/commands/perf.py
@@ -876,9 +876,20 @@ def _perf_modules(
                         )
 
                 if hw_ctx:
+                    # Drive the same live chart single-model mode uses so
+                    # --monitor renders a per-module HW utilization chart
+                    # instead of silently dumping metrics to JSON (issue #654).
                     with session.perf(warmup=warmup) as stats, hw_ctx as hw:
-                        for _ in range(total_iters):
-                            session.run(inputs)
+                        _run_monitored_loop(
+                            session,
+                            inputs,
+                            stats,
+                            hw,
+                            total_iterations=total_iters,
+                            warmup=warmup,
+                            model_id=label,
+                            device=resolved_device,
+                        )
                         hw_metrics = hw.to_dict()
                 else:
                     with session.perf(warmup=warmup) as stats:
--- a/tests/unit/commands/test_perf_module.py
+++ b/tests/unit/commands/test_perf_module.py
@@ -316,6 +316,108 @@ class TestPerfModuleParameterForwarding:
         assert instance["running_model_path"] == str(running_model_path)
 
 
+class TestPerfModuleMonitor:
+    """--monitor must drive the live HW utilization chart in --module mode.
+
+    Regression guard for #654: previously the module path created an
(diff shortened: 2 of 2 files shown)


src/winml/modelkit/commands/perf.py
```diff
@@ -875,9 +876,20 @@ def _perf_modules(
...
+                            session,
+                            inputs,
+                            stats,
+                            hw,
+                            total_iterations=total_iters,
+                            warmup=warmup,
+                            model_id=label,
+                            device=resolved_device,
+                        )
                         hw_metrics = hw.to_dict()
```
DingmaomaoBJTU, line 896: **Info — correct but subtle ordering**: `hw.to_dict()` is called here while `hw_ctx` is still live (inside the `with` block), which is correct — the HWMonitor hasn't been stopped yet. If `_run_monitored_loop` raises, `hw_metrics` stays `None` and the exception propagates to the outer `try/except`. The placement is intentional and mirrors the single-model path at line 637, but it's worth a brief comment (e.g. `# collect before __exit__ stops the monitor`) since the ordering is load-bearing.

tests/unit/commands/test_perf_module.py
```diff
@@ -316,6 +316,108 @@ def test_running_model_path_in_module_result(self, tmp_path: Path) -> None:
...
+                    "--warmup",
+                    "0",
+                    "-o",
+                    str(out_path),
+                ],
+            )
+
+        assert result.exit_code == 0, result.output
+        # The live-chart loop must be driven once for the single module instance.
+        mock_loop.assert_called_once()
```
DingmaomaoBJTU: **Warning — assertion doesn't validate forwarded arguments**: `assert_called_once()` only confirms the loop was invoked; it doesn't verify that `total_iterations`, `warmup`, `model_id`, or `device` are forwarded correctly. For example, if someone accidentally passed `total_iterations=iterations` (dropping warmup), this assertion would still pass. Consider tightening to: `mock_loop.assert_called_once_with(ANY, ANY, ANY, ANY, total_iterations=1, warmup=0, model_id=ANY, device=ANY)` — this is the highest-value regression guard the test can add with minimal extra effort.

tests/unit/commands/test_perf_module.py
```diff
@@ -316,6 +316,108 @@ def test_running_model_path_in_module_result(self, tmp_path: Path) -> None:
...
+            patch(
+                "winml.modelkit.session.WinMLSession",
+                return_value=fake_session,
+            ),
+            patch(
+                "winml.modelkit.commands.perf.generate_random_inputs",
+                return_value={},
+            ),
+            patch(
+                "winml.modelkit.session.monitor.hw_monitor.HWMonitor",
```
DingmaomaoBJTU, line 390: **Info — patch target is correct but non-obvious**: Patching `winml.modelkit.session.monitor.hw_monitor.HWMonitor` (source module) rather than `winml.modelkit.commands.perf.HWMonitor` (call site) is correct here because the import is lazy — it happens inside the function body at line 869 of perf.py. At call time Python fetches the attribute from the already-patched module object, so `fake_hw_cls` is what the function receives. A one-line comment would help future readers avoid confusion (e.g. `# lazy import inside fn — must patch source module, not call site`).

xieofxie:
Thanks for the review — addressed all three in e8015a7e:

1. **`hw.to_dict()` ordering** — added a comment explaining the collect-before-`__exit__` ordering is load-bearing.
2. **Assertion tightening** — switched to `assert_called_once_with(..., total_iterations=1, warmup=0, model_id=ANY, device="cpu")` so the forwarded benchmark params are validated, not just the call.
3. **Patch target** — added a one-liner noting the lazy import means we patch the source module rather than the call site.

DingmaomaoBJTU (approved):
LGTM! All three findings from the previous review are addressed:

- **WARNING resolved**: assert_called_once_with now tightly validates all positional args and keyword params (including device="cpu" rather than ANY), catching parameter-forwarding regressions.
- **INFO resolved**: Comment on hw.to_dict() explains why it must stay inside the with hw_ctx as hw: block, guarding against future reordering.
- **INFO resolved**: Comment on the HWMonitor patch explains the lazy-import rationale, preventing a fix to the wrong patch target.

The core fix (routing --module --monitor through _run_monitored_loop) is clean and mirrors the single-model path correctly.

PR #922, opened 2026-06-19
fix analyze coverage bugs

fangyangci:


src/winml/modelkit/analyze/utils/rule_loader.py
```diff
@@ -61,18 +61,20 @@ def _get_env_rules_dirs(env_name: str) -> list[Path]:
...
+        2. If :data:`WINMLCLI_RULES_DIR` is unset/empty, use the embedded default
+            directory (``src/winml/modelkit/analyze/rules/runtime_check_rules/``).
 
     Returns:
         List of directory Paths (may include non-existent ones; callers filter).
     """
     dirs = _get_env_rules_dirs(WINMLCLI_RULES_DIR_ENV)
-    dirs.append(_DEFAULT_RUNTIME_RULES_DIR)
-    return dirs
+    if dirs:
```
DingmaomaoBJTU: **Bug: WINMLCLI_RULES_DIR with multiple paths now only searches the first directory.** The old code iterated _all_ search dirs and returned the first existing file. Now 
esolve_rule_parquet_path returns search_dirs[0] / relative_path unconditionally — the existence check in _load_parquet_rule_table only probes that one path. If a user sets WINMLCLI_RULES_DIR=/a:/b and the parquet exists in /b but not /a, it was found before (silently falls through to rules_not_found now). The behavior change from 'env supplements default' to 'env replaces default' makes this worse: users who depended on the fallback to embedded rules when their custom dir doesn't have a file are also broken.

src/winml/modelkit/analyze/utils/rule_loader.py
```diff
@@ -108,21 +109,22 @@ def _infer_ep_device_subdir(filename: str) -> str | None:
...
+    )
 
     if for_debug:
-        for debug_dir in get_runtime_rules_debug_search_dirs():
-            candidate_in_subdir = debug_dir / ep_device_subdir / parquet_filename
-            if candidate_in_subdir.exists():
-                return candidate_in_subdir
+        debug_dirs = get_runtime_rules_debug_search_dirs()
+        if debug_dirs:
+            return debug_dirs[0] / relative_path
```
DingmaomaoBJTU, line 125: **Bug: or_debug=True no longer falls back to normal search dirs.** The old code searched debug dirs first and, if not found, continued to normal search dirs. Now if any debug dirs are configured (WINMLCLI_RULES_DIR_FOR_DEBUG is set), the function returns the first debug dir candidate without checking existence and without falling back. If the file doesn't exist in the debug dir but does in the normal rules dir, the result is 
ules_not_found instead of finding the file. The test 	est_resolve_parquet_for_debug_falls_back_to_rules_dir was specifically catching this case and is now removed.

src/winml/modelkit/analyze/core/runtime_checker_query.py
```diff
@@ -1935,41 +1937,42 @@ def _load_parquet_rule_table(
...
+                The resolved or expected parquet path for lookup,
+                loaded dataframe when available, otherwise None,
                 and optional pre-built condition tree.
         """
         parquet_name = (
             f"{op_name}_{self.ep_name}_{self.device_type.upper()}_{op_domain.name}"
             f"_opset{op_since_version}{'_qdq' if is_qdq else ''}.parquet"
         )
         parquet_path = resolve_rule_parquet_path(parquet_name, for_debug=for_debug)
+        resolved_parquet_path = parquet_path if parquet_path.exists() else None
```
DingmaomaoBJTU: Nit: 
esolved_parquet_path is used only in the if resolved_parquet_path is None branch (three lines below). Consider inlining it as if not parquet_path.exists() to keep the variable count low, since parquet_path is always a Path now.

DingmaomaoBJTU (commented):
Review of PR #922 (fix analyze coverage bugs). Three separate bug fixes are bundled here. The 
ode_stable_key fix and return-tuple reorder are correct. The 
esolve_rule_parquet_path refactor introduces regressions described in inline comments.

src/winml/modelkit/analyze/core/runtime_checker_query.py
```diff
@@ -1935,41 +1937,42 @@ def _load_parquet_rule_table(
...
 
         cache_key = (op_name, op_domain.value, op_since_version, is_qdq)
         if cache_key in self._parquet_rule_table_cache:
-            if parquet_path is not None:
+            if self._parquet_rule_table_cache[cache_key] is not None:
```
DingmaomaoBJTU, line 1961: **Correctness concern on cache hit path:** When the cache already has None (i.e., the file was previously not found), parquet_path is recomputed via 
esolve_rule_parquet_path on every call. If the environment changes between calls (e.g., WINMLCLI_RULES_DIR set after first run), the cache will return stale None but parquet_path will reflect the new env. The caller then reports the new path in debug details while still returning no data — misleading. This existed before but is now more visible. Consider caching parquet_path too, or documenting that the cache is instance-scoped and env must not change mid-run.

src/winml/modelkit/analyze/core/runtime_checker.py
```diff
@@ -221,12 +221,11 @@ def op_support(
         run_for_node_total_ms = 0
         callback_total_ms = 0
 
-        # Get all nodes from model
-        model_proto = self._model.get_model()
         # Get cached RuntimeCheckerQuery
         query = self._get_query()
+        # Use the same graph snapshot as RuntimeCheckerQuery (post shape inference).
+        nodes = query.model_proto.graph.node
```
DingmaomaoBJTU, line 227: Fix looks correct: using query.model_proto.graph.node ensures the iterated node objects are the same Python instances that query._node_key_by_node_id was built from (via id(node) keys). The old code used the raw (pre-inference) model's nodes, whose Python object identities would differ from the post-inference nodes held by the query, causing id() lookups to miss and falling back to slower/potentially incorrect name resolution for unnamed nodes.

tests/unit/analyze/models/test_rule_loader.py
```diff
@@ -646,4 +645,4 @@ def test_resolve_parquet_for_debug_falls_back_to_rules_dir(self, monkeypatch):
             monkeypatch.setenv(WINMLCLI_RULES_DIR_FOR_DEBUG_ENV, debug_tmp)
```
DingmaomaoBJTU, line 661: **Missing regression test for multi-dir search.** The PR removes 	est_resolve_parquet_for_debug_falls_back_to_rules_dir, which caught a real user scenario. Even under the new semantics (no fallback), there is no test verifying that when WINMLCLI_RULES_DIR=/a:/b and the file exists in /b but not /a, the system correctly reports 
ules_not_found (current behavior) rather than silently returning a wrong path. A test for the multi-dir case would make the new behavior explicit.

src/winml/modelkit/analyze/utils/rule_loader.py
fangyangci: Resolved in 6a4d1e29. Single-dir is intentional: WINMLCLI_RULES_DIR now resolves exactly one directory and is no longer split on os.pathsep, and it replaces (not supplements) the embedded default. The half-supported 'parse many, probe only the first' behavior is removed so lookup is deterministic with no surprising fallback. Docs/tests updated to match.

src/winml/modelkit/analyze/utils/rule_loader.py
fangyangci: Resolved in 6a4d1e29. The debug->normal fallback removal is intentional — --debug should only consult the debug dir, never silently fall through to the embedded/normal rules. Single-dir + no-fallback is now the documented, tested behavior.

src/winml/modelkit/analyze/core/runtime_checker_query.py
fangyangci: Done in 6a4d1e29 — dropped resolved_parquet_path and inlined to 'if not parquet_path.exists()'. As a bonus this moves the existence probe after the cache-hit early return, so cache hits no longer do a filesystem stat.

src/winml/modelkit/analyze/core/runtime_checker_query.py
fangyangci: Addressed in 6a4d1e29 with a doc comment: the per-instance cache assumes the rule-dir env vars are stable for the query's lifetime. Mid-run env changes are out of scope (an instance is short-lived within a single analyze invocation), so caching the path was not worth the added complexity.

src/winml/modelkit/analyze/core/runtime_checker.py
fangyangci: Thanks — confirming, no change needed. Resolving.

tests/unit/analyze/models/test_rule_loader.py
fangyangci: Added in 6a4d1e29: test_env_var_not_split_on_pathsep and test_debug_env_var_not_split_on_pathsep make the new single-dir semantics explicit (a value containing os.pathsep is treated as one literal dir; embedded default is not consulted as fallback).

src/winml/modelkit/analyze/utils/rule_loader.py
```diff
@@ -46,54 +46,59 @@ def _resolve_env_rules_dir_entry(entry: str) -> Path:
...
+        Single-element list with the selected directory (the embedded default
+        when the env var is unset). The directory may not exist; callers filter.
     """
-    dirs = _get_env_rules_dirs(WINMLCLI_RULES_DIR_ENV)
-    dirs.append(_DEFAULT_RUNTIME_RULES_DIR)
-    return dirs
+    env_dir = _get_env_rules_dir(WINMLCLI_RULES_DIR_ENV)
+    if env_dir is not None:
+        return [env_dir]
+    return [_DEFAULT_RUNTIME_RULES_DIR]
```
DingmaomaoBJTU, line 79: **Bug 1 resolved.** The multi-dir ambiguity is eliminated: _get_env_rules_dir (singular) no longer splits on os.pathsep, making the single-directory contract explicit and deterministic. The embedded default is still used when the env var is unset — correct. README and CHANGELOG are updated to match. ✅

src/winml/modelkit/analyze/utils/rule_loader.py
```diff
@@ -108,21 +113,22 @@ def _infer_ep_device_subdir(filename: str) -> str | None:
...
+    )
 
     if for_debug:
-        for debug_dir in get_runtime_rules_debug_search_dirs():
-            candidate_in_subdir = debug_dir / ep_device_subdir / parquet_filename
-            if candidate_in_subdir.exists():
-                return candidate_in_subdir
+        debug_dirs = get_runtime_rules_debug_search_dirs()
+        if debug_dirs:
+            return debug_dirs[0] / relative_path
```
DingmaomaoBJTU, line 125: **Bug 2 — intentional behavioral change, documented.** When WINMLCLI_RULES_DIR_FOR_DEBUG is set, or_debug=True now returns the debug-dir candidate without checking existence and without falling back to normal rules. The previous fallback is intentionally removed in favor of deterministic lookup. This is documented in the commit message, the test was renamed to make the intent explicit, and the README reflects it. One note: if someone sets WINMLCLI_RULES_DIR_FOR_DEBUG to a partial set of rules expecting normal rules to fill gaps, they'll get 
ules_not_found silently. The new behavior is clear but could be worth calling out in the README under 'What happens if parquet rules are missing' to set expectations.

src/winml/modelkit/analyze/core/runtime_checker_query.py
```diff
@@ -1950,26 +1952,30 @@ def _load_parquet_rule_table(
         )
         parquet_path = resolve_rule_parquet_path(parquet_name, for_debug=for_debug)
 
+        # This per-instance cache assumes a stable rules location for the query's
+        # lifetime: the rule-dir env vars must not change between calls. The path
+        # is recomputed each call (so reporting reflects the current location),
+        # but a cached None is reused without re-probing the filesystem.
```
DingmaomaoBJTU, line 1958: **Warning addressed.** The 
esolved_parquet_path temporary is gone; parquet_path.exists() is now called after the cache-hit early return (line 1969), avoiding a filesystem probe on every cache hit. The comment at line 1955–1958 documents the stable-env-vars assumption clearly. ✅

DingmaomaoBJTU (commented):
Follow-up review of PR #922. Both previously flagged bugs in resolve_rule_parquet_path are addressed — Bug 1 via an explicit single-directory design, Bug 2 via an intentional semantic change (documented and tested). The cache warning is addressed with an explanatory comment. One minor note inline.

DingmaomaoBJTU (approved):
All previously flagged bugs are addressed. Bug 1 (multi-dir search) is resolved by making single-directory support explicit — _get_env_rules_dir no longer splits on os.pathsep, and the README/CHANGELOG are updated. Bug 2 (for_debug fallback removal) is an intentional semantic change: when WINMLCLI_RULES_DIR_FOR_DEBUG is set, lookup returns the debug-dir candidate exclusively; this is documented, tested, and explained in the commit message. The cache warning is addressed with a comment and the parquet_path.exists() probe is correctly moved after the cache-hit return. The 
ode_stable_key fix and tuple-reorder fix are correct and well-tested. LGTM.

PR #926, opened 2026-06-22
fix analyze e2e test

fangyangci:


tests/e2e/test_analyze_e2e.py
```diff
@@ -321,50 +335,46 @@ def test_optim_config_writes_valid_json(
...
+        With a synthetic rule present the run must complete cleanly; the auto
+        axis resolves from the real ORT device map (CPU EP is always available
+        as a fallback), so only documented exit codes are asserted."""
         _write_supported_rule(rules_dir, "QNNExecutionProvider", "NPU")
         result = _invoke(["-m", str(onnx_model_path), "--quiet"])
-        # Aggregate result depends on whether every probed EP is fully
+        # Aggregate result depends on whether the resolved EP is fully
         # supported; only assert documented exit codes.
-        assert result.exit_code in {0, 1}
+        assert result.exit_code in {0, 1, 2}
```
xieofxie: a little unuseful.. could we make sure only 1 of these exit codes selected?

fangyangci: fixed

tests/e2e/test_analyze_e2e.py
```diff
@@ -349,18 +363,28 @@ def test_default_device_auto_resolves_single_best_device_for_pinned_ep(
...
         """Omitting ``--ep`` resolves a single best EP from local availability.
 
-        With a synthetic rule present the run must complete cleanly; the auto
-        axis resolves from the real ORT device map (CPU EP is always available
-        as a fallback), so only documented exit codes are asserted."""
+        The test is hardware-agnostic: local availability is controlled via the
+        ORT device->EP map monkeypatch rather than real machine capabilities.
+        With this synthetic map, auto resolves to QNN on NPU and should be fully
+        supported."""
+        monkeypatch.setattr(
```
xieofxie: given it is in e2e, I think we should not patch codes.

Consider using `_require_npu, require_ep` help function for qnn + npu to make it return 0

Also even could check not 0 if not satisfied ?

fangyangci: Good call — done in 6c172161. Both auto-resolution tests now drop the _get_device_ep_map_from_ort monkeypatch and gate on real hardware via require_ep("qnn"), then assert exit 0. On a QNN host NPU is the highest-priority device and QNN its highest-priority EP, so both pinned-ep-auto-device and bare-auto resolve to (qnn, NPU); non-QNN hosts skip cleanly. The hardware-agnostic auto-resolution logic stays covered at the unit level (test_static_analyzer_cli.py::test_selection_matrix). Kept it to the positive exit-0 path; the 'assert non-0 when not satisfied' negative case is really require_not_ep territory and out of scope here.

DingmaomaoBJTU (approved):

PR #928, opened 2026-06-22
clean up unused codes

### LINKED ISSUE #913: Clean up unused inject_hub_metadata in utils/hub_utils.py

xieofxie:
Close https://github.com/microsoft/winml-cli/issues/913


DingmaomaoBJTU (approved):
Clean removal of dead code. No issues found.

Issue #449, opened 2026-05-06
Implement --compare for winml perf: side-by-side benchmark across all device+EP combinations


hi-brenda:
## Summary

Add `--compare` to `winml perf` to run a benchmark across all available device+EP combinations in a single invocation and display a side-by-side comparison table with a recommendation. Currently users must run `winml perf` separately for each device/EP and mentally compare results.

## Context

A GPU may expose multiple EPs (e.g. OpenVINO and TensorRT RTX). A simple device-level flag would miss these variants — the comparison must be at the **device+EP** level. `--compare` captures this correctly.

Note: `winml perf` already has a `--compare-devices` stub at [perf.py:1126](src/winml/modelkit/commands/perf.py) that prints `"not yet implemented"`. `--compare` supersedes this stub.

## Current Pain

```bash
# User must run three (or more) separate commands and mentally compare:
winml perf -m ProsusAI/finbert --device npu
winml perf -m ProsusAI/finbert --device gpu --ep openvino
winml perf -m ProsusAI/finbert --device gpu --ep nv_tensorrt_rtx
winml perf -m ProsusAI/finbert --device cpu
```

## Proposed UX

```bash
winml perf -m ProsusAI/finbert --compare          # auto-detect all available device+EP combos
winml perf -m ProsusAI/finbert --compare npu,gpu  # limit to specific devices
```

### Side-by-side output table

```
  Model   ProsusAI/finbert · text-classification · 100 iterations
  ┌─────────┬─────────────────┬───────────┬──────────┬──────────┬────────────┬─────────┬─────────────────┐
  │ Device  │ EP              │ Mean (ms) │ P50 (ms) │ P99 (ms) │ Throughput │ Memory  │                 │
  ├─────────┼─────────────────┼───────────┼──────────┼──────────┼────────────┼─────────┼─────────────────┤
  │ NPU     │ QNN             │    18.8   │   18.1   │   24.3   │  53 sps    │ 142 MB  │ ✓ Recommended   │
  │ GPU     │ OpenVINO        │    27.4   │   26.9   │   61.2   │  36 sps    │ 389 MB  │                 │
  │ GPU     │ TensorRT RTX    │    22.1   │   21.8   │   38.5   │  45 sps    │ 512 MB  │                 │
  │ CPU     │ CPU EP          │    94.3   │   93.1   │  102.8   │  11 sps    │ 210 MB  │                 │
[...]

PR #930, opened 2026-06-22
perf: scale throughput by effective batch size, not requested --batch-size

### LINKED ISSUE #155: P1-FEATURE-002: wmk perf Command — Improve Existing
## Summary

Improve the existing `wmk perf` command with expanded metrics (latency/throughput/memory), batch size sweep, and cross-EP comparison report.

## Context

`wmk perf` currently provides basic performance benchmarking. For the May 1 delivery, it needs to be production-ready with expanded metrics, batch size support, and cross-EP comparison capability to serve as the foundation for the cross-EP benchmarking effort (P1-EP-011).

From `plans/release/0501_release_plan/P0_CHECKLIST.md` (P1-FEATURE-002).

## Current State

- `wmk perf` command exists and runs basic benchmarks
- Limited metrics (latency only)
- Single EP per run; no cross-EP comparison
- Related tracking: #305 (need more perf results)

## Desired State

- `wmk perf` reports: latency (mean/p50/p95/p99 ms), throughput (fps/tokens/s), memory (peak RSS)
- Batch size sweep: runs with batch sizes 1, 4, 8, 16 (configurable)
- `--ep all` option: runs same model on all available EPs and produces comparison table
- Output: JSON + human-readable terminal table

## Acceptance Criteria

- [ ] Latency statistics: mean, p50, p95, p99 in milliseconds
- [ ] Throughput: samples/s or tokens/s depending on task type
- [ ] Memory: peak RSS measured per EP run
- [ ] Batch size sweep: `--batch-sizes 1,4,8,16`
- [ ] Cross-EP comparison: `--ep all` runs all available EPs and outputs comparison
- [ ] Output: `artifacts/perf_results.json` + formatted terminal table
- [ ] Warm-up runs configurable (default 10), measurement runs configurable (default 50)
- [ ] All P0 built-in models pass `wmk perf` after this improvement

## Technical Notes

- Build on existing `wmk perf` implementation
- Use `time.perf_counter_ns` for high-resolution timing
- Memory: use `psutil.Process().memory_info().rss` for peak measurement
- Warm-up runs prevent cold-start skew

## Related Files

- `plans/release/0501_release_plan/feature-scale.md` — P1.2 wmk perf improvement
- `plans/release/0501_release_plan/P0_CHECKLIST.md` — P1-FEATURE-002
- #305 — need more perf results (related)

xieofxie:
## Summary

Fixes incorrect `samples_per_sec` in `wmk perf` when the requested `--batch-size` cannot be applied to the model.

The requested `--batch-size` only lands on inputs whose **leading dimension is dynamic** (see `_resolve_shape`). For a model with a statically-fixed batch dim (e.g. `[1, 3, 224, 224]`), input generation silently keeps the static value, so the session runs batch=1 — but throughput was still computed as `config.batch_size / latency`, inflating `samples_per_sec` by the requested batch factor (e.g. reporting 800 sps when the model actually processed 100). Latency stats were always correct; only the derived per-sample throughput was wrong.

## Changes

- **New helper `effective_batch_size()`** — reads the actual batch back from the first batched (rank ≥ 1) generated input, matching the module's existing "first dim is batch" convention. Falls back to the requested value when all inputs are scalar.
- **`_collect_results()`** now scales `samples_per_sec` by the effective batch the session ran, not the requested `config.batch_size`. (`batches_per_sec` was already correct — it is per-call.)
- **`_generate_inputs()`** logs a warning when the requested `--batch-size` could not be applied (static batch dim).
- **`BenchmarkResult`** gains an `effective_batch_size` field, surfaced in the JSON `benchmark_info` alongside the requested `batch_size`, and shown in the console throughput line with a `Note:` when the two differ.

## Tests

New `TestEffectiveBatchSize` in `tests/unit/commands/test_perf_cli.py` covers the helper (dynamic / static / scalar / fallback), throughput scaling for both static-batch (regression guard: 100 not 800 sps) and dynamic-batch cases, the warning on mismatch, and the JSON field. Full `test_perf_cli.py`, `test_perf_module.py`, and `test_perf_composite.py` suites pass.

## Scope note

This addresses the batch-size correctness aspect of the epic. The broader items (batch-size **sweep** `--batch-sizes 1,4,8,16` and `--ep all` cross-EP comparison) are not part of this PR.

Closes #155

### DIFF
--- a/src/winml/modelkit/commands/perf.py
+++ b/src/winml/modelkit/commands/perf.py
@@ -570,7 +570,7 @@ class PerfBenchmark:
             self.config.batch_size,
         )
         if self.config.batch_size != 1 and self._effective_batch != self.config.batch_size:
-            logger.info(
+            logger.warning(
                 "Requested --batch-size %d could not be applied: the model's "
                 "leading input dimension is statically %d. Throughput is scaled "
                 "by the actual batch (%d), not the requested value.",
--- a/src/winml/modelkit/commands/perf.py
+++ b/src/winml/modelkit/commands/perf.py
@@ -130,6 +130,12 @@ class BenchmarkResult:
     samples_per_sec: float = 0.0
     batches_per_sec: float = 0.0
 
+    # Batch dimension the session actually ran. Equals config.batch_size when
+    # the model's leading input dim is dynamic; falls back to the model's
+    # static batch (often 1) otherwise. samples_per_sec is scaled by this, not
+    # by the requested config.batch_size.
+    effective_batch_size: int = 1
+
     # Actual values used (after auto-detection)
     actual_device: str = ""
     actual_task: str = ""
@@ -159,6 +165,7 @@ class BenchmarkResult:
                 "iterations": self.config.iterations,
                 "warmup": self.config.warmup,
                 "batch_size": self.config.batch_size,
+                "effective_batch_size": self.effective_batch_size,
                 "timestamp": self.timestamp,
             },
             "model_info": {
@@ -279,6 +286,32 @@ def _resolve_shape(
     return tuple(resolved)
 
 
+def effective_batch_size(
+    inputs: dict[str, np.ndarray],
+    input_names: list[str],
+    requested: int,
+) -> int:
+    """The batch dimension actually present in the generated inputs.
+
+    The requested ``--batch-size`` only lands on inputs whose leading
+    dimension is dynamic; a model with a statically-fixed batch dim ignores
+    it (see :func:`_resolve_shape`). Throughput (samples/sec) must be scaled
+    by what the session actually ran, not by what was asked, or a static-batch
+    model reports ``requested / latency`` while only processing one batch per
+    call -- inflating samples/sec by ``requested``.
+
+    Reads the leading dim back from the first batched (rank >= 1) input,
+    matching the "first dim is batch" convention used throughout this module.
+    Falls back to ``requested`` when no batched input exists (e.g. all-scalar
+    inputs), which preserves the prior behavior for that edge case.
+    """
+    for name in input_names:
+        arr = inputs.get(name)
+        if arr is not None and arr.ndim >= 1:
+            return int(arr.shape[0])
+    return requested
+
+
 # =============================================================================
 # Benchmark Engine
 # =============================================================================
@@ -302,6 +335,7 @@ class PerfBenchmark:
         self.config = config
         self._model: WinMLPreTrainedModel | WinMLCompositeModel | None = None
         self._inputs: dict[str, np.ndarray] | None = None
+        self._effective_batch: int = config.batch_size
         self._memory: dict[str, float] | None = None
 
     @property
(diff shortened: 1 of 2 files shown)


xieofxie:
For the issue 155, only the remaining `--ep all` is tracked via https://github.com/microsoft/winml-cli/issues/449, so close it

src/winml/modelkit/commands/perf.py
```diff
@@ -525,10 +559,25 @@ def _load_model(self) -> None:
...
-            io_config=self._single.io_config,
+            io_config=io_config,
             batch_size=self.config.batch_size,
         )
+        self._effective_batch = effective_batch_size(
+            self._inputs,
+            io_config["input_names"],
+            self.config.batch_size,
+        )
+        if self.config.batch_size != 1 and self._effective_batch != self.config.batch_size:
```
DingmaomaoBJTU: The `self.config.batch_size != 1` guard suppresses this logger warning for the edge case where `batch_size=1` (the default) but `effective_batch != 1` (e.g. a model with a static batch of 2). Throughput is still computed correctly (uses `_effective_batch`), and `display_console_report` shows the `Note:` regardless — so there is no correctness issue. But the asymmetry is subtle.

Consider simplifying the guard to just `self._effective_batch != self.config.batch_size`: the second condition already handles the no-mismatch case, so removing the `!= 1` check makes the logger fire consistently with the console Note.

DingmaomaoBJTU (commented):
Overall this is a clean, well-motivated fix. The core bug (samples_per_sec inflated by requested batch on static-batch models) is correctly diagnosed and fixed. The effective_batch_size helper is clear and its fallback behavior is well-documented. Tests cover the important scenarios including the regression guard. A few minor observations below.

src/winml/modelkit/commands/perf.py
```diff
@@ -279,6 +286,32 @@ def _resolve_shape(
...
+    call -- inflating samples/sec by ``requested``.
+
+    Reads the leading dim back from the first batched (rank >= 1) input,
+    matching the "first dim is batch" convention used throughout this module.
+    Falls back to ``requested`` when no batched input exists (e.g. all-scalar
+    inputs), which preserves the prior behavior for that edge case.
+    """
+    for name in input_names:
+        arr = inputs.get(name)
+        if arr is not None and arr.ndim >= 1:
```
DingmaomaoBJTU, line 315: Returns the leading dim of the **first batched input**. For multimodal or encoder-decoder models where different inputs may have different leading dims (e.g., image batch of 4 alongside a scalar token-count), this is still correct under the existing 'first dim is batch' convention. But if two batched inputs have different leading dims, the reported `effective_batch_size` silently reflects only the first. Worth adding a note to the docstring about this single-input assumption, or optionally asserting that all batched inputs agree on the leading dim.

src/winml/modelkit/commands/perf.py
xieofxie: Done — dropped the redundant `!= 1` check so the logger warning now fires consistently with the console Note whenever `_effective_batch != config.batch_size`. (3776265d)

src/winml/modelkit/commands/perf.py
xieofxie: Added a note to the docstring documenting the single-input assumption (first batched input only; multimodal/encoder-decoder inputs that disagree on the leading dim reflect only the first). (3776265d)

DingmaomaoBJTU (approved):
Both prior findings addressed in the follow-up commit:

1. The redundant `batch_size != 1` guard was dropped — `logger.warning` now fires consistently with the console Note whenever effective and requested batch differ.
2. The `effective_batch_size()` docstring now explicitly documents the single-input assumption for multimodal models.

Implementation is correct, well-tested, and the code is clean. LGTM.

PR #937, opened 2026-06-23
fix: declare psutil as a runtime dependency

### LINKED ISSUE #936: winml perf crashes on clean install — psutil imported but not a runtime dependency
Found during the v0.2.0 Gate 3 bug bash.

## Symptom
On a clean `pip install winml-cli`, any `winml perf` invocation (with or without `--monitor` / `--memory`) fails immediately:

```
Error: Benchmark failed: No module named 'psutil'
```

## Root cause
`src/winml/modelkit/session/monitor/memory_tracker.py:13` has a module-level `import psutil`, and the perf flow imports `memory_tracker` unconditionally. `psutil` is **not** declared in `[project].dependencies` in `pyproject.toml` — only the dev type stub `types-psutil` is present. As a result the published wheel's `Requires-Dist` omits `psutil`, so it is never installed for end users.

## Regression
Introduced by #861 (`feat: add --memory flag to measure RAM and VRAM per phase`). `memory_tracker.py` did not exist at v0.1.0, so `winml perf` was unaffected there.

## Impact
Blocks v0.2.0 — `winml perf` is a headline command and is broken out-of-the-box for every clean install. Installing `psutil` manually confirms perf/build/`--monitor` otherwise work correctly, so the only defect is the missing dependency declaration.

## Fix
Add `psutil` to `[project].dependencies` in `pyproject.toml`.

timenick:
## Summary

`winml perf` crashes on a clean install with `No module named 'psutil'` because `src/winml/modelkit/session/monitor/memory_tracker.py` imports `psutil` at module level, and the perf flow imports that module unconditionally. `psutil` was never declared in `[project].dependencies` — only the dev type stub `types-psutil` is present — so the published wheel's `Requires-Dist` omits it, breaking `winml perf` (and `--monitor` / `--memory`) out-of-the-box for every user.

Regression from #861 (`feat: add --memory flag`); `memory_tracker.py` did not exist in v0.1.0, so `winml perf` was unaffected there. Installing `psutil` manually confirms perf/build/`--monitor` otherwise work correctly — the only defect is the missing dependency declaration.

## Change

Add `psutil>=7` to `[project].dependencies` (aligns with the existing `types-psutil>=7.2.2` stub).

Targeting `release/v0.2.0` directly as a release hotfix; `main` will pick it up via the post-release merge-back.

Closes #936

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -56,6 +56,7 @@ dependencies = [
   # non-functional but huge diffs across every generated rule artifact. If you
   # truly must bump this, regenerate ALL parquet artifacts in the same change.
   "pandas==2.3.3",
+  "psutil>=7",
   "pydantic>=2",
   "python-multipart>=0.0.22",
   "rapidfuzz>=3.9",


DingmaomaoBJTU (approved):
The fix is correct and complete. `memory_tracker.py` imports `psutil` unconditionally at module level (line 13), so any user running `winml perf` / `--monitor` / `--memory` on a clean install hits a hard `ModuleNotFoundError`. The dependency was previously declared only as a dev type stub (`types-psutil>=7.2.2` in `[dependency-groups].dev`), so the published wheel's `Requires-Dist` omitted it entirely.

The single-line addition of `psutil>=7` to `[project].dependencies` is the right surgical fix:
- Placement is alphabetically correct (between `pandas` and `pydantic`).
- The `>=7` lower bound aligns with the existing `types-psutil>=7.2.2` stub and the stable `psutil.Process.memory_info().rss` API used in the tracker.
- No other runtime code paths need updating; the underlying logic was always correct.

Approve.

PR #940, opened 2026-06-23
Pipeline(E2E): align ModelKitArtifacts with release branch at runtime

KayMKM:
## Problem

The E2E Test pipeline is supposed to check out the **same-named `release/*` branch**
of `ModelKitArtifacts` when it runs on a release branch, so the artifact content
matches the modelkit release. This never actually worked:

- The artifacts repo was selected via the repository resource's `ref`
  (`ref: ${{ variables.artifactsRef }}`, driven by a compile-time `variables`
  block). A repository resource `ref` is resolved **before** the pipeline starts
  and only accepts template expressions over `parameters` and **system predefined**
  variables — not YAML/UI-defined variables. So the expression was silently
  ignored and ModelKitArtifacts always checked out its default branch (`main`),
  even on `release/*` runs.
- Pointing `ref` directly at `Build.SourceBranch` does compile, but then the
  resource tries to check out *whatever* branch the pipeline runs on, which fails
  outright when ModelKitArtifacts has no matching branch (feature/ad-hoc runs):
  `No commit found for SHA: refs/heads/<branch>`.

In short, the branch of a repository resource cannot be conditionally chosen at
resource-resolution time.

## Fix

Select the artifacts branch **at runtime** instead:

- Pin the `ModelKitArtifacts` resource to `ref: main` (always exists), and remove
  the unusable `variables` indirection.
- Add a job step **"Align ModelKitArtifacts with release branch"** that, when
  `Build.SourceBranch` is `refs/heads/release/*`, switches the already-checked-out
  artifacts repo to the same-named release branch (`git fetch` + `git checkout` +
  `git lfs pull`). It falls back to `main` with a warning when no matching branch
  exists, so non-release runs (main schedule, manual, feature branches) keep
  working instead of failing at init.
- Enable `persistCredentials: true` on the artifacts checkout so the runtime
  fetch / lfs pull is authenticated.

## Result

- `release/*` run → ModelKitArtifacts switches to the same-named release branch ✓
- `main` / manual / feature run → stays on `main`, no init failure ✓

Verified on a live pipeline run.

### DIFF
--- a/.pipelines/Modelkit E2E Test.yml
+++ b/.pipelines/Modelkit E2E Test.yml
@@ -24,10 +24,10 @@
 #     an automatic E2E run. (Subsequent pushes to a release branch will
 #     also trigger a run, which is acceptable since release branches
 #     typically see few commits.) When triggered from a release branch,
-#     the ModelKitArtifacts repo is also checked out from the same-named
-#     release branch (see `resources.repositories.ModelKitArtifacts.ref`
-#     below); main is used for all other triggers (scheduled, manual on
-#     main, etc.).
+#     the job switches the checked-out ModelKitArtifacts repo to the
+#     same-named release branch at runtime (see the "Align ModelKitArtifacts
+#     with release branch" step in templates/e2e-test-jobs.yml); main is used
+#     for all other triggers (scheduled, manual on main, feature branches).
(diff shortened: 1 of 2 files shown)


xieofxie (approved):

xieofxie:
please update title

Issue #939, opened 2026-06-23
refactor: fold _perf_modules into PerfBenchmark


xieofxie:
## Summary

`winml perf --module` is handled by a standalone `_perf_modules()` function in `src/winml/modelkit/commands/perf.py`, separate from the `PerfBenchmark` class used for single-model and composite runs. The two paths duplicate build + benchmark orchestration (input generation, session compile, monitored/simple loops, HW monitor wiring, result collection) and have already drifted (e.g. device/EP resolution lives in different places).

## Motivation

While fixing #931 (perf with no `--ep` should target one concrete EP, not aggregate all), device/EP resolution was moved **inside** `PerfBenchmark` (it resolves `config.device`/`config.ep` at the start of `_load_model`, failing fast before the build). The `--module` path still resolves the EP in the CLI `perf()` function because `_perf_modules` is not part of `PerfBenchmark`. This leaves a transitional duplication:

- `PerfBenchmark._resolve_device_ep()` — single-model / composite
- inline `resolve_device`/`resolve_eps` block in `perf()`'s module branch — per-module

## Proposal

Fold per-module benchmarking into `PerfBenchmark` so there is a single orchestration path:

- A per-module mode on `PerfBenchmark` (or a small subclass) that builds each submodule ONNX and benchmarks it through the same input-gen / compile / loop / collect machinery used for single models.
- Remove the standalone `_perf_modules()` and the duplicated device/EP resolution in the CLI module branch — the CLI just constructs the benchmark and runs it.
- Reuse `_run_monitored_loop` / `_run_simple_loop` and result collection instead of the parallel implementations currently in `_perf_modules`.

## Acceptance

- `winml perf --module <Class>` produces the same per-instance table + JSON report.
- Device/EP resolution happens in exactly one place (`PerfBenchmark`).
- Existing `tests/unit/commands/test_perf_module.py` passes (adjusted only for the new call path).

## References

- Follow-up to #931.
- Code marker: comment in `perf()`'s `--module` branch in `src/winml/modelkit/commands/perf.py` points here.

DingmaomaoBJTU:
**Triage Summary** (auto-generated 2026-07-27)

| Field | Value |
|-------|-------|
| **Priority** | P2 |
| **Type** | refactor |
| **Component** | perf |
| **Owner** | @vortex-captain |

_This issue has been triaged. Please update priority/assignment if needed._

PR #941, opened 2026-06-23
fix(perf): resolve concrete EP for the analyzer instead of ep=None (#931)

### LINKED ISSUE #931: bug: in perf, no ep means analyze all but actually only run one ep
We should pass the resolved ep + device down to analyzer

`[15:59:58 WARNING winml.modelkit.analyze.analyzer] analyze_onnx called with ep=None — results will aggregate all EPs. For the build pipeline, always pass an explicit ep.`

xieofxie:
## Summary

Fixes #931.

In `winml perf` without `--ep`, the EP was never resolved before the build, so the static analyzer ran with `ep=None` and aggregated across **all** EPs (logging `analyze_onnx called with ep=None — results will aggregate all EPs`), even though inference runs on a single device's EP.

This resolves a concrete device + EP from the request and passes it down to the build:

- **`PerfBenchmark`** resolves device + EP internally (`_resolve_device_ep`), at the start of `_load_model` so an unavailable/invalid combo **fails fast** before the export/optimize/quantize/compile pipeline runs (previously it only surfaced at `session.compile()`). `BenchmarkConfig` keeps only the raw request; the resolved values live on the instance and drive `from_pretrained`/`from_onnx`.
- **`_perf_modules`** derives a concrete EP from the resolved device when none is given. Explicit EPs are kept verbatim (downstream stages normalize aliases).
- The CLI `perf()` no longer pre-resolves — it just builds the config and dispatches.

`WinMLAutoModel` stays permissive: `ep=None` remains a valid library mode (aggregate across EPs). The fix makes `perf` — which targets one device — always pass an explicit EP, which is what the analyzer warning asked for.

## Verification

End-to-end A/B on a real CPU build (`--no-skip-build`, no `--ep`):

| | analyzer target | `ep=None` warning |
|---|---|---|
| before | `None on cpu` | **fires** |
| after | `OpenVINOExecutionProvider on cpu` | **gone** |

Unit tests (`tests/unit/commands/test_perf_cli.py`, `test_perf_module.py`, 51 passed) cover: derived concrete EP reaches `from_onnx`; explicit EP passes through verbatim; an unavailable combo raises before the build is kicked off.

## Follow-up

#939 tracks folding `_perf_modules` into `PerfBenchmark` so the two resolution sites become one.

### DIFF
--- a/src/winml/modelkit/commands/perf.py
+++ b/src/winml/modelkit/commands/perf.py
@@ -303,6 +303,46 @@ class PerfBenchmark:
         self._model: WinMLPreTrainedModel | WinMLCompositeModel | None = None
         self._inputs: dict[str, np.ndarray] | None = None
         self._memory: dict[str, float] | None = None
+        # Concrete device + EP resolved from the config's request, populated by
+        # _resolve_device_ep() on the first call (before the build). The config
+        # keeps the raw request (e.g. "auto"); these hold what actually drives
+        # the build and inference.
+        self._resolved_device: str | None = None
+        self._resolved_ep: EPNameOrAlias | None = None
+
+    def _resolve_device_ep(self) -> None:
+        """Resolve the concrete target device + EP, failing fast on a bad combo.
+
+        Idempotent: resolves once, then returns cached values. Called at the
+        start of model loading so an unavailable/invalid device+EP raises here —
+        before the export/optimize/quantize/compile pipeline runs — instead of
+        only surfacing at session.compile(). Deriving a concrete EP also lets the
+        build's static analyzer target one EP instead of aggregating across all
+        of them (WinMLAutoModel itself stays permissive: ep=None is a valid
+        library mode).
+
+        Raises:
+            ValueError: If the requested device/EP combination is unavailable
+                or invalid (propagated from ``resolve_device``).
+        """
+        if self._resolved_device is not None:
+            return
+
+        from ..sysinfo import resolve_device, resolve_eps
+
+        # resolve_device() availability-checks even when --ep is explicit, so a
+        # named-but-absent EP is caught here too.
+        resolved_device, _ = resolve_device(device=self.config.device, ep=self.config.ep)
+        if self.config.ep is not None:
+            # Keep the user's EP (alias or canonical) verbatim — downstream
+            # stages normalize it. Only derive one when the user gave none.
+            resolved_ep: EPNameOrAlias | None = self.config.ep
+        else:
+            device_eps = resolve_eps(resolved_device)
+            resolved_ep = device_eps[0] if device_eps else None
+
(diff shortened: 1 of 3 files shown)


src/winml/modelkit/commands/perf.py
```diff
@@ -303,6 +303,46 @@ def __init__(self, config: BenchmarkConfig) -> None:
...
+        # resolve_device() availability-checks even when --ep is explicit, so a
+        # named-but-absent EP is caught here too.
+        resolved_device, _ = resolve_device(device=self.config.device, ep=self.config.ep)
+        if self.config.ep is not None:
+            # Keep the user's EP (alias or canonical) verbatim — downstream
+            # stages normalize it. Only derive one when the user gave none.
+            resolved_ep: EPNameOrAlias | None = self.config.ep
+        else:
+            device_eps = resolve_eps(resolved_device)
+            resolved_ep = device_eps[0] if device_eps else None
```
DingmaomaoBJTU, line 342: Nit: if 
esolve_eps returns [], 
esolved_ep stays None and p=None still flows to the analyzer — the exact behavior the fix is avoiding. In practice 
esolve_device() succeeding implies at least one EP is available for that device, so this branch is unreachable under normal conditions. A short inline comment (e.g. # shouldn't happen: resolve_device guarantees at least one EP for the device) would make that invariant explicit.

src/winml/modelkit/commands/perf.py
```diff
@@ -752,10 +797,17 @@ def _perf_modules(
...
     from .build import _instantiate_parent_model
 
     resolved_device, _ = resolve_device(device=device, ep=ep)
+    # Derive a concrete EP when none was given so each per-module build's static
+    # analyzer targets one EP instead of ep=None (which aggregates across all
+    # EPs and warns; see #931). An explicit EP is kept verbatim — downstream
+    # stages normalize it.
+    if ep is None:
+        device_eps = resolve_eps(resolved_device)
+        ep = device_eps[0] if device_eps else None
```
DingmaomaoBJTU, line 810: Same pattern: lse None silently falls back to p=None when 
esolve_eps returns empty. Same reasoning applies — unreachable in practice since 
esolve_device already validated the device — a clarifying comment here too would be nice.

DingmaomaoBJTU (approved):
Clean fix. The EP resolution logic is correct, idempotent, and placed at the right point (before the expensive build pipeline). Tests cover all the important cases: derived EP reaches from_onnx, explicit EP passes verbatim, unavailable combo fails fast. One minor nit: when resolve_eps returns [], _resolved_ep silently stays None — a comment making that invariant explicit would help future readers.

src/winml/modelkit/commands/perf.py
xieofxie: no need

PR #947, opened 2026-06-23
fix(build): honor --ep; require explicit device+ep in resolve_device; support cross-compile

xieofxie:
## Summary

Several related EP/device-handling fixes in `winml build`, found after `winml build --ep openvino --device npu --compile` **silently ignored `--ep`**.

### 1. `winml build` now honors `--ep`
The command passed only `device`/`precision` to `generate_build_config` — never `ep` — so the requested EP was silently dropped (and, with an explicit absent `--device`, it failed for the wrong reason). Now `ep` is forwarded.

### 2. `resolve_device` requires explicit `device` + `ep`
Both are now required (no `device="auto"` default; `ep` was already required), so an omission is a **call error** instead of a silent `ep=None` / `device="auto"`. Several callers relied on the old defaults (latent `TypeError`s + 8 failing `test_device.py` tests) — fixed to pass both explicitly:
- analyzer auto-device resolution
- build's auto-EP (`ep is None`) block
- eval device resolution
- compile-stage device resolution

This closes the class of bug behind #931.

### 3. Cross-compile support via `resolve_check_device_ep`
`winml build` now resolves through `resolve_check_device_ep`, which validates an explicit device+ep **statically** (no availability cross-check). A build can target a device that isn't on the build machine; the outcome depends on whether compilation is actually needed:

| Scenario | Result |
|---|---|
| `--no-compile`, target device absent | ✅ builds — portable, analyzed/quantized ONNX (cross-compile) |
| `--compile`, EP has no EPContext (e.g. MIGraphX, DML) | ✅ builds — compile is a no-op |
| `--compile`, EP has EPContext, device absent (e.g. OpenVINO+NPU) | ⚠️ fails **at the compile stage** (late) — tracked in #950 |

## Verification

Empirical builds of `microsoft/resnet-50` on a machine without an NPU confirm all three rows (exit 0 / 0 / 1, failing only at the compile stage for row 3). Unit suites green: `tests/unit/{commands,sysinfo,compiler,analyze,config}` (2500+ passed); `test_build.py` + `test_device.py` updated for the new resolver contract.

## Follow-up

- #950 — make the `--compile`-on-absent-device case fail fast (before export) instead of at the compile stage. A gate was intentionally deferred so it can be conditioned correctly on `config.compile` (must not regress cross-compile / no-EPContext builds).

Related: #931 (fixed in #941).

### DIFF
--- a/src/winml/modelkit/commands/build.py
+++ b/src/winml/modelkit/commands/build.py
@@ -545,20 +545,19 @@ def build(
     # hardware -- analyzing for the wrong EP leaves black nodes that block a
     # later build targeting the actual device (#663).
     #
-    # resolve_device() either returns a device with >=1 available EP (auto-mode
-    # walks the priority list, falls back to cpu which is always valid), or
-    # raises ValueError for an explicit device with no compatible EP. So the
-    # following resolve_eps()[0] is safe whenever resolve_device returns.
+    # resolve_check_device_ep() either returns a device with >=1 available EP
+    # (auto-mode walks the priority list, falls back to cpu which is always
+    # valid), or raises ValueError for an explicit device with no compatible EP.
+    # So the following available_eps[0] is safe whenever it returns.
     if ep is None:
-        from ..sysinfo import resolve_device as _resolve_device
-        from ..sysinfo import resolve_eps as _resolve_eps
+        from ..sysinfo import resolve_check_device_ep
 
         try:
-            resolved_device, _ = _resolve_device(device=device, ep=None)
+            resolved_device, _, available_eps = resolve_check_device_ep(device=device, ep=None)
         except ValueError as e:
             raise click.UsageError(str(e)) from e
         device = resolved_device
-        ep = _resolve_eps(resolved_device)[0]
+        ep = available_eps[0]
         logger.info("Auto-resolved device=%s, EP=%s", resolved_device, ep)
 
     try:
@@ -1013,11 +1012,15 @@ def _run_optimize_stage(
             _header_shown[0] = False
 
         # Resolve "auto" to a concrete device once so that has_rule_data_for_ep
-        # doesn't search for non-existent "*_AUTO_*.parquet" files.
+        # doesn't search for non-existent "*_AUTO_*.parquet" files. Use
+        # resolve_check_device_ep so an explicit device+ep is validated
+        # statically (no availability cross-check): a --no-compile build may
+        # target a device absent on this machine (cross-compile), and this call
+        # only needs a concrete device name for the rule-data lookup.
         from ..analyze.utils.ep_utils import has_rule_data_for_ep
-        from ..sysinfo import resolve_device as _resolve_device
+        from ..sysinfo import resolve_check_device_ep
 
-        _resolved_device, _ = _resolve_device(device=device or "auto", ep=ep)
+        _resolved_device, _, _ = resolve_check_device_ep(device=device or "auto", ep=ep)
 
         def _on_ep_start(ep_name: EPName, operator_counts: dict) -> None:
             nonlocal _current_ep
--- a/src/winml/modelkit/commands/build.py
+++ b/src/winml/modelkit/commands/build.py
@@ -579,6 +579,7 @@ def build(
                 trust_remote_code=trust_remote_code,
                 device=device,
                 precision=precision,
+                ep=ep,
             )
             if not quant:
                 config_or_configs.quant = None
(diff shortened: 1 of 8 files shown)


src/winml/modelkit/compiler/stages/compile.py
```diff
@@ -176,7 +176,7 @@ def _compile_multiple(self, context: CompileContext) -> None:
         sess_options = context.shared_session_options
         if sess_options is None:
             register_execution_providers(ort=True)
-            resolved_device, _ = resolve_device(context.config.get("device", "auto"))
+            resolved_device, _ = resolve_device(context.config.get("device", "auto"), ep=None)
```
xieofxie: if not set, `--ep cpu` will be resolved to gpu

src/winml/modelkit/commands/eval.py
```diff
@@ -364,7 +364,7 @@ def _resolve_device(cfg: WinMLEvaluationConfig) -> None:
 
     console = Console(stderr=True)
     console.print("[bold]Detecting available devices...[/bold]")
-    resolved, _ = resolve_device(cfg.device)
+    resolved, _ = resolve_device(cfg.device, ep=None)
```
xieofxie: if not set, `--ep cpu` will be resolved to gpu

timenick (approved):
Solid, well-scoped fix. The root cause (build dropping `--ep` into `generate_build_config`) is addressed and the surrounding cleanup is consistent: making `device` + `ep` required on `resolve_device` turns the silent `ep=None` / `device="auto"` defaults into call errors, and every caller that relied on those defaults is updated to pass both explicitly. I checked all `resolve_device` / `resolve_check_device_ep` call sites at this head — they all forward `ep`, so there’s no latent `TypeError` from a missed caller. The auto-EP block’s `available_eps[0]` is safe since `resolve_device` only returns a device with ≥ 1 available EP (else it raises), and routing `build` through `resolve_check_device_ep` for the static device+ep validation is the right call for the cross-compile scenarios. Nice regression test for the `--ep` forwarding. LGTM.

src/winml/modelkit/sysinfo/device.py
```diff
@@ -233,7 +233,7 @@ def resolve_eps(resolved_device: str) -> list[EPName]:
 
 
 def resolve_check_device_ep(
-    *, device: str = "auto", ep: EPNameOrAlias | None = None
+    *, device: str, ep: EPNameOrAlias | None
```
DingmaomaoBJTU, line 236: The docstring says 'Ideal for commands that do not need the device + ep actually exists on the system.' But static validation (no availability probe) only applies when BOTH device is explicit (not auto) AND ep is non-None. When either is absent the function delegates to resolve_device which does a full availability cross-check. A caller passing --device npu without --ep on a machine without NPU will still get a ValueError. The docstring should clarify this for users relying on this for cross-compile.

src/winml/modelkit/commands/build.py
```diff
@@ -545,20 +545,19 @@ def build(
...
+    # valid), or raises ValueError for an explicit device with no compatible EP.
+    # So the following available_eps[0] is safe whenever it returns.
     if ep is None:
-        from ..sysinfo import resolve_device as _resolve_device
-        from ..sysinfo import resolve_eps as _resolve_eps
+        from ..sysinfo import resolve_check_device_ep
 
         try:
-            resolved_device, _ = _resolve_device(device=device)
+            resolved_device, _, available_eps = resolve_check_device_ep(device=device, ep=ep)
```
DingmaomaoBJTU, line 556: Minor nit: inside the 'if ep is None:' block, passing ep=ep (which is provably None here) is technically correct but slightly opaque. Writing ep=None explicitly would remove any ambiguity.

DingmaomaoBJTU (commented):
Overall this is a well-motivated three-part fix: (1) forward --ep into generate_build_config, (2) make both params required in resolve_device/resolve_check_device_ep to turn silent bugs into call errors, (3) route the optimize stage through resolve_check_device_ep so cross-target --no-compile builds work. All 138 tests pass.

General: In the static path of resolve_check_device_ep (explicit device + explicit ep), the second tuple element is list(EP_SUPPORTED_DEVICES[ep_name]) -- the EP's statically declared supported devices. In the dynamic path it returns runtime-filtered devices. All callers use _ so no active bug, but future callers will see inconsistent semantics. Consider documenting or unifying the two paths.

tests/unit/commands/test_build.py
```diff
@@ -1711,3 +1709,34 @@ def test_returns_compiled_path_when_file_exists(
 
         # current_path should be updated to compiled_path
         assert result == compiled_path
+
+
+class TestBuildEpResolution:
+    """--ep forwarding into config generation + the compile EP-availability gate."""
+
+    def _base_args(self, cfg: str, tmp_path: Path) -> list[str]:
+        return ["-c", cfg, "-m", "microsoft/resnet-50", "-o", str(tmp_path / "out")]
```
DingmaomaoBJTU, line 1718: _base_args is never called by any test in this class -- dead helper. If intended for future tests add a TODO comment; otherwise remove it. The args it builds also mix -c and -m which would not match the auto-config path the class is testing.

tests/unit/commands/test_build.py
```diff
@@ -1711,3 +1709,34 @@ def test_returns_compiled_path_when_file_exists(
...
+            patch(
+                "winml.modelkit.commands.build._validate_loader_tasks_for_model",
+                return_value=None,
+            ),
+        ):
+            result = _invoke(
+                ["-m", "microsoft/resnet-50", "--ep", "openvino", "-o", str(tmp_path / "out")]
+            )
+        assert result.exit_code == 0, result.output
+        assert mock_gen.call_args.kwargs["ep"] == "openvino"
```
DingmaomaoBJTU, line 1742: Missing regression test for the cross-compile static path: the PR's headline feature (--device npu --ep openvino --no-compile completing on a machine without NPU) is verified empirically but has no unit test. A test that mocks resolve_check_device_ep to return a static result and confirms the optimize stage proceeds would guard against future regressions (e.g. accidentally restoring resolve_device in the optimize path).

PR #949, opened 2026-06-23
feat(keypoint-detection): add COCO OKS-AP evaluation

jeon185:
Adds the eval stage for keypoint-detection (ViTPose), so the COCO-keypoint models from #284 now go through `config -> build -> perf -> eval`. Stacked on #905 (the config/build/perf enablement) - that one should go in first.

What's here:

- `metrics/keypoint.py` - KeypointAPMetric. Computes the COCO keypoint score (OKS-based AP over 0.50:0.95) with pycocotools COCOeval, the same way object-detection already reuses the COCO mAP protocol.
- `keypoint_detection_evaluator.py` - top-down evaluator. transformers has no keypoint-detection pipeline, so it runs the image processor and ONNX model directly: for each ground-truth person box it does preprocess -> model -> post_process_pose_estimation and scores against the GT keypoints. ViTPose is exported with a static batch of 1, so each person crop runs separately and the heatmaps are stacked back together for post-processing. It uses the GT person boxes (standard COCO top-down protocol - keeps the score about pose accuracy, not detection).
- `scripts/build_coco_keypoints.py` - builds a local COCO val keypoints dataset. COCO has no script-free HF mirror for person keypoints, so this downloads the annotations once and fetches images individually, which means a small subset doesn't need the full image zip.
- Schema, evaluator registry, default dataset, and unit tests for the metric and evaluator.

Verified on the five COCO 17-keypoint models (vitpose-base-simple and vitpose-plus-{small,base,large,huge}): config -> build -> perf -> eval all pass and return COCO AP/AR. AP rises with model size as you'd expect. Absolute numbers are on the low side right now because the build quantizes with random calibration data, but relative comparison holds.

## synthpose-vitpose-huge-hf - not covered yet

This is the one model from #284 that this PR does not evaluate. It predicts 52 anatomical keypoints instead of COCO's 17, so it can't be scored against COCO ground truth - the keypoint sets don't line up and OKS is only defined when they do.

How it's handled for now: the metric checks the keypoint count up front and raises a clear, actionable error instead of failing with a numpy broadcast error deep inside pycocotools.

Idea for finishing it: KeypointAPMetric already takes `sigmas` and `keypoint_names` as arguments, so the main missing piece is a dataset with SynthPose's 52-keypoint ground truth plus the matching OKS sigmas. I'd rather agree on the dataset and sigmas in review before adding that - happy to land it in this PR or as a follow-up, whichever you prefer.

Refs #284.

### DIFF
--- a/scripts/build_coco_keypoints.py
+++ b/scripts/build_coco_keypoints.py
@@ -1,0 +1,154 @@
+# -------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+# --------------------------------------------------------------------------
+"""Build a local COCO keypoints dataset for ``winml eval`` keypoint-detection.
+
+Downloads the COCO person-keypoints annotations (cached once) and a chosen
+number of validation images individually (so a small subset does not require
+the full ~780 MB image zip), then writes an Arrow dataset to disk via
+``datasets.Dataset.save_to_disk``. Point ``winml eval --dataset-path`` at the
+output directory.
+
+Each record has:
+    - ``image``: the RGB image (datasets ``Image`` feature)
+    - ``objects``: dict with parallel per-person lists ``keypoints`` (flat
+      ``[x, y, v]`` triplets), ``bbox`` (COCO ``[x, y, w, h]``) and ``area``.
+
+Only images containing at least one labeled-keypoint person are included.
+
+Usage:
+    uv run python scripts/build_coco_keypoints.py --output-dir ~/.cache/winml/datasets/coco_keypoints_val2017
+    uv run python scripts/build_coco_keypoints.py --output-dir <dir> --num-images 100
+    uv run python scripts/build_coco_keypoints.py --output-dir <dir> --num-images 0  # all images
+"""
+
+import argparse
+import io
+import json
+import shutil
+import sys
+import urllib.request
+import zipfile
+from pathlib import Path
+
+
+ANNOTATIONS_URL = "http://images.cocodataset.org/annotations/annotations_trainval2017.zip"
+ANNOTATION_MEMBER = "annotations/person_keypoints_val2017.json"
+IMAGE_URL_TEMPLATE = "http://images.cocodataset.org/val2017/{file_name}"
+
+DEFAULT_CACHE = Path.home() / ".cache" / "winml" / "coco_build"
+
+
+def _download(url: str, dest: Path) -> None:
+    """Download ``url`` to ``dest`` (skips if it already exists)."""
+    if dest.exists():
+        return
+    dest.parent.mkdir(parents=True, exist_ok=True)
+    print(f"Downloading {url}")
+    with urllib.request.urlopen(url) as resp, dest.open("wb") as fh:  # noqa: S310
+        shutil.copyfileobj(resp, fh)
+
+
+def _load_annotations(cache_dir: Path) -> dict:
+    """Return the parsed person-keypoints annotation JSON, downloading once."""
+    ann_zip = cache_dir / "annotations_trainval2017.zip"
+    _download(ANNOTATIONS_URL, ann_zip)
+    print("Reading keypoint annotations...")
+    with zipfile.ZipFile(ann_zip) as zf, zf.open(ANNOTATION_MEMBER) as fh:
+        return json.load(fh)
+
+
+def _group_annotations_by_image(annotations: list[dict]) -> dict[int, list[dict]]:
+    """Group person annotations by image id, keeping only labeled-keypoint people."""
+    by_image: dict[int, list[dict]] = {}
+    for ann in annotations:
+        if ann.get("num_keypoints", 0) <= 0 or ann.get("iscrowd", 0):
+            continue
+        by_image.setdefault(ann["image_id"], []).append(ann)
+    return by_image
+
+
+def _fetch_image(file_name: str) -> bytes:
+    """Download one validation image and return its raw bytes."""
+    url = IMAGE_URL_TEMPLATE.format(file_name=file_name)
+    with urllib.request.urlopen(url) as resp:  # noqa: S310
+        return resp.read()
+
+
+def build(output_dir: Path, num_images: int, cache_dir: Path) -> None:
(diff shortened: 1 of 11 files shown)


src/winml/modelkit/eval/keypoint_detection_evaluator.py
```diff
@@ -0,0 +1,195 @@
...
+if TYPE_CHECKING:
+    from transformers.image_processing_utils import BaseImageProcessor
+
+    from ..models.winml.base import WinMLPreTrainedModel
+    from .config import WinMLEvaluationConfig
+
+logger = logging.getLogger(__name__)
+
+
+class WinMLKeypointDetectionEvaluator(WinMLEvaluator):
```
zhenchaoni, line 34: Is `synthpose-vitpose-huge-hf` not supported mainly because we haven't found a good dataset yet? The reason why I asked is that we need to ensure the `WinMLKeypointDetectionEvaluator` is extensible for various key point detection models. Otherwise, we would have to implement different evaluators.

scripts/build_coco_keypoints.py
```diff
@@ -0,0 +1,154 @@
...
+
+
+ANNOTATIONS_URL = "http://images.cocodataset.org/annotations/annotations_trainval2017.zip"
+ANNOTATION_MEMBER = "annotations/person_keypoints_val2017.json"
+IMAGE_URL_TEMPLATE = "http://images.cocodataset.org/val2017/{file_name}"
+
+DEFAULT_CACHE = Path.home() / ".cache" / "winml" / "coco_build"
+
+
+def _download(url: str, dest: Path) -> None:
```
zhenchaoni, line 44: Is this dataset very enormous? If so, you can consider downloading 5000 samples in steaming, shuffle and take 1000 samples. We did this for other large dataset in build_ai4privacy.py

It would be fine if the dataset is not huge like below 5GB.

scripts/build_coco_keypoints.py
```diff
@@ -0,0 +1,154 @@
+# -------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+# --------------------------------------------------------------------------
+"""Build a local COCO keypoints dataset for ``winml eval`` keypoint-detection.
```
zhenchaoni, line 5: For the models which you verified and can be evaluated, can you help to update C:\Users\zhenni\repos\wmk\scripts\e2e_eval\testsets\models_with_acc.json

and post the evaluation data you run locally?

src/winml/modelkit/eval/evaluate.py
```diff
@@ -172,6 +174,21 @@ def get_evaluator_class(config: WinMLEvaluationConfig) -> type[WinMLEvaluator]:
         # the legacy `nyu_depth_v2.py` loader script.
         "revision": "refs/convert/parquet",
     },
+    "keypoint-detection": {
+        # Built locally by scripts/build_coco_keypoints.py (COCO has no
+        # script-free HF mirror for person keypoints). Run that script first,
+        # or pass --dataset-path to point at your own build.
+        "path": "~/.cache/winml/datasets/coco_keypoints_val2017",
```
zhenchaoni, line 182: If no existing HF dataset can be used directly, it is OK not to document a default dataset

src/winml/modelkit/eval/metrics/keypoint.py
```diff
@@ -0,0 +1,207 @@
...
+            coco_dt = coco_gt.loadRes(detections)
+            coco_eval = COCOeval(coco_gt, coco_dt, iouType="keypoints")
+            coco_eval.params.kpt_oks_sigmas = np.array(sigmas, dtype=np.float64)
+            coco_eval.evaluate()
+            coco_eval.accumulate()
+            coco_eval.summarize()
+
+        stats = coco_eval.stats
+        return {
+            "AP": float(stats[0]),
```
zhenchaoni: Do you think we can have consistent metric as mean_average_precision.py? It is coco based object detection metrics. The main different is that it uses a predicted box. But I think we can also report the map in keypoint.

src/winml/modelkit/eval/keypoint_detection_evaluator.py
jeon185: Right, it's the dataset and its OKS sigmas, not the evaluator. I wired `sigmas` and `keypoint_names` through the dataset config, so a non-COCO layout runs through this same evaluator — defaults stay COCO-17. SynthPose just needs a 52-keypoint dataset plus its sigmas, no separate evaluator. Added a test that scores a custom 5-keypoint layout to keep that honest.

scripts/build_coco_keypoints.py
jeon185: It's small — we fetch images one by one instead of the full image zip, so the default 100-image subset is only tens of MB. Good point on sampling though: it now shuffles (seeded) before taking N, so the subset is representative instead of the lowest image ids.

scripts/build_coco_keypoints.py
jeon185: Done — added the five COCO-17 models. Local AP (CPU, 100 images):
- vitpose-base-simple: 0.4485
- vitpose-plus-small: 0.4699
- vitpose-plus-base: 0.4505
- vitpose-plus-large: 0.5222
- vitpose-plus-huge: 0.5184

Absolute numbers are low because the build calibrates on random data; relative order tracks model size as expected.

src/winml/modelkit/eval/evaluate.py
jeon185: I kept a default pointing at the local build path with a note to run the build script first (or pass `--dataset-path`). Happy to drop it if you'd rather not show a default that isn't directly loadable.

src/winml/modelkit/eval/metrics/keypoint.py
jeon185: Renamed the outputs to `map`/`map_50`/`map_75` (plus `map_medium`/`map_large` and `mar`/...) so they line up with `MAPMetric`. Kept pycocotools underneath since it's what does OKS keypoint scoring, but the reported keys now match object detection.

zhenchaoni (dismissed):

zhenchaoni (approved):

PR #953, opened 2026-06-24
fix(logging): drop dead _HFSymlinksInfoFilter and redundant TasksManager demote filter

### LINKED ISSUE #909: `_HFSymlinksInfoFilter` is dead code; demote-to-INFO log filters relabel but don't suppress
## Summary

`_warnings.py` has "demote `WARNING`→`INFO`" logging filters that, given how the CLI gates logging, **relabel records but don't actually hide them**. One of them — `_HFSymlinksInfoFilter` (added in #647) — is now fully dead code, shadowed by a hard `filterwarnings("ignore")` added later in #709.

## Details

**1. `_HFSymlinksInfoFilter` never runs.**

- #647 (`405272d7f`) added `_HFSymlinksInfoFilter` on the `py.warnings` logger to demote the `huggingface_hub` symlinks `UserWarning` to INFO ("hidden by default, visible at `-v`").
- #709 (`f12db93a5`) later added `warnings.filterwarnings("ignore", message=r".*huggingface_hub.*cache-system.*symlinks.*")`, which drops the warning at the Python `warnings` layer **before** `captureWarnings` routes it to `py.warnings`. The demote filter never executes.
- Net: the warning is fully suppressed (default **and** `-v`); the filter does nothing.

**2. Demote-to-INFO doesn't suppress — it only relabels.**

The CLI uses the root logger level as the sole gate (handler is `NOTSET`, see `utils/logging.py`). That gate is evaluated at **emit time** on the record's original level; a filter that later rewrites `levelno` `WARNING`→`INFO` runs *after* the gate has passed, and the `NOTSET` handler doesn't re-filter — so the record still prints, just labeled INFO. Empirically, `_TasksManagerFilter` (`"TasksManager returned …"`) still prints in **default** verbosity as an INFO line; it is not hidden.

## Impact

Cosmetic / low — no functional bug. But `_HFSymlinksInfoFilter` is misleading dead code, and `_TasksManagerFilter` doesn't hide what it appears to.

## Suggested cleanup

- Remove the dead `_HFSymlinksInfoFilter` (its effect is now provided by the `filterwarnings("ignore")` in #709), **or** decide the HF notice should be visible at `-v` and implement it via a verbosity-conditional logger floor.
- For logging-layer noise that should be "hidden by default, visible at `-v`", use a verbosity-conditional `setLevel` floor (as in #904) rather than a demote filter.

timenick:
Closes #909.

`_warnings.py` had two "demote `WARNING`→`INFO`" logging filters that relabel records but don't actually suppress them. The CLI uses the root logger level as the sole emit-time gate (handler is `NOTSET`), so a filter that rewrites `levelno` *after* the gate has passed cannot hide a record — it only changes the printed label.

### `_HFSymlinksInfoFilter` was dead code
- #647 added it on the `py.warnings` logger to demote the huggingface_hub symlinks `UserWarning` to INFO.
- #709 later added `warnings.filterwarnings("ignore", message=r".*huggingface_hub.*cache-system.*symlinks.*")`, which drops the warning at the Python `warnings` layer **before** `captureWarnings` could route it to `py.warnings`. The demote filter never executed.
- Net: the warning is fully suppressed in every verbosity mode; the filter did nothing.

### `_TasksManagerFilter` was redundant
- It demoted optimum's "TasksManager returned …" `WARNING` on `optimum.exporters.tasks`.
- #904 added a verbosity-conditional ERROR floor on the parent `optimum` logger in `configure_logging`. Because `optimum.exporters.tasks` inherits its effective level from `optimum`, the notice is already gated at the source: hidden by default, shown at `-v`/`-vv`. The demote filter only relabeled the record at `-v`; it provided no suppression.

### Changes
- Remove both filter classes and their `addFilter` registrations in `src/winml/modelkit/_warnings.py`.
- Keep the `filterwarnings("ignore")` suppressor; refresh its comment and add a NOTE that optimum's informational WARNINGs are gated by the ERROR floor in `utils/logging.py` (so no demote filter is needed and one must not be re-added).
- Replace the obsolete `_HFSymlinksInfoFilter` test in `tests/unit/test_warnings_configure.py` with a regression test that the HF symlinks `UserWarning` is suppressed (plus a specificity check that unrelated `symlinks` warnings pass through).
- Add `test_optimum_child_logger_gated_by_parent_floor` to `tests/unit/utils/test_logging.py` proving `optimum.exporters.tasks` is `isEnabledFor(WARNING)` False at default and True at `-v` — locking in floor-based gating.

### Behavior
- HF symlinks warning: unchanged (fully suppressed before and after).
- TasksManager notice: hidden by default (unchanged); at `-v` it now prints at its native WARNING level instead of a relabeled INFO line — more truthful.

### DIFF
--- a/src/winml/modelkit/_warnings.py
+++ b/src/winml/modelkit/_warnings.py
@@ -101,73 +101,26 @@ def _configure() -> None:
     )
 
     # =========================================================================
-    # py.warnings logger filters (for warnings routed via logging.captureWarnings)
+    # huggingface_hub: suppress the Windows "symlinks not supported" notice
     # =========================================================================
-
-    class _HFSymlinksInfoFilter(logging.Filter):
-        r"""Downgrade the huggingface_hub symlinks UserWarning from WARNING to INFO.
-
-        On Windows without Developer Mode, huggingface_hub warns that symlinks
-        are unsupported and the cache will use copies instead. This is cosmetic —
-        the cache still works, just without deduplication. WARNING is misleading
-        here; INFO is the appropriate level.
-
-        When warnings are routed via logging.captureWarnings(True), Python's
-        warnings.formatwarning() embeds the source filename in the log message
-        body ("path/to/huggingface_hub/file_download.py:1: UserWarning: ..."),
-        so we match against getMessage() rather than record.pathname (which
-        is always warnings.py in that path).
-
-        Before (WARNING level, always visible):
-            [09:12:34] WARNING  C:\\...\\huggingface_hub\\file_download.py:1:
-                                UserWarning: `huggingface_hub` cache-system uses
-                                symlinks by default to efficiently store
-                                duplicated files but your machine does not
-                                support them
-
-        After (INFO level, only visible with --verbose or -v):
-            [09:12:34] INFO     C:\\...\\huggingface_hub\\file_download.py:1:
-                                UserWarning: `huggingface_hub` cache-system uses
-                                symlinks by default to efficiently store
-                                duplicated files but your machine does not
-                                support them
-        """
-
-        def filter(self, record: logging.LogRecord) -> bool:
-            msg = record.getMessage()
-            if "symlinks" in msg and "huggingface_hub" in msg:
-                record.levelno = logging.INFO
-                record.levelname = "INFO"
-            return True
-
-    logging.getLogger("py.warnings").addFilter(_HFSymlinksInfoFilter())
-
-    # Suppress the huggingface_hub symlinks warning at the Python warnings level
-    # so it is hidden even before captureWarnings(True) is activated in build.py.
-    # When captureWarnings(True) is active, _HFSymlinksInfoFilter above handles
-    # the same message for loggers that need it at INFO (e.g. with --verbose).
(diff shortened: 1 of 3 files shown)


DingmaomaoBJTU (approved):
Both removed filters are verified dead/redundant after tracing the execution path.

_HFSymlinksInfoFilter: The filterwarnings(ignore) added in #709 suppresses the warning at the Python warnings layer, before captureWarnings(True) can route it to the py.warnings logger. The demote filter never had a record to process. Dead code confirmed.

_TasksManagerFilter: optimum.exporters.tasks inherits its effective level from the optimum parent. configure_logging (#904) pins optimum at ERROR at default verbosity, so WARNING records are rejected by level check before reaching any filter. At -v the floor drops, but the filter only relabeled records anyway. Redundant confirmed.

Remaining references: Only a clarifying comment in test_logging.py -- no live code references to either symbol.

Tests: Old tests covered the dead demote-filter behavior; new tests correctly exercise the actual suppression mechanism. All 7 tests pass.

Minor observation: _restore_logging_filters does not restore warnings module state, but both new tests use catch_warnings(record=True) + resetwarnings() which fully own the warnings stack inside their context.

PR #923, opened 2026-06-22
feat(cli): unify quant/optimize/analyze/max-optim flags across build, config, perf, eval

xieofxie:
## Summary

Unifies the five build-pipeline CLI flags — `--quant/--no-quant`, `--optimize/--no-optimize`, `--analyze/--no-analyze`, and `--max-optim-iterations` — across `build`, `config`, `perf`, and `eval` so they share one definition and identical semantics, defined in `utils/cli.py`.

### Shared helpers (`utils/cli.py`)
- `quant_option()` → `--quant/--no-quant`, with `--quantize/--no-quantize` kept as an **alias** so existing `perf` invocations keep working.
- `optimize_option()` → `--optimize/--no-optimize`
- `analyze_option()` → `--analyze/--no-analyze`
- `max_optim_iterations_option()` → `--max-optim-iterations`
- `build_pipeline_extra_kwargs()` centralizes the `(optimize, analyze, max_optim_iterations)` → `(skip_optimize, hack_max_optim_iterations)` mapping shared by build/perf/eval.

### Per-command changes
| Command | Change |
|---|---|
| `build` | Refactored to use the shared helpers (behavior unchanged). |
| `perf` | `--quantize` → canonical `--quant` (alias retained); `optimize`/`analyze`/`max-optim` added and plumbed through `BenchmarkConfig`, the single-model load path, and the per-module build path. |
| `eval` | `quant`/`optimize`/`analyze`/`max-optim` added to `WinMLEvaluationConfig` and applied in `_load_model`. |
| `config` | Flags added for CLI parity; the build-only flags **warn** when passed, since `config` only emits a static config. |

### Plumbing fix
`WinMLAutoModel.from_pretrained` now forwards `skip_optimize` / `hack_max_optim_iterations` to `build_hf_model` so the HF build path honors these flags (`from_onnx` already forwarded them).

## Testing
- `tests/unit/utils/test_cli.py` — new coverage for the four option helpers (incl. the `--quantize` alias) and `build_pipeline_extra_kwargs`.
- `tests/unit/commands/test_boolean_flag_pairs.py` — updated to the unified surface (canonical `--quant` + alias, new flags on config/perf/eval).
- `tests/unit/eval/test_eval.py` — updated `_load_model` call assertion + new test that `--no-quant`/`--no-optimize`/`--max-optim-iterations` reach `from_pretrained`.
[...]

PR #956, opened 2026-06-24
fix(quantize): suppress ORT pre-processing warnings in quantize flow

### LINKED ISSUE #557: [winml quantize] [P2] Two near-duplicate ORT WARNING lines per run
**Severity**: P2  
**Command**: `winml quantize`

---
**Category**: Quiet at Default — library WARNING noise should be suppressed or de-duplicated.

**Repro**:
```
uv run winml quantize -m temp\cli-audit\resnet.onnx --samples 3 -o temp\ux\q.onnx
```

**Actual**:
```
WARNING: Please consider to run pre-processing before quantization. Refer to example: …
WARNING: Please consider pre-processing before quantization. See …
```
Two ORT layers emit the same advice with slightly different wording, ~7 s apart.

**Expected**: Silence both at default verbosity, or de-duplicate so only one fires per run.

**Why it matters**: Duplicate warnings make users assume something is wrong twice.

xieofxie:
## Summary

Two related cleanups in the quantize/eval CLI surface.

### 1. Suppress duplicate ORT pre-processing warnings (fixes #557)

`winml quantize` emitted two near-duplicate ORT warnings per run:

```
WARNING: Please consider to run pre-processing before quantization. Refer to example: …
WARNING: Please consider pre-processing before quantization. See …
```

Both originate in ORT's `quantize_static`, gated on `model_has_pre_process_metadata()` of the model it reloads from the input path. Previously the pre-process metadata was injected in `optimize_onnx`, which only suppressed the warnings for models that went through optimize first — a standalone `winml quantize` on a raw model still warned twice.

Moved the tagging into the quantizer: load the input model, tag it via ORT's own `add_pre_process_metadata()`, and hand the in-memory `ModelProto` to `quantize()`. Passing the in-memory model (rather than the path) is required because ORT checks the flag on the copy it reloads from `model_input`, and it avoids mutating the user's pristine input file. ORT externalizes weights into its own temp dir, so external-data handling is unaffected.

### 2. Address deferred review comments from #923

- `max_optim_iterations_option`: use `. ` before `optional_message` to match the `optimize`/`analyze` sibling helpers (drop the redundant trailing period from the base help so the joined output stays clean).
- `eval`: warn when build-pipeline flags (`--no-quant`/`--no-optimize`/`--no-analyze`/`--max-optim-iterations`) are passed with a pre-built ONNX path and `skip_build`, where they're silently forwarded to `from_onnx` as no-ops — mirrors the existing `--precision`-ignored warning.

## Verification

- `tests/unit/test_quantizer.py` — passes
- `tests/integration/test_quantization.py` (3 tests, real ORT) — pass
- `tests/unit/commands/test_eval.py` + `tests/unit/utils/test_cli.py` (79 tests) — pass

Fixes #557

### DIFF
--- a/src/winml/modelkit/quant/quantizer.py
+++ b/src/winml/modelkit/quant/quantizer.py
@@ -152,13 +152,20 @@ def quantize_onnx(
             extra_options=extra_options,
         )
 
-        # Step 2: Capture metadata before ORT quantization (it rebuilds the graph)
+        # Step 2: Load the input model, capture its metadata snapshot (ORT
+        # rebuilds the graph during quantization, so we restore afterwards),
+        # and tag it as pre-processed so quantize_static() does not emit the
+        # "run pre-processing before quantization" warning.  We hand this
+        # in-memory ModelProto to ORT directly rather than mutating the user's
+        # input file on disk.
+        from onnxruntime.quantization.quant_utils import add_pre_process_metadata
+
         from ..onnx import capture_metadata, load_onnx, restore_metadata, save_onnx
         from .qdq_fix import fix_qdq_dtype_info
 
-        pre_quant_model = load_onnx(model_path, load_weights=False, validate=False)
-        metadata_snapshot = capture_metadata(pre_quant_model)
-        del pre_quant_model
+        input_model = load_onnx(model_path, validate=False)
+        metadata_snapshot = capture_metadata(input_model)
+        add_pre_process_metadata(input_model)
 
         # Step 3: Apply quantization
         if use_external_data:
@@ -171,9 +178,8 @@ def quantize_onnx(
         # output directory rather than the process CWD.  Without this, a stale
         # .onnx.data sidecar in the process CWD from a previous build triggers
         # a false-positive FileExistsError even when the output dir is clean.
-        # Use absolute paths so the chdir does not break relative input/output
+        # Use an absolute output path so the chdir does not break its
         # resolution.  output_path.parent is guaranteed to exist (caller mkdir).
-        abs_model_input = str(Path(model_path).resolve())
         abs_model_output = str(Path(output_path).resolve())
         # Remove stale output artifacts from a previous build.  ORT/onnx refuse
         # to overwrite an existing external-data sidecar (e.g. quantized.onnx.data),
@@ -187,7 +193,7 @@ def quantize_onnx(
         try:
             os.chdir(output_path.parent)
             quantize(
-                model_input=abs_model_input,
+                model_input=input_model,
                 model_output=abs_model_output,
                 quant_config=qdq_config,
             )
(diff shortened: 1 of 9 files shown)


src/winml/modelkit/quant/quantizer.py
```diff
@@ -152,13 +152,20 @@ def quantize_onnx(
...
+        # input file on disk.
+        from onnxruntime.quantization.quant_utils import add_pre_process_metadata
+
         from ..onnx import capture_metadata, load_onnx, restore_metadata, save_onnx
         from .qdq_fix import fix_qdq_dtype_info
 
-        pre_quant_model = load_onnx(model_path, load_weights=False, validate=False)
-        metadata_snapshot = capture_metadata(pre_quant_model)
-        del pre_quant_model
+        input_model = load_onnx(model_path, validate=False)
```
DingmaomaoBJTU, line 166: Memory note (not a blocker): Switching from load_weights=False to the default (True) now loads full model weights into a Python ModelProto. For large models with external data (multi-GB LLMs) this doubles peak memory compared to the prior approach where ORT streamed directly from disk. This is the necessary trade-off for the in-memory tagging approach - worth a comment or doc note so future readers don't revert it without understanding why.

tests/unit/test_quantizer.py
```diff
@@ -74,9 +79,13 @@ def _install_fake_ort_quantization(monkeypatch: pytest.MonkeyPatch, *, quantize_
     )
     quant_module.get_qdq_config = lambda **_: SimpleNamespace(use_external_data_format=False)
     quant_module.quantize = quantize_impl
+    quant_utils_module = _FakeOrtQuantUtilsModule("onnxruntime.quantization.quant_utils")
+    quant_utils_module.add_pre_process_metadata = lambda _model: None
```
DingmaomaoBJTU, line 83: Weak mock coverage: lambda _model: None does not track whether add_pre_process_metadata was actually called. If that call were accidentally removed, this unit test would still pass (the fake_quantize assertion only checks object identity, not that the metadata was set). Consider MagicMock() with .assert_called_once_with(input_model) for stronger coverage of the core fix. Integration tests cover the real behaviour, so not a blocker.

DingmaomaoBJTU (approved):
Clean implementation. The approach of loading the in-memory ModelProto, tagging it with add_pre_process_metadata, and passing it directly to ORT's quantize() is correct and avoids mutating the user's input file. Two minor notes inline - neither is a blocker.

src/winml/modelkit/quant/quantizer.py
xieofxie: we already have comment for this for add_pre_process_metadata

tests/unit/test_quantizer.py
xieofxie: in next pr

PR #960, opened 2026-06-24
feat(perf): add --runtime [ort|openvino] to compare ORT vs OpenVINO

### LINKED ISSUE #948: feat: implement openvino run in perf to compare ?

xieofxie:
## What

Adds a `--runtime [ort|openvino]` flag to `winml perf` so the same ONNX file can be benchmarked on **ONNX Runtime** vs **OpenVINO Runtime** for a side-by-side comparison.

```bash
winml perf -m model.onnx --runtime openvino --device gpu
winml perf -m model.onnx --runtime ort --ep cpu   # ORT-native baseline
```

- Default is `ort` — existing behavior is unchanged.
- `--runtime openvino` reads the **raw ONNX directly** via OpenVINO Runtime (no quantize/optimize/compile build), which is the fair, simple comparison on the same graph. ONNX input only.

## How

- **`OpenVINOSession`** (`session/openvino/openvino_session.py`) mirrors the subset of `WinMLSession` the perf engine uses — `compile()` / `run()` / `perf()` plus `io_config` / `device` / `ep_name` / `running_model_path`. It reuses `get_io_config`, `load_onnx`, `PerfStats`, and `WinMLSession._get_precision`, so I/O metadata and reports match the ORT path. No model-specific logic.
- **`_OpenVINOModel`** adapter in `perf.py` exposes the `_single` surface the benchmark engine reads, so `_run_single` / `_run_benchmark*` / reporting are **untouched**. `PerfBenchmark._load_model()` branches to it and skips `WinMLAutoModel` + ORT EP resolution entirely (OpenVINO is independent of ORT's EPs).
- **`--device`** maps `cpu/gpu/npu/auto` → OpenVINO `CPU/GPU/NPU/AUTO`. `compile()` fails fast against `Core().available_devices` with a readable message instead of a raw backend stack trace.
- **`RuntimeName`** Literal + `RUNTIME_NAMES` in `constants.py` (mirrors `CompilerName`) — the CLI choice list and the typed config field derive from one source.
- CLI guards: `--runtime openvino` requires a `.onnx` input and rejects `--module`.

## Verified locally

- `--runtime openvino` runs on **CPU** and **GPU** end-to-end; latency/throughput populated.
- `--monitor` works on CPU and GPU (HW utilization via PDH; falls back to `NullEPMonitor` like most EPs — no OV-specific `ep_proof` telemetry yet).
- Absent device (NPU here) → friendly error: `OpenVINO device 'NPU' ... is not available. OpenVINO sees: ['CPU', 'GPU']`.
- New unit tests (`tests/unit/session/test_openvino_session.py`, gated on `importorskip("openvino")`) + CLI guard tests; all existing perf tests pass; ruff clean.

## Notes / follow-ups

- `--ep` and quant/optimize flags are intentional no-ops under `--runtime openvino` (raw ONNX) — documented in the flag help.
- On machines where the WinML registry installs the OpenVINO **EP**, `--runtime ort --device cpu` already routes ORT→OpenVINO EP; use `--runtime ort --ep cpu` for a true ORT-native baseline.
- Possible follow-up: surface `EXECUTION_DEVICES` in the report so AUTO-mode fallbacks are visible.

Closes #948

### DIFF
--- a/src/winml/modelkit/commands/perf.py
+++ b/src/winml/modelkit/commands/perf.py
@@ -28,7 +28,7 @@ from rich.console import Console
 from rich.table import Table
 
 from ..utils import cli as cli_utils
-from ..utils.constants import EPName, EPNameOrAlias
+from ..utils.constants import RUNTIME_NAMES, EPName, EPNameOrAlias, RuntimeName
 from ..utils.logging import configure_logging
 from ._live_chart import LiveMonitorDisplay
 
@@ -94,7 +94,7 @@ class BenchmarkConfig:
     shape_config: dict | None = None
     # Inference runtime backend: "ort" (ONNX Runtime, default) or "openvino"
     # (OpenVINO Runtime on the raw ONNX file, for ORT-vs-OV comparison).
-    runtime: str = "ort"
+    runtime: RuntimeName = "ort"
 
 
 @dataclass
@@ -1540,7 +1540,7 @@ def _run_simple_loop(
 )
 @click.option(
     "--runtime",
-    type=click.Choice(["ort", "openvino"]),
+    type=click.Choice(list(RUNTIME_NAMES)),
     default="ort",
     show_default=True,
     help="Inference runtime backend. 'ort' = ONNX Runtime (InferenceSession). "
@@ -1632,7 +1632,7 @@ def perf(
     precision: str,
     ep: EPNameOrAlias | None,
     ep_options: tuple[str, ...],
-    runtime: str,
+    runtime: RuntimeName,
     output: Path | None,
     batch_size: int,
     shape_config_path: Path | None,
--- a/src/winml/modelkit/commands/perf.py
+++ b/src/winml/modelkit/commands/perf.py
@@ -92,6 +92,9 @@ class BenchmarkConfig:
     ep: EPNameOrAlias | None = None
     ep_options: dict[str, str] | None = None
     shape_config: dict | None = None
+    # Inference runtime backend: "ort" (ONNX Runtime, default) or "openvino"
+    # (OpenVINO Runtime on the raw ONNX file, for ORT-vs-OV comparison).
+    runtime: str = "ort"
 
 
 @dataclass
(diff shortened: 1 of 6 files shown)


src/winml/modelkit/commands/perf.py
```diff
@@ -1613,6 +1692,17 @@ def perf(
...
+    # OpenVINO runner: ONNX input only, and no per-module path (submodule
+    # discovery walks a live nn.Module graph, which OpenVINO never builds).
+    if runtime == "openvino":
+        if Path(hf_model).suffix.lower() != ".onnx":
+            raise click.UsageError(
+                "--runtime openvino requires an ONNX (.onnx) model file "
+                f"(not a HuggingFace model ID), got: {hf_model}"
+            )
+        if module_class:
+            raise click.UsageError("--runtime openvino does not support --module benchmarking.")
```
DingmaomaoBJTU, line 1704: **Silent discard of `--ep` / `--ep-options` when `--runtime openvino` — no user feedback**

When `--runtime openvino` is combined with `--ep cuda` (or any EP), the value is forwarded into BenchmarkConfig but silently ignored because `_load_model()` returns before `_resolve_device_ep()` runs. A pattern already used in this file (--shape-config warning in --module mode) would work here — emit a yellow console warning so users know the flag had no effect.

src/winml/modelkit/session/openvino/openvino_session.py
```diff
@@ -0,0 +1,266 @@
...
+
+            model = load_onnx(self._onnx_path, load_weights=False, validate=False)
+            self._io_config = get_io_config(model)
+            # Reuse the operator-schema-based precision estimate (no
+            # architecture assumptions) so reports match the ORT path.
+            self._io_config["precision"] = WinMLSession._get_precision(model)
+        return self._io_config
+
+    @property
+    def device(self) -> str:
```
DingmaomaoBJTU, line 243: **`device` property returns input string, not resolved OpenVINO device**

After AUTO compilation `self._ov_device` holds the concrete target (e.g. `'GPU.0'`), but this property still returns `self._device` (e.g. `'auto'`). Since the perf engine reads `model.device` for report labelling, an AUTO-mode run will appear as `device='auto'` in JSON output rather than the true hardware. Consider `return self._ov_device or self._device` once compiled.

src/winml/modelkit/session/openvino/openvino_session.py
```diff
@@ -0,0 +1,266 @@
...
+        if self._perf_stats is not None:
+            result = self._perf_stats.record(lambda: compiled(ov_inputs))
+        else:
+            result = compiled(ov_inputs)
+
+        # Map outputs back to graph order. Index keying avoids OpenVINO
+        # output-name normalization mismatches (the order of model.outputs
+        # matches the ONNX graph output order get_io_config reads).
+        out_names = self.io_config["output_names"]
+        return {name: np.asarray(result[i]) for i, name in enumerate(out_names)}
```
DingmaomaoBJTU, line 181: **Output index mapping assumes io_config and compiled output count agree**

If the two disagree (e.g. optional outputs interpreted differently), the dict comprehension silently truncates or raises an IndexError with no context. A defensive length check before the comprehension would make failures easier to diagnose.

DingmaomaoBJTU (commented):
Overall the PR is well-structured: the adapter pattern cleanly separates OpenVINO from the ORT pipeline, error handling is thoughtful (lazy import, device pre-check, file existence), and test coverage is solid. Three findings below.

xieofxie:
Pros
- Internal check our top200 models on ORT VS Native
    - should cert team do this?
- From ISV case, they want to know the difference between ORT VS Native

Cons
- What if user finds that Native if better than ORT?
- Are we measuring correctly?
- It will lead to many different implementations for IHVs

PR #962, opened 2026-06-24
implement quantization debug

zhenchaoni:
# Design: `winml debug` — Per-Op Quantization Error Debugging

| | |
|---|---|
| **Status** | Implemented |
| **Area** | `debug`, `commands` |
| **Surface** | `winml debug` CLI · `winml.modelkit.debug.debug_quantization` API |

---

## 1. The problem

A user quantizes a model with the WinML CLI and the quantized artifact shows an accuracy drop in `winml eval`. The existing signals only say *that* accuracy dropped:

- `winml eval` — a single end-to-end metric.
- `winml eval --mode compare` — per-**output** parity (ONNX vs PyTorch).

Neither says **which tensors or weights** the quantization damaged. Without per-tensor attribution the user cannot tell whether the loss is concentrated in a few ops or spread across the whole graph, and has no starting point for mixed-precision tuning.

`winml debug` fills that gap: it compares the float ONNX model against its quantized counterpart **tensor by tensor** and reports where the quantized model diverges.

---

## 2. ONNX Runtime provides the measurement

We do **not** invent the augment / collect / match / compare mechanism. It is ONNX Runtime's officially documented quantization-debugging recipe, shipped as [`onnxruntime.quantization.qdq_loss_debug`](https://onnxruntime.ai/docs/performance/model-optimizations/quantization.html#quantization-debugging).

The idea: two graphs should compute the same function — the **float** reference and the **quantized** model (the float graph with QuantizeLinear/DequantizeLinear "QDQ" pairs inserted). Feed both the same input, capture *every* intermediate tensor from each, and compare them.

```mermaid
flowchart LR
    X[same input] --> F[Float model]
    X --> Q[Quantized model]
    F -->|t1_f, t2_f, ...| C[compare each tensor]
    Q -->|t1_q, t2_q, ...| C
    C --> R[per-tensor SQNR]
```

ORT does this in a handful of calls:

| Function | Role |
|---|---|
| `modify_model_output_intermediate_tensors` | Augment a model so every intermediate tensor becomes a graph output. |
| `collect_activations` | Run an augmented model over a calibration reader → `{tensor: [values]}`. |
| `create_activation_matching` | Align float/quant tensors and pre/post-QDQ pairs by name. |
| `compute_activation_error` | Reduce each match to local and cumulative SQNR. |
| `create_weight_matching` + `compute_weight_error` | Compare each weight float vs dequantized → weight SQNR. |

### The metric: SQNR (dB)

Each comparison reduces two tensors to a single quality number — the signal-to-quantization-noise ratio, in decibels:

```
SQNR = 20 * log10( norm(x) / norm(x - y) )   dB
```

where `x` is the float (reference) tensor and `x - y` is the quantization error. Higher dB means the error is small relative to the signal.

Because it is `20 * log10` of an amplitude ratio, **every 20 dB is a 10x change** in signal-to-error:

- **40 dB** — signal is ~100x the error (error ~1%): clean.
- **20 dB** — signal is ~10x the error (error ~10%): suspect.
- **< 20 dB** — error is more than a tenth of the signal: likely culprit.

SQNR is scale-invariant, so it is comparable across tensors of very different magnitude.

### Three signals

| Signal | Compares | Question it answers |
|---|---|---|
| **Local** | pre-QDQ vs post-QDQ (both in the quant model) | How lossy is quantizing **this tensor itself**? Upstream error is present on both sides and cancels. |
| **Cumulative** | float model vs quant model | How much **total** drift has reached this tensor, including error inherited from upstream? Tracks real end-to-end accuracy loss. |
| **Weight** | a weight's float value vs its dequantized value | How lossy is quantizing **this weight**? |

Reading local and cumulative together separates originators from inheritors: a tensor with high local but low cumulative quantizes fine and merely inherited upstream damage, whereas low local **and** low cumulative means the tensor's own rounding is lossy.

---

## 3. Integration: `debug_quantization`

The engine ([`src/winml/modelkit/debug/debugger.py`](../../src/winml/modelkit/debug/debugger.py)) wraps the ORT calls with WinML's calibration-data plumbing and result shaping. It is import-light and Click/Rich-free, mirroring the `quant/quantizer.py` ↔ `commands/quantize.py` split.

```python
def debug_quantization(
    float_model_path, quant_model_path, *,
    samples=8, model_id=None, task=None,
) -> dict
```

What WinML adds on top of ORT:

- **Calibration inputs** via `DatasetCalibrationReader` — task-aware and correctly preprocessed when `model_id`/`task` are given, otherwise random inputs synthesized from the model's I/O spec (self-contained, no downloads). The reader is `rewind()`-ed between the float and quant collection passes so both models see the identical sample sequence.
- **CPU execution**, matching ORT's quantization-debugging guidance.
- **A scalar-tolerant SQNR wrapper** (`_sqnr_db`) passed as `err_func` to `compute_weight_error`: ORT's SQNR calls `len()` on its inputs, which fails for a weight that dequantizes to a numpy scalar; `numpy.atleast_1d` keeps such weights as length-1 arrays.
- **Result shaping + summary statistics**: non-finite SQNR values (from overflow or zero-difference tensors) are dropped from the summary so `mean`/`std` stay well-defined.

### Return shape

```python
{
    "activations": [
        # cumulative_sqnr_db is None when there is no float reference
        {"tensor_name", "local_sqnr_db", "cumulative_sqnr_db"},
        ...
    ],
    "weights": [
        {"weight_name", "weight_sqnr_db"},
        ...
    ],
    "model_outputs": [
        # cumulative SQNR at each graph output
        {"output_name", "cumulative_sqnr_db"},
        ...
    ],
    "summary": {
        "local":      {"count", "mean", "std", "min", "max"},
        "cumulative": {"count", "mean", "std", "min", "max"},
        "weight":     {"count", "mean", "std", "min", "max"},
    },
}
```
---

## 4. CLI surface and output

```bash
winml debug --float-model FLOAT.onnx --quant-model QUANT.onnx [options]
```

| Flag | Default | Purpose |
|---|---|---|
| `--float-model` | *(required)* | Pre-quantization float ONNX model. |
| `--quant-model` | *(required)* | Quantized (QDQ) artifact to debug. |
| `--samples` | `8` | Input samples to average over. |
| `--model-id` | — | HF id for task-aware calibration inputs. |
| `--task` | — | Task for task-aware calibration (else random inputs). |
| `--output` | — | Write the full per-tensor results as JSON. |

The command ([`src/winml/modelkit/commands/debug.py`](../../src/winml/modelkit/commands/debug.py)) prints four tables — the cumulative SQNR at every graph output (in full), and the top-10 worst **local**, **cumulative**, and **weight** SQNR — each followed by a one-line summary `(count, mean, std, min, max)`. SQNR colouring: **> 40 dB** green · **20–40 dB** yellow · **< 20 dB** red.

```
Float model: temp\fairface\optimized.onnx
Quant model: temp\fairface\quantized.onnx
Samples: 8

      Model outputs
┏━━━┳━━━━━━━━━━━┳━━━━━━━━┓
┃ # ┃ SQNR (dB) ┃ Output ┃
┡━━━╇━━━━━━━━━━━╇━━━━━━━━┩
│ 1 │     22.71 │ logits │
└───┴───────────┴────────┘
Local      = error from quantizing this tensor alone, excluding upstream.
Cumulative = error at this tensor, including error inherited from upstream.

                               Top 10 worst local SQNR
┏━━━━┳━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┓
┃  # ┃ SQNR (dB) ┃ Tensor                                                            ┃
┡━━━━╇━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┩
│  1 │     32.20 │ gemm_output_reshape_arg_token_238                                 │
│  2 │     34.65 │ gemm_output_reshape_arg_token_202                                 │
│  3 │     36.88 │ gemm_output_reshape_arg_token_388                                 │
│  4 │     39.98 │ gemm_output_reshape_arg_token_412                                 │
│  5 │     42.03 │ gemm_output_reshape_arg_token_304                                 │
│  6 │     47.60 │ gemm_output_reshape_arg_token_292                                 │
│  7 │     52.25 │ /vit/encoder/layer.0/layernorm_before/LayerNormalization_output_0 │
│  8 │     53.10 │ gemm_output_reshape_arg_token_340                                 │
│  9 │     53.82 │ gemm_output_reshape_arg_token_64                                  │
│ 10 │     58.91 │ gemm_output_reshape_arg_token_172                                 │
└────┴───────────┴───────────────────────────────────────────────────────────────────┘
                               Top 10 worst cumulative SQNR
┏━━━━┳━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┓
┃  # ┃ SQNR (dB) ┃ Tensor                                                                ┃
┡━━━━╇━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┩
│  1 │      8.29 │ gemm_output_reshape_arg_token_388                                     │
│  2 │      8.29 │ /vit/encoder/layer.10/output/dense/Add_output_0                       │
│  3 │      9.56 │ gemm_input_reshape_arg_token_385                                      │
│  4 │      9.56 │ /vit/encoder/layer.10/intermediate/intermediate_act_fn/Mul_1_output_0 │
│  5 │     10.66 │ /vit/encoder/layer.11/attention/attention/Softmax_output_0            │
│  6 │     11.23 │ /vit/encoder/layer.11/attention/attention/Transpose_output_0          │
│  7 │     11.23 │ /vit/encoder/layer.11/attention/attention/Reshape_1_output_0          │
│  8 │     11.23 │ gemm_output_reshape_arg_token_394                                     │
│  9 │     11.34 │ /vit/encoder/layer.11/intermediate/intermediate_act_fn/Mul_1_output_0 │
│ 10 │     11.34 │ gemm_input_reshape_arg_token_421                                      │
└────┴───────────┴───────────────────────────────────────────────────────────────────────┘
       Top 10 worst weight SQNR
┏━━━━┳━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━┓
┃  # ┃ SQNR (dB) ┃ Weight            ┃
┡━━━━╇━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━┩
│  1 │    -34.67 │ onnx::MatMul_1310 │
│  2 │    -34.58 │ onnx::MatMul_1426 │
│  3 │    -34.19 │ onnx::MatMul_1281 │
│  4 │    -33.88 │ onnx::MatMul_1279 │
│  5 │    -33.34 │ onnx::MatMul_1455 │
│  6 │    -33.33 │ onnx::MatMul_1397 │
│  7 │    -32.09 │ onnx::MatMul_1454 │
│  8 │    -32.03 │ onnx::MatMul_1339 │
│  9 │    -29.76 │ onnx::MatMul_1368 │
│ 10 │    -29.64 │ onnx::MatMul_1484 │
└────┴───────────┴───────────────────┘
```

---

### DIFF
--- a/src/winml/modelkit/commands/debug.py
+++ b/src/winml/modelkit/commands/debug.py
@@ -1,0 +1,211 @@
+# -------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+# --------------------------------------------------------------------------
+r"""Measure per-op quantization error (local and cumulative SQNR).
+
+Usage:
+    winml debug --float-model float.onnx --quant-model quantized.onnx
+
+Examples:
+    # Random inputs (self-contained, no downloads)
+    winml debug --float-model model_optimized.onnx --quant-model model_quantized.onnx
+
+    # Real, task-aware calibration inputs
+    winml debug --float-model float.onnx --quant-model qdq.onnx \\
+        --model-id microsoft/swinv2-tiny-patch4-window16-256 \\
+        --task image-classification --samples 16
+"""
+
+from __future__ import annotations
+
+import json
+import logging
+from pathlib import Path
+
+import click
+from rich.console import Console
+from rich.table import Table
+
+from ..utils import cli as cli_utils
+from ..utils.logging import configure_logging
+
+
+logger = logging.getLogger(__name__)
+console = Console()
+
+
+@click.command("debug")
+@click.option(
+    "--float-model",
+    "float_model",
+    required=True,
+    type=click.Path(exists=True, path_type=Path),
+    help="Float (pre-quantization) ONNX model.",
+)
+@click.option(
+    "--quant-model",
+    "quant_model",
+    required=True,
+    type=click.Path(exists=True, path_type=Path),
+    help="Quantized (QDQ) ONNX model — the build artifact to debug.",
+)
+@click.option(
+    "--samples",
+    type=int,
+    default=2,
+    show_default=True,
+    help="Number of input samples to average over.",
+)
+@click.option(
+    "--model-id",
+    type=str,
+    default=None,
+    help="HuggingFace model id for real, task-aware calibration inputs.",
+)
+@click.option(
+    "--task",
+    type=str,
+    default=None,
+    help="Task for task-aware calibration (e.g. 'image-classification'). "
+    "Falls back to random inputs when omitted.",
+)
+@cli_utils.output_option("Write the full per-tensor results to this JSON file.")
+@cli_utils.verbosity_options()
+@click.pass_context
+def debug(
+    ctx: click.Context,
+    float_model: Path,
+    quant_model: Path,
+    samples: int,
+    model_id: str | None,
+    task: str | None,
+    output: Path | None,
+    verbose: int,
+    quiet: bool,
+) -> None:
+    """Measure per-op quantization error, op by op.
+
+    Runs the float and quantized models over the same inputs and reports, per
+    activation, the local SQNR and the cumulative SQNR. Lower dB == more damage.
+
+    Local SQNR is the error from quantizing this tensor alone, excluding
+    upstream. Cumulative SQNR is the error at this tensor, including error
+    inherited from upstream.
+    """
+    verbose, quiet = cli_utils.resolve_verbosity(ctx, verbose, quiet)
+    configure_logging(verbosity=verbose, quiet=quiet)
+
+    from ..debug import debug_quantization
+
+    console.print(f"[bold blue]Float model:[/bold blue] {float_model}")
+    console.print(f"[bold blue]Quant model:[/bold blue] {quant_model}")
+    console.print(f"[bold blue]Samples:[/bold blue] {samples}\n")
+
+    result = debug_quantization(
+        float_model,
+        quant_model,
+        samples=samples,
+        model_id=model_id,
+        task=task,
+    )
+
+    print_result(result)
+
+    if output is not None:
+        output.parent.mkdir(parents=True, exist_ok=True)
+        output.write_text(json.dumps(result, indent=2), encoding="utf-8")
+        console.print(f"\n[dim]Full per-tensor results written to {output}[/dim]")
+
+# Number of worst-ranked rows shown per table.
+TOP_N = 10
+
+
+def print_result(result: dict) -> None:
+    """Render the debug result dict as console tables."""
+    activations = result["activations"]
+    weights = result["weights"]
+    model_outputs = result["model_outputs"]
+    summary = result["summary"]
+
+    console.print(
+        "Local SQNR      = error from quantizing this tensor alone, excluding upstream."
+    )
+    console.print(
+        "Cumulative SQNR = error at this tensor, including error inherited from upstream."
+    )
+
+    # Model outputs: cumulative SQNR at every graph output (shown in full).
+    _render_table(
+        "Outputs cumulative SQNR",
+        "Output",
+        [(o["output_name"], o["cumulative_sqnr_db"]) for o in model_outputs],
+    )
+
+    local_sorted = sorted(activations, key=lambda a: a["local_sqnr_db"])
+    _render_table(
+        f"Top {TOP_N} worst local SQNR",
+        "Tensor",
+        [(a["tensor_name"], a["local_sqnr_db"]) for a in local_sorted],
+        top=TOP_N,
+    )
+    _print_stats(summary["local"])
+
+    cumulative_sorted = sorted(
+        activations,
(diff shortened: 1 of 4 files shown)


xieofxie:
have you tried this? I tried this before and found that even the diff for internal nodes are big, the final result could be good. So it is not very useful for identifying bugs..

PR #965, opened 2026-06-25
test(quantize): use fake model for image-segmentation e2e calibration

KayMKM:
The image-segmentation quantize e2e test exported and calibrated the real nvidia/segformer-b0-finetuned-ade-512-512 model, whose heavy backbone caused random hangs on QNN hosts. Replace it with a tiny FP32 conv model that has identical segmentation I/O (pixel_values [batch, 3, H, W] -> logits [batch, num_labels, H/4, W/4]) and random weights, built via the new _build_fake_segmentation_onnx() helper.

The ImageSegmentationDataset path is still fully exercised: the test keeps --model-name so the real image processor and dataset samples feed real pixel_values into calibration; only the heavy model is swapped out.


tests/e2e/test_quantize_e2e.py
```diff
@@ -62,6 +62,72 @@ def _build_tiny_onnx(path: Path, *, with_metadata: bool = True) -> None:
...
 
 
+# Image-segmentation I/O contract shared by HF semantic-segmentation exports
+# (e.g. nvidia/segformer-*-ade-*): pixel_values [batch, 3, height, width] ->
+# logits [batch, num_labels, height/4, width/4]. 150 = ADE20K class count.
+_SEG_NUM_CHANNELS = 3
+_SEG_NUM_LABELS = 150
+
+
+def _build_fake_segmentation_onnx(path: Path) -> None:
```
zhenchaoni: You can consider putting this logic inside fixtures\create_test_models.py

KayMKM: updated

zhenchaoni (approved):

PR #970, opened 2026-06-25
feat(cli): add --overwrite/--no-overwrite output safety to all output commands

### LINKED ISSUE #561: [cross-cutting] [P2] No safe-overwrite control on any output-producing command
**Severity**: P1  
**Command**: `cross-cutting`

---
**Category**: Output File Safety — destructive overwrite must be opt-in.

**Repro**: Run any output-producing command twice with the same `-o`:
```
uv run winml export -m microsoft/resnet-50 -o temp\ux\out.onnx
"hello world" | Out-File temp\ux\out.onnx -Force
uv run winml export -m microsoft/resnet-50 -o temp\ux\out.onnx
Get-Item temp\ux\out.onnx | Select-Object Length
```

**Actual**: The second invocation silently overwrites the file the user just wrote. No warning, no prompt, no `--force` requirement. Reproduced for: `export`, `optimize`, `quantize`, `compile`, `perf -o`, `build -o`, `config -o`, `eval -o`, `analyze --output`, `hub -o`. None of these commands have a `--force` flag.

**Expected**: Add a shared `--force / -f` flag (default `False`) to every output-producing command. At entry:
```python
if output.exists() and not force:
    raise click.ClickException(
        f"Output '{output}' already exists. Re-run with --force to overwrite."
    )
```

**Why it matters**: Users keep multiple variant outputs in one folder (`q1.onnx`, `q1_int8.onnx`, etc.) and easily clobber yesterday's good model with today's broken one. There is no recovery — the old file is gone.

xieofxie:
## Summary

Output-producing commands silently overwrote an existing file/dir on a re-run — no warning, no prompt, no opt-in (#561). This adds a shared safe-overwrite control so a re-run can't destroy a previous result.

### Flag name
Chose **`--overwrite/--no-overwrite`** (default `--no-overwrite`) over the issue body's `--force/-f` because it matches the dominant `--x/--no-x` toggle house style, is precise to the file-overwrite concern, avoids the `-f` collision (`run --file`), and won't be confused with the existing `--rebuild`/cache controls.

### Shared helpers (`utils/cli.py`)
- `overwrite_option()` — `--overwrite/--no-overwrite`, default no-overwrite.
- `guard_output(path, overwrite, *, label)` — fail-fast existence check. Files block when they exist; **directories block only when non-empty** (a fresh/empty output dir doesn't false-trigger); a `None` path (stdout) is a no-op.

### Wired into 10 commands
`export`, `optimize`, `quantize`, `compile`, `config`, `eval`, `analyze`, `perf`, `catalog`, `run`.

- Guard runs **before** any mkdir/cleanup/work — notably before the quantizer's destructive stale-sidecar `unlink()`.
- Covers **default-derived** output paths (e.g. `{stem}_qdq.onnx`), not just explicit `-o`.
- `analyze` guards both `--output` and `--optim-config`.
- `compile` guards the resolved `-o` file or `--output-dir` (non-empty).

### `build` intentionally excluded
`build` already has `--rebuild` plus incremental resume — without `--rebuild`, existing stage artifacts are *reused*, not clobbered (e.g. quantize stage skips when QDQ nodes are present). A non-empty-output-dir guard there would break legitimate incremental rebuilds and overlap confusingly with `--rebuild`.

## Tests
- `guard_output` / `overwrite_option` unit tests (file/dir/empty-dir/overwrite-bypass/None/custom-label).
- Real-command wiring: `catalog` (file), `quantize` (file + default-derived path + verifies `quantize_onnx` is not called when blocked), `compile` (empty dir OK, non-empty blocked, `--overwrite` proceeds).
- Full `tests/unit/commands/` + `tests/unit/utils/test_cli.py` suite green (916 + new tests).

Fixes #561


DingmaomaoBJTU (approved):
Overall the design is clean and the coverage is thorough — approve with two minor notes left as inline comments.

- guard_output() utility is well-scoped (non-empty dir semantics, None is a no-op, str coercion).
- Fail-fast placement is correct in the majority of commands (before model load / eval / compile).
- Test suite covers the happy path, both flag polarities, default derivation, str vs Path, custom label, and directory cases.

Issue #510, opened 2026-05-09
[winml] [P2] Multiple brand spellings in use (`WML`, `WinML`, `winml`, `Windows ML`, `ModelKit`)


zhenchaoni:
**Severity**: P2  
**Command**: `winml`

---
**Category**: Naming Consistency — pick one canonical product name and use it everywhere.

**Repro**:
```
uv run winml --help
uv run winml --version
uv run winml export --help
```

**Actual**: Five different forms appear in user-visible output:
- `winml --help` first line: `WML ModelKit - Accelerate Model Deployment on WinML.` (uses **WML** and **WinML** in the same sentence).
- `winml --version` prints: `winml, version 0.0.2` (lowercase **winml**).
- `winml export --help` first line: `Export HuggingFace model to ONNX format with HTP.` (no brand).
- `winml inspect` panel header reads `WinML Inference Class`.
- `winml sys` table headers read `Windows ML` for the OS service and `winml` for the CLI.

**Expected**: Pick one canonical capitalization (e.g. **WinML ModelKit** for the product, `winml` only as the literal command name) and use it in every place a user can read it: top-level docstring, every subcommand docstring, `--version` output, every panel/table title, `README.md`, and `pyproject.toml`. Concretely:
- Replace `WML ModelKit` → `WinML ModelKit` in the top-level docstring.
- Make `winml --version` print `WinML ModelKit, version 0.0.2 (winml CLI)`.

**Why it matters**: Users searching docs/issues for the product name need one term. Marketing/legal usually require one canonical capitalization too.

DingmaomaoBJTU:
**Triage summary**

- **Component**: Other
- **Priority**: P2

Multiple brand spellings in use (WML, WinML, winml, Windows ML, ModelKit) across the codebase. Standardize to a single canonical name.

DingmaomaoBJTU:
**Triage summary**

- **Component**: Other
- **Priority**: P2

Multiple brand spellings (`WML`, `WinML`, `winml`, `Windows ML`) are used inconsistently across the CLI. Next step: standardize on `WinML` for display text and `winml` for CLI commands throughout.

PR #978, opened 2026-06-25
Fix multiple consistency bugs

### LINKED ISSUE #565: [cross-cutting] [P2] `-d` short used only by `compile --device`; everyone else uses `--device` with no short
**Severity**: P2  
**Command**: `cross-cutting`

---
**Category**: Reserved Short Flags — each short flag means one thing, used the same way across commands.

**Repro**:
```
uv run winml compile --help | Select-String '^\s+-d,'
uv run winml build --help   | Select-String '^\s+-d,'
uv run winml eval --help    | Select-String '^\s+-d,'
uv run winml perf --help    | Select-String '^\s+-d,'
```

**Actual**: Only `compile` declares `-d, --device`. Other commands take `--device` long-form only.

**Expected**: Either drop `-d` from `compile` (so `--device` is consistently long-form everywhere), or add `-d, --device` to all four commands. The shared `device_option` decorator from another audit finding should make this trivial.

**Why it matters**: A user who runs `winml compile -d npu` then tries `winml build -d npu` gets "no such option" with no hint that `--device` is the long form.

zhenchaoni:
Fixes  #565, #510

For #565 ,
For all command, use the shared `device_op[tion` and `ep_option` util. Therefore, all commands can have consistent CLI option name.

For #510 ,
Some URL in our project is still referencing the old wording "modelkit". Now change them to "winml"

### DIFF
--- a/src/winml/modelkit/analyze/core/runtime_checker.py
+++ b/src/winml/modelkit/analyze/core/runtime_checker.py
@@ -206,7 +206,7 @@ class RuntimeChecker:
             if not self._has_any_rule_data:
                 logger.warning(
                     "No runtime check data found. Follow "
-                    "https://github.com/microsoft/WinML-ModelKit/blob/main/CONTRIBUTING.md "
+                    "https://github.com/microsoft/winml-cli/blob/main/CONTRIBUTING.md "
                     "to set up runtime check files."
                 )
             else:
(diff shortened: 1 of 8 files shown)


xieofxie:
There are already two prs 
https://github.com/microsoft/winml-cli/pull/818/changes
https://github.com/microsoft/winml-cli/pull/816/changes

Which is better? Should we merge the others first?

zhenchaoni:
> There are already two prs https://github.com/microsoft/winml-cli/pull/818/changes https://github.com/microsoft/winml-cli/pull/816/changes
> 
> Which is better? Should we merge the others first?

@xieofxie The changes are basically the same. To avoid conflict, I will remove those duplicate changes in my PR. Then, this PR will not overlap with the other 2.

xieofxie (approved):

PR #984, opened 2026-06-26
Rename quantization option from --model-name to --model-id

### LINKED ISSUE #563: [cross-cutting] [P2] `--model` semantic drift across 10 commands

zhenchaoni:
Fix #563 

This is part of consistency work. Because "model ID" is the correct term for a hugging face model identifier like `microsoft/resnet-50`. So I make the user facing options consistent.

In this PR, I
- renamed `winml quantize` option from `--model-name` to `--model-id`
- renamed the config json's `quantize`->`model_name` property to `model_id`
- renamed the recipes

Caveat: This changes the user facing option name. I think it is fine since we are still in early releases.

About the consistency of our commands. Currently, all our commands follow the practice:
- -m or --model can be used for either onnx model path or huggingface model id
- --model-id is used for huggingface model id only, and only when the -m option is used for onnx model. There are cases where our command requires both an onnx path and model ID. For this case, the model ID goes to --model-id option.

### DIFF
--- a/src/winml/modelkit/commands/quantize.py
+++ b/src/winml/modelkit/commands/quantize.py
@@ -98,11 +98,8 @@ console = Console()
     default=None,
     help="Task for calibration dataset selection (e.g., 'image-classification').",
 )
-@click.option(
-    "--model-id",
-    type=str,
-    default=None,
-    help="HuggingFace model id (e.g., 'microsoft/resnet-50'). When provided "
+@cli_utils.model_id_option(
+    help_text="HuggingFace model id (e.g., 'microsoft/resnet-50'). When provided "
     "with --task, enables task-aware calibration datasets using the model's preprocessor.",
 )
 @cli_utils.build_config_option()
--- a/src/winml/modelkit/commands/quantize.py
+++ b/src/winml/modelkit/commands/quantize.py
@@ -99,10 +99,10 @@ console = Console()
     help="Task for calibration dataset selection (e.g., 'image-classification').",
 )
 @click.option(
-    "--model-name",
+    "--model-id",
     type=str,
     default=None,
-    help="HuggingFace model name (e.g., 'microsoft/resnet-50'). When provided "
+    help="HuggingFace model id (e.g., 'microsoft/resnet-50'). When provided "
     "with --task, enables task-aware calibration datasets using the model's preprocessor.",
 )
 @cli_utils.build_config_option()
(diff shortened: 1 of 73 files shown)


src/winml/modelkit/commands/quantize.py
```diff
@@ -99,10 +99,10 @@
     help="Task for calibration dataset selection (e.g., 'image-classification').",
 )
 @click.option(
```
xieofxie: there is also model-id flag in eval, consider share them

zhenchaoni: Fixed

timenick (approved):

xieofxie (approved):

Issue #964, opened 2026-06-25
Refactor quantizer into Quantizer class with BaseQuantPass pipeline


DingmaomaoBJTU:
## Context

Following the architecture pattern established by `optim/optimizer.py` (pipe-based pipeline), refactor the quantizer from standalone functions into a class-based pass pipeline for better extensibility and custom multi-pass support.

## Current State

- `quant/quantizer.py` has `quantize_onnx()` function with `_quantize_fp16`, `_quantize_rtn`, `_quantize_qdq` helper functions
- Multi-pass logic (`w4a16 → [int4, fp16]`) is handled internally via `_run_multi_pass()`
- Users cannot customize pass ordering or per-pass parameters in multi-pass scenarios

## Proposed Design

```python
# quant/passes/base.py
class BaseQuantPass:
    def should_run(self, config) -> bool: ...
    def build_config(self, precision, config) -> PassConfig: ...
    def run(self, model_path, config) -> Path: ...

# quant/passes/rtn.py, fp16.py, qdq.py
class RTNPass(BaseQuantPass): ...
class FP16Pass(BaseQuantPass): ...
class QDQPass(BaseQuantPass): ...

# quant/quantizer.py
class Quantizer:
    passes = [RTNPass, QDQPass, FP16Pass]

    def quantize(self, model_path, *, precision=None, config=None, passes=None):
        resolved_passes = passes or self._resolve_passes(precision, config)
        for qpass in resolved_passes:
            model_path = qpass.run(model_path, ...)
        return result
```

## Benefits

1. **Extensibility** — adding a new quantization algorithm = adding a new Pass class
2. **Custom multi-pass** — users can pass explicit `passes=[RTNPass(bits=4), FP16Pass()]`
3. **Consistency** — mirrors `optim/optimizer.py` pipe architecture
4. **Per-pass config** — each pass can have different parameters in multi-pass scenarios
5. **Pass coordination** — passes can inspect previous pass outputs (e.g., FP16 skips already-quantized ops)

## Scope

- Extract `_quantize_fp16`, `_quantize_rtn`, `_quantize_qdq` into Pass classes
- Create `BaseQuantPass` ABC
- Refactor `quantize_onnx()` into `Quantizer.quantize()`
- Keep backward-compatible `quantize_onnx()` as a thin wrapper
- Also addresses #963 (cross-pass op coordination)

## Related

- PR #872 (precision-driven quantization — current implementation)
[...]

PR #985, opened 2026-06-26
feat(quant): Quantizer class with BaseQuantPass pipeline (#964)

DingmaomaoBJTU:
## Summary

Refactors the quantizer from a flat dispatch function into an extensible pass-based pipeline, as tracked in #964.

### New: `passes/` sub-package

| File | Class | What it does |
|---|---|---|
| `passes/base.py` | `BaseQuantPass` | ABC — `__init__(config)` + abstract `run(model_path, output_path) -> QuantizeResult` |
| `passes/fp16.py` | `FP16Pass` | Reads `fp16_keep_io_types`, `fp16_op_block_list` from config |
| `passes/rtn.py` | `RTNPass` | Reads `rtn_bits`, `rtn_block_size`, `rtn_symmetric`, `rtn_accuracy_level` from config |
| `passes/static.py` | `StaticPass` | Reads all QDQ/calibration fields from config |

All passes accept a single `WinMLQuantizationConfig` — each reads only its relevant fields.

### Refactored: `quantizer.py`

- **`Quantizer(passes)`** — chains passes sequentially; single-pass takes the direct path, multi-pass routes intermediates through a `TemporaryDirectory`; merges `QuantizeResult` stats across passes
- **`expand_precision(mode, config)`** — maps mode strings to pass lists; `mode` is optional and falls back to `config.mode`:
  - `"fp16"` → `[FP16Pass(config)]`
  - `"rtn"` → `[RTNPass(config)]`
  - `"static"` / `"dynamic"` → `[StaticPass(config)]`
- **`quantize_onnx()`** — kept as backward-compatible entry point; now delegates to `Quantizer`

### Updated: `commands/quantize.py`

- `--precision` now accepts multiple values to compose a pass pipeline (e.g. `--precision int4 --precision fp16` runs RTN then FP16)
- Default output name for multi-pass: `{stem}_int4_fp16.onnx`

## Tests

`tests/unit/test_quant_passes.py` — 19 tests, all passing:

- `TestExpandPrecision` (7) — mapping correctness, unknown mode, None config, mode-from-config fallback
- `TestQuantizerSinglePass` (4) — path routing, missing model, exception handling, empty passes guard
- `TestQuantizeOnnxKwargsGuard` (1) — unexpected kwargs raise `TypeError`
- `TestQuantizerMultiPass` (4) — chaining, stat merging, abort on failure, warning accumulation
- `TestFP16PassConfig` (1) — config field wiring
- `TestRTNPassConfig` (2) — config field wiring, accuracy_level=0 → None

`tests/e2e/test_quantize_e2e.py` — `TestMultiPrecision` (2 tests):

- `test_int4_then_fp16_pipeline` — verifies `MatMulNBits` nodes (RTN) + `FLOAT16` initializers (FP16) are both present
- `test_pipeline_default_output_path` — verifies auto-named output file

### DIFF
--- a/src/winml/modelkit/commands/quantize.py
+++ b/src/winml/modelkit/commands/quantize.py
@@ -330,8 +330,6 @@ def _run_multi_precision(
     console: Console,
 ) -> None:
     """Execute a multi-pass quantization pipeline from ordered precision strings."""
-    from typing import cast
-
     from ..config.precision import extract_weight_bits
     from ..quant import Quantizer, WinMLQuantizationConfig, expand_precision
 
--- a/src/winml/modelkit/commands/quantize.py
+++ b/src/winml/modelkit/commands/quantize.py
@@ -13,7 +13,7 @@ Usage:
 Examples:
     winml quantize -m model.onnx
     winml quantize -m model.onnx --precision int8
-    winml quantize -m model.onnx -o model_qdq.onnx --samples 100
+    winml quantize -m model.onnx -o model_quantized.onnx --samples 100
     winml quantize -m model.onnx --weight-type int8 --activation-type uint8
 """
 
@@ -47,7 +47,7 @@ console = Console()
     type=click.Path(exists=True, path_type=Path),
     help="Input ONNX model file",
 )
-@cli_utils.output_option("Output path (default: {input}_qdq.onnx)")
+@cli_utils.output_option("Output path (default: {input}_quantized.onnx)")
 @cli_utils.overwrite_option()
 @cli_utils.precision_option(
     default=(),
@@ -243,7 +243,7 @@ def quantize(
             single, weight_type, activation_type
         )
         if output is None:
-            output = model.parent / f"{model.stem}_qdq.onnx"
+            output = model.parent / f"{model.stem}_quantized.onnx"
         config = WinMLQuantizationConfig(
             samples=samples,
             calibration_method=cast('Literal["minmax", "entropy", "percentile"]', method),
--- a/src/winml/modelkit/commands/quantize.py
+++ b/src/winml/modelkit/commands/quantize.py
@@ -347,14 +347,22 @@ def _run_multi_precision(
         4,
     )
 
+    # Resolve weight/activation types from the first static precision in the list
+    # (same logic as single-pass path) so -p int16 -p fp16 uses int16, not uint8.
+    first_static = next(
+        (p for p in precision if _cli_precision_to_mode(p) == "static"),
+        None,
+    )
+    resolved_weight, resolved_activation = _resolve_quant_types(
+        first_static, weight_type, activation_type
+    )
+
     config = WinMLQuantizationConfig(
         rtn_bits=rtn_bits,
         samples=samples,
         calibration_method=cast('Literal["minmax", "entropy", "percentile"]', method),
-        weight_type=cast('Literal["uint8", "int8", "uint16", "int16"]', weight_type or "uint8"),
-        activation_type=cast(
-            'Literal["uint8", "int8", "uint16", "int16"]', activation_type or "uint8"
-        ),
+        weight_type=cast('Literal["uint8", "int8", "uint16", "int16"]', resolved_weight),
+        activation_type=cast('Literal["uint8", "int8", "uint16", "int16"]', resolved_activation),
         per_channel=per_channel,
         symmetric=symmetric,
         task=task,
--- a/src/winml/modelkit/commands/quantize.py
+++ b/src/winml/modelkit/commands/quantize.py
@@ -50,11 +50,13 @@ console = Console()
 @cli_utils.output_option("Output path (default: {input}_qdq.onnx)")
 @cli_utils.overwrite_option()
 @cli_utils.precision_option(
-    default=None,
-    help_text="Quantization precision: auto, fp16, int4, int8, int16, or w{x}a{y} where "
-    "x in {4,8,16}, y in {8,16} (e.g., w4a16, w8a8, w8a16). "
-    "int4/w4a16 uses RTN weight-only quantization; "
-    "fp16 converts all FP32 tensors to FP16 (no QDQ)",
+    default=(),
+    multiple=True,
+    help_text="Quantization precision: fp16, int4, int8, int16, or w{x}a{y} where "
+    "x in {4,8,16}, y in {8,16} (e.g., w8a8, w8a16). "
+    "int4 uses RTN weight-only quantization; "
+    "fp16 converts all FP32 tensors to FP16 (no QDQ). "
+    "Repeat to chain passes in order (e.g. -p int4 -p fp16)",
     optional_message="Overridden by explicit --weight-type/--activation-type",
 )
 @click.option(
@@ -111,7 +113,7 @@ def quantize(
     model: Path,
     output: Path | None,
     overwrite: bool,
-    precision: str | None,
+    precision: tuple[str, ...],
     samples: int,
     method: str,
     weight_type: str | None,
@@ -127,8 +129,11 @@ def quantize(
     r"""Quantize ONNX model by inserting QDQ nodes, RTN weight-only, or convert to FP16.
 
     This command applies quantization to an ONNX model. The algorithm is
-    auto-selected from the precision: int4/w4a16 → RTN weight-only,
+    auto-selected from the precision: int4 → RTN weight-only,
     int8/int16/w8a8 → static QDQ, fp16 → FP16 conversion.
+
+    Repeat --precision to chain passes in order:
+    ``-p int4 -p fp16`` runs RTN int4 quantization then FP16 conversion.
 
     \b
     Examples:
@@ -140,6 +145,9 @@ def quantize(
 
         # RTN 4-bit weight-only quantization (no calibration data needed)
         winml quantize -m model.onnx --precision int4
+
+        # RTN int4 followed by FP16 conversion (two-pass pipeline)
+        winml quantize -m model.onnx --precision int4 --precision fp16
 
         # Int16 quantization
         winml quantize -m model.onnx --precision int16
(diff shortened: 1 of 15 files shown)


src/winml/modelkit/quant/passes/base.py
```diff
@@ -0,0 +1,59 @@
...
+
+    @property
+    def config(self) -> WinMLQuantizationConfig:
+        """Return the shared quantization configuration."""
+        return self._config
+
+    @abstractmethod
+    def run(
+        self,
+        model_path: Path,
```
xieofxie, line 44: should we use in memory model proto directly? to be faster

src/winml/modelkit/quant/passes/static.py
```diff
@@ -0,0 +1,209 @@
+# -------------------------------------------------------------------------
```
xieofxie, line 1: given we don't support dynamic qdq, how about name is qdq_static.py? leave qdq_dynamic.py as future work

src/winml/modelkit/quant/quantizer.py
```diff
@@ -7,398 +7,253 @@
...
 
+# Precision strings that expand to multiple sequential passes.
+_COMPOSITE_PRECISIONS: dict[str, list[str]] = {}
 
-def quantize_onnx(
-    model_path: str | Path,
-    output_path: str | Path | None = None,
+
+def expand_precision(
+    mode: str,
```
xieofxie: why we have mode parameter? since config already have it

src/winml/modelkit/quant/passes/static.py
DingmaomaoBJTU: Let me rename to static, align with options

timenick (approved):

src/winml/modelkit/quant/quantizer.py
```diff
@@ -7,422 +7,266 @@
...
-    logger.info(
-        "Quantization complete: %s -> %s (%.2fs)",
-        model_path.name,
-        output_path.name,
-        total_time,
-    )
+    if output_path is not None:
+        output_path = Path(output_path)
+    else:
+        output_path = model_path.parent / f"{model_path.stem}_qdq.onnx"
```
xieofxie: should we use `_quantized` instead?

src/winml/modelkit/quant/quantizer.py
```diff
@@ -7,422 +7,266 @@
...
 
+# Precision strings that expand to multiple sequential passes.
+_COMPOSITE_PRECISIONS: dict[str, list[str]] = {}
 
-def quantize_onnx(
-    model_path: str | Path,
-    output_path: str | Path | None = None,
+
+def expand_precision(
+    mode: str | None = None,
```
xieofxie: should we rename it to precision? As mode maps to one of the four modes, but precision could be in _COMPOSITE_PRECISIONS etc.

src/winml/modelkit/quant/passes/static.py
```diff
@@ -0,0 +1,209 @@
...
+
+        qdq_start = time.perf_counter()
+
+        weight_type = weight_type_map[self._config.weight_type]
+        activation_type = activation_type_map[self._config.activation_type]
+        calibrate_method = calibration_method_map[self._config.calibration_method]
+
+        extra_options = {
+            "ActivationSymmetric": self._config.symmetric,
+            "WeightSymmetric": self._config.symmetric,
```
xieofxie: **Should fix — regression: per-target symmetric override dropped.**

The previous `_quantize_qdq` resolved weight/activation symmetry independently:

```python
weight_symmetric = config.weight_symmetric if config.weight_symmetric is not None else config.symmetric
activation_symmetric = config.activation_symmetric if config.activation_symmetric is not None else config.symmetric
extra_options = {"ActivationSymmetric": activation_symmetric, "WeightSymmetric": weight_symmetric}
```

This version collapses both to `config.symmetric`, so `config.weight_symmetric` / `config.activation_symmetric` are now silently ignored. That's exactly the w8a16 case (symmetric int8 weights + asymmetric uint16 activations) — this looks like a quiet accuracy regression for the build path. Please restore the independent override (falling back to `config.symmetric` when unset).

src/winml/modelkit/commands/quantize.py
```diff
@@ -275,6 +302,102 @@ def quantize(
...
+        4,
+    )
+
+    config = WinMLQuantizationConfig(
+        rtn_bits=rtn_bits,
+        samples=samples,
+        calibration_method=cast('Literal["minmax", "entropy", "percentile"]', method),
+        weight_type=cast('Literal["uint8", "int8", "uint16", "int16"]', weight_type or "uint8"),
+        activation_type=cast(
+            'Literal["uint8", "int8", "uint16", "int16"]', activation_type or "uint8"
```
xieofxie: **Should fix — multi-pass static ignores the precision bit-width.**

In the multi-pass path the precision token only maps to a *mode* (`_cli_precision_to_mode`), and the static config hardcodes `weight_type or "uint8"` / `activation_type or "uint8"`. So `-p int16 -p fp16` runs the static pass as **uint8**, whereas the single-pass path (`-p int16`) correctly derives int16 via `_resolve_quant_types`. RTN bits are extracted, but static width is silently lost.

Either thread the precision token through `_resolve_quant_types` for static passes here too, or reject the unsupported combinations explicitly.

src/winml/modelkit/quant/quantizer.py
```diff
@@ -7,422 +7,266 @@
...
-        warnings=warnings,
-    )
+    use_external_data: bool = kwargs.pop("use_external_data", True)
+    # Apply model-type-specific quant finalizer if registered. Some model types
+    # finalize calibration reader / nodes-to-exclude / dtypes only once the
+    # exported ONNX exists.
+    if config.model_type and config.calibration_data is None:
+        from .calibration import get_quant_finalizer
+
+        finalizer = get_quant_finalizer(config.model_type)
```
xieofxie, line 266: **Consider — model-type finalizer only runs on the `quantize_onnx` path.**

The calibration finalizer now lives inside `quantize_onnx`, so any caller that builds a pipeline directly via `Quantizer(...).run()` (including the new CLI `_run_multi_precision`) bypasses it. Low impact today since CLI multi-precision has no `model_type`, but it makes the pipeline the wrong seam. Consider moving finalization into `expand_precision`/`Quantizer` so every pipeline consumer gets it.

src/winml/modelkit/commands/quantize.py
```diff
@@ -275,6 +302,102 @@ def quantize(
...
+    """Map a CLI precision string to a quantizer pass mode."""
+    p = precision.lower()
+    if p == "fp16":
+        return "fp16"
+    if is_weight_only_precision(p):
+        return "rtn"
+    return "static"
+
+
+def _run_multi_precision(
```
xieofxie, line 315: **Consider — duplication with the single-pass `quantize()` body.**

`_run_multi_precision` re-implements config construction, default output naming, `guard_output`, `mkdir`, and the console/try-except reporting block that already exist in the single-pass path. Worth factoring out the shared "resolve output → run → report" tail so the two paths can't drift (e.g. the bit-width handling above).

xieofxie (commented):
A few correctness concerns (2 should-fix, 2 to consider) on the pass-pipeline refactor. Overall the design is clean and well-tested — these are mostly about parity with the previous single-dispatch behavior.

src/winml/modelkit/quant/quantizer.py
```diff
@@ -7,422 +7,266 @@
...
-    )
+    _pass_factories: dict[str, BaseQuantPass] = {
+        "fp16": FP16Pass(config),
+        "rtn": RTNPass(config),
+        "static": StaticPass(config),
+        "dynamic": StaticPass(config),
+    }
 
+    if effective_mode in _pass_factories:
+        return [_pass_factories[effective_mode]]
```
xieofxie: **Nit — `_pass_factories` holds instances, not factories (misnomer), and is built eagerly.**

Every call to `expand_precision` constructs all four passes (including `StaticPass` twice — `static` and `dynamic`) just to return one. It's cheap since the constructors only stash the config, but the name and eager construction are slightly misleading. Consider mapping mode → class and instantiating only the selected one, e.g. `_PASS_TYPES = {"fp16": FP16Pass, "rtn": RTNPass, "static": StaticPass, "dynamic": StaticPass}` then `[_PASS_TYPES[effective_mode](config)]`.

src/winml/modelkit/quant/passes/fp16.py
```diff
@@ -0,0 +1,87 @@
...
+    - ``fp16_op_block_list`` — op types that must not be cast to FP16
+
+    Example::
+
+        pass_ = FP16Pass(config)
+        result = pass_.run("model.onnx", "model_fp16.onnx")
+    """
+
+    def __init__(self, config: WinMLQuantizationConfig) -> None:
+        super().__init__(config)
```
xieofxie: **Nit — redundant `__init__`.** This only calls `super().__init__(config)`, which is exactly what `BaseQuantPass.__init__` already does — drop it (same for `RTNPass.__init__`).

xieofxie (commented):
Two minor nits to go with the earlier review.

src/winml/modelkit/quant/quantizer.py
DingmaomaoBJTU: Agreed that moving the finalizer closer to the pipeline would be cleaner. However, the finalizer depends on the ONNX model path already existing (it inspects the exported model), which ties it to the backward-compat wrapper's setup code. Moving it into `expand_precision()` would require passing `model_path` there, changing the signature. I'll keep it in `quantize_onnx()` for now and address in a follow-up refactor. Happy to change if preferred.

src/winml/modelkit/commands/quantize.py
DingmaomaoBJTU: Fair point on the duplication. The two paths diverge enough (single uses `quantize_onnx` with model_type finalizer, multi uses `Quantizer.run` directly) that factoring them fully would require more design. I'll track this as a follow-up cleanup. The current duplication is isolated to `_run_multi_precision` and the report block.

xieofxie (approved):

PR #986, opened 2026-06-26
fix: reconcile task registry into a single source of truth (#724)

### LINKED ISSUE #724: [task] Source-of-truth inconsistency: 13 task names in TASK_ABBREV are rejected by validate_task

timenick:
Fixes #724.

## Problem

`loader/task.py` held two hand-maintained task tables, created by different PRs from different authorities, that silently drifted apart:

- `TASK_ABBREV` (PR #354) — canonical→abbreviation map for cache keys; intentionally broad (full HF taxonomy).
- `KNOWN_TASKS` + `validate_task` (PR #717) — a hand-copied snapshot of optimum's exportable tasks for a fast `winml inspect --list-tasks`, never cross-checked against `TASK_ABBREV`.

PR #717's `validate_task` gate consults only `KNOWN_TASKS`, so the pre-existing divergence became a user-visible bug: 13 task names that lived only in `TASK_ABBREV` were rejected — including ones WinML genuinely supports through a *third* independent table, `inference/tasks.py` (`TaskInputSpec`). Root cause: no single source of truth for the canonical task set.

## Fix

Both public views now derive from one private registry, so they can no longer desync:

- `_TASK_REGISTRY: dict[str, str | None]` — the single authoritative table of canonical task → abbreviation (`None` keeps the existing 8-char truncation fallback).
- `KNOWN_TASKS = frozenset(_TASK_REGISTRY)` and `TASK_ABBREV` (non-`None` abbrevs + 5 collapsed aliases) are both *derived*.

Audit of the 13 divergent names:

- **5 wired → added to the canonical set:** `video-classification`, `keypoint-matching`, `table-question-answering`, `zero-shot-audio-classification`, `any-to-any` — each has a `TaskInputSpec`, a composite-model registration, and/or an e2e testset tag.
- **8 stale → dropped:** `instance-segmentation`, `universal-segmentation`, `masked-image-modeling`, `text-encoding`, `audio-tokenization`, `multimodal-lm`, `backbone`, `time-series-prediction` — unreferenced anywhere outside the old table; `masked-image-modeling` / `time-series-prediction` were long-form duplicates of the existing `masked-im` / `time-series-forecasting`.
- **5 collapsed aliases** (`pretraining`, `sequence-classification`, `summarization`, `translation`, `zero-shot-classification`) stay out of `KNOWN_TASKS` but keep their abbreviations, because some callers use the alias name directly as the resolved task (composite models register `summarization`/`translation`; `inference.tasks` defines a spec for `zero-shot-classification`).

**Zero cache churn:** every abbreviation still in use is byte-for-byte unchanged. Net: `KNOWN_TASKS` 35→40, `TASK_ABBREV` 47→39, and no abbreviation value modified — so no existing cache key changes.

## Drift guards added

`tests/unit/loader/test_known_tasks.py` locks the invariant going forward:

- `KNOWN_TASKS` and `TASK_ABBREV` must derive from `_TASK_REGISTRY`; abbreviations must be unique (the `serve` reverse-decode map stays lossless).
- The 5 wired tasks must pass both `validate_task` and the click `--task` callback.
- The 8 dropped tasks must stay out of every view.
- Every `inference.tasks` `TaskInputSpec` key must be a known task — the cross-source check that would have caught this bug.


src/winml/modelkit/loader/task.py
```diff
@@ -37,96 +54,76 @@
...
+KNOWN_TASKS: frozenset[str] = frozenset(_TASK_REGISTRY)
+
+
+# Task -> abbreviation for cache keys. Derived from `_TASK_REGISTRY`: canonical
+# tasks whose abbreviation is `None` are omitted here (and truncated to 8 chars
+# by `get_task_abbrev`); the collapsed aliases are appended for cache-key and
+# `serve` reverse-decode stability.
+TASK_ABBREV: dict[str, str] = {
+    **{task: abbrev for task, abbrev in _TASK_REGISTRY.items() if abbrev is not None},
+    **_TASK_ALIAS_ABBREV,
```
DingmaomaoBJTU, line 125: Info: Tasks with None abbreviations (masked-im, inpainting, text-to-image, reinforcement-learning, time-series-forecasting) are absent from TASK_ABBREV and use 8-char truncation fallback. The serve/app.py:698 reverse-decode falls back to the raw truncated key for user-cached models of these tasks. Pre-existing behaviour now made explicit. Consider assigning short abbreviations as follow-up.

tests/unit/loader/test_known_tasks.py
```diff
@@ -74,5 +112,93 @@ def test_covers_optimum_tasks(self) -> None:
...
+
+    def test_inference_tasks_are_known_or_alias(self) -> None:
+        from winml.modelkit.inference.tasks import TASK_REGISTRY as INFERENCE_TASKS
+
+        recognized = KNOWN_TASKS | set(_TASK_ALIAS_ABBREV) | self.INFERENCE_ONLY_ALIASES
+        missing = set(INFERENCE_TASKS) - recognized
+        assert not missing, (
+            f"inference.tasks defines tasks unknown to the loader registry: {sorted(missing)}. "
+            "If these are real canonical tasks, add them to _TASK_REGISTRY in "
+            "src/winml/modelkit/loader/task.py (do not extend INFERENCE_ONLY_ALIASES)."
```
DingmaomaoBJTU, line 203: Info: The assertion message only suggests adding to _TASK_REGISTRY. For pipeline-sugar aliases (like sentiment-analysis, ner), the correct fix is INFERENCE_ONLY_ALIASES. Suggest extending the message to mention this path.

DingmaomaoBJTU (approved):
Review: Approve - clean reconciliation of task registry into single source of truth.

PR #987, opened 2026-06-26
fix(build): surface clear disk-full error instead of misleading opset ValueError

### LINKED ISSUE #259: bug: winml build shows misleading ValueError (opset domain) when disk is full during quantize step
## Summary

When disk space is exhausted during `winml build`, the optimize step writes a corrupted/truncated ONNX file. The subsequent quantize step then fails with `ValueError: Failed to find proper ai.onnx domain`, giving no indication that the real cause is an out-of-disk-space condition. Users will chase a phantom code bug.

## Repro

```bash
# Fill disk to near-zero free space, then:
winml build -c config.json -m microsoft/resnet-50 -o model_a_config/
```

## Error

```
ValueError: Failed to find proper ai.onnx domain in
onnxruntime.quantization.quant_utils.get_opset_version
```

Stack trace: `quantizer.py:166 → quantize.py:909 → quantize.py:702 → quant_utils.py:984 → quant_utils.py:977 → ValueError`

## Root Cause

The optimize step writes a truncated/zero-byte ONNX output due to `OSError` (disk full). The corrupted file is passed to the quantizer, which fails trying to parse the opset domain. The `OSError` is swallowed or not surfaced to the user.

## Expected behavior

ModelKit should either:
1. Check available disk space before `winml build` starts and warn if below a threshold, or
2. Catch `OSError` during ONNX file writes and surface a clear error: \"Insufficient disk space — unable to write output file.\"

## Notes

- Severity: P2
- Reporter: Agency bug bash feedback
- Context: Full bug bash consumes significant disk — venv (~1.4 GB), HuggingFace cache (~14.4 GB), per-model artifacts (~200–400 MB each). Minimum ~20 GB free disk recommended.

timenick:
Closes #259

## Problem

When the disk fills during the **optimize** step of `winml build`, the ONNX write fails: `onnx.save_model` truncates the target to 0 bytes and then `write()` raises `OSError(ENOSPC)`, leaving a corrupt/zero-byte `optimized.onnx` behind. The **quantize** step then loads that file — `onnx.load` parses the empty bytes into an empty `ModelProto` with no `opset_import`, and ORT raises the opaque:

```
ValueError: Failed to find proper ai.onnx domain
```

which surfaces as `Quantization failed: ...`. The real cause (out of disk space) is never reported, so users chase a phantom code bug.

## Fix

Root-cause fix at the ONNX **write boundary**, plus a defensive guard:

- **`ONNXSaveError`** (`onnx/persistence.py`) — new `OSError` subclass with `path` and `disk_full` attributes. Exported from `onnx/__init__.py`.
- **`save_onnx`** — both write paths (external-data and inline) now catch `OSError`, remove the partial `.onnx`/`.data` artifact, and raise `ONNXSaveError`. Disk-full (`errno.ENOSPC` / Windows `ERROR_DISK_FULL` 112) gets a clear "Insufficient disk space" message; other errors get a generic write-failure message.
- **`copy_onnx_model`** (`onnx/external_data.py`) — same treatment for the copy boundary, cleaning up `dst` (+ sidecar).
- **Quantizer guard** (`quant/quantizer.py`) — `_quantize_single_pass` now detects an empty/opset-less/unparseable input model up front and returns a clear `QuantizeResult` failure instead of letting ORT raise its opaque opset error.
- **Build hint** (`commands/build.py`) — disk-space errors map to an actionable hint (free up disk / clear `~/.cache/winml`).
- **Docs** (`docs/troubleshooting.md`) — note on the new behavior and automatic partial-file cleanup.

Subclassing `OSError` keeps existing `except OSError` callers working and lets the build command's top-level handler surface the clear message verbatim. The write-boundary cleanup guarantees no truncated artifact is left for a later stage to misread.

Out of scope (decided during triage): no proactive free-disk pre-check before build.


src/winml/modelkit/onnx/persistence.py
```diff
@@ -25,6 +27,75 @@
...
+    """
+
+    def __init__(
+        self,
+        message: str,
+        *,
+        path: str | Path | None = None,
+        disk_full: bool = False,
+    ) -> None:
+        super().__init__(message)
```
DingmaomaoBJTU, line 68: **Bug: `ONNXSaveError.errno` is always `None`.**

`super().__init__(message)` with a single string sets `self.errno = None`. The class docstring (line 39) explicitly says *"errno is preserved"*, but that is false — any caller catching this as an `OSError` and checking `e.errno == errno.ENOSPC` will always see `None`.

Fix: preserve errno at the call site in `_raise_save_error`:
```python
exc = ONNXSaveError(message, path=path, disk_full=disk_full)
exc.errno = error.errno   # preserve so callers can inspect e.errno
raise exc from error
```
Or accept `errno_code` in `__init__` and call `super().__init__(errno_code, message)`. The docstring should also be updated to match whichever approach is chosen.

src/winml/modelkit/quant/quantizer.py
```diff
@@ -22,6 +22,41 @@
...
+    because it ran out of disk space). Detecting it here lets us surface the
+    real cause instead of ORT's opaque "Failed to find proper ai.onnx domain".
+
+    Reads only the graph (no external weights) directly via ``onnx.load_model``
+    so the check stays cheap and never trips over a missing ``.data`` sidecar.
+    """
+    from onnx import load_model
+
+    try:
+        model = load_model(str(model_path), load_external_data=False)
```
DingmaomaoBJTU, line 254: **Warning: full proto parse on every quantize call.**

`load_model(..., load_external_data=False)` still deserialises the entire `.onnx` proto from disk on every call, even for healthy models. For large LLMs with multi-MB proto headers this is measurable overhead with no benefit in the success path.

Consider a cheap fast-path before the full parse:
```python
if model_path.stat().st_size == 0:
    return (
        f"Input ONNX model is empty (zero bytes): {model_path}. "
        "A previous build stage may have run out of disk space. "
        "Free up disk space and rebuild."
    )
```
This handles the most common disk-full artefact (zero-byte file) without any I/O beyond a `stat` call, and the full proto parse becomes a fallback for the rarer truncated-but-nonzero case.

DingmaomaoBJTU (commented):
The disk-full fix is well-structured and the test coverage is thorough. Two issues to address before merging:

1. **Bug** — `ONNXSaveError.errno` is always `None`, contradicting the docstring that says "errno is preserved". Any caller that checks `e.errno == errno.ENOSPC` will silently get the wrong answer.

2. **Warning** — `_check_input_model_opset` fully parses the ONNX proto on every single quantize call, even for perfectly healthy models. For large models with multi-MB proto headers, this is unnecessary latency.

src/winml/modelkit/onnx/persistence.py
timenick: Good catch — fixed in `3b632749`. `ONNXSaveError.__init__` now takes an `errno_code` and sets `self.errno` explicitly (kept the single-arg `super().__init__(message)` so `str(e)` stays the clean message instead of gaining an `[Errno N]` prefix), and `_raise_save_error` threads `error.errno` through. Added `err.errno == errno.ENOSPC` / `errno.EACCES` assertions to the disk-full tests so it stays preserved, and updated the docstring to match.

Generated with Claude Code

src/winml/modelkit/quant/quantizer.py
timenick: Done in `3b632749` — added a zero-byte `stat()` fast-path before the full parse, so a healthy model's success path no longer deserialises the whole proto (and a zero-byte file now gets a more specific "empty (zero bytes)" message). The full `onnx.load_model(..., load_external_data=False)` parse remains the fallback for the rarer truncated-but-nonzero case.

Generated with Claude Code

DingmaomaoBJTU (approved):
Both issues from the previous review are resolved.

**Bug fix confirmed** (persistence.py): ONNXSaveError now accepts errno_code and sets self.errno = errno_code explicitly after super().__init__(message). _raise_save_error passes errno_code=error.errno, and the new tests explicitly assert err.errno == errno.ENOSPC / err.errno == errno.EACCES.

**Warning addressed** (quantizer.py): _check_input_model_opset now has a stat().st_size == 0 fast-path as suggested. The most common disk-full artefact (zero-byte file) is caught cheaply without a proto parse; non-zero truncated files fall through to load_model.

Minor note (non-blocking): the docstring at lines 232-233 says "the healthy success path never pays for a full proto parse" which is slightly inaccurate — a non-zero healthy model still does load_model. Consider rewording to "a zero-byte file is caught before the full parse".

PR #990, opened 2026-06-26
refactor(cli): unify -m/--model option across all commands

### LINKED ISSUE #563: [cross-cutting] [P2] `--model` semantic drift across 10 commands

xieofxie:
@
## Summary

Unifies the `-m/--model` option across every command in `src/winml/modelkit/commands/` by routing them all through the shared `model_option` / `model_path_option` helpers in `utils/cli.py`, instead of ad-hoc inline `click.option` definitions. The flag spelling, ordering, type, and validation now have a single source of truth.

Closes #563.

## Changes

**Helpers (`utils/cli.py`)**
- `model_path_option` (validated ONNX `Path`) — added `multiple` and `help_text` parameters.
- `model_option` (any model reference, no validation) — added `multiple` and `help_text` parameters.

**Commands converted from inline `click.option` to the shared helpers**

| Command | Now uses |
|---|---|
| `compile` | `model_path_option(required=False, multiple=True, help_text=…)` |
| `optimize` | `model_path_option(required=False, help_text=…)` |
| `quantize` | `model_path_option(required=True, help_text=…)` |
| `export` | `model_option(required=True, help_text=…)` |
| `eval` | `model_option(required=False, multiple=True, help_text=…)` |
| `build` | `model_option(required=False, help_text=…)` |
| `inspect` | `model_option(required=False, help_text=…)` |

(`analyze`, `run`, `serve`, `perf`, `config` already used the helpers.)

**Destination standardized to `model`** — `build` and `inspect` previously bound `--model` to `model_id`; the CLI binding is now `model`. Downstream keyword args that are genuinely the domain concept (e.g. `WinMLQuantizationConfig.model_id`) are unchanged.

This is purely structural — each command keeps its existing accept-types (`PATH`-validated vs flexible `TEXT`), `required` semantics, `multiple`, and help wording.
@


src/winml/modelkit/utils/cli.py
```diff
@@ -92,47 +92,77 @@ def warn_ignored_calibration_options(
...
     Returns:
-        Decorator function
+        Decorator function.
     """
-    help = "Model: HF model ID, build output directory, or .onnx file path"
+    help = help_text or "Model: HF model ID, build output directory, or .onnx file path"
     if optional_message:
         help = f"{help}. {optional_message}"
+    # ``multiple`` options default to an empty tuple; single-valued ones to None.
+    kwargs: dict[str, Any] = {"multiple": True} if multiple else {"default": None}
```
DingmaomaoBJTU, line 160: **Nit: missing `type=str` for the `multiple=True` path.**

The original `eval.py` decorator explicitly set `type=str`:
```python
@click.option("-m", "--model", type=str, multiple=True, default=(), ...)
```
The new helper omits `type` entirely when `multiple=True`. Click defaults to `STRING` (== `str`) so behaviour is identical, but the intent is less obvious. Consider:
```python
kwargs: dict[str, Any] = {"multiple": True, "type": str} if multiple else {"default": None}
```

src/winml/modelkit/utils/cli.py
```diff
@@ -92,47 +92,77 @@ def warn_ignored_calibration_options(
...
+        multiple: Accept the flag repeatably; the value becomes a tuple
+            (default: False).
+        help_text: Override for the help string (default: a generic
+            ONNX-file description).
 
     Returns:
-        Decorator function
+        Decorator function.
     """
     return click.option(
```
DingmaomaoBJTU, line 119: **Nit: flag-ordering change in --help output.**

Both helpers emit `--model, -m` (long first). Commands like `eval.py`, `build.py`, and `inspect.py` previously showed `-m, --model` (short first) in `--help`. Not a bug, but it is a user-visible cosmetic change not mentioned in the PR description.

DingmaomaoBJTU (commented):
This PR cleanly consolidates the -m/--model option across 7 command files into two shared helpers. All 410 relevant unit/CLI/regression tests pass.

Overall: Refactoring is behaviorally equivalent and well-documented. One nit on missing type=str for the multiple=True path (see inline).

src/winml/modelkit/utils/cli.py
xieofxie: no need

src/winml/modelkit/utils/cli.py
xieofxie: Create an issue to track https://github.com/microsoft/winml-cli/issues/1001

DingmaomaoBJTU (approved):

PR #992, opened 2026-06-29
Implement no-color option

### LINKED ISSUE #567: [cross-cutting] [P2] No `--no-color` flag (NO_COLOR works only via Rich's tty auto-detection)
**Severity**: P2  
**Command**: `cross-cutting`

---
**Category**: Color & Theming — color must be controllable both per-invocation (flag) and per-environment (env var).

**Repro**:
```
uv run winml --help        | Select-String 'no-color'
uv run winml export --help | Select-String 'no-color'
$env:NO_COLOR='1'; uv run winml sys --format compact > out.txt 2>&1; Remove-Item Env:NO_COLOR
Get-Content out.txt -Raw | Select-String -Pattern '\x1b\['
```

**Actual**:
- No `--no-color` flag exists in `winml --help` or any subcommand.
- `NO_COLOR=1` does suppress ANSI in piped output (verified — 0 escape sequences in capture). This works because Rich auto-detects no-tty / NO_COLOR.
- But there is no per-invocation flag, so a user running interactively in a terminal cannot disable color without setting an env var.

**Expected**: Add a top-level `--no-color` flag that sets the Rich `Console(no_color=True, force_terminal=False)` for the rest of the run. Document `NO_COLOR=1` and `CI=true` in the help epilog. Do not regress the current Rich auto-detection.

**Why it matters**: Some terminals (older Windows consoles, log viewers, screen-reader sessions) render the color codes as garbage. A per-invocation flag is the standard escape hatch.

zhenchaoni:
Fix #567 

## Add `--no-color` flag for color-free output

`--no-color` is a common accessibility option in CLI tools such as `git`. It lets users disable ANSI color codes for a single invocation so output renders plainly — useful for older consoles, log viewers, screen readers, and piped output.

This PR adds `--no-color` support to winml CLI. The flag is available on the root `winml` command and on every subcommand, and it disables colored output for the current process only (the next invocation is colored again). `NO_COLOR=1` and `CI=true` environment variables continue to disable color as before.

**Usage:**
```
winml sys --no-color
```

### DIFF
--- a/docs/commands/overview.md
+++ b/docs/commands/overview.md
@@ -63,6 +63,10 @@ differ per command (e.g., `-p` is a short form for `--precision` only on
 `config` and `quantize`); check the **Flags** section of each command page
 rather than assuming they transfer.
 
+`--no-color` is accepted by every command and disables colored output for
+that invocation. Color is also auto-disabled when `NO_COLOR=1` or `CI=true`
+is set in the environment.
+
 ## See also
 
 - [How winml-cli Works](../concepts/how-it-works.md) — end-to-end pipeline overview
--- a/src/winml/modelkit/cli.py
+++ b/src/winml/modelkit/cli.py
@@ -34,7 +34,7 @@ if TYPE_CHECKING:
 from . import __version__
 from .telemetry import ActionGroup
 from .telemetry import telemetry as _telemetry_mod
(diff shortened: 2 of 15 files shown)


xieofxie (approved):

xieofxie:
cool

PR #996, opened 2026-06-29
feat(qwen3): genai bundle generation

DingmaomaoBJTU:
## Summary

Adds `onnxruntime-genai` **bundle generation** for winml-exported Qwen3 transformer-only models. A single `scripts/qwen3.py export` command builds all four bundle components (transformer `ctx`/`iter`, `embeddings`, `lm_head`) and assembles them into an onnxruntime-genai directory (`genai_config.json` + HF tokenizer files). The bundle-assembly machinery is split into an **EP-agnostic core** and a **Qwen3/QNN-specific layer**.

Inference over the assembled bundle is handled by the `winml-genai` runtime (`winml perf --runtime winml-genai`, already on `main`); this PR is generation-only. A dedicated genai inference session lands in a follow-up.

### What's new

| File | Change |
|---|---|
| `src/winml/modelkit/utils/genai.py` | **New** generic, EP-agnostic core: `PipelineStage`, `DecoderIOMapping`, `build_genai_config()`, `build_decoder_pipeline_stages()`, `write_genai_bundle()` |
| `src/winml/modelkit/models/hf/qwen3/genai.py` | **New** Qwen3 + QNN layer: `build_qwen3_transformer_only_stages()`, `write_genai_bundle()` wrapper, `qnn_stage_session_options()` (QNN HTP routing) |
| `scripts/qwen3.py` | **New** unified export CLI (`export` subcommand) |
| `scripts/export_qwen3_transformer_only.py`, `scripts/export_qwen3_embeddings_lm_head.py` | **Deleted** — superseded by `scripts/qwen3.py` |
| `src/winml/modelkit/onnx/utils.py` | Adds `strip_node_attrs()` (drops exporter-injected default GQA attributes) |
| `src/winml/modelkit/models/hf/qwen3/qwen3_modeling.py` | `max_rope_len` export hook (prepared; defaults to current behavior) |
| `tests/unit/models/qwen3/test_genai_config.py`, `test_qwen3_modeling.py`, `tests/unit/onnx/test_utils.py` | New unit tests |

### Design

**`utils/genai.py` is architecture-agnostic.** `build_genai_config` takes a `list[PipelineStage]` and a `DecoderIOMapping` — no tensor names or EP details are hardcoded. `build_decoder_pipeline_stages` introspects the built `ctx.onnx`/`iter.onnx` (`_introspect_onnx_io` + `_detect_format_patterns`) to discover `past_keys_%d` / `present_values_%d`-style patterns from the actual graph I/O, so tensor names can never drift from what the ONNX really contains.

**`models/hf/qwen3/genai.py` is the Qwen3/QNN-specific layer.** `build_qwen3_transformer_only_stages` wires the four Qwen3 stages, and `qnn_stage_session_options` emits the QNN HTP `session_options` for the transformer stages — all QNN specifics live here, not in the generic core. Its `write_genai_bundle` wraps the generic assembler so the one-shot API is unchanged.

```
build_genai_config(hf_config, ..., pipeline=stages, decoder_io=decoder_io)   <- generic (utils/genai.py)
build_qwen3_transformer_only_stages(ctx_onnx, iter_onnx, num_layers)          <- Qwen3/QNN (models/hf/qwen3/genai.py)
```

### Usage

```bash
# Build all four components + assemble the genai bundle (transformer on NPU/QNN):
uv run python scripts/qwen3.py export --model-id Qwen/Qwen3-0.6B --device npu --output out/bundle

# Reuse pre-built companion ONNX files instead of rebuilding them:
uv run python scripts/qwen3.py export --model-id Qwen/Qwen3-0.6B \
  --embeddings path/to/embeddings.onnx --lm-head path/to/lm_head.onnx --output out/bundle

# Benchmark / run the assembled bundle (runtime already on main):
uv run winml perf -m out/bundle --runtime winml-genai
```

### Notes
- `embeddings.onnx` (fp32) and `lm_head.onnx` (w4a32 `MatMulNBits`) are **built by the script** on CPU; `--embeddings` / `--lm-head` override them with pre-built files.
- Transformer `ctx`/`iter` stages route to the QNN HTP (NPU) via per-stage `session_options` in `genai_config.json`; embeddings/lm_head stay on CPU.
- Verified end-to-end: `winml perf -m out/bundle --runtime winml-genai` installs the WinML QNN EP, runs the transformer stages on the NPU, and generates correctly.

Follows from #836.


src/winml/modelkit/session/genai_session.py
```diff
@@ -0,0 +1,699 @@
...
+
+    from ..winml import add_ep_for_device
+    from .ep_registry import WinMLEPRegistry
+
+    registry = WinMLEPRegistry.get_instance()
+    registry.register_execution_providers()
+    so = ort.SessionOptions()
+    so.add_session_config_entry("ep.context_enable", "1")
+    so.add_session_config_entry("ep.context_file_path", dst)
+    add_ep_for_device(so, "QNNExecutionProvider", ort.OrtHardwareDeviceType.NPU, qnn_options)
```
DingmaomaoBJTU: Support dynamic EP input here, not only for QNN

src/winml/modelkit/session/genai_session.py
```diff
@@ -0,0 +1,699 @@
...
+    mc = ort.ModelCompiler(so, src, embed_compiled_data_into_model=False)
```
DingmaomaoBJTU: Could it leverage the origin compile logic in session.py?

src/winml/modelkit/utils/genai.py
```diff
@@ -0,0 +1,724 @@
...
+from typing import Any
+
+
+logger = logging.getLogger(__name__)
+
+# Default filenames inside the bundle directory.
+DEFAULT_EMBEDDINGS_FILENAME = "embeddings.onnx"
+DEFAULT_CONTEXT_FILENAME = "ctx.onnx"
+DEFAULT_ITERATOR_FILENAME = "iter.onnx"
+DEFAULT_LM_HEAD_FILENAME = "lm_head.onnx"
```
DingmaomaoBJTU, line 88: why do we need this

scripts/qwen3.py
```diff
@@ -0,0 +1,379 @@
...
+
+def _cmd_export(args: argparse.Namespace) -> int:
+    """Build all components and write the genai bundle."""
+    from winml.modelkit.models.hf.qwen3.genai import write_genai_bundle
+
+    t0 = time.monotonic()
+
+    # --- Transformer (ctx + iter) ---
+    print(f"\n=== building transformer stages (device={args.device}) ===")
+    transformer = WinMLQwen3TransformerOnlyModel.from_pretrained(
```
xieofxie: also use `WinMLAutoModel.from_pretrained(model_type=?)` ?

scripts/qwen3.py
```diff
@@ -0,0 +1,379 @@
...
+# --------------------------------------------------------------------------
+r"""Unified Qwen3 genai pipeline: full model export and inference.
+
+Sub-commands
+------------
+export
+    Build (or reuse) all four components of the Qwen3 genai bundle and
+    assemble them into an onnxruntime-genai directory:
+
+      - ``ctx.onnx``        — transformer prefill graph (QNN-quantized)
```
xieofxie: QDQ-quantized ?

scripts/qwen3.py
DingmaomaoBJTU: Good idea, but `WinMLAutoModel.from_pretrained` can't select the transformer-only variant here. With `task="text-generation"` it takes the composite path, which delegates to `WinMLCompositeModel.from_pretrained` **without forwarding `model_type`**. The base then re-derives `model_type` from the HF config (`"qwen3"`) and resolves `("qwen3", "text-generation")` → the full `WinMLQwen3Model`, silently ignoring the `qwen3_transformer_only` override — the exact fallback that `WinMLQwen3TransformerOnlyModel.from_pretrained` guards against.

(The companions already go through `WinMLAutoModel` because `feature-extraction` isn't a composite task, so they take the single-model build path that *does* honor `model_type`.)

Unifying this would require the composite dispatch to honor an explicit `model_type` override (forward it from `WinMLAutoModel` and prefer it over `hf_config.model_type` in the base lookup), which touches all composites. I'll handle that in the follow-up PR and keep the direct `WinMLQwen3TransformerOnlyModel.from_pretrained(...)` call here for now.

src/winml/modelkit/session/genai_session.py
DingmaomaoBJTU: This file (`genai_session.py`) has been removed from this PR — genai inference/session support is deferred to a dedicated follow-up PR, so this comment no longer applies here. Resolving.

src/winml/modelkit/session/genai_session.py
DingmaomaoBJTU: `genai_session.py` has been removed from this PR — the genai inference/session layer (including this compile worker) is deferred to a dedicated follow-up PR, where the dynamic / multi-EP support noted here will be handled. Resolving as it no longer applies to this PR.

src/winml/modelkit/utils/genai.py
DingmaomaoBJTU: These `DEFAULT_*_FILENAME` constants are the canonical onnxruntime-genai bundle filenames (`ctx.onnx` / `iter.onnx` / `embeddings.onnx` / `lm_head.onnx`). They're the default values for the `context_filename` / `iterator_filename` / `embeddings_filename` / `lm_head_filename` parameters of `write_genai_bundle()` and drive the on-disk layout that `og.Config()` expects — so filenames aren't hardcoded at each call site while staying overridable. Resolving; happy to inline if preferred.

scripts/qwen3.py
DingmaomaoBJTU: Good catch — the ctx/iter transformer stages are QDQ-format quantized (not QNN-specific), and the docstring now reads `QDQ-quantized`. Resolving.

src/winml/modelkit/utils/genai.py
```diff
@@ -0,0 +1,686 @@
...
+
+
+def _introspect_onnx_io(onnx_path: Path) -> tuple[list[str], list[str]]:
+    """Return ``(input_names, output_names)`` from an ONNX model graph header.
+
+    External data is intentionally not loaded — only the graph topology is read,
+    so this is fast even for large quantized models.
+    """
+    try:
+        import onnx
```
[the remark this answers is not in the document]
DingmaomaoBJTU: False positive from the local `winml.modelkit.onnx` subpackage sharing its short name with the pip `onnx` package. In this module the runtime `import onnx` calls (here and at lines 511/584) are deliberate lazy/optional imports used for `onnx.load` / `onnx.save`, while `from ..onnx import copy_onnx_model` pulls from the local package — CodeQL resolves both as the same `onnx` module and flags `import-and-import-from`. This same rule is already open on non-PR files (e.g. `external_data.py`, alerts 593/594), so this is consistent existing behavior, not something introduced here. Resolving as a known false positive.

src/winml/modelkit/models/hf/qwen3/genai.py
```diff
@@ -0,0 +1,220 @@
...
+    Returns:
+        Dict suitable for the ``session_options`` key of a pipeline stage in
+        ``genai_config.json``.
+    """
+    return {
+        "log_id": log_id,
+        "provider_options": [
+            {
+                "qnn": {
+                    "backend_path": "QnnHtp.dll",
```
xieofxie, line 69: this does not work for winml ?

src/winml/modelkit/onnx/utils.py
```diff
@@ -13,6 +13,41 @@
...
 
+def strip_node_attrs(
+    model: onnx.ModelProto,
+    op_type: str,
+    keep_attrs: frozenset[str] | set[str],
+    domain: str = "",
+) -> onnx.ModelProto:
+    """Remove all attributes from matching nodes except those in *keep_attrs*.
+
+    Useful for stripping default-valued optional attributes that an exporter
```
xieofxie, line 24: when will this happen?

xieofxie:
The PR description is stale and should be refreshed before merge. The "What's new" table references files that aren't in the actual diff: `scripts/export_qwen3_transformer_only.py` is actually **deleted** (not modified), `scripts/infer_genai.py` is **not in this PR**, and the `--genai-bundle` / `--embeddings` / `--lm-head` flags on the export script don't exist. The real change instead adds `scripts/qwen3.py` and `src/winml/modelkit/utils/genai.py`. The Usage and Notes sections also describe an inference script that isn't part of this PR. A reviewer relying on the description will be misled — please update it to match the actual file set.

xieofxie:
`main()` in `scripts/qwen3.py` ignores the parsed subcommand — it always hard-dispatches to `_cmd_export(args)` regardless of `args.command`. It works today with a single `export` subcommand, but the `add_subparsers` scaffolding implies more commands are coming, and the next one added would silently run export instead. Consider wiring dispatch via `p.set_defaults(func=_cmd_export)` (and `return args.func(args)`) so each subcommand routes to its own handler.

src/winml/modelkit/models/hf/qwen3/genai.py
DingmaomaoBJTU: It works for winml — I just re-verified end-to-end with `winml perf -m out/bundle --runtime winml-genai` (no `--device`, so `mixed`):

- WinML auto-installs and registers the QNN EP straight from this bundle config: `[WinML] Installing Execution Provider: QNNExecutionProvider … EP installed successfully` (package `MicrosoftCorporationII.WinML.Qualcomm.QNN.EP`, `onnxruntime_providers_qnn.dll`).
- The `context`/`iterator` stages then compile to and run on the QNN HTP NPU — the log shows the QNN HTP graph-compiler stages (VTCM Allocation, Parallelization Optimization, DDR bandwidth summary).
- Generation succeeds: 128 tokens, decode ~30.7 tok/s, TTFT ~213 ms (P50).

The `winml-genai` runtime (`GenaiSession`) reads these per-stage `provider_options` and forwards `backend_path`/`htp_performance_mode`/`soc_model` verbatim to the QNN EP for both EPContext compilation and `og.Model` inference. `QnnHtp.dll` is the standard QNN-HTP backend name; WinML places it on the EP search path when it installs the QNN EP package, so the bare filename resolves.

src/winml/modelkit/onnx/utils.py
DingmaomaoBJTU: Concretely, it happens when exporting the Qwen3 transformer to ONNX. The exporter emits `com.microsoft::GroupQueryAttention` nodes with every schema-optional attribute filled in at its default value (`k_quant_type`, `local_window_size`, `qk_output`, `smooth_softmax`, `v_quant_type`) — none of which are needed at inference. `scripts/qwen3.py::_strip_gqa_default_attrs` (lines 60-70) calls this to drop them, keeping only the three Qwen3 actually uses (`do_rotary`, `kv_num_heads`, `num_heads`), so the emitted GQA nodes match the expected minimal attribute set. The util stays generic (op-type + keep-set are parameters); the GQA-specific keep-set lives in the Qwen3 script.

DingmaomaoBJTU:
> `main()` in `scripts/qwen3.py` ignores the parsed subcommand — it always hard-dispatches to `_cmd_export(args)` regardless of `args.command`. It works today with a single `export` subcommand, but the `add_subparsers` scaffolding implies more commands are coming, and the next one added would silently run export instead. Consider wiring dispatch via `p.set_defaults(func=_cmd_export)` (and `return args.func(args)`) so each subcommand routes to its own handler.

Sure~ Fixed. Btw in next pr I will remove this script~

DingmaomaoBJTU:
Refreshed the PR title + description to match the actual diff (thanks for the catch @xieofxie). Dropped the stale bits — `scripts/export_qwen3_transformer_only.py` is deleted, and there's no `scripts/infer_genai.py` / `--genai-bundle` flag / inference script in this PR. The **What's new** table, **Usage**, and **Notes** now reflect the real files: `scripts/qwen3.py` (the unified `export` CLI), the generic `src/winml/modelkit/utils/genai.py` core, and the Qwen3/QNN `src/winml/modelkit/models/hf/qwen3/genai.py` layer. Also noted that inference is handled by the `winml-genai` runtime on `main`, not a script in this PR.

The `main()` dispatch fix is pushed in 272126e7 (`p.set_defaults(func=_cmd_export)` + `return args.func(args)`), so each subcommand routes to its own handler and a missing subcommand errors out.

xieofxie (approved):

PR #1004, opened 2026-06-30
feat(perf): cache per-module builds and honor --rebuild/--ignore-cache

### LINKED ISSUE #918: feat: perf --module should also use cache folder instead of temp dir

xieofxie:
## Summary

`winml perf --module` built every submodule into a throwaway `tempfile.TemporaryDirectory` with no `cache_key` and no `rebuild` flag, so per-module artifacts were rebuilt on every run and `--rebuild` / `--ignore-cache` were silently ignored. The single-model path already honored these flags via `WinMLAutoModel` → `auto.py`.

This mirrors `auto.py`'s cache logic in `_perf_modules`:

- **Default**: each submodule builds into the model's persistent cache dir (`~/.cache/winml/artifacts/<model_slug>/`) and is reused on the next run.
- **`--rebuild`**: reuses the cache dir but overwrites artifacts.
- **`--ignore-cache`**: falls back to a temp dir + forced rebuild (prior behavior, now opt-in).

## Why per-module caching is collision-free

Each submodule config carries a unique `loader.module_path` (plus its traced I/O shapes). `cfg.generate_cache_key()` hashes the full config, so every instance gets a distinct key — even shape-identical siblings — and `build_hf_model`'s `cache_key` filename prefixing keeps each instance's `model.onnx` separate within one shared model dir.

## Behavior change worth noting

Per-module artifacts now **persist by default** (more disk use), which is the intended consequence of matching the single-model / `auto.py` behavior.

## Tests

Added `TestPerfModuleCache` (default → persistent dir + `rebuild=False`; `--rebuild` → `rebuild=True` in cache dir; `--ignore-cache` → temp dir + `rebuild=True`). Full `tests/unit/commands/test_perf_module.py`: 15 passed.

Closes #918

### DIFF
--- a/src/winml/modelkit/cache/path.py
+++ b/src/winml/modelkit/cache/path.py
@@ -12,8 +12,19 @@ Dependencies: stdlib only (os, pathlib). ZERO internal imports.
 
 from __future__ import annotations
 
+import hashlib
+import json
 import os
 from pathlib import Path
+from typing import Any
+
+
+# Build-control kwargs that change the produced artifact but are NOT part of
+# WinMLBuildConfig (they are forwarded to build_hf_model / build_onnx_model as
+# kwargs). Folding them into the cache key prevents --no-optimize / --no-analyze
+# / --max-optim-iterations from silently reusing an artifact built with
+# different settings. Keep in sync with utils.cli.build_pipeline_extra_kwargs.
+_CACHE_KEY_BUILD_CONTROLS = ("skip_optimize", "hack_max_optim_iterations")
 
 
 __all__ = [
@@ -63,7 +74,11 @@ def get_artifacts_dir(cache_dir: Path | None = None) -> Path:
     return cache_dir / "artifacts"
 
 
-def get_cache_key(task_abbrev: str, config_hash: str) -> str:
+def get_cache_key(
+    task_abbrev: str,
+    config_hash: str,
+    build_controls: dict[str, Any] | None = None,
+) -> str:
     """Assemble the cache key prefix from task abbreviation and config hash.
 
     Args:
@@ -71,11 +86,26 @@ def get_cache_key(task_abbrev: str, config_hash: str) -> str:
             Obtained from ``modelkit.loader.task.get_task_abbrev()``.
         config_hash: Deterministic config hash (e.g., ``"a1b2c3d4e5f67890"``).
             Obtained from ``WinMLBuildConfig.generate_cache_key()``.
+        build_controls: Optional mapping of build-pipeline kwargs (as produced
+            by ``utils.cli.build_pipeline_extra_kwargs``). Recognized
+            artifact-affecting keys (``skip_optimize``,
+            ``hack_max_optim_iterations``) are folded into the key so that
+            ``--no-optimize`` / ``--no-analyze`` / ``--max-optim-iterations``
+            do not silently reuse an artifact built with different settings.
+            Unrecognized keys are ignored, and a default (empty) mapping leaves
+            the key unchanged for backward compatibility.
 
     Returns:
-        Cache key string: ``"{task_abbrev}_{config_hash}"``
+        Cache key string: ``"{task_abbrev}_{config_hash}"``, optionally suffixed
+        with ``"_{controls_hash}"`` when non-default build controls are present.
     """
-    return f"{task_abbrev}_{config_hash}"
+    key = f"{task_abbrev}_{config_hash}"
+    if build_controls:
+        overrides = {k: build_controls[k] for k in _CACHE_KEY_BUILD_CONTROLS if k in build_controls}
+        if overrides:
+            blob = json.dumps(overrides, sort_keys=True)
+            key = f"{key}_{hashlib.sha256(blob.encode()).hexdigest()[:8]}"
+    return key
 
 
 def get_artifact_path(
--- a/src/winml/modelkit/commands/perf.py
+++ b/src/winml/modelkit/commands/perf.py
@@ -831,6 +831,8 @@ def _perf_modules(
     ep_options: dict[str, str] | None = None,
     precision: str = "auto",
     allow_unsupported_nodes: bool = False,
+    rebuild: bool = False,
+    ignore_cache: bool = False,
 ) -> None:
     """Run per-module build and benchmark for matching submodules.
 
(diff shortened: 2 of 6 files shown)


src/winml/modelkit/commands/perf.py
```diff
@@ -960,12 +982,26 @@ def _perf_modules(
         if no_compile:
             cfg.compile = None
 
-        with tempfile.TemporaryDirectory(ignore_cleanup_errors=True) as tmpdir:
+        # Compute the cache key AFTER the quant/compile mutations above so it
+        # reflects what is actually built.
+        cache_key = get_cache_key(task_abbrev, cfg.generate_cache_key())
```
DingmaomaoBJTU: Good that `no_quantize`/`no_compile` mutate `cfg` before this line, so they're folded into `cache_key`. Worth calling out the asymmetry: `--no-optimize`/`--no-analyze`/`--max-optim-iterations` are forwarded as `build_hf_model` kwargs (`skip_optimize`, `hack_max_optim_iterations`) and aren't part of `cfg`, so they don't affect `cache_key`.

Now that per-module caching is on by default, `perf --module X` followed by `perf --module X --no-optimize` will hit the cached optimized `model.onnx` and silently ignore `--no-optimize` unless `--rebuild` is passed. This matches the single-model/`auto.py` limitation, so it's consistent rather than a regression vs. that path — but it's a new footgun for `--module`, which previously rebuilt in a temp dir every run and thus always honored these flags.

Consider folding the optimize/analyze toggles into the cache key, or documenting that `--rebuild` is needed when changing them.

tests/unit/commands/test_perf_module.py
```diff
@@ -543,3 +543,113 @@ def test_no_quantize_clears_only_quant(self, tmp_path: Path) -> None:
...
+        persistent build dir is deterministic. The benchmark is short-circuited
+        via a failing ``session.perf()`` — build_hf_model is already called by
+        then, so its kwargs are captured.
+        """
+        cache_root = tmp_path / "cache"
+
+        fake_cfg = MagicMock()
+        fake_cfg.loader.model_type = "bert"
+        fake_cfg.loader.module_path = "encoder.layer.0"
+        fake_cfg.generate_cache_key.return_value = "deadbeefdeadbeef"
```
DingmaomaoBJTU, line 571: `generate_cache_key` is mocked to a constant here, so the PR's central claim — distinct key per sibling instance → collision-free within the shared model dir — isn't directly exercised by these tests (they verify the rebuild/`output_dir`/`cache_key` plumbing, which is the important part). Optional: add a case with two configs whose `loader.module_path` differ and assert `build_hf_model` is called with distinct `cache_key`s, to guard the collision-free property at this layer too.

DingmaomaoBJTU (commented):
Reviewed the per-module cache changes. The logic cleanly mirrors `auto.py`/the single-model path (`use_cache = not ignore_cache`, `force_rebuild = rebuild or ignore_cache`, persistent model dir + per-module `cache_key`, temp-dir fallback on `--ignore-cache`), and `build_hf_model` creates `output_dir` and scopes its rebuild-glob to `{cache_key}_*.onnx`, so sibling modules coexist and `--rebuild` won't wipe each other's artifacts. Ran `tests/unit/commands/test_perf_module.py` locally: 15 passed. Two notes inline — one behavioral nuance and one optional test-coverage suggestion; neither is blocking.

src/winml/modelkit/commands/perf.py
xieofxie: Good catch. This was actually a latent bug in the single-model/`auto.py` path too (no existing issue filed), not just `--module` — both derive the key from `config.generate_cache_key()` and forward `skip_optimize`/`hack_max_optim_iterations` as `build_hf_model` kwargs that never entered the key.

Fixed universally: `get_cache_key()` now folds recognized build-control overrides (`skip_optimize`, `hack_max_optim_iterations`) into the key. Default builds keep their existing key (backward compatible); only `--no-optimize` / `--no-analyze` / `--max-optim-iterations` runs get a distinct suffix. Applied at every call site: `auto.py` (HF + ONNX), the perf module path, and `build` single-model. Added tests in `test_path.py` and `test_perf_module.py`.

tests/unit/commands/test_perf_module.py
xieofxie: Added `test_sibling_instances_get_distinct_cache_keys` in `TestPerfModuleCache`: two configs with distinct `generate_cache_key()` values (as real siblings with different `loader.module_path` have) now assert `build_hf_model` is called with distinct `cache_key`s, guarding the collision-free property at this layer.

src/winml/modelkit/cache/path.py
```diff
@@ -12,8 +12,19 @@
...
 from pathlib import Path
+from typing import Any
+
+
+# Build-control kwargs that change the produced artifact but are NOT part of
+# WinMLBuildConfig (they are forwarded to build_hf_model / build_onnx_model as
+# kwargs). Folding them into the cache key prevents --no-optimize / --no-analyze
+# / --max-optim-iterations from silently reusing an artifact built with
+# different settings. Keep in sync with utils.cli.build_pipeline_extra_kwargs.
+_CACHE_KEY_BUILD_CONTROLS = ("skip_optimize", "hack_max_optim_iterations")
```
DingmaomaoBJTU, line 27: Two minor, non-blocking thoughts on this allowlist:

1. **Coupling.** This tuple must stay in sync with `build_pipeline_extra_kwargs`'s output keys (the comment already flags it 👍). If a future artifact-affecting flag is added there but not here, the silent-reuse bug quietly returns for that flag. Since perf/build already pass exactly `build_pipeline_extra_kwargs(...)`, a shared constant — or having auto.py filter its full `kwargs` down to that same set so every caller passes an already-scoped dict — would remove this as a second source of truth.
2. **`--max-optim-iterations 3` vs default.** `build_pipeline_extra_kwargs` emits `hack_max_optim_iterations=3` when the flag is passed explicitly but omits it by default, yet `build_hf_model`'s own default is also 3. So `perf` and `perf --max-optim-iterations 3` build identically but hash to different keys → a redundant rebuild + extra cached artifact. Harmless (never wrong reuse, always the safe direction), just a minor cache miss; normalizing out keys equal to the pipeline default before hashing would collapse them.

DingmaomaoBJTU (commented):
Re-reviewed the new commit (`53a8f26`) — it directly resolves both points from my previous review:

1. **Optimize/analyze toggles now affect the cache key.** `get_cache_key` folds `skip_optimize` / `hack_max_optim_iterations` in, applied at all four call sites (auto.py HF + ONNX, perf module, build single-model). I checked key/build consistency: the same controls folded into the key are exactly what reach `build_hf_model` / `build_onnx_model`, so a later `--no-optimize` run no longer silently reuses an optimized artifact. Nice that path.py's stdlib-only / zero-internal-imports contract is preserved (hashlib/json/typing are all stdlib).
2. **Collision-free claim now has a real test.** `test_sibling_instances_get_distinct_cache_keys` uses two distinct config hashes (no longer mocked to a constant) and asserts distinct `cache_key`s reach `build_hf_model`; `test_no_optimize_changes_cache_key` guards the flag→key coupling.

Verified locally: applied the PR head and ran `tests/unit/cache/test_path.py` + `tests/unit/commands/test_perf_module.py` → **33 passed**. One optional, non-blocking note inline. LGTM otherwise.

DingmaomaoBJTU (approved):
Approving. The follow-up commit cleanly resolves both points from my earlier reviews: optimize/analyze toggles now fold into the cache key at all four call sites (verified key/build consistency), and the collision-free claim has real test coverage. Verified locally — tests/unit/cache/test_path.py + tests/unit/commands/test_perf_module.py: 33 passed. The two inline notes are minor/optional and not blocking.

src/winml/modelkit/cache/path.py
xieofxie: Create an issue to track https://github.com/microsoft/winml-cli/issues/1016

PR #1005, opened 2026-06-30
fix: migrate EPAlias to EPNameOrAlias for better alignment; fix infer_ihv_from_ep_name

### LINKED ISSUE #646: migrate EPAlias to EPName where possible
We should use EPName

xieofxie:
## Summary

Close #646 (migrate `EPAlias` to `EPName` where possible). `EPNameOrAlias` is the public-facing input type; `EPName` is the internal canonical form. This PR removes places where a bare alias was used where a canonical name belongs, and stops comparing against single alias literals.

### Changes

- **`analyze/utils/ep_utils.py`** — `infer_ihv_from_ep_name` now does a direct, exact `dict[EPName, IHVType]` lookup covering every member of the `EPName` Literal. An unknown name now **raises `ValueError`** instead of silently defaulting to `IHVType.MICROSOFT` (the callers that matter already handle it: `InformationEngine` catches `ValueError` and falls back to loading all rules; the model validator catches it defensively). Also fixes a latent miss where `CUDAExecutionProvider` resolved to `MICROSOFT` instead of `NVIDIA`.
- **`serve/schema.py`** — `EpSwitchRequest.ep` was the only `ep` field still typed bare `EPAlias`; every sibling is `EPNameOrAlias` and `manager.switch_ep` already accepts `EPNameOrAlias`. Widened it for consistency so it accepts both `"qnn"` and `"QNNExecutionProvider"`.
- **`compiler/utils.py`** — `needs_format_conversion` compared `ep == "qnn"`, a single alias literal. An EP can have multiple aliases (e.g. `nv_tensorrt_rtx` / `nvtensorrtrtx`), and the canonical name itself wouldn't match. Now normalizes via `normalize_ep_name` and compares against the canonical `"QNNExecutionProvider"`; parameter widened to `EPNameOrAlias`.

### Notes

- The compiler's `EPConfig.provider` deliberately remains alias-typed: every value it holds is an alias (`provider="qnn"`, etc.), so retyping to `EPName` would require a value migration with serialization back-compat — out of scope here.
- An audit of all ~90 `EPNameOrAlias` call sites confirmed every one either normalizes before a canonical sink or forwards to a callee that does — no other un-normalized canonical-sink usages were found.

### Tests

- Replaced the per-EP `infer_ihv` unit tests with one that verifies every `EP_NAMES` member resolves to an `IHVType` (enforcing map/Literal parity) plus an unknown-raises test.
- Updated fixtures that relied on the old lenient behavior to use canonical EP names (`ACEExecutionProvider` → `VitisAIExecutionProvider`; alias `"openvino"` → `"OpenVINOExecutionProvider"` in the validator tests).
- Added `test_qlinear_for_qnn_canonical_name` covering the canonical-name case that previously failed in `needs_format_conversion`.

### DIFF
--- a/src/winml/modelkit/analyze/utils/ep_utils.py
+++ b/src/winml/modelkit/analyze/utils/ep_utils.py
@@ -21,18 +21,18 @@ logger = logging.getLogger(__name__)
 
 
 def infer_ihv_from_ep_name(ep_name: EPName) -> IHVType:
-    """Infer IHVType from Execution Provider name.
+    """Infer IHVType from a canonical Execution Provider name.
 
-    Maps an execution provider name to its corresponding IHV type.
-    Supports multiple name variations for each provider.
-    Unknown EPs (e.g., CPUExecutionProvider, DmlExecutionProvider) resolve
-    to IHVType.MICROSOFT.
+    ``EPName`` is a closed set of canonical EP names, so this is a direct,
+    exact lookup. Any name without an explicit IHV owner (e.g.
+    ``CPUExecutionProvider``, ``DmlExecutionProvider``) resolves to
+    ``IHVType.MICROSOFT``.
 
     Args:
-        ep_name: Execution Provider name (e.g., QNNExecutionProvider, OpenVINOExecutionProvider)
+        ep_name: Canonical Execution Provider name (see ``utils.constants.EPName``).
 
     Returns:
-        IHVType: Inferred IHV type (QC, INTEL, AMD, NVIDIA, or MICROSOFT)
+        IHVType: Inferred IHV type (QC, INTEL, AMD, NVIDIA, or MICROSOFT).
 
     Examples:
         >>> infer_ihv_from_ep_name("QNNExecutionProvider")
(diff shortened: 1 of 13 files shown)


DingmaomaoBJTU (approved):

DingmaomaoBJTU (commented):
Reviewed the EPAlias -> EPNameOrAlias migration and the `infer_ihv_from_ep_name` rework. Overall this is a clean, well-motivated change — the map/Literal-parity test is a nice touch, and catching `CUDAExecutionProvider` -> `NVIDIA` (was silently `MICROSOFT`) is a good latent-bug fix.

One concern about the new `raise` on unknown EPs and one un-audited caller (`output_aggregator.build_ep_support`) — left inline. Not necessarily blocking, but worth resolving for consistency with the other two call sites.

src/winml/modelkit/analyze/utils/ep_utils.py
```diff
@@ -46,32 +48,25 @@ def infer_ihv_from_ep_name(ep_name: EPName) -> IHVType:
...
+        "CUDAExecutionProvider": IHVType.NVIDIA,
+        "CPUExecutionProvider": IHVType.MICROSOFT,
+        "DmlExecutionProvider": IHVType.MICROSOFT,
+    }
+
+    canonical = normalize_ep_name(ep_name)
+    try:
+        return ep_name_to_ihv[canonical]  # type: ignore[index]
+    except KeyError:
+        raise ValueError(f"Cannot infer IHV for unknown EP name: {ep_name!r}") from None
```
DingmaomaoBJTU: This new `raise` has a third caller the PR description's audit doesn't cover. `output_aggregator.build_ep_support()` calls `infer_ihv_from_ep_name(ep_type)` completely unguarded (`output_aggregator.py:173`), whereas the two callers listed in the description both swallow the error — `InformationEngine.__init__` catches `ValueError`, and `BatchedConstMatMulValidator._is_enabled` catches `Exception`.

Reachability: `analyze_from_proto()` / `analyze()` run `ep` through `normalize_ep_name`, which returns unknown/misspelled names **unchanged** (`constants.py:149`), and `eps_to_analyze = [ep_normalized]` flows straight into the aggregator. So a programmatic call like `analyzer.analyze_from_proto(model, ep="FooExecutionProvider")` now propagates an unhandled `ValueError` out of aggregation, where before it degraded to `IHVType.MICROSOFT`. The CLI path is safe (`click.Choice`), but these public methods aren't.

Since the map covers every `EPName`, this only bites on genuinely-unknown input — but the behavior is now inconsistent across the three call sites. Suggest either guarding the aggregator call the same way, or validating the EP once upstream in `analyze_from_proto` so unknown names fail fast with a clear message instead of deep in aggregation.

xieofxie: use IHVType.UNKNOWN instead of raise

DingmaomaoBJTU: Confirmed resolved by the `IHVType.UNKNOWN` approach — nice, this is cleaner than either option I suggested. Making inference total removes the inconsistency across all three call sites at once: `output_aggregator.build_ep_support` (the unguarded one) now just gets `UNKNOWN` and only uses it for a debug log, and `InformationEngine` maps `UNKNOWN -> None` to load all rules. The now-redundant `try/except` in `BatchedConstMatMulValidator._is_enabled` is correctly dropped too. Also spotted you fixed the docstring's `INTEL` value (`'INTEL'` -> `'Intel'`) along the way.

DingmaomaoBJTU (commented):
Re-reviewed the two new commits (`use IHVType.UNKNOWN instead of raise`, `resolve mypy union-attr and assignment narrowing`).

**My earlier concern is fully resolved.** Switching `infer_ihv_from_ep_name` to return `IHVType.UNKNOWN` instead of raising makes inference total, so all three call sites — including the previously-unguarded `output_aggregator.build_ep_support` — are now safe. Verified end to end: `UNKNOWN` only flows into a debug log there; `InformationEngine` maps `UNKNOWN -> None` (loads all rules); the redundant `try/except` in `BatchedConstMatMulValidator` is correctly removed since `UNKNOWN != INTEL`. Tests updated accordingly and CI is green (lint/mypy + all test shards + CodeQL).

Two minor, non-blocking notes:

1. **`rule_loader.load_runtime_rules(None)`** iterates `for ihv in IHVType`, which now includes `UNKNOWN`, producing one extra `No rule files found for Unknown ... skipping` warning on the load-all path. It's harmless and exactly mirrors the existing behavior for `MICROSOFT` (also absent from `prefix_map`), so no change needed — just flagging that adding an enum member has this ripple effect.

2. **`models/auto.py`** is unrelated to the EP migration — it's a legit mypy narrowing fix (`_model_type: str | None`) bundled into the mypy commit, and it also defers `AutoConfig.from_pretrained` until there's no explicit `model_type` override (a small behavioral improvement). Fine to keep; noting it only because it's outside the PR's stated scope.

LGTM from my side.

DingmaomaoBJTU (approved):
Approving. My earlier concern about the unguarded `output_aggregator.build_ep_support` caller is fully resolved by making `infer_ihv_from_ep_name` total (returns `IHVType.UNKNOWN` instead of raising), which cleanly removes the inconsistency across all three call sites. Verified end to end, and CI is green (lint/mypy + all test shards + CodeQL). The two remaining items I noted (the harmless `Unknown` warning in `rule_loader`'s load-all path, and the unrelated `auto.py` mypy fix) are non-blocking.

PR #1006, opened 2026-06-30
refactor(optracing): per-sample QNN CSV parsing with warmup skip and clean output

xieofxie:
## Summary

Reworks the QNN basic-mode op-tracing CSV pipeline so metadata is tracked per inference sample, durations are computed correctly, and the JSON output is clean.

- **Per-sample parsing.** `parse_qnn_profiling_csv` now returns a list of `{"metadata": {...}, "samples": [...]}` — one entry per inference, each carrying its *own* ROOT metadata (HVX threads, accelerator execute cycles/US) instead of a single first-occurrence snapshot. Samples are delimited by the `"Number of HVX threads used"` marker that begins each inference.
- **Correct duration aggregation.** Operator metric construction moved into the profiler's `_csv_operator_metrics` helper, which computes each operator's `duration_us`/`percent_of_total` against the metadata of the *same* sample (the cycle→US factor differs slightly per inference) before averaging across samples.
- **Warmup handling.** `warmup` is threaded through `run → _collect_results → _from_csv`. The CSV records every execute call (warmup included), so `_from_csv` drops the first `warmup` samples, asserts the remainder equals `iterations`, and builds metrics only from the measured samples.
- **No null fields.** `OperatorMetrics.to_dict()` now omits unset (`None`) fields, so basic-mode traces no longer serialize the many detail-only fields (DMA/VTCM/roofline) as `null`.

## Tests

- `test_basic_pipeline_csv_to_json` now validates the full CSV→JSON output against a code-generated golden fixture (`basic_pipeline_expected.json`) with a pinned timestamp.
- Added per-sample metadata tests, warmup-skip and sample-count-mismatch tests; updated callers and assertions for the new shapes.
- `uv run pytest tests/unit/optracing/` — 98 passed.


DingmaomaoBJTU (changes requested):
One real bug to fix before merge, plus two warnings and an info note. See inline comments for details.

src/winml/modelkit/optracing/qnn/profiler.py
```diff
@@ -303,27 +361,25 @@ def _from_csv(
...
-                name=op["name"],
-                op_path=op["name"],
-                op_id=op["op_id"],
-                duration_us=op["cycles"] * cycle_to_us,
-                percent_of_total=(op["cycles"] / total_cycles * 100 if total_cycles > 0 else 0),
-            )
-            for op in parsed["operators"]
-        ]
+        measured = samples[warmup:]
+        assert len(measured) == iterations, (
```
DingmaomaoBJTU: **Bug: `assert` used for production runtime validation**

`assert` is disabled when Python runs with the `-O` flag. Replace with a proper `ValueError`.

src/winml/modelkit/optracing/qnn/profiler.py
```diff
@@ -58,6 +58,64 @@ def _resolve_shape(shape: list, default_dim: int = 1) -> list[int]:
...
+
+def _csv_summary(samples: list[dict[str, Any]]) -> dict[str, Any]:
+    """Headline metadata across samples (HVX threads constant; cycles/US averaged)."""
+    if not samples:
+        return {"hvx_threads": 0, "accel_execute_cycles": 0, "accel_execute_us": 0}
+
+    n = len(samples)
+    metas = [s["metadata"] for s in samples]
+    return {
+        "hvx_threads": metas[0]["hvx_threads"],
```
DingmaomaoBJTU: **Warning: `hvx_threads` silently taken from sample[0] only**

If HVX thread count ever varies across samples, the discrepancy is silently dropped. Log a warning when values differ.

src/winml/modelkit/optracing/result.py
```diff
@@ -48,8 +48,13 @@ class OperatorMetrics:
...
     def to_dict(self) -> dict[str, Any]:
-        """Serialize to dict, preserving None for unavailable fields."""
-        return asdict(self)
+        """Serialize to dict, omitting fields left unset (``None``).
+
+        Basic-mode traces only populate identity + timing, so dropping ``None``
+        keeps the output free of the many detail-only fields (DMA/VTCM/roofline)
+        that would otherwise serialize as ``null``.
+        """
+        return {k: v for k, v in asdict(self).items() if v is not None}
```
DingmaomaoBJTU, line 57: **Warning: Breaking serialization change**

`to_dict()` now omits `None` fields entirely instead of serializing as `null`. Downstream callers accessing missing keys will get `KeyError`. A migration note is needed.

tests/unit/optracing/fixtures/basic_pipeline_expected.json
```diff
@@ -0,0 +1,573 @@
...
+    "accel_execute_cycles": 3455548,
+    "accel_execute_us": 1041
+  },
+  "operators": [
+    {
+      "name": "/resnet/embedder/embedder/convolution/Conv",
+      "op_path": "/resnet/embedder/embedder/convolution/Conv",
+      "op_id": 24,
+      "duration_us": 197.87246478452283,
+      "percent_of_total": 19.10347271565341
```
DingmaomaoBJTU, line 22: **Info: High-precision floats make the golden fixture brittle**

Consider rounding to a fixed precision or using `pytest.approx` comparisons.

src/winml/modelkit/optracing/result.py
xieofxie: no downstream users yet

tests/unit/optracing/fixtures/basic_pipeline_expected.json
xieofxie: it works as json comparision, so I think it is fine?

src/winml/modelkit/optracing/qnn/profiler.py
xieofxie: Fixed: replaced the ssert with a ValueError so the check holds under python -O.

src/winml/modelkit/optracing/qnn/profiler.py
xieofxie: Fixed: now logs a warning listing the per-sample HVX thread counts when they differ before falling back to the first sample's value.

DingmaomaoBJTU (approved):
Both blocking issues addressed: `assert` replaced with `ValueError`, and HVX thread variance now logs a warning. The two remaining items (serialization migration note and golden fixture float precision) are non-blocking. LGTM.

PR #1009, opened 2026-07-01
feat(e2e_eval): unify example-tests into recipe-driven run_eval

KayMKM:
## Summary

Unifies the two e2e evaluation scripts (`run_eval.py` and the example-only
`run_example_tests.py`) into a single recipe-driven runner: build a model, run
`winml perf`, and — when perf passes — run `winml eval`, writing one unified
`eval_result.json` per (model, task, precision). Also adds two `--clean-cache`
disk-hygiene fixes so long matrix runs don't fill the disk.

## Recipe-driven runner

- Add `utils/recipes.py`: data-driven discovery of `examples/recipes/<slug>`
  configs (fp16/w8a16/w8a8), single and composite (encoder/decoder,
  image/text-encoder).
- `run_eval.py` expands the registry into one job per recipe precision variant;
  builds via `winml build -c`, falling back to `winml config` when a model has
  no recipe. Perf-gated accuracy reuses the existing both-mode gating.
- Accuracy phase now only runs `winml eval` and records the full metrics +
  dataset; delta/verdict grading vs the PyTorch baseline is left to the report
  site. Inline baseline removed; offline `--update-baseline` maintains the
  baseline cache.
- Remove runner-side report generation (`generate_report.py`,
  `analyze_results.py`); slim `reporter.py`/`accuracy.py` (keep `compute_delta`
  + thresholds for the site). Drop the in-pipeline `generate_report` step.

## Disk hygiene under `--clean-cache`

- **Per-job build artifacts:** remove a job's `export/optimized/model.onnx`
  (+ `.data`/`.bin`) once its `eval_result.json` is written, keeping only the
  JSON the report site reads — peak disk stays at ~one job.
- **Stray ORT/QNN scratch models:** sweep temporary models ORT/QNN leak into the
  working directory (UUID-named `<uuid>.onnx`/`.data`, QNN EP-context dumps
  `<uuid>.onnx_<EP>.bin`, and `sym_shape_infer_temp.onnx`). Matches only
  UUID-prefixed or the fixed temp name so real files are never touched; runs at
  startup and after each job.

## Notes

- Rebased onto latest `main`; the prerequisite build-only EP-matrix work landed
  separately in #845, so this branch carries only the 3 commits for this change.


DingmaomaoBJTU (approved):
Well-structured refactor: recipe-driven eval with clean separation of concerns. All 97 tests pass. The runner now records facts only (metrics + dataset maps) and defers baseline grading to the report site. Recipe discovery is data-driven — drop a new precision config and it's picked up automatically. LGTM.

scripts/e2e_eval/run_eval.py
```diff
@@ -2264,43 +2745,62 @@ def main() -> None:
...
             onnx_paths = build_result["onnx_paths"] if build_result["success"] else {}
             onnx_size = _compute_onnx_size(onnx_paths)
 
-            if not build_result["success"]:
+            if backfill_existing is not None:
+                # Accuracy backfill: perf was already recorded (and passed), so
+                # skip run_model and only (re)build + run accuracy. A build that
+                # now fails records a skipped accuracy without touching perf.
+                if not build_result["success"]:
+                    accuracy_result = {"skipped": True, "skip_reason": "build_failed"}
```
DingmaomaoBJTU, line 2851: When a backfilled (re)build fails here, accuracy is recorded as `{skipped: build_failed}` while the reused perf stays `passed`, so the job still exits 0. And because `_should_skip_existing` skips a passed-perf result whose accuracy is `skipped`, a later `--continue` / `--retry-failed` won't re-attempt it — the only recovery is deleting `eval_result.json`. This mirrors the accuracy-only path at ~L2813, so it may be intentional, but for a resume-oriented feature a transient build failure being permanently masked (and not counted as a failure) is surprising. Consider leaving `accuracy` as `None` on build failure so the next `--continue` backfills it again.

scripts/e2e_eval/run_eval.py
```diff
@@ -1581,39 +1826,60 @@ def _run_winml_eval(
...
-        num_samples = ds_config.get("num_samples", _DEFAULT_SAMPLES)
-        metric = _parse_metric_from_winml_output(output_path, winml_key, num_samples)
-    status = "PASS" if (proc["exit_code"] == 0 and metric is not None) else "FAIL"
+        metrics, dataset = _read_eval_output(output_path)
+        # Best-effort single value for log lines; the site reads the full
+        # metrics map by its own key, so a miss here is non-fatal.
+        winml_key = ds_config.get("winml_metric_key") or ds_config.get("metric")
+        if winml_key and metrics and winml_key in metrics:
+            num_samples = (dataset or {}).get("samples", ds_config.get("num_samples"))
+            metric = {"metric": winml_key, "value": float(metrics[winml_key]), "num_samples": num_samples}
```
DingmaomaoBJTU, line 1913: Minor: this branch isn't `ruff format`-clean. `ruff format` rewraps this line (106 chars) into a multi-line dict, and also touches the `_build_jobs` signature, the BACKFILL `safe_print`, and two spots in `tests/unit/eval/test_run_eval_script.py`. CI's lint job only runs `ruff check` (which passes), so it won't block — but the repo's pre-commit includes `ruff-format`, so worth running `uvx ruff format scripts/e2e_eval/run_eval.py tests/unit/eval/test_run_eval_script.py` before merge.

DingmaomaoBJTU (commented):
Nice consolidation of the two eval scripts. The recipe-driven job expansion, perf-gated accuracy, and the `--clean-cache` disk-hygiene sweeps read cleanly and are well covered (103 new tests pass locally; `ruff check` is clean). I verified `_read_eval_output` matches `EvalResult.to_dict()`'s top-level `metrics`/`dataset` keys, and that the removed `generate_report.py`/`analyze_results.py` plus the trimmed reporter/accuracy symbols have no remaining references anywhere. A couple of minor, non-blocking notes inline.

zhenchaoni (commented):
**1. Build artifacts go to the wrong place.**
Recipe builds run `winml build -c … -o eval_results/<date>/models/<slug>`. Since `-o` and `--use-cache` are mutually exclusive, artifacts **never** reach `~/.cache/winml/artifacts`. The fallback path still uses `--use-cache`, so within one run, recipe jobs and fallback jobs store artifacts in two different locations. This contradicts your goal of keeping everything in the winml cache.

**2. Build failure silently mislabels precision.**
When a recipe build fails, `onnx_paths` is empty, so perf/eval fall back to `-m <hf_id>` and rebuild at the *default* precision — then record the result under the recipe's precision slug (e.g. a failed `fp16` build gets tested and stored as `fp16` but actually ran `w8a16`). A failed recipe build should be a hard failure, not a silent fallback.

KayMKM:
> **1. Build artifacts go to the wrong place.** Recipe builds run `winml build -c … -o eval_results/<date>/models/<slug>`. Since `-o` and `--use-cache` are mutually exclusive, artifacts **never** reach `~/.cache/winml/artifacts`. The fallback path still uses `--use-cache`, so within one run, recipe jobs and fallback jobs store artifacts in two different locations. This contradicts your goal of keeping everything in the winml cache.
> 
> **2. Build failure silently mislabels precision.** When a recipe build fails, `onnx_paths` is empty, so perf/eval fall back to `-m <hf_id>` and rebuild at the _default_ precision — then record the result under the recipe's precision slug (e.g. a failed `fp16` build gets tested and stored as `fp16` but actually ran `w8a16`). A failed recipe build should be a hard failure, not a silent fallback.

**Re: 1 — build artifacts go to two different places**

Good catch, agreed — fixed. Recipe builds now use `--use-cache` instead of
`-o <job_dir>`, matching the `winml config` fallback, so every job's build
artifacts land in `~/.cache/winml` regardless of recipe-vs-fallback:

- `_run_recipe_build` swaps `-o <job_dir>` → `--use-cache` and locates the built
  ONNX via `_extract_onnx_path` (the same output-parsing + cache lookup the
  fallback already uses) instead of `_find_built_onnx`.
- Dropped the now-redundant `_clean_job_artifacts` — it only existed to sweep the
  ONNX that `-o` wrote into the job dir; cache artifacts are already bounded by
  `_clear_disk_caches` under `--clean-cache`.
- Updated the `--clean-cache` help text / log line and the affected unit tests.

**Re: 2 — build failure silently rebuilds at default precision and mislabels it**

The fallback you're describing (empty `onnx_paths` → `winml perf`/`eval -m <hf_id>`
rebuilding at the default precision) does exist inside `run_model` /
`_run_winml_eval`, but it's unreachable from the runner, so the mislabel can't
actually happen:

- The build always runs first. When it fails we take the
  `elif not build_result["success"]` branch, which synthesizes a failed result
  (perf records the failed build proc; accuracy is `skipped` / `build_failed`)
  and never calls `run_model` / `_run_winml_eval`.
- Those functions are only reached on the success branches, and `success` is
  `len(onnx_paths) == len(components)` with `components >= 1`, so a successful
  build always yields a non-empty `onnx_paths`.
- Net: they only ever receive a non-empty `onnx_paths`, so the `if not onnx_paths`
  fallback is dead code — a failed build is already a hard failure, not a silent
  default-precision rebuild.

zhenchaoni (approved):

zhenchaoni (approved):

zhenchaoni (approved):

PR #1015, opened 2026-07-01
feat(session): add GenaiSession for onnxruntime-genai pipeline inference

### LINKED ISSUE #1011: feat: add genai inference to perf command
We could do it separately and verify by downloading existing genai models or convert from model builder, FTK etc.

Related https://github.com/microsoft/winml-cli/pull/996

DingmaomaoBJTU:
Closes https://github.com/microsoft/winml-cli/issues/1011

Introduces `GenaiSession`, a generic wrapper around `onnxruntime-genai` (`og.Model` + `og.Generator`) for autoregressive text generation using multi-model decoder pipelines (ctx + iter + embeddings + lm_head), plus a `winml-genai` perf runtime so `winml perf` can benchmark these bundles.

## Usage

Benchmark a genai bundle **directory** (one containing `genai_config.json`, as produced by a winml-cli export) with `winml perf --runtime winml-genai`:

```bash
# Mixed execution (default): ctx/iter route to the accelerator via each stage's
# session_options, embeddings/lm_head stay on CPU.
winml perf -m path/to/bundle --runtime winml-genai

# Pre-compile EPContext-capable stages (e.g. QNN) before benchmarking:
winml perf -m path/to/bundle --runtime winml-genai --compile \
  --prompt "Explain the theory of relativity in simple terms." \
  --max-new-tokens 128 --iterations 10 --warmup 2
```

image

genai-specific flags:

| Flag | Default | Description |
| --- | --- | --- |
| `--runtime winml-genai` | `winml` | Benchmark an onnxruntime-genai bundle (LLM generation) instead of single-shot ONNX inference. |
| `--device` | `auto` | `auto`→mixed, `npu`→qnn, `gpu`→dml, `cpu`→CPU-only. |
| `--compile` / `--no-compile` | off | Pre-compile EPContext-capable stages before the run (JIT otherwise). |
| `--compile-timeout` | `300` | Max seconds to compile each EPContext stage before falling back to the original ONNX. |
| `--prompt` | built-in | Prompt to generate from (wrapped in the bundle's own chat template before timing). |
| `--max-new-tokens` | `128` | New tokens generated per iteration. |
| `--iterations` / `--warmup` | `10` / `2` | Timed / warmup generations (genai-specific low defaults; a full generation is far costlier than one `session.run()`). |
| `-o, --output` | auto | JSON report path (`~/.cache/winml/perf/<bundle>/<ts>.json` otherwise). |

Reported metrics come straight from `onnxruntime-genai` call boundaries — TTFT/prefill measured around `append_tokens`, decode tokens/sec and per-output-token latency (TPOT) around `generate_next_token`.

The benchmark prompt is wrapped in the bundle's own chat template (via `og.Tokenizer.apply_chat_template`) before timing, so prefill reflects realistic chat usage; bundles that ship no template benchmark the raw prompt.

Programmatic use:

```python
from winml.modelkit.session import GenaiSession, GenerationConfig

with GenaiSession("path/to/bundle", ep="mixed") as sess:
    for tok in sess.generate_streaming("Hello, world!", GenerationConfig(max_new_tokens=64)):
        print(tok, end="", flush=True)
```

## Changes

**`src/winml/modelkit/session/genai_session.py`** (new)
- `GenaiSession` — context manager wrapping `og.Model` + `og.Generator`
- `generate_streaming()` / `generate()` — streaming and full-string generation; `generate_timed()` returns a `GenerationTiming` with TTFT/TPOT sourced from ort-genai call boundaries
- EPContext pre-compilation via `_prepare_compiled_bundle()`; each pipeline stage's EP is resolved **generically** from its `genai_config.json` `provider_options` via `WinMLCompileConfig.for_provider()` (QNN, OpenVINO, VitisAI, …) — no provider name is hardcoded — and compiled in an isolated subprocess that delegates to the shared compiler component (`compile_onnx`) with a configurable `compile_timeout` (default 300 s). Stages on non-EPContext EPs (CPU, DML) stay on JIT. A `.meta.json` marker invalidates a cached stage when its provider options change or the source weights (`.onnx` / `.data` sidecar) are newer.
- WinML EP auto-registration for QNN/DML/mixed bundles; static KV cache length read from `genai_config.json`
- `apply_chat_template()` renders the prompt through the bundle's own chat template (model-driven via `og.Tokenizer`; no chat format hardcoded)
- `_mirror_non_onnx_files()` symlinks non-compiled assets into the `_compiled/` subdir
- `GenerationConfig` / `GenerationTiming` dataclasses and `GenaiSessionError` / `GenaiNotInstalledError` / `GenaiLoadError` hierarchy

**`src/winml/modelkit/commands/_perf_genai.py`** (new) — the `winml-genai` perf runtime: `GenaiPerfConfig`, the generation benchmark loop, LLM metric aggregation (TTFT percentiles, decode tokens/sec, avg output-token latency), and rich/JSON reporting.

**`src/winml/modelkit/commands/perf.py`** — adds `--runtime {winml,winml-genai}`, `--prompt`, `--max-new-tokens`, and `--compile-timeout`. `winml-genai` takes a bundle **directory** and defaults `--device auto` to mixed execution so each stage's per-stage EP routing (ctx/iter on the accelerator, embeddings/lm_head on CPU) is honored. winml-only flags are warned-and-ignored; genai imports are function-local so `perf --help` stays fast.

**`src/winml/modelkit/session/__init__.py`** — exports `GenaiSession`, `GenerationConfig`, `GenerationTiming`, and the exception classes.

**`pyproject.toml`** — adds `onnxruntime-genai-winml==0.14.1`.

**Tests** — `tests/unit/session/test_genai_session.py` and `tests/unit/commands/test_perf_genai.py` (mock `onnxruntime_genai`, no hardware needed): init/load lifecycle, EP registration, streaming, chat-template application, generic per-EP compile dispatch, EPContext cache invalidation, `_prepare_compiled_bundle`, `_mirror_non_onnx_files`, `_patch_stage_filename`, `compile_timeout` propagation, timing math with an injected clock, and full CLI dispatch/validation.


timenick (approved):

src/winml/modelkit/session/genai_session.py
```diff
@@ -0,0 +1,703 @@
...
+                yield stream.decode(new_token)
+                n += 1
+                if n >= cfg.max_new_tokens:
+                    break
+        finally:
+            # Explicit deletion releases the KV cache buffer held by the generator.
+            # This runs whether the caller exhausts the iterator normally, breaks
+            # out early, or the generator is garbage-collected — preventing the NPU
+            # memory from being held until a later GC cycle.
+            del generator
```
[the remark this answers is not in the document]
DingmaomaoBJTU: Fixed in 20ead267 — removed the redundant del generator and its now-purposeless try/finally in generate_streaming (the local is released when the generator frame is torn down). Same cleanup applied to the identical block in generate_timed.

tests/unit/commands/test_perf_genai.py
```diff
@@ -0,0 +1,483 @@
...
+        run_genai_perf(cfg, console=Console(stderr=True), json_mode=True)
+
+        out = capsys.readouterr().out
+        assert json.loads(out)["benchmark_info"]["runtime_type"] == "winml-genai"
+
+    def test_not_installed_becomes_click_error(self, monkeypatch) -> None:
+        import click
+
+        class _Boom:
+            def __init__(self, _cfg: object) -> None: ...
```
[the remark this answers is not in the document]
DingmaomaoBJTU: Fixed in 20ead267 — replaced the no-op ... body with pass in the test stub's __init__.

src/winml/modelkit/session/genai_session.py
```diff
@@ -0,0 +1,948 @@
...
+                yield stream.decode(new_token)
+                n += 1
+                if n >= cfg.max_new_tokens:
+                    break
+        finally:
+            # Explicit deletion releases the KV cache buffer held by the generator.
+            # This runs whether the caller exhausts the iterator normally, breaks
+            # out early, or the generator is garbage-collected — preventing the
+            # device memory from being held until a later GC cycle.
+            del generator
```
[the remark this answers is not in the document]
DingmaomaoBJTU: Fixed in 20ead267 — dropped the unnecessary del generator / try/finally here too; relying on normal frame teardown to release the generator.

src/winml/modelkit/commands/perf.py
```diff
@@ -1490,8 +1490,133 @@ def _run_simple_loop(
...
+        compile_timeout=p["compile_timeout"],
+        output_path=output,
+    )
+    run_genai_perf(config, console=console, json_mode=json_mode)
+
+
 @click.command("perf")
 @cli_utils.model_option(required=False)
+@click.option(
+    "--runtime-type",
```
xieofxie: should we use --runtime? Like this pr https://github.com/microsoft/winml-cli/pull/960 maybe use ort / ort-genai

src/winml/modelkit/commands/_perf_genai.py
```diff
@@ -0,0 +1,458 @@
...
+
+
+# =============================================================================
+# Constants
+# =============================================================================
+
+RUNTIME_TYPE = "winml-genai"
+
+# Built-in benchmark prompt used when the user does not pass --prompt.
+_DEFAULT_PROMPT = "Explain the theory of relativity in simple terms."
```
xieofxie, line 60: should be put into click.option

src/winml/modelkit/commands/perf.py
```diff
@@ -1490,8 +1490,133 @@ def _run_simple_loop(
...
+    type=click.Choice(["winml", "winml-genai"]),
```
xieofxie: use enum

src/winml/modelkit/session/genai_session.py
```diff
@@ -0,0 +1,939 @@
...
+    @classmethod
+    def _write_compile_marker(cls, ctx_onnx: Path, ep_alias: str, ep_opts: dict) -> None:
+        """Record the EP + options a compiled stage was built with (cache key)."""
+        cls._compile_marker_path(ctx_onnx).write_text(
+            json.dumps({"ep": ep_alias, "provider_options": ep_opts}, sort_keys=True),
+            encoding="utf-8",
+        )
+
+    @classmethod
+    def _epcontext_is_fresh(cls, src_onnx: Path, ctx_onnx: Path, ep_opts: dict) -> bool:
```
xieofxie, line 812: we also have a function to check if the file is already compiled and to skip

src/winml/modelkit/session/genai_session.py
```diff
@@ -0,0 +1,939 @@
...
+# ---------------------------------------------------------------------------
+# Valid EP short names.
+# "mixed" = use genai_config.json as-is (embeddings/lm_head on CPU,
+#           ctx/iter on the target accelerator).
+# EP routing is driven entirely by per-stage session_options in the bundle's
+# genai_config.json — GenaiSession never calls clear_providers/append_provider.
+# ---------------------------------------------------------------------------
+_VALID_EPS: frozenset[str] = frozenset({"cpu", "mixed", "qnn", "dml"})
+# EPs that require WinML EP discovery + registration before og.Model() init.
+_NEEDS_WINML_EPS: frozenset[str] = frozenset({"mixed", "qnn", "dml"})
```
xieofxie: I am thinking we could remove this and given we don't alter the genai_config for its EP used, we could detect from the config for the ep to register and compile

src/winml/modelkit/session/genai_session.py
```diff
@@ -0,0 +1,939 @@
...
+            import onnxruntime_genai as og
+
+            return og
+        except ImportError as exc:
+            raise GenaiNotInstalledError(
+                "onnxruntime_genai is not installed. "
+                "Install it with: pip install onnxruntime-genai-winml"
+            ) from exc
+
+    def _register_eps(self, og: object) -> None:
```
xieofxie: og parameter is not needed

src/winml/modelkit/session/genai_session.py
```diff
@@ -0,0 +1,939 @@
...
+            if dst.exists():
+                continue
+            if src.is_file():
+                try:
+                    dst.symlink_to(src.resolve())
+                except (OSError, NotImplementedError):
+                    shutil.copy2(src, dst)
+
+    @staticmethod
+    def _import_og() -> Any:
```
xieofxie, line 956: no need? already in dependency

src/winml/modelkit/session/genai_session.py
```diff
@@ -0,0 +1,939 @@
...
+            first_token_s=marks[2] - marks[1],
+            decode_s=[marks[i + 1] - marks[i] for i in range(2, 1 + generated)],
+        )
+
+    # ------------------------------------------------------------------
+    # Chat-template helpers
+    # ------------------------------------------------------------------
+
+    @staticmethod
+    def apply_chatml_template(
```
xieofxie: there is tokenizer.apply_chat_template

xieofxie: should we apply it so the latency reflects more on the real usage? But the prefill time will be longer than prompt

src/winml/modelkit/commands/_perf_genai.py
```diff
@@ -0,0 +1,458 @@
...
+    """Map a ``--device`` value to a :class:`GenaiSession` EP short name."""
+    return _DEVICE_TO_GENAI_EP.get(device.lower(), "mixed")
+
+
+def default_genai_prompt() -> str:
+    """Return the built-in benchmark prompt."""
+    return _DEFAULT_PROMPT
+
+
+def genai_output_path(bundle_dir: str | Path) -> Path:
```
xieofxie: should we update generate_output_path in perf.py to unify the logic in one function?

src/winml/modelkit/commands/perf.py
```diff
@@ -1490,8 +1490,133 @@ def _run_simple_loop(
 # =============================================================================
 
 
+# perf() param names for WinML-only options that a prebuilt genai bundle
+# ignores. Mapped to the user-facing flag for the warning message.
+_GENAI_IGNORED_FLAGS: dict[str, str] = {
```
xieofxie, line 1502: should we use white list instead?

src/winml/modelkit/commands/perf.py
DingmaomaoBJTU: Renamed `--runtime-type` → `--runtime` (values `winml`|`winml-genai`), and the `benchmark_info` key `runtime_type` → `runtime`, to converge with #960's `--runtime`. Kept the values winml-prefixed for now: #960 is on a different axis (`ort`/`openvino`, backend on the same ONNX), so aligning the value vocabulary is best done when #960 merges. ef01cdce

src/winml/modelkit/commands/_perf_genai.py
DingmaomaoBJTU: Keeping `default=None` + runtime fallback here. `_DEFAULT_PROMPT` is surfaced via `default_genai_prompt()`, which lives in `_perf_genai` and pulls in the heavy `..session` package. That module is imported lazily *inside* the command body specifically so `winml perf --help` doesn't pay the import cost (guarded by tests/cli/test_import_time.py). Putting the default in the `click.option` decorator would evaluate it at import time and reintroduce that startup cost. `default=None` + a runtime fallback is the idiomatic click pattern for a default that can't be computed at import time.

src/winml/modelkit/commands/perf.py
DingmaomaoBJTU: Kept `click.Choice(["winml", "winml-genai"])`. The codebase has no enum-backed click options (no `from enum import` anywhere under `commands/`); the established convention is plain `click.Choice([...])`, and #960 uses `click.Choice` too. A one-off enum here would diverge from both.

src/winml/modelkit/session/genai_session.py
DingmaomaoBJTU: Added a guard in `_prepare_compiled_bundle` that reuses `onnx.is_compiled_onnx` to skip stages whose source ONNX is already an EPContext model — it points ort-genai at the original file instead of recompiling. A parse failure is treated as "not compiled" so a malformed source still falls through to the normal compile path. ef01cdce

src/winml/modelkit/session/genai_session.py
DingmaomaoBJTU: Removed `_NEEDS_WINML_EPS`. EP registration and stage pre-compilation are now derived from `genai_config.json` via `_bundle_uses_hardware_ep` — a bundle needs WinML EP registration iff some stage's `session_options.provider_options` names a non-CPU provider (the only referenced name is the universal `cpu` alias, so no hardware EP short name is hardcoded). The `ep` argument is informational now; using `--ep`/`--device` to override the config is tracked in #1025. ef01cdce

src/winml/modelkit/session/genai_session.py
DingmaomaoBJTU: Removed the unused `og` parameter from `_register_eps`. ef01cdce

src/winml/modelkit/session/genai_session.py
DingmaomaoBJTU: Keeping `_import_og`. `onnxruntime-genai-winml` is a declared dependency, but `import onnxruntime_genai` can still raise `ImportError` when the native module fails to load (a real failure mode on Windows/ARM64 hardware-EP builds even with the wheel installed). The `try/except` converts that into an actionable `GenaiNotInstalledError`, which the CLI path consumes (`_perf_genai.py`) and tests assert (`test_genai_not_installed_raises`). Without it the user gets a raw native ImportError.

src/winml/modelkit/session/genai_session.py
```diff
@@ -0,0 +1,1005 @@
...
+# Valid EP short names accepted by the ``ep`` argument.
+# "mixed" = use genai_config.json as-is (embeddings/lm_head on CPU,
+#           ctx/iter on the target accelerator).
+# EP routing is driven entirely by per-stage session_options in the bundle's
+# genai_config.json — GenaiSession never calls clear_providers/append_provider,
+# and it decides whether to register WinML EPs / pre-compile stages by reading
+# that config (see GenaiSession._bundle_uses_hardware_ep), not from ``ep``.
+# Using ``ep``/``device`` to override the config is tracked in #1025.
+# ---------------------------------------------------------------------------
+_VALID_EPS: frozenset[str] = frozenset({"cpu", "mixed", "qnn", "dml"})
```
xieofxie: since we register all eps, we could remove this

src/winml/modelkit/session/genai_session.py
```diff
@@ -0,0 +1,1005 @@
...
+            ONNX.  Defaults to 300 seconds (5 minutes).
+    """
+
+    # Sub-directory within the bundle that holds pre-compiled EPContext ONNX files.
+    _COMPILED_SUBDIR: str = "_compiled"
+
+    def __init__(
+        self,
+        bundle_dir: str | Path,
+        ep: str = "cpu",
```
xieofxie, line 286: Consider changing to EPNameOrAlias typed in future iterations

src/winml/modelkit/commands/perf.py
xieofxie: oh, sorry, I mean follow like
```
@click.option(
    "--compiler",
    type=click.Choice(list(COMPILER_NAMES)),
    default="ort",
    help="Compiler backend (default: ort). 'ort_session' compiles via "
    "ort.InferenceSession (ep.context_enable) — required for shared-context multi-model.",
)
```
We should use type for a closed set when possible instead of plain str

src/winml/modelkit/session/genai_session.py
xieofxie: but the error and message are not related to the error you stated. As you said, user already installs it but a ImportError throws. Any better error message ?

src/winml/modelkit/commands/_perf_genai.py
DingmaomaoBJTU: You're right that the default is a plain string with no import cost — my earlier concern conflated it with calling `default_genai_prompt()` (which does pull in `..session`). Moved the default into the option: `--prompt` now uses `default="Explain the theory of relativity in simple terms."` with `show_default=True`, and I dropped the `None` fallback plus the `default_genai_prompt()` helper. `GenaiPerfConfig.prompt` keeps `_DEFAULT_PROMPT` as its field default, and a test asserts the CLI default and the field default stay in sync.

src/winml/modelkit/commands/perf.py
DingmaomaoBJTU: Followed the `--compiler`/`COMPILER_NAMES` pattern: added `RuntimeName = Literal["winml", "winml-genai"]` and `RUNTIME_NAMES = get_args(RuntimeName)`, so `--runtime` is now `type=click.Choice(list(RUNTIME_NAMES))` and the handler param is typed `runtime: RuntimeName` instead of plain `str`. The closed set lives in one place now.

src/winml/modelkit/session/genai_session.py
DingmaomaoBJTU: Reworded the error. It no longer claims "not installed" — it reports that the import failed and names both possible causes, and includes the underlying exception: `Could not import onnxruntime_genai: {exc}. It may not be installed, or it is installed but its native module failed to load (...)`. Updated the `GenaiNotInstalledError` docstring to match. The message also no longer hard-codes the `pip install` line — the CLI handler already appends it, so the old message double-printed it.

src/winml/modelkit/session/genai_session.py
DingmaomaoBJTU: Left the benchmark on the raw prompt, deliberately:

- It mirrors ort-genai's official `benchmark_e2e.py` (the timing reference this path is modeled on), which generates from a raw prompt / synthetic token counts, not a chat-templated one.
- og's `Tokenizer` has no `apply_chat_template` (that's a HF `transformers` API), and the bundle doesn't ship the HF tokenizer, so there's no generic way to read the model's real template at runtime.
- Auto-applying one would require hard-coding a specific format (ChatML vs Llama `[INST]` vs Phi `<|user|>` all differ), which conflicts with the architecture-agnostic rule.

As you note, a template only adds a fixed model-specific delta to prefill and doesn't change decode throughput (the headline number). `GenaiSession.apply_chatml_template` stays as an opt-in helper for ChatML models, and a user can pass a pre-templated `--prompt`. If you'd like a realism mode, I can file a tracking issue for an opt-in `--apply-chat-template` flag.

src/winml/modelkit/commands/_perf_genai.py
DingmaomaoBJTU: Unified: `genai_output_path` now delegates to `perf.generate_output_path` (function-local import to avoid a module-level cycle, since `perf` imports this module lazily inside its command body). Both runtimes now share the one `~/.cache/winml/perf/<slug>/<ts>.json` implementation; passing `bundle_dir.name` as the slug keeps the previous path.

src/winml/modelkit/commands/perf.py
DingmaomaoBJTU: Kept the deny-list. It enumerates this command's own WinML build-pipeline options that a prebuilt bundle can't use (task/precision/quant/optimize/analyze/rebuild/…) and maps each to its user-facing flag for the warning text. A whitelist would additionally have to list the shared/global flags injected by common decorators (`--device`, `--output`, `--overwrite`, `--json`, `--verbose`, `--quiet`, `--config`, …); a new one added upstream would then produce a *false* "ignored" warning even though the genai path honors it — coupling this warning to decorator internals. The deny-list's failure mode (a future WinML-only flag not warned about) seems quieter than falsely warning on a flag that works. Happy to flip it if you'd prefer that trade-off.

src/winml/modelkit/session/genai_session.py
DingmaomaoBJTU: Done — removed `_VALID_EPS` and the `ep` validation in `__init__`. Registration/compilation is decided entirely from `genai_config.json`, not `ep`, so the check was decorative, and the frozenset was itself a hard-coded EP-alias list. `ep` is now a free-form reporting label; dropped `test_unknown_ep_raises` to match.

src/winml/modelkit/session/genai_session.py
DingmaomaoBJTU: Agreed as a future iteration — tracking it in #1025 (the ep/device redesign). One thing to reconcile there: the current `mixed` value (means "register all WinML EPs" for mixed-EP pipelines) isn't a member of `EPNameOrAlias`, so typing `ep` needs that handled rather than a drop-in annotation. Noted on the issue.

src/winml/modelkit/session/genai_session.py
DingmaomaoBJTU: Went ahead and implemented this. The benchmark now wraps the prompt in the bundle's own chat template before measuring, so the prefill reflects realistic chat usage.

Kept it generic by using `og.Tokenizer.apply_chat_template` (available in onnxruntime-genai) rather than hardcoding a format: `GenaiSession.apply_chat_template(prompt, system=None)` builds the `[{role, content}]` messages, reads the bundle's `chat_template.jinja` sidecar when present (and otherwise lets ort-genai fall back to the template embedded in the tokenizer config), then renders via the model's own template — so ChatML / Llama `[INST]` / Phi etc. all work with no model-specific code here. Bundles that ship no template fall back to the raw prompt (logged), so nothing breaks.

I also replaced the old hardcoded `apply_chatml_template` helper with this model-driven `apply_chat_template`, and it's applied by default in `winml perf --runtime winml-genai`.

xieofxie (approved):

PR #1017, opened 2026-07-01
Fix ep_registry spawn-progress stderr bug breaking child EP discovery

### LINKED ISSUE #1007: Bug: check_patterns compile always false during pattern check_results generation
﻿## Summary
check_patterns appears to have a regression in the pattern check-results generation flow: compilation is failing for almost all cases, including cases that previously passed with compile=true.

Affected example:
- File: Gelu2Pattern_QNNExecutionProvider_NPU_ai.onnx_opset13.json
- case_index: 69j0c5030116e46901067eec339d93375b72

## Reproduction Paths Tested
Both paths below were tested, and both produced compile=false:
1. Run dry_run first, then run 
un_all_dry_run
2. Directly invoke winml.modelkit.analyze.pattern.check_patterns

## Observed Behavior
- For a large number of newly created patterns, there are almost no cases with compile=true.
- Previously passing cases are now failing compilation.

## Expected Behavior
- Previously passing cases should continue to compile successfully.
- Newly generated pattern check results should not regress broadly to compile=false.

## Request
Please investigate and identify the concrete root cause, then provide a fix.

fangyangci:
## Summary
This PR fixes a spawn-worker progress rendering bug in EP readiness.

When tqdm writes to an unavailable sys.stderr (seen in spawned child workers or non-console hosts), _ensure_provider_ready can fail before provider readiness finishes. That leaves child ep_paths empty and causes check_patterns/check_ops to report compile=false incorrectly.

## Changes
- Harden _make_progress_bar in ep_registry:
  - fallback to _NoopBar when sys.stderr is missing or non-writable
  - wrap tqdm initialization in a safe fallback to _NoopBar
  - keep debug logging for fallback reasons
- Add explicit in-code note describing the spawn/stderr failure mode and EP path impact.
- Add regression unit test: test_ensure_provider_ready_works_when_stderr_is_none

## Validation
- uv run pytest tests/unit/session/test_ep_registry.py -k stderr_is_none -q
- Verified that tqdm stderr failures no longer block EP path readiness flow.

Fixes #1007


xieofxie (approved):

xieofxie:
need fix test

xieofxie (approved):

PR #1020, opened 2026-07-02
Align pattern support with node-level checks (scoped 15-pattern rules)

fangyangci:
Artifact pattern generate:https://github.com/gim-home/ModelKitArtifacts/pull/156
## Summary
- Enable rules checks for the scoped 15 patterns (alternative target patterns + check patterns).
- For pattern cases with node count > 1, run node-level precheck first; if all nodes are supported, directly use the supported result. To keep pattern check results consistent with split node-level check results on supported cases.
- Unify and stabilize `check_result` output via centralized sanitization and backward-compatible import path.

## Scope
- `available_ops_all.json` has 19 pattern definitions in total.
- This PR intentionally enables rules checks for the scoped 15 patterns, not all 19.

## Consistency Note
- Helper path without node-level precheck exists in:
  - ModelKitArtifacts/scripts/check_ops_run_helps/debug_case_helper/debug_pattern_help.bat
- Results between the helper path and the node-level precheck path have been verified as consistent.

## Validation
- `uv run ruff check src/winml/modelkit/analyze src/winml/modelkit/pattern tests/unit/analyze`
- `uv run mypy -p winml.modelkit`
- `uv run pytest tests/unit/analyze/runtime_checker/test_ep_checker.py tests/unit/analyze/test_check_ops.py -q`

### DIFF
--- a/src/winml/modelkit/analyze/core/rules_prefilter.py
+++ b/src/winml/modelkit/analyze/core/rules_prefilter.py
@@ -1,0 +1,94 @@
+# -------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+# --------------------------------------------------------------------------
+"""Rules-based prefilter utilities for runtime-check skip decisions."""
+
+from __future__ import annotations
+
+import logging
+from typing import TYPE_CHECKING, Any
+
+from .runtime_checker_query import RuntimeCheckerQuery
+
+
+if TYPE_CHECKING:
+    import onnx
+
+    from ...utils.constants import EPName
+
+
+logger = logging.getLogger(__name__)
+
+
+class RuntimeCheckerRulesPrefilter:
+    """Fixed rules-prefilter service bound to one EP+device pair."""
+
+    def __init__(self, ep_name: EPName, device_type: str) -> None:
+        self._ep_name = ep_name
+        self._device_type = device_type
+
+    def build_skip_check_result_for_rules_all_nodes_compile_run_pass(
+        self,
+        onnx_model: onnx.ModelProto,
+    ) -> dict[str, Any] | None:
+        """Build synthetic check_result when rules indicate all nodes pass."""
+        return build_skip_check_result_for_rules_all_nodes_compile_run_pass(
+            onnx_model,
+            self._ep_name,
+            self._device_type,
+        )
+
+
+def build_skip_check_result_for_rules_all_nodes_compile_run_pass(
+    onnx_model: onnx.ModelProto,
+    ep_name: EPName,
+    device_type: str,
+) -> dict[str, Any] | None:
+    """Return synthetic check_result when all nodes have rules compile/run pass.
+
+    Returns None when rules data is missing, any node is unsupported,
+    or prefilter evaluation fails. Caller should then run real compile/run.
+    """
+    try:
+        query = RuntimeCheckerQuery(
+            model_proto=onnx_model,
+            ep_name=ep_name,
+            device_type=device_type,
+        )
+        node_results = [
+            query.run_for_node(node, run_unknown_op=False) for node in query.model_proto.graph.node
+        ]
+
+        if not node_results:
+            return None
+
+        all_supported = all(
+            result.result.compile and result.result.run and not result.result.no_data
+            for result in node_results
+        )
+        if not all_supported:
+            return None
+
+        reason = f"rules_all_nodes_compile_run_pass_{len(node_results)}_nodes"
+        return {
+            "compile": {
+                "result": {"success": True, "reason": reason},
+                "stdout": "skipped_by_rules_all_nodes_compile_run_pass",
+                "stderr": "",
+            },
+            "run": {
+                "result": {"success": True, "reason": reason},
+                "stdout": "skipped_by_rules_all_nodes_compile_run_pass",
+                "stderr": "",
+            },
+        }
+    except Exception as exc:
+        logger.warning("Rules prefilter failed for one case (%s); fallback to ep_checker.", exc)
+        return None
+
+
+__all__ = [
+    "RuntimeCheckerRulesPrefilter",
+    "build_skip_check_result_for_rules_all_nodes_compile_run_pass",
+]
--- a/src/winml/modelkit/analyze/pattern/check_patterns.py
+++ b/src/winml/modelkit/analyze/pattern/check_patterns.py
@@ -27,7 +27,6 @@ from ...pattern.base import (
     get_pattern_input_generator,
     get_registered_pattern_input_generators,
 )
-from ...sysinfo import SysInfo
 
 
 if TYPE_CHECKING:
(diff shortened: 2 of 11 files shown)


src/winml/modelkit/analyze/core/rules_prefilter.py
```diff
@@ -0,0 +1,110 @@
...
+        return copy.deepcopy(result)
+
+
+def build_skip_check_result_for_rules_all_nodes_compile_run_pass(
+    onnx_model: onnx.ModelProto,
+    ep_name: EPName,
+    device_type: str,
+) -> dict[str, Any] | None:
+    """Return synthetic check_result when all nodes have rules compile/run pass.
+
```
DingmaomaoBJTU, line 49: This treats node-level parquet rules as a proxy for whole-pattern support: when every node is "supported" we write a synthetic success and skip the real EP compile/run for the fused pattern. Two things worth confirming:

1. A newly-enabled pattern now emits a "supported" result **without ever running on the EP**, so an incorrect/stale node rule silently produces a wrong artifact. Is there a periodic full (non-prefiltered) run to catch drift?
2. This re-derives `all_compile / all_run / any_no_data`, which `RuntimeCheckerQuery._run_for_subgraph_per_node` already computes. Could it reuse that path so the two don't diverge over time?

src/winml/modelkit/analyze/core/rules_prefilter.py
```diff
@@ -0,0 +1,110 @@
...
+                "stderr": "",
+            },
+        }
+    except Exception as exc:
+        logger.warning("Rules prefilter failed for one case (%s); fallback to ep_checker.", exc)
+        return None
+
+
+__all__ = [
+    "RuntimeCheckerRulesPrefilter",
```
DingmaomaoBJTU, line 92: Catching bare `Exception` means any programming error (not just missing rules) is swallowed and we silently fall back to the expensive real compile/run — exactly the path this is meant to avoid — with only a WARNING to show for it. Consider narrowing to the expected failure modes, or at least `logger.warning(..., exc_info=True)` so regressions stay debuggable.

src/winml/modelkit/analyze/core/rules_prefilter.py
```diff
@@ -0,0 +1,110 @@
...
+        onnx_model: onnx.ModelProto,
+    ) -> dict[str, Any] | None:
+        """Build synthetic check_result when rules indicate all nodes pass."""
+        model_hash = hashlib.sha256(onnx_model.SerializeToString()).hexdigest()
+        cache_key = f"{self._ep_name}|{self._device_type}|{model_hash}"
+        if cache_key in self._cache:
+            cached_result = self._cache.pop(cache_key)
+            self._cache[cache_key] = cached_result
+            return copy.deepcopy(cached_result)
+
```
DingmaomaoBJTU: The cache key is the full-model SHA-256. Each generated pattern case is a distinct model, so the hit rate is effectively zero while every multi-node case pays an extra full `SerializeToString()` + `deepcopy`. Is the cache buying anything for this workload? If duplicate models are rare it may be simpler and faster to drop it.

src/winml/modelkit/analyze/pattern/check_patterns.py
```diff
@@ -234,6 +241,19 @@ def __init__(self, device_type: ort.OrtHardwareDeviceType) -> None:
         super().__init__(ep_name="QNNExecutionProvider", device_type=device_type)
 
 
+class RTXChecker(EPChecker):
```
DingmaomaoBJTU, line 244: `RTXChecker` and `get_ep_checker` now exist both here and in `check_ops.py`, and the two `get_ep_checker`s have already drifted — this one omits VitisAI/MIGraphX. Every new EP has to be added in two places. Worth pulling the checker wrappers + `get_ep_checker` into one shared module.

src/winml/modelkit/analyze/runtime_checker/result_sanitizer.py
```diff
@@ -0,0 +1,10 @@
+# -------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+# --------------------------------------------------------------------------
+"""Backward-compatible re-export for runtime checker result sanitizers."""
+
+from ...utils.result_sanitizer import sanitize_check_result_payload, sanitize_result_text
```
DingmaomaoBJTU, line 7: I couldn't find any in-repo importer of this module — `op_input_gen.py` imports `sanitize_check_result_payload` directly from `...utils.result_sanitizer`. If this shim is for the external ModelKitArtifacts scripts, a one-line comment saying so would help; otherwise it looks removable.

DingmaomaoBJTU (commented):
Reviewed the logic statically (didn't run the suite). Overall the rules-first prefilter plus the `sys_info` / `yield_skipped` cleanups look coherent — I verified no remaining callers pass the removed `sys_info`/`yield_skipped` args, `device_type.name` is the right value to feed `RuntimeCheckerQuery`, and the synthetic `check_result` shape matches the real `{compile, run}` one. Left inline notes; the main one is a design question about skipping the real EP run purely on node-level rules.

Two nits not worth their own threads: `available_ops_all.json` gained two trailing blank lines at EOF, and `_RulesPrefilterProtocol`'s method body reads better as `...` than `pass`.

src/winml/modelkit/analyze/core/rules_prefilter.py
fangyangci: Thanks for calling this out. This is also by design: prefilter only answers a binary question (all nodes supported or not). We intentionally keep prefilter semantics separate from pattern/node exception semantics, and fall back to the real checker path when prefilter cannot provide a definitive supported result.

src/winml/modelkit/analyze/core/rules_prefilter.py
fangyangci: Good point, and agreed. model_bytes is designed to be unique per generated case, so cache hit rate is not expected to be meaningful here. I removed the cache/hash/deepcopy path from RuntimeCheckerRulesPrefilter in this PR.

src/winml/modelkit/analyze/runtime_checker/result_sanitizer.py
fangyangci: Good catch. This shim is intentionally kept for compatibility with ModelKitArtifacts imports (for example: ModelKitArtifacts/scripts/check_ops_run_helps/run_all_dry_run/src/op_case_runner.py). So it is not removable at this point.

src/winml/modelkit/analyze/pattern/check_patterns.py
fangyangci: Thanks for the suggestion. By design, pattern support scope is not fully unified with op-level support scope. In this change we intentionally keep pattern checker EP scope limited, and VitisAI/MIGraphX are out of scope for this PR.

src/winml/modelkit/analyze/core/rules_prefilter.py
fangyangci: Thanks for the review. This is by design for the scoped 15-pattern rules migration. The prefilter is a rules-based gate and is not intended to replace full EP validation. For real-check vs prefilter comparisons, we use targeted manual case-by-case checks; random sampling is often not representative for this issue space.

DingmaomaoBJTU (approved):

PR #1021, opened 2026-07-02
Add --op-tracing flag to e2e run_eval

xieofxie:
## Summary

Adds an `--op-tracing {basic,detail}` flag to `scripts/e2e_eval/run_eval.py` so operator-level profiling can be driven end-to-end through the eval runner, and copies the resulting op-trace JSON into each model's output folder.

## Changes

- **`scripts/e2e_eval/run_eval.py`**
  - New `--op-tracing {basic,detail}` CLI flag, plumbed through `run_model` for both the HF-model-id and pre-built-ONNX (composite) perf paths.
  - When enabled, passes `--op-tracing <level>` to `winml perf`, parses the `Op-trace saved to:` line from perf's output (rejoining Rich-wrapped path fragments), and copies the file into the model's output dir as `op_tracing.json` (label-suffixed for composite sub-models). Missing traces are ignored (op-tracing may be unsupported for a given EP/device/level).

- **`src/winml/modelkit/optracing/qnn/profiler.py`**
  - Gate `ep.context_enable` / `ep.context_embed_mode` session options to `detail` level only.
  - Resolve `profiling_file_path` to an absolute path.

- **Tooling**
  - Bump pinned ruff in `.pre-commit-config.yaml` (`v0.15.0` → `v0.15.13`) to pick up the S603 false-positive fix; align `pyproject.toml` dev dependency to `ruff>=0.15`.


KayMKM:
### Review notes

Nice work — CI is green and the plumbing through `run_model` is clean: `shutil`/`Path` are already imported, `model_dir` is always defined before the call, and copy failures are guarded with `OSError`. A few things worth addressing before merge.

#### 🔴 `--op-tracing` makes `winml perf` exit non-zero when tracing is unsupported → run_eval marks perf as failed

`is_profiling_available()` only returns `True` for **QNN + NPU + `basic`**. When op-tracing is requested but unsupported, `perf.py` raises `click.ClickException` *after* the benchmark has already run and its results were written, so the command exits non-zero.

In `run_eval.py` this has two consequences:

- **`--op-tracing detail` is never supported** by `is_profiling_available`, so every non-composite perf run exits non-zero and is marked failed. In `both` mode this also skips accuracy (the `perf_proc["exit_code"] != 0` → `perf_failed` branch). Yet argparse advertises `detail` as a valid choice.
- **`--op-tracing basic`**: any model that doesn't resolve to QNN+NPU (e.g. `--device auto` landing on CPU/GPU) also exits non-zero and is falsely marked failed — even though the benchmark numbers themselves are fine.

The graceful degradation described in the PR only happens at the copy layer (`_copy_op_trace`), not at the perf exit-code layer. Suggested fixes (any one):

1. Drop `detail` from `choices` until perf actually supports it.
2. Mirror the `is_profiling_available` condition in run_eval and only append `--op-tracing` when the resolved ep is qnn, device is npu, and level is `basic`.
3. (perf.py, out of scope here) make unsupported op-tracing warn instead of changing the command exit code.

#### 🔴 Output filename won't match the downstream report reader

This PR writes the trace as **`op_tracing.json`** (`op_tracing_<label>.json` for composite sub-models), but the report generator that renders these traces looks for **`op_trace.json`**. With both in place, the report will never find real pipeline artifacts. Note that `winml perf`'s own native output uses `<stem>_op_trace.json` as well. Please settle on a single filename across the producer and the consumer (aligning with perf's `op_trace` spelling seems most consistent).

#### 🟡 `_extract_op_trace_path` edge cases

- `joined.find(".json")` takes the *first* `.json`; a parent directory containing `.json` would truncate the path.
- If Rich wraps the path exactly at a space (e.g. `C:\Users\First Last\...`), stripping `\n` drops the space and `exists()` silently fails, so the trace is skipped. Both are narrow, and the docstring already flags the fragility — just calling it out.

#### 🟡 Nit

PR description says ruff `v0.15.0 → v0.15.13`, but the diff is `v0.14.0 → v0.15.13`.

#### 🟢 Nice

- `profiling_file_path = str(csv_path.resolve())` correctly anchors the QNN write path to the original cwd before the profiler `chdir`s into `output_dir`, matching where `_collect_results` reads it back.
- The `ep.context_*` gating on `detail` and the `optrace`/`detailed` provider-option split are sensible, and the unit tests were updated to match.

xieofxie:
> ### Review notes
> Nice work — CI is green and the plumbing through `run_model` is clean: `shutil`/`Path` are already imported, `model_dir` is always defined before the call, and copy failures are guarded with `OSError`. A few things worth addressing before merge.
> 
> #### 🔴 `--op-tracing` makes `winml perf` exit non-zero when tracing is unsupported → run_eval marks perf as failed
> `is_profiling_available()` only returns `True` for **QNN + NPU + `basic`**. When op-tracing is requested but unsupported, `perf.py` raises `click.ClickException` _after_ the benchmark has already run and its results were written, so the command exits non-zero.
> 
> In `run_eval.py` this has two consequences:
> 
> * **`--op-tracing detail` is never supported** by `is_profiling_available`, so every non-composite perf run exits non-zero and is marked failed. In `both` mode this also skips accuracy (the `perf_proc["exit_code"] != 0` → `perf_failed` branch). Yet argparse advertises `detail` as a valid choice.
> * **`--op-tracing basic`**: any model that doesn't resolve to QNN+NPU (e.g. `--device auto` landing on CPU/GPU) also exits non-zero and is falsely marked failed — even though the benchmark numbers themselves are fine.
> 
> The graceful degradation described in the PR only happens at the copy layer (`_copy_op_trace`), not at the perf exit-code layer. Suggested fixes (any one):
> 
> 1. Drop `detail` from `choices` until perf actually supports it.
> 2. Mirror the `is_profiling_available` condition in run_eval and only append `--op-tracing` when the resolved ep is qnn, device is npu, and level is `basic`.
> 3. (perf.py, out of scope here) make unsupported op-tracing warn instead of changing the command exit code.
> 
> #### 🔴 Output filename won't match the downstream report reader
> This PR writes the trace as **`op_tracing.json`** (`op_tracing_<label>.json` for composite sub-models), but the report generator that renders these traces looks for **`op_trace.json`**. With both in place, the report will never find real pipeline artifacts. Note that `winml perf`'s own native output uses `<stem>_op_trace.json` as well. Please settle on a single filename across the producer and the consumer (aligning with perf's `op_trace` spelling seems most consistent).
> 
> #### 🟡 `_extract_op_trace_path` edge cases
> * `joined.find(".json")` takes the _first_ `.json`; a parent directory containing `.json` would truncate the path.
> * If Rich wraps the path exactly at a space (e.g. `C:\Users\First Last\...`), stripping `\n` drops the space and `exists()` silently fails, so the trace is skipped. Both are narrow, and the docstring already flags the fragility — just calling it out.
> 
> #### 🟡 Nit
> PR description says ruff `v0.15.0 → v0.15.13`, but the diff is `v0.14.0 → v0.15.13`.
> 
> #### 🟢 Nice
> * `profiling_file_path = str(csv_path.resolve())` correctly anchors the QNN write path to the original cwd before the profiler `chdir`s into `output_dir`, matching where `_collect_results` reads it back.
> * The `ep.context_*` gating on `detail` and the `optrace`/`detailed` provider-option split are sensible, and the unit tests were updated to match.

for the 1st one, it is OK as it will only run on supported device now
for the 2nd one, fixed

KayMKM (approved):

PR #1023, opened 2026-07-02
Fix issues in depth-estimation and keypoint-detection

zhenchaoni:
This PR contains
1. Fix bugs found while running evaluation for depth-estimation and keypoint-detection. The issues causes evaluation errors during inference.
2. Run the evaluation and record the baseline cache. The baseline cache aligns with published or community shared metric values

### DIFF
--- a/scripts/build_coco_keypoints.py
+++ b/scripts/build_coco_keypoints.py
@@ -79,6 +79,10 @@ def _fetch_image(file_name: str) -> bytes:
 
 def build(output_dir: Path, num_images: int, cache_dir: Path, seed: int = 42) -> None:
     """Build and save the COCO keypoints dataset to ``output_dir``."""
+    if (output_dir / "dataset_info.json").exists():
+        print(f"Dataset already exists at {output_dir}, skipping build.")
+        return
+
     from datasets import Dataset, Features, Image, Sequence, Value
     from PIL import Image as PILImage
 
--- a/scripts/build_coco_keypoints.py
+++ b/scripts/build_coco_keypoints.py
@@ -141,7 +141,7 @@ def main() -> int:
     parser.add_argument(
         "--num-images",
         type=int,
-        default=100,
+        default=500,
         help="Number of images to include (0 = all images with keypoints).",
     )
     parser.add_argument(
--- a/scripts/e2e_eval/testsets/models_with_acc.json
+++ b/scripts/e2e_eval/testsets/models_with_acc.json
@@ -1918,6 +1918,7 @@
       "build_script": "scripts/build_coco_keypoints.py",
       "path": "~/.cache/winml/datasets/coco_keypoints_val2017",
       "split": "validation",
+      "samples": 500,
       "metric": "map",
       "winml_metric_key": "map",
       "columns_mapping": {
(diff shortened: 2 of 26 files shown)


timenick (approved):

DingmaomaoBJTU (approved):

scripts/build_coco_keypoints.py
```diff
@@ -79,6 +79,10 @@ def _fetch_image(file_name: str) -> bytes:
 
 def build(output_dir: Path, num_images: int, cache_dir: Path, seed: int = 42) -> None:
     """Build and save the COCO keypoints dataset to ``output_dir``."""
+    if (output_dir / "dataset_info.json").exists():
```
DingmaomaoBJTU, line 82: This early-return keys the cache only on `dataset_info.json` existing, not on `num_images`/`seed`. An earlier run (e.g. the 100-sample build from a previous commit) leaves a dataset here, so bumping the default to 500 — or requesting any different count — silently reuses the stale set at the wrong sample size. Could we validate the existing dataset length against `num_images` and rebuild on mismatch, or fold `num_images` into the cache path?

src/winml/modelkit/eval/keypoint_detection_evaluator.py
```diff
@@ -178,6 +204,21 @@ def _predict_poses(
...
+        convolution layout is not transposed back at the output boundary). This
+        might be a bug in MIGraphX; transpose to NCHW to work around it. It is a
+        no-op on every other EP.
+        """
+        if getattr(self.model, "ep_name", None) != "MIGraphXExecutionProvider":
+            return heatmaps
+        # The buffer is NHWC-ordered under the declared (N, C, H, W) shape:
+        # reinterpret as (N, H, W, C) and transpose back to (N, C, H, W).
+        n, c, h, w = heatmaps.shape
+        return heatmaps.reshape(n, h, w, c).permute(0, 3, 1, 2).contiguous()
```
DingmaomaoBJTU, line 220: This reinterpret-via-reshape is only correct when `heatmaps` is C-contiguous in the declared NCHW layout. If `_extract_heatmaps` ever returns a non-contiguous view, `reshape(n, h, w, c)` copies in logical NCHW order rather than reinterpreting the raw buffer, silently producing wrong keypoints on MIGraphX — and the trailing `.contiguous()` would *not* fix it (it just bakes in the wrong order). Worth an `assert heatmaps.is_contiguous()` before the reshape so a future change fails loudly instead of corrupting metrics.

scripts/e2e_eval/testsets/models_with_acc.json
```diff
@@ -1802,9 +1786,11 @@
       "revision": "refs/convert/parquet",
       "split": "validation",
       "metric": "abs_rel",
+      "samples": 200,
       "columns_mapping": {
         "input_column": "image",
-        "depth_column": "depth_map"
+        "depth_column": "depth_map",
+        "depth_kind": "disparity"
```
DingmaomaoBJTU, line 1793: This change also removes `Intel/zoedepth-nyu-kitti` and `apple/DepthPro-hf` from the accuracy testset (the two metric-depth `align: none` entries). Is dropping that coverage intentional? If they were failing or too slow rather than deliberately out of scope, a `timeout_skip_list` entry would keep them visible instead of silently disappearing from the suite. The orphaned `Intel/zoedepth-nyu-kitti` baseline_cache entry is now unused too.

DingmaomaoBJTU (commented):
A few points from reviewing the depth-estimation / keypoint-detection fixes. The disparity-space affine alignment (align + invert against 1/gt) is the right MiDaS-style approach and the added tests cover it well. Three inline notes below: a cache foot-gun in the dataset builder, a contiguity precondition in the MIGraphX workaround, and a question about dropped model coverage.

PR #1026, opened 2026-07-02
fix: remove hidden `qnn` default from EPConfig and compile configs

### LINKED ISSUE #954: chore: compiler config / context uses qnn as execution_provider default which hides potential code bug that ep is actually not set
We should not have a default
<del>Also should migrate EPAlias to EPName</del>

xieofxie:
## Summary

Remove the hidden `"qnn"` default for `execution_provider` in `EPConfig`, `CompileContext`, and `WinMLCompileConfig.from_dict` so that a missing EP is surfaced immediately instead of silently falling back to QNN.

Closes https://github.com/microsoft/winml-cli/issues/954

## Approach

Changed `EPConfig.provider` from `EPAlias = "qnn"` to `EPAlias | None = None`. This preserves the existing API surface — `EPConfig()`, `WinMLCompileConfig()`, and `WinMLBuildConfig()` all still work with defaults — but the default is now `None` instead of a hidden `"qnn"`. Downstream validation in `WinMLBuildConfig.validate()` already catches `provider=None` when compile is enabled.

## Changes

| File | Change |
|---|---|
| `compiler/configs.py` | `EPConfig.provider` default: `"qnn"` → `None` (`EPAlias \| None`) |
| `compiler/configs.py` | `device` property: guards against `None` provider |
| `compiler/configs.py` | `from_dict`: returns `WinMLCompileConfig` with `provider=None` when key is absent (no `"qnn"` fallback); returns `None` for empty dicts |
| `compiler/context.py` | `execution_provider` property: returns `None` when key is absent (no `"qnn"` fallback) |
| `compiler/stages/optimize.py` | Guard `get_transforms_for_ep` against `None` EP |
| `compiler/stages/qformat.py` | Guard `needs_format_conversion` against `None` EP |
| `commands/build.py` | Guard `.upper()` call against `None` provider |
| Tests | Updated default-value assertions to expect `None`; tests that need a real EP use `for_qnn()` or `compile=None` |

### DIFF
--- a/tests/unit/compiler/test_compiler_configs.py
+++ b/tests/unit/compiler/test_compiler_configs.py
@@ -25,11 +25,6 @@ class TestEPConfig:
         assert config.compiler == "ort"
         assert config.qnn_sdk_root is None
 
-    def test_provider_is_required(self):
-        """Test that provider defaults to None (not a hidden EP like 'qnn')."""
-        config = EPConfig()
-        assert config.provider is None
-
     def test_custom_values(self):
         """Test custom EP configuration."""
         config = EPConfig(
--- a/tests/unit/compiler/test_compiler_configs.py
+++ b/tests/unit/compiler/test_compiler_configs.py
@@ -16,9 +16,9 @@ class TestEPConfig:
     """Test EPConfig dataclass."""
 
     def test_default_values(self):
-        """Test default EP configuration (non-provider fields)."""
-        config = EPConfig(provider="qnn")
-        assert config.provider == "qnn"
+        """Test default EP configuration."""
+        config = EPConfig()
+        assert config.provider is None
         assert config.provider_options == {}
         assert config.enable_ep_context is True
         assert config.embed_context is False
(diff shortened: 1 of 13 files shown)


tests/unit/compiler/test_compiler_configs.py
```diff
@@ -18,13 +18,18 @@ class TestEPConfig:
...
         config = EPConfig()
-        assert config.provider == "qnn"
+        assert config.provider is None
         assert config.provider_options == {}
         assert config.enable_ep_context is True
         assert config.embed_context is False
         assert config.compiler == "ort"
         assert config.qnn_sdk_root is None
 
+    def test_provider_is_required(self):
```
timenick: `test_provider_is_required` is misnamed and redundant: it constructs `EPConfig()` and asserts `provider is None`, which proves the field is *optional* (defaults to `None`) — the opposite of "required" — and duplicates the assertion already in `test_default_values` above. If it's meant to lock in the new no-hidden-default behavior, consider renaming to `test_provider_defaults_to_none`; otherwise it can be dropped as redundant. (Side note: the PR description still describes an earlier "make the fields required / raise `KeyError` on missing EP" design — `ep_config`/`compile` still have `default_factory`, and `from_dict` returns `None` rather than raising — so the description is worth updating to match this narrower change.)

xieofxie: drop and update

timenick (approved):

PR #1031, opened 2026-07-03
feat(export): support exporting sub-models from composite models

### LINKED ISSUE #1014: feat: Composite model: support export, build, perf etc. one or any of it sub models

xieofxie:
Add --submodel option to `winml export` for selecting a specific sub-model from composite models (e.g., encoder, decoder).

When a composite model is detected:
- --submodel <name>: exports only that component as {stem}_{name}.onnx
- No --submodel: exports all sub-models automatically
- Non-composite models: unchanged single-model export

Extract _export_single_model() helper shared by both paths and add _resolve_composite_model_components() for registry/auto-detection.

Resolves microsoft/winml-cli#1014


src/winml/modelkit/commands/export.py
```diff
@@ -243,6 +254,12 @@ def export(
 
         # Dynamic dimensions from a dedicated JSON file
         winml export -m bert-base-uncased -o bert.onnx --dynamic-axes dynamic_axes.json
+
+        # Export all sub-models of a composite model (auto-detected)
+        winml export -m google-t5/t5-small --task translation -o t5.onnx
+
+        # Export only the encoder sub-model
+        winml export -m google-t5/t5-small --task translation -o t5.onnx --submodel encoder
```
DingmaomaoBJTU, line 264: `test_export_help_examples_run` auto-runs every `winml export ...` line in this docstring, but it only patches `load_hf_model`, `export_pytorch` and `resolve_export_config` — not `resolve_composite_components`. So this example really calls `AutoConfig.from_pretrained('google-t5/t5-small')`.

Every other example still exits 0 if that resolution fails, because composite detection swallows the `OSError` and falls back to a single-model export. This `--submodel encoder` example is the only one that hard-fails (`--submodel ... is not a composite model`) the moment the Hub is unreachable — offline CI, a rate limit, or the model being renamed. That turns a Hub hiccup into a red test.

Suggest patching `resolve_composite_components` in that test, or not exercising `--submodel` from an auto-run docstring example.

src/winml/modelkit/commands/export.py
```diff
@@ -527,6 +544,19 @@ def _run_component_export(component_task: str | None, out_path: Path) -> None:
         # producing a single-model artifact for a possibly-composite model.
         raise click.ClickException(f"Composite model detection failed unexpectedly: {e}") from e
 
+    # ── --submodel validation ──────────────────────────────────────────
+    if submodel is not None:
+        if components is None:
+            raise click.ClickException(
```
DingmaomaoBJTU: Consistency nit: everywhere else this command raises `click.UsageError` for bad input — ONNX/dir passed to `-m` (~L269/276), model/task mismatch (~L530), `--input-specs` on a composite (~L565). These two `--submodel` failures are the same category (an invalid option value), so `click.UsageError` — or `click.BadParameter(..., param_hint="--submodel")` — would be more consistent and give the usual usage hint + exit code 2.

src/winml/modelkit/commands/export.py
```diff
@@ -527,6 +544,19 @@ def _run_component_export(component_task: str | None, out_path: Path) -> None:
...
+                f"--submodel '{submodel}' was specified, but '{model}' "
+                f"is not a composite model (no sub-models detected)."
+            )
+        if submodel not in components:
+            raise click.ClickException(
+                f"Unknown sub-model '{submodel}'. Available: {', '.join(components.keys())}"
+            )
+        components = {submodel: components[submodel]}
```
DingmaomaoBJTU, line 562: After narrowing to a single sub-model here, execution still flows into the composite branch below, which rejects `--input-specs` and tells the user to "export a sub-model individually with --task". But `--submodel` *is* that, and `--task` alone won't collapse a genuine composite down to one model — so that suggested workaround is a dead end. `--submodel encoder --input-specs specs.json` ends up with no valid path. Consider allowing `--input-specs` when exactly one sub-model is selected, or at least updating that message to mention `--submodel`.

src/winml/modelkit/commands/export.py
```diff
@@ -177,6 +177,16 @@ def _load_json_object(path: Path, option_name: str) -> dict:
         '(e.g., {"input_ids": {"0": "batch", "1": "sequence"}}).'
     ),
 )
+@click.option(
+    "--submodel",
+    type=str,
+    default=None,
+    help=(
+        "Export a specific sub-model from a composite model "
```
DingmaomaoBJTU, line 185: Minor: worth stating in the help that a selected sub-model is still written to `{stem}_{name}.onnx` (`-o t5.onnx --submodel encoder` -> `t5_encoder.onnx`), not to the `-o` path verbatim. The suffix is easy to miss when you have asked for a single sub-model.

DingmaomaoBJTU (commented):
Took a look — small, focused feature and the tests cover the happy path plus the two error cases. A few things worth addressing before merge: a test-hermeticity concern (the new `--submodel` docstring example is auto-run and uniquely depends on live Hub resolution), an exception-type consistency nit, and a `--submodel` + `--input-specs` dead end. Details inline.

Also, the PR description mentions extracting `_export_single_model()` and adding `_resolve_composite_model_components()`, but neither is in the diff — the code uses the existing `_run_component_export()` / `resolve_composite_components()`. Worth updating the description so it matches what actually changed.

src/winml/modelkit/commands/export.py
xieofxie: good point, we should support when one sub-model

src/winml/modelkit/commands/export.py
xieofxie: yes, we should patch resolve_composite_components  and verify it is called

src/winml/modelkit/commands/export.py
xieofxie: Done — the --submodel help now spells out that the sub-model is written to {stem}_{name}.onnx (e.g. -o t5.onnx --submodel encoder -> 	5_encoder.onnx), not the -o path verbatim.

src/winml/modelkit/commands/export.py
xieofxie: Good call — both --submodel failures now raise `click.BadParameter(..., param_hint='--submodel')` so they get the usual usage hint and exit code 2, consistent with the other invalid-option-value errors in this command.

src/winml/modelkit/commands/export.py
xieofxie: Fixed — `--input-specs` is now allowed when `--submodel` narrows the export to exactly one sub-model; the specs apply to that single component. The rejection only fires for a genuine multi-sub-model fan-out, and its message now points users at `--submodel` instead of the dead-end `--task` suggestion. Added `test_submodel_allows_input_specs` to cover it.

src/winml/modelkit/commands/export.py
xieofxie: Fixed — `test_export_help_examples_run` now patches `resolve_composite_components` (t5 -> encoder/decoder, everything else single-model) so the `--submodel encoder` example never reaches the Hub, and asserts the mock was called to prove the sub-model path is actually exercised.

DingmaomaoBJTU (commented):
Re-reviewed the `fix(export): address PR #1031 review comments` commit (`3bf05092`). All four earlier points are resolved correctly:

1. **Test hermeticity** — `resolve_composite_components` is now patched in `test_export_help_examples_run` (with a `.called` assertion), so the `--submodel` example no longer depends on the live Hub. Nice bonus: this makes *all* the docstring examples hermetic, not just the composite one.
2. **Exception type** — both `--submodel` validations now raise `click.BadParameter(param_hint="--submodel")`, consistent with the rest of the command (usage error + exit code 2).
3. **`--submodel` + `--input-specs` dead end** — the specs are now allowed for a single selected sub-model (`input_specs and submodel is None` guard), and the multi-component message points at `--submodel` instead of the `--task` dead end. I checked `test_composite_rejects_input_specs` still holds (single-component mock, no `--submodel` → still rejected, message still contains "composite").
4. **Help text** — now documents the `{stem}_{name}.onnx` output.

One follow-up inline (test-strength, non-blocking). Also very minor: the reworded rejection says "exporting multiple sub-models", but the guard fires for any composite without `--submodel`; it reads slightly off against the single-component mock in `test_composite_rejects_input_specs` (real composites are all ≥2, so it's only cosmetic). LGTM once the test point is considered.

tests/unit/commands/test_export.py
```diff
@@ -1129,3 +1145,174 @@ def fake_export_onnx(**kwargs):
...
+                    "--submodel",
+                    "decoder_prefill",
+                    "--input-specs",
+                    str(specs_file),
+                ],
+                obj={"debug": False},
+            )
+
+        assert result.exit_code == 0, result.output
+        assert mock_export_onnx.call_count == 1
```
DingmaomaoBJTU, line 1316: This verifies `--input-specs` is *not rejected* (exit 0 + output path), but not that it actually takes effect on the selected sub-model — which is the whole point of the change. I traced the flow: `_run_component_export` turns this `inputs.json` into `InputTensorSpec(name='input_ids', dtype='int64', shape=(1, 8))` and passes it via `config_kwargs['input_tensors']` into the `export_config` handed to `export_pytorch`, so it's directly assertable here, e.g.:

```python
spec_names = [t.name for t in mock_export_onnx.call_args.kwargs["export_config"].input_tensors]
assert "input_ids" in spec_names
```

Otherwise this test would still pass even if a future refactor silently dropped `--input-specs` for the single-sub-model path.

DingmaomaoBJTU (approved):
Approving. All four points from the earlier review are addressed correctly and I confirmed there is no regression (`test_composite_rejects_input_specs` still holds; changes are within the 100-char line limit). The one remaining inline note (strengthening `test_submodel_allows_input_specs` to assert the specs actually apply) is non-blocking and can be picked up as a follow-up.

Note: I could not execute the suite locally (win_arm64 has no torch wheel), so this is based on code-path analysis; please make sure CI is green before merging.

tests/unit/commands/test_export.py
xieofxie: fixed in next pr

xieofxie: Good catch. Strengthened the test to assert the spec actually reaches the selected sub-model rather than just being accepted:

```python
export_config = mock_export_onnx.call_args.kwargs["export_config"]
spec_names = [t.name for t in (export_config.input_tensors or [])]
assert "input_ids" in spec_names
```

Since #1031 is already merged, this landed on branch `hualxie/merge-classify-model-input` (PR #1079) in commit 23167ab8.

PR #1032, opened 2026-07-03
Auto-enable op-tracing via per-model op_tracing_targets

xieofxie:
## Summary

Adds per-model auto-enable of op-tracing in the e2e eval runner. When `run_eval.py` runs without an explicit `--op-tracing`, a model whose registry entry opts into the current `<EP>_<device>` target now gets `basic` op-tracing automatically.

## Changes

- **`scripts/e2e_eval/utils/registry.py`** — add `op_tracing_targets: list[str]` field to `ModelEntry` and load it in `load_registry`.
- **`scripts/e2e_eval/run_eval.py`**
  - `_op_tracing_target_key(ep, device)` builds the match key, e.g. `QNNExecutionProvider_npu`.
  - `_resolve_op_tracing(cli, entry, ep, device)` resolves the effective level: `disabled` forces off (even for opted-in models); `basic`/`detail` used as-is; unset defaults to `basic` when the target key is in the model's `op_tracing_targets`, otherwise off.
  - Add `disabled` to `--op-tracing` choices with expanded help text.
  - Compute effective `op_tracing` per model in the run loop and pass it to both `run_model` call sites.
- **`scripts/e2e_eval/testsets/models_all.json`** — add `"op_tracing_targets": ["QNNExecutionProvider_npu"]` to the 10 profiled models.

## Behavior

| `--op-tracing` | Model opts in (`op_tracing_targets` matches) | Result |
| --- | --- | --- |
| omitted | yes | `basic` |
| omitted | no | off |
| `basic`/`detail` | any | as specified |
| `disabled` | any | off |

### DIFF
--- a/scripts/e2e_eval/run_eval.py
+++ b/scripts/e2e_eval/run_eval.py
@@ -1379,6 +1379,39 @@ def _run_build_only(entries: list[ModelEntry], args: argparse.Namespace) -> None
 # ---------------------------------------------------------------------------
 
 
+def _op_tracing_target_key(ep: str | None, device: str) -> str | None:
+    """Return the ``<EP>_<device>`` key used to match ``op_tracing_targets``.
+
+    Example: ep=``qnn``, device=``npu`` -> ``QNNExecutionProvider_npu``. Returns
+    None when no EP is set (the target list only makes sense per EP/device).
+    """
+    if not ep:
+        return None
+    from winml.modelkit.utils.constants import normalize_ep_name
+
+    return f"{normalize_ep_name(ep)}_{device.lower()}"
+
+
+def _resolve_op_tracing(
+    cli_op_tracing: str | None, entry: ModelEntry, ep: str | None, device: str
+) -> str | None:
+    """Resolve the effective op-tracing level for one model run.
+
+    - ``disabled`` (explicit): never trace, even if the model opts in.
+    - ``basic``/``detail`` (explicit): use as-is.
+    - unset (None): default to ``basic`` when the model's ``op_tracing_targets``
+      includes the current ``<EP>_<device>`` target, otherwise no tracing.
+    """
+    if cli_op_tracing == "disabled":
+        return None
+    if cli_op_tracing in ("basic", "detail"):
+        return cli_op_tracing
+    key = _op_tracing_target_key(ep, device)
+    if key and key in entry.op_tracing_targets:
+        return "basic"
+    return None
+
+
 def _extract_op_trace_path(text: str) -> Path | None:
     """Parse the ``Op-trace saved to: <path>`` line from winml perf output.
 
@@ -1987,11 +2020,15 @@ def parse_args() -> argparse.Namespace:
     parser.add_argument(
         "--op-tracing",
         dest="op_tracing",
-        choices=["basic", "detail"],
+        choices=["basic", "detail", "disabled"],
         default=None,
         help=(
-            "Enable operator-level profiling in winml perf (requires onnxruntime-qnn). "
-            "The resulting op_trace.json is copied into each model's output folder."
+            "Operator-level profiling in winml perf (requires onnxruntime-qnn). "
+            "'basic'/'detail' force the tracing level; 'disabled' turns it off even "
+            "for models that opt in via 'op_tracing_targets'. When omitted, tracing "
+            "defaults to 'basic' for a model whose 'op_tracing_targets' includes the "
+            "current <EP>_<device> target (e.g. QNNExecutionProvider_npu). The "
+            "resulting op_trace.json is copied into each model's output folder."
         ),
     )
     parser.add_argument(
@@ -2353,6 +2390,7 @@ def main() -> None:
         try:
             perf_proc: dict | None = None
             accuracy_result: dict | None = None
+            op_tracing = _resolve_op_tracing(args.op_tracing, entry, args.ep, args.device)
 
             # Build phase: winml config + winml build → list of ONNX paths
             # Build is shared by perf and eval, avoiding redundant builds.
@@ -2394,7 +2432,7 @@ def main() -> None:
                     args.timeout,
                     onnx_paths,
                     ep=args.ep,
-                    op_tracing=args.op_tracing,
+                    op_tracing=op_tracing,
                     model_dir=model_dir,
                 )
             else:
@@ -2405,7 +2443,7 @@ def main() -> None:
                     args.timeout,
                     onnx_paths,
                     ep=args.ep,
-                    op_tracing=args.op_tracing,
+                    op_tracing=op_tracing,
                     model_dir=model_dir,
                 )
                 if perf_proc["exit_code"] != 0:
--- a/scripts/e2e_eval/run_eval.py
+++ b/scripts/e2e_eval/run_eval.py
@@ -57,7 +57,13 @@ from utils.accuracy import (
     format_delta,
 )
 from utils.dataset_config import get_dataset_config, register_from_registry
-from utils.registry import ModelEntry, filter_registry, load_registry, make_adhoc_entry
+from utils.registry import (
+    ModelEntry,
+    filter_registry,
+    load_registry,
+    make_adhoc_entry,
+    op_tracing_target_key,
+)
 from utils.reporter import (
     build_eval_result,
     classify_result,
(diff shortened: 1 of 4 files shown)


DingmaomaoBJTU (commented):
Reviewed the per-model auto-enable op-tracing change. The approach is clean overall: tracing runs as a separate additive pass so it doesn't skew the benchmark, and `disabled` gives an escape hatch. A few things worth addressing, the main one being the `device=auto` matching gap. Details inline.

scripts/e2e_eval/run_eval.py
```diff
@@ -1379,6 +1379,39 @@ def _record(
...
+    """Return the ``<EP>_<device>`` key used to match ``op_tracing_targets``.
+
+    Example: ep=``qnn``, device=``npu`` -> ``QNNExecutionProvider_npu``. Returns
+    None when no EP is set (the target list only makes sense per EP/device).
+    """
+    if not ep:
+        return None
+    from winml.modelkit.utils.constants import normalize_ep_name
+
+    return f"{normalize_ep_name(ep)}_{device.lower()}"
```
DingmaomaoBJTU: The match key is built from the raw `device` string, but `--device` defaults to `auto` (and we normalize `auto` -> `npu` elsewhere, e.g. `eval_device = "npu" if device == "auto"` around L1615). So a local `run_eval.py --ep qnn` with the default device builds `QNNExecutionProvider_auto`, which never matches `QNNExecutionProvider_npu`, and auto-tracing silently stays off. It only fires in CI because the pair catalog passes an explicit `--device npu`. Could we resolve `auto` to the concrete device before building the key, or at least document that an explicit `--device` is required for auto-enable to work?

scripts/e2e_eval/run_eval.py
```diff
@@ -1379,6 +1379,39 @@ def _record(
...
+
+
+def _resolve_op_tracing(
```
DingmaomaoBJTU, line 1388: No unit tests cover this new resolver, even though it branches on several cases (disabled / basic / detail / unset x match / no-ep / auto-device) and the sibling `_resolve_precision` is already tested in `tests/unit/eval/test_run_eval_script.py`. Could you add a parametrized test here, especially for the `disabled`-overrides-opt-in path and the `device=auto` no-match case noted above?

scripts/e2e_eval/testsets/models_all.json
```diff
@@ -19,7 +19,10 @@
     "downloads": 1258716,
     "last_update_time": "2024-12-03T14:14:31+00:00",
     "optimum_supported": true,
-    "order": 5
+    "order": 5,
+    "op_tracing_targets": [
```
DingmaomaoBJTU, line 23: The target is spelled with the full normalized EP name (`QNNExecutionProvider_npu`), but the rest of the eval tooling keys the same EP/device concept by the short alias: the pipeline catalog uses `qnn_npu`, and the build-only subdir label is `f"{ep or ''auto''}_{device}"` (run_eval.py ~L1212). Two conventions for the same concept is easy to get wrong. Worth standardizing on one form and centralizing the key builder in a single helper.

scripts/e2e_eval/utils/registry.py
```diff
@@ -75,6 +76,7 @@ def load_registry(path: Path) -> list[ModelEntry]:
                 downloads=item.get("downloads", 0) or 0,
                 last_update_time=item.get("last_update_time"),
                 optimum_supported=item.get("optimum_supported", False),
+                op_tracing_targets=item.get("op_tracing_targets", []) or [],
```
DingmaomaoBJTU: These target strings are matched by exact equality with no validation. Since `winml perf` hard-fails with a `ClickException` when op-tracing is enabled on anything other than QNN NPU at `basic` (perf.py ~L2012-2020), a target that matches a non-NPU device (e.g. adding `QNNExecutionProvider_gpu`) would turn a passing perf run into a failure, while a typo or wrong case silently disables the feature. Consider validating entries against the known EP/device set when loading the registry.

scripts/e2e_eval/run_eval.py
xieofxie: yes, we should document that ep and device should both be explicitly set

scripts/e2e_eval/testsets/models_all.json
xieofxie: yes, we could split and then normalize the input op_tracing_targets

scripts/e2e_eval/utils/registry.py
xieofxie: we only added validated ones

scripts/e2e_eval/run_eval.py
xieofxie: Documented in the `--op-tracing` help and `_resolve_op_tracing` docstring that auto-enable requires both `--ep` and `--device` set explicitly — `auto` is intentionally not resolved to a concrete device here, so it won't match a device-specific target. Added a unit test covering the `device=auto` no-match case.

scripts/e2e_eval/run_eval.py
xieofxie: Added parametrized tests in tests/unit/eval/test_run_eval_script.py (TestResolveOpTracing / TestOpTracingTargetKey): disabled-overrides-opt-in, explicit basic/detail, unset auto-enable, no-match, device=auto no-match, and no-ep.

scripts/e2e_eval/testsets/models_all.json
xieofxie: Standardized on the short alias: models_all.json now uses `qnn_npu` and the loader normalizes each entry to the canonical form via a single shared helper (registry.op_tracing_target_key / normalize_op_tracing_target), which run_eval reuses. Both the alias and the full name map to `QNNExecutionProvider_npu`.

scripts/e2e_eval/utils/registry.py
xieofxie: Per discussion we only add validated targets, and load-time normalization now makes matching case/alias-insensitive (qnn_npu, QNN_NPU, QNNExecutionProvider_npu all canonicalize the same), so typos/case no longer silently disable the feature.

scripts/e2e_eval/utils/registry.py
```diff
@@ -29,12 +29,50 @@ class ModelEntry:
...
+    explicit EP). Callers should pass a concrete device — ``auto`` is kept
+    verbatim and will not match a device-specific target such as
+    ``QNNExecutionProvider_npu``.
+    """
+    if not ep:
+        return None
+    return _canonical_ep_device_key(ep, device)
+
+
+def normalize_op_tracing_target(target: str) -> str:
```
DingmaomaoBJTU, line 63: Nice — normalizing alias vs full-name and case here removes most of the footguns from the previous form. One gap remains from the earlier validation point: `normalize_ep_name` returns unknown EPs verbatim (see its "let validation catch invalid names" branch), so this normalizer doesn't reject bad entries. Two latent cases:

- A typo like `qnnn_npu` normalizes to `qnnn_npu`, silently never matches, and auto-enable quietly stays off.
- A supported-looking but wrong device such as `qnn_gpu` normalizes cleanly to `QNNExecutionProvider_gpu`; if a run ever targets `--ep qnn --device gpu`, it would match and enable tracing, and `winml perf` then hard-fails with a ClickException (op-tracing is QNN-NPU/basic only). That turns a passing perf run into a failure.

Non-blocking since every current target is `qnn_npu` and CI pins `--device npu`, but validating each entry against the known EP/device set (or at least the op-tracing-supported set) at load time would make this self-defending. Up to you.

DingmaomaoBJTU (commented):
Re-reviewed the follow-up commit (bbef45c). Thanks for the quick turnaround — all four points are addressed well:

1. **device=auto gap** — not silently ignored anymore: documented in the `_resolve_op_tracing` docstring, the `--op-tracing` help text, and pinned by `test_unset_device_auto_does_not_match`. Reasonable to require explicit `--ep`/`--device` rather than guessing.
2. **Tests** — `TestResolveOpTracing` + `TestOpTracingTargetKey` cover disabled-overrides-optin, explicit levels, matching/non-matching, auto-device, and no-ep. I ran the file locally: 57 passed, ruff clean.
3. **Naming** — centralizing on `op_tracing_target_key` / `_canonical_ep_device_key` and accepting both `qnn_npu` and the full form (JSON now uses the short alias, consistent with the pair catalog) resolves the two-conventions concern. Key builder and normalizer share one code path, so they can't drift.
4. **Validation** — normalization handles case/alias variance; one residual gap noted inline (non-blocking).

Looks good to merge once you've had a look at the inline note.

DingmaomaoBJTU (approved):
Approving. All four review points from the prior round are addressed; op-tracing runs as a separate additive pass so it does not skew the benchmark, `disabled` gives a clean override, and the key builder/normalizer share one code path. Verified locally: 57 tests pass in tests/unit/eval/test_run_eval_script.py and ruff is clean. The remaining registry-validation note is non-blocking (every current target is qnn_npu and CI pins --device npu).

PR #1034, opened 2026-07-03
refactor(cli): centralize -m/--model classification with classify_model_input

### LINKED ISSUE #959: chore: resolving of model flag should be unified
Currently there are complex ones like _resolve_model_path in eval, and simple ones like check path in perf

xieofxie:
Close https://github.com/microsoft/winml-cli/issues/959

## Summary

Introduces a centralized `classify_model_input()` in `utils/cli.py` and adopts it across the CLI commands so all commands share one existence-first classification of the `-m/--model` input (ONNX file vs HuggingFace id vs folder), replacing per-command ad-hoc discriminators.

## Classifier

`classify_model_input(raw) -> ModelInput` with `ModelInputKind` = `HF_ID | ONNX_FILE | FOLDER`:
- **Not on local disk:** `.onnx` suffix → raises "ONNX file not found" (#553 guard); path-shaped → raises "path does not exist"; valid HF-id format → `HF_ID`; else → raises "not a valid HuggingFace id".
- **Exists + file:** `.onnx` → `ONNX_FILE`; else → raises "Unsupported model file".
- **Exists + dir:** `FOLDER` with flags `is_hf_folder` (`config.json` present), `folder_has_onnx` (`*.onnx`), `is_winml_cli_folder` (`folder_has_onnx` and `*build_manifest.json`).

## Wiring

Adopted in `build.py`, `config.py`, `perf.py`, `eval.py`, `inspect.py` — each classifies once and threads the result down. eval converts the classifier's `UsageError`→`BadParameter` and inspect converts it→`ClickException` to preserve each command's existing error contract.

## Behavior change

A missing `.onnx` path now raises a clean error everywhere instead of silently falling through to HuggingFace loading. Three tests that encoded the old fallthrough were updated to the unified messages.

## Notes

Library dispatchers (`inference/engine.py`, `models/auto.py`) and `perf.py._load_model` are intentionally left unwired — they must not raise click exceptions (layering).

Pushing to let CI run the full suite with timeouts, since the local combined run appeared to hang.

### DIFF
--- a/src/winml/modelkit/utils/cli.py
+++ b/src/winml/modelkit/utils/cli.py
@@ -8,6 +8,9 @@ from __future__ import annotations
 
 import json
 import os
+import re
+from dataclasses import dataclass
+from enum import Enum
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, Literal, TypeAlias, TypeVar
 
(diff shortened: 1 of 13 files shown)


src/winml/modelkit/utils/cli.py
```diff
@@ -1028,14 +1031,116 @@ def load_build_config(config_path: Path) -> tuple[WinMLBuildConfig, dict]:
...
+
+    if path.is_file():
+        if path.suffix.lower() == ".onnx":
+            return ModelInput(kind=ModelInputKind.ONNX_FILE, raw=raw)
+        raise click.UsageError(
+            f"Unsupported model file: {raw} (expected a .onnx file or a directory)."
+        )
+
+    # Existing directory: report what it contains; callers apply policy.
+    folder_has_onnx = any(path.glob("*.onnx"))
```
timenick, line 1135: These folder flags (`is_hf_folder`, `folder_has_onnx`, `is_winml_cli_folder`) and the `FOLDER` kind aren't read by any wired command yet — build/config/perf/eval/inspect all branch only on `kind is ONNX_FILE`, so HF_ID vs FOLDER is never distinguished downstream and the flags are exercised only by the unit tests. The docstring frames them as groundwork for run/serve; is that the intent (wired in a follow-up)? If so, no change needed. If not, this could be trimmed until a caller needs it, so tested-but-unused public API doesn't read as already-in-use.

zhenchaoni (approved):

src/winml/modelkit/utils/cli.py
xieofxie: updated in one place. During investigation, run / server needs this separation but it is not in command but deep into implementation, so not applied

timenick (approved):

PR #1037, opened 2026-07-03
Fix composite model export (Qwen3 and seq2seq models)

xieofxie:
## Summary

Refactors `winml export` to support **composite (seq2seq) models** such as Qwen3, T5, and BART. Previously, running `winml export` on a composite model would fail because the command treated every model as a single-component export. Now, composite models are automatically detected and each sub-component (e.g. encoder / decoder / decoder-with-past) is exported separately into its own ONNX file.

### `src/winml/modelkit/loader/resolution.py`
- Added `resolve_composite_components()` — a shared entry point that resolves a composite model's `_SUB_MODEL_CONFIG` (sub-name → task mapping). Works with both explicit `--task` and auto-detected task paths.

### `src/winml/modelkit/commands/export.py`
- Extracted per-component export logic into `_run_component_export()` inner function (I/O tensor resolution, config assembly, model load, and ONNX export).
- Added composite detection: when a model is composite, the command fans out into one export per sub-component, writing each to a stem-suffixed flat path `<output_stem>_<sub_name>.onnx` (matching the sibling `winml config` fan-out and the runtime's `*_model.onnx` discovery).
- Non-composite models continue through the single-export path unchanged.
- Intentional loud guards during composite resolution (empty registry `RuntimeError` / model-task incompatibility `ValueError`) are re-raised (the latter as a `UsageError`, mirroring `winml config`) instead of being masked as "not composite".
- Each composite sub-component is loaded and exported independently because a composite's sub-tasks map to distinct model classes/heads; the per-component reload is intentional and documented inline.

### `src/winml/modelkit/commands/config.py`
- Switched to the new shared `resolve_composite_components()` helper, removing duplicated resolution logic.

### `src/winml/modelkit/export/htp/exporter.py`
- Added `clear_sidecar()` method to `HTPExporter` for cleaning up external data sidecar files after export.

### `src/winml/modelkit/onnx/__init__.py`
- Exported the new `clear_sidecar` symbol.

### Tests
- Added unit tests for composite export fan-out (flat stem-suffixed layout), the `--input-specs` rejection, that resolution errors surface instead of being swallowed, and `resolve_composite_components`.

### DIFF
--- a/src/winml/modelkit/commands/export.py
+++ b/src/winml/modelkit/commands/export.py
@@ -24,7 +24,6 @@ Examples:
 
 from __future__ import annotations
 
-import contextlib
 import json
 import logging
 from pathlib import Path
@@ -59,19 +58,31 @@ def _delete_onnx_with_external_data(onnx_path: Path) -> None:
         logger.debug("Could not parse external data from %s", onnx_path, exc_info=True)
 
     if onnx_path.exists():
-        with contextlib.suppress(OSError):
-            onnx_path.unlink()
+        onnx_path.unlink()
 
 
-def _cleanup_partial_composite(written: list[Path]) -> None:
-    """Remove sub-model ONNX outputs written by a composite export that failed mid-run.
+def _warn_partial_composite(completed: list[Path]) -> None:
+    """Warn that a composite export failed mid-run, listing what was written.
 
-    Best-effort: a failed run should not leave a half-exported composite on disk.
-    Each path is deleted together with its external-data sidecars.
+    We deliberately do NOT delete anything: the targets may be pre-existing files
+    the user chose to ``--overwrite``, and a component can fail before touching its
+    file, so auto-deleting could destroy artifacts this run never actually wrote.
+    Instead we surface the completed sub-models and let the user decide whether to
+    keep or remove the partial composite.
     """
-    for onnx_path in written:
-        _delete_onnx_with_external_data(onnx_path)
-        logger.debug("Cleaned up partial composite sub-model: %s", onnx_path)
+    if not completed:
+        return
+    console.print(
+        "\n[yellow]Warning:[/yellow] composite export did not finish; "
+        f"{len(completed)} sub-model(s) were written/updated by this run:"
+    )
+    for onnx_path in completed:
+        console.print(f"  • {onnx_path}")
+    console.print(
+        "[yellow]The export did not complete for every sub-model.[/yellow] "
+        "Review these files and remove them if you don't want to keep the "
+        "partial export."
+    )
 
 
 @click.command()
@@ -507,20 +518,22 @@ def export(
             for sub_out in sub_outputs.values():
                 cli_utils.guard_output(sub_out, overwrite)
 
-            # If a component fails, remove the sub-models this invocation already
-            # wrote (plus the partial output of the failing one) so we don't leave
-            # a half-exported composite behind.
-            written: list[Path] = []
+            # Track sub-models this invocation actually completes. On a mid-run
+            # failure we do NOT delete anything (the targets may be pre-existing
+            # files the user chose to --overwrite, and a component can fail before
+            # touching its file). Instead we warn and list what was written so the
+            # user decides whether to keep or remove the partial composite.
+            completed: list[Path] = []
             try:
                 for name, component_task in components.items():
                     sub_out = sub_outputs[name]
                     console.print(
                         f"\n[bold blue]Sub-model:[/bold blue] {name} (task={component_task})"
                     )
-                    written.append(sub_out)
                     _run_component_export(component_task, sub_out)
+                    completed.append(sub_out)
             except BaseException:
-                _cleanup_partial_composite(written)
+                _warn_partial_composite(completed)
                 raise
         else:
             cli_utils.guard_output(output_path, overwrite)
--- a/src/winml/modelkit/commands/export.py
+++ b/src/winml/modelkit/commands/export.py
@@ -431,15 +431,27 @@ def export(
                 console.print(f"  JSON: {json_metadata}")
 
     # Detect a composite pipeline (registry-driven). A composite fans out into one
-    # ONNX per sub-component written to <output>/<component>/model.onnx; a plain
-    # model exports to the single output path as before. Detection is best-effort:
-    # if the HF config can't be resolved we fall through to the single-model path.
+    # ONNX per sub-component, each written next to <output> with a _<component>
+    # stem suffix; a plain model exports to the single output path as before.
+    # Detection is best-effort for unresolvable HF configs (we fall through to the
+    # single-model path), but intentional loud guards (empty registry / model-task
+    # incompatibility) are re-raised below rather than masked.
     components = None
     try:
         from ..loader.resolution import resolve_composite_components
 
         components = resolve_composite_components(model, task=task)
     except click.ClickException:
+        raise
+    except ValueError as e:
+        # A genuine (model, task) incompatibility, or an ambiguous composite that
+        # needs an explicit --task, is surfaced as a usage error (mirroring
+        # `winml config`) instead of being silently downgraded to a single export.
+        raise click.UsageError(str(e)) from e
+    except RuntimeError:
+        # The empty-COMPOSITE_MODEL_REGISTRY guard is meant to fail loudly (a
+        # registrations moved/renamed refactor mistake); never mask it as
+        # "not composite".
         raise
     except Exception as e:
         logger.debug("Composite detection unavailable, treating as single model: %s", e)
@@ -455,10 +467,18 @@ def export(
                     "you need custom input specs."
                 )
             console.print(
-                f"[dim]Composite model: {len(components)} sub-models -> {output_path}/[/dim]"
+                f"[dim]Composite model: {len(components)} sub-models "
+                f"(fanned out from {output_path.name} with a _<component> suffix)[/dim]"
             )
+            # Fan out flat, suffixing the output stem per component to match the
+            # sibling `winml config` layout and the runtime's `*_model.onnx`
+            # discovery: e.g. `model.onnx` -> `model_decoder_prefill.onnx`.
+            # Each component is loaded and exported independently because a
+            # composite's sub-tasks map to distinct model classes/heads (e.g.
+            # decoder-prefill vs. decoder-with-past), so the full model is
+            # reloaded per component — this N-load cost is intentional.
             for name, component_task in components.items():
-                sub_out = output_path / name / "model.onnx"
+                sub_out = output_path.with_stem(f"{output_path.stem}_{name}")
                 cli_utils.guard_output(sub_out, overwrite)
                 console.print(f"\n[bold blue]Sub-model:[/bold blue] {name} (task={component_task})")
                 _run_component_export(component_task, sub_out)
--- a/src/winml/modelkit/commands/export.py
+++ b/src/winml/modelkit/commands/export.py
@@ -237,14 +237,9 @@ def export(
     if export_config:
         console.print(f"[bold blue]Export config:[/bold blue] {export_config}")
 
-    # Refuse to clobber an existing output unless the user opted in.
     output_path = Path(output)
-    cli_utils.guard_output(output_path, overwrite)
 
-    # Create output directory if needed
-    output_path.parent.mkdir(parents=True, exist_ok=True)
-
-    # Load export configuration from JSON file if provided, or create default
+    # Load export configuration from JSON file if provided (task-independent).
     export_config_dict: dict = {}
     if export_config:
         try:
@@ -255,17 +250,7 @@ def export(
             console.print(f"[bold red]Failed to load export config:[/bold red] {e}")
             raise click.ClickException(f"Failed to load export config: {e}") from e
 
-    # Load input/output specifications.
-    #
-    # We ALWAYS run Optimum auto-resolution because it provides authoritative
-    # output_tensors (names that match the actual ONNX graph). --input-specs
-    # then overrides input_tensors only; output_tensors stays from Optimum so
-    # tasks like feature-extraction don't trip torch.onnx.export with extra
-    # dataclass field names that aren't in the traced graph.
-    input_tensors: list[InputTensorSpec] | None = None
-    output_tensors: list[OutputTensorSpec] | None = None
-
-    # Load shape overrides from JSON
+    # Load shape overrides from JSON (task-independent).
     shape_overrides = None
     if shape_config:
         try:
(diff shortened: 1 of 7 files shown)


src/winml/modelkit/commands/export.py
```diff
@@ -446,6 +430,44 @@ def export(
...
+                raise click.UsageError(
+                    "--input-specs is not supported for composite models; each sub-model "
+                    "resolves its own I/O. Export a sub-model individually with --task if "
+                    "you need custom input specs."
+                )
+            console.print(
+                f"[dim]Composite model: {len(components)} sub-models -> {output_path}/[/dim]"
+            )
+            for name, component_task in components.items():
+                sub_out = output_path / name / "model.onnx"
```
DingmaomaoBJTU: `--output` is documented as a file (`Output ONNX file path (e.g., model.onnx)`), but for a composite it's silently reinterpreted as a *directory* here: `output_path / name / "model.onnx"`. So `winml export -m Qwen/Qwen3-0.6B -o qwen.onnx` writes `qwen.onnx/decoder_prefill/model.onnx` — a directory literally named `qwen.onnx`.

Two concerns:
- It contradicts the PR description, which says composites write to `<output_dir>/<sub_name>_model.onnx`.
- It diverges from the sibling `winml config` fan-out, which keeps `--output` a file and suffixes the stem (`output.with_stem(f"{output.stem}_{component_name}")` -> `build_<component>.json`). The runtime's own multi-model discovery (`_find_build_artifacts`) also expects a flat `*_model.onnx` layout, not `<name>/model.onnx` subdirs.

Suggest committing to one convention: either document + validate that `--output` must be a directory for composites, or fan out flat (`output.with_stem(f"{stem}_{name}")`) to match `config` and the runtime loader.

src/winml/modelkit/commands/export.py
```diff
@@ -446,6 +430,44 @@ def export(
...
+    # model exports to the single output path as before. Detection is best-effort:
+    # if the HF config can't be resolved we fall through to the single-model path.
+    components = None
+    try:
+        from ..loader.resolution import resolve_composite_components
+
+        components = resolve_composite_components(model, task=task)
+    except click.ClickException:
+        raise
+    except Exception as e:
```
DingmaomaoBJTU, line 487: This broad `except Exception` also swallows the intentional `RuntimeError` that `_composite_registry()` raises when `COMPOSITE_MODEL_REGISTRY` is empty. That guard was added specifically to be *loud* — its docstring says it "turns a registrations moved/renamed refactor mistake into a loud failure instead of silently disabling the composite feature." Catching it here downgrades it to a `debug` log and silently exports a composite as a single model.

Same for a genuine `(model, task)` incompatibility surfaced during resolution: it gets swallowed and mis-routed to the single-model path, whereas `winml config` reports it as a `UsageError`. Consider narrowing the except (e.g. let `RuntimeError` / `ValueError` propagate) so real failures aren't masked as "not composite".

DingmaomaoBJTU (commented):
Reviewed the composite-export fan-out and the external-data cleanup. The refactor is clean overall — extracting `_run_component_export` and the shared `resolve_composite_components` reads well, and `_cleanup_stale_external_data` correctly keys off the *final* on-disk references, so it's safe whether the consolidated re-save ends up inline or external.

A few things worth considering (see inline):
1. For composites, `--output` is silently treated as a directory, which is inconsistent with the `winml config` fan-out convention and with this PR's own description.
2. The broad `except Exception` around composite detection swallows the intentional loud `RuntimeError` from `_composite_registry()`.
3. The full model is reloaded once per sub-component.

Nits (no inline):
- The PR description says a `clear_sidecar()` method and a `clear_sidecar` export were added, but the code actually adds `HTPExporter._cleanup_stale_external_data()` and exports `get_external_data_files`. Please sync the description.
- The composite `--input-specs` `UsageError` is raised *after* the `Starting HTP export...` banner prints; hoisting the check above the banner would read cleaner.
- The new composite tests mock `load_hf_model` / `resolve_export_config` / `export_onnx`, so they verify the fan-out plumbing (paths, call counts) but not that the sub-models actually differ — worth a follow-up integration check.

(Note: I couldn't run the suite locally — this machine is `win_arm64` and torch ships no matching wheel, so this is a static review.)

src/winml/modelkit/commands/export.py
```diff
@@ -401,20 +288,126 @@ def export(
...
+            cfg = WinMLExportConfig.from_dict(config_kwargs)
+        except Exception as e:
+            console.print(f"[bold red]Configuration error:[/bold red] {e}")
+            logger.exception("Failed to create export config")
+            raise click.ClickException(f"Configuration error: {e}") from e
 
         # Load model with task detection (CLI is the orchestration layer)
-        pytorch_model, _, detected_task = load_hf_model(model, task=task)
-        if task:
+        pytorch_model, _, detected_task = load_hf_model(model, task=component_task)
```
DingmaomaoBJTU, line 426: For an N-component composite this loads the full model once per component (and `resolve_export_config` re-reads the config each time, on top of the initial `resolve_composite_components` call). For large seq2seq checkpoints that's N full loads.

Is loading once and reusing across sub-tasks feasible, or does each sub-task genuinely need a distinct model class/head? If the latter, a short comment noting the cost is intentional would help.

src/winml/modelkit/commands/export.py
xieofxie: yes you are right. we should follow config to generate with stem

xieofxie: Done — switched to a flat, stem-suffixed layout matching `winml config`: output.with_stem(f"{output.stem}_{name}"), so -o qwen.onnx now writes qwen_decoder_prefill.onnx / qwen_decoder_gen.onnx (no directory named after the file). Updated the PR description and the composite test accordingly.

src/winml/modelkit/commands/export.py
xieofxie: Good catch — narrowed the except so the intentional loud guards propagate: RuntimeError (empty COMPOSITE_MODEL_REGISTRY) is re-raised as-is, and a genuine (model, task) ValueError is surfaced as a click.UsageError (mirroring winml config). Only truly unresolvable HF-config errors still fall through to the single-model path. Added a test covering the ValueError case.

src/winml/modelkit/commands/export.py
xieofxie: Each sub-task genuinely needs a distinct model class/head (e.g. decoder-prefill vs. decoder-with-past), so the per-component reload is required rather than incidental. Added an inline comment documenting that the N full loads are intentional.

src/winml/modelkit/commands/export.py
```diff
@@ -446,6 +430,64 @@ def export(
...
+        # A genuine (model, task) incompatibility, or an ambiguous composite that
+        # needs an explicit --task, is surfaced as a usage error (mirroring
+        # `winml config`) instead of being silently downgraded to a single export.
+        raise click.UsageError(str(e)) from e
+    except RuntimeError:
+        # The empty-COMPOSITE_MODEL_REGISTRY guard is meant to fail loudly (a
+        # registrations moved/renamed refactor mistake); never mask it as
+        # "not composite".
+        raise
+    except Exception as e:
```
DingmaomaoBJTU, line 487: This catch-all turns any unexpected failure in `resolve_composite_components()` into “not composite” and proceeds with a single-model export. For composite-capable models, that can silently produce the wrong artifact shape instead of failing. Can we only suppress explicitly expected non-composite/unavailable cases and re-raise unexpected exceptions, or wrap them as a `ClickException`?

src/winml/modelkit/commands/export.py
```diff
@@ -446,6 +430,64 @@ def export(
...
+            # Fan out flat, suffixing the output stem per component to match the
+            # sibling `winml config` layout and the runtime's `*_model.onnx`
+            # discovery: e.g. `model.onnx` -> `model_decoder_prefill.onnx`.
+            # Each component is loaded and exported independently because a
+            # composite's sub-tasks map to distinct model classes/heads (e.g.
+            # decoder-prefill vs. decoder-with-past), so the full model is
+            # reloaded per component — this N-load cost is intentional.
+            for name, component_task in components.items():
+                sub_out = output_path.with_stem(f"{output_path.stem}_{name}")
+                cli_utils.guard_output(sub_out, overwrite)
```
DingmaomaoBJTU, line 519: Because this guard runs inside the component loop, an earlier component can already be exported before a later component path fails the overwrite check, leaving a partial composite export even though `--overwrite` was not set. Can we precompute all `sub_out` paths and guard them all before starting any export work? If export can fail mid-run, it may also be worth cleaning up outputs written by this invocation.

tests/unit/commands/test_export.py
```diff
@@ -795,3 +795,109 @@ def test_export_inherits_debug_mode(
...
+            mock_load.return_value = (MagicMock(), None, "text2text-generation")
+            result = runner.invoke(
+                export,
+                ["--model", "Qwen/Qwen3-0.6B", "--output", str(output_path)],
+                obj={"debug": False},
+            )
+
+        assert result.exit_code == 0, result.output
+        # One export per component, each to <output-stem>_<name>.onnx (flat layout).
+        assert mock_export_onnx.call_count == len(components)
```
DingmaomaoBJTU, line 843: This verifies the fan-out count and output paths, but not that each component’s task is actually propagated. A regression could still pass this test while calling `resolve_export_config()`, `load_hf_model()`, or `export_pytorch()` with the outer task/`None` for every sub-model. Can we assert those calls receive the expected task per component?

tests/unit/commands/test_export.py
```diff
@@ -795,3 +795,109 @@ def test_export_inherits_debug_mode(
...
+        ):
+            result = runner.invoke(
+                export,
+                ["--model", "Qwen/Qwen3-0.6B", "--output", str(tmp_path / "qwen3.onnx")],
+                obj={"debug": False},
+            )
+
+        # Must not silently fall back to a single-model export.
+        assert result.exit_code != 0
+        assert "multiple composite exports" in result.output
```
DingmaomaoBJTU, line 923: This would still pass if the `ValueError` were caught by the generic export wrapper and printed as `Export failed: ...`. Since the behavior being protected is surfacing the resolver error as a usage error, can we also assert `"Export failed" not in result.output`?

tests/unit/export/test_pytorch_export.py
```diff
@@ -359,3 +359,72 @@ def test_mismatched_input_order_exports_successfully(self, tmp_path) -> None:
...
+        assert referenced, "expected the consolidated model to reference a .data sidecar"
+
+        # Simulate leftover per-tensor sidecars from the raw TorchScript export.
+        orphans = ["onnx__MatMul_1", "linear.weight", "linear.bias"]
+        for name in orphans:
+            (tmp_path / name).write_bytes(b"stale-weights")
+
+        # The exporter records raw sidecars *plus* whatever the consolidated save
+        # references; only the unreferenced ones must be deleted.
+        HTPExporter._cleanup_stale_external_data(str(model_path), orphans + referenced)
```
DingmaomaoBJTU, line 415: This exercises `_cleanup_stale_external_data()` directly, but it would still pass if `HTPExporter.export()` stopped recording the pre-consolidation sidecars or stopped invoking cleanup after tag injection. Can we add an export-path test around `export_pytorch()`/`HTPExporter.export()` that verifies the final output directory contains only the ONNX file and sidecars still referenced by `get_external_data_files()`?

DingmaomaoBJTU (commented):
I left a few inline comments around composite export error handling, output preflight, and test coverage.

src/winml/modelkit/commands/export.py
xieofxie: Narrowed the handling: only the expected "unavailable" case (OSError — the reference isn't a resolvable HF config, e.g. not a hub id / no local config.json) falls through to single-model. ValueError -> UsageError, the empty-registry RuntimeError re-raises as-is, and anything else is now surfaced as a ClickException ("Composite model detection failed unexpectedly: ...") rather than silently producing a single-model artifact. Added a test for the unexpected-error case.

src/winml/modelkit/commands/export.py
xieofxie: Good point. All sub_out paths are now precomputed and every one is passed through guard_output before any export starts, so an overwrite collision on a later component can no longer leave an earlier one written. I also added best-effort cleanup: if a component fails mid-run, the sub-models this invocation already wrote (plus the partial output of the failing one) are deleted along with their external-data sidecars, so a failure never leaves a half-exported composite. Covered by a new test.

tests/unit/commands/test_export.py
xieofxie: Added assertions that each component's own task is propagated: the test now echoes the per-component task back through load_hf_model and verifies resolve_export_config, load_hf_model, and export_onnx each receive the expected task per sub-model (mapping output path -> task), so a regression that passes the outer task/None would fail.

tests/unit/commands/test_export.py
xieofxie: Added assert "Export failed" not in result.output so the test also guards that the resolver error is surfaced as a usage error and not swallowed then re-wrapped by the generic export handler.

tests/unit/export/test_pytorch_export.py
xieofxie: Added an export-path test (test_export_path_leaves_only_referenced_sidecars) that drives the real HTPExporter.export() via export_pytorch, with the raw ONNX step patched to emit per-tensor external sidecars (as the TorchScript exporter does). It asserts that after export no orphaned per-tensor sidecars remain — every external-data file left in the directory is still referenced by get_external_data_files(). This fails if export() stops recording pre-consolidation sidecars or stops invoking cleanup.

src/winml/modelkit/commands/export.py
```diff
@@ -446,6 +443,91 @@ def export(
...
+            # wrote (plus the partial output of the failing one) so we don't leave
+            # a half-exported composite behind.
+            written: list[Path] = []
+            try:
+                for name, component_task in components.items():
+                    sub_out = sub_outputs[name]
+                    console.print(
+                        f"\n[bold blue]Sub-model:[/bold blue] {name} (task={component_task})"
+                    )
+                    written.append(sub_out)
```
DingmaomaoBJTU: This adds `sub_out` to `written` before `_run_component_export()` has actually written anything. With `--overwrite`, the target may already exist, and `_run_component_export()` can fail during config resolution or model loading before touching the file; the cleanup path would then delete that pre-existing ONNX and its sidecars. Can we track only outputs created/modified by this invocation, or snapshot pre-existing paths/sidecars and skip deleting unchanged artifacts?

DingmaomaoBJTU (commented):
I left one follow-up inline comment on the latest update.

DingmaomaoBJTU (approved):
Approved with the follow-up cleanup comment left as non-blocking.

src/winml/modelkit/commands/export.py
xieofxie: Good call! if onnx exist and overwrite is set, we will overwrite one by one. If fail in the middle, we may warn user that xxx files are updated but given the whole export failed, user should decide whether removed them or keep them

xieofxie: Good catch — you're right that appending before the write (and blindly deleting) could destroy a pre-existing file the user chose to --overwrite. Reworked per @xieofxie's suggestion: instead of deleting anything, on a mid-run failure we now warn and list the sub-models that were written/updated by this run, and tell the user to review/remove them if they don't want the partial export. A sub_out is only recorded as "completed" after _run_component_export returns, so a component that fails during config resolution / model loading contributes nothing to delete or mis-report. Updated the test to assert the completed file is kept and the warning is shown.

xieofxie:
Failure log will be like

```
Success! Model exported to: temp\export_qwen3_decoder_prefill

Warning: composite export did not finish; 1 sub-model(s) were written/updated by this run:
  • temp\export_qwen3_decoder_prefill
The export did not complete for every sub-model. Review these files and remove them if you don't want to keep the partial export.

Export failed: No active exception to reraise
[14:34:44 ERROR   winml.modelkit.commands.export] Export failed: No active exception to reraise
Error: Export failed: No active exception to reraise
```

DingmaomaoBJTU (approved):

src/winml/modelkit/commands/export.py
```diff
@@ -446,6 +454,93 @@ def export(
...
+            # user decides whether to keep or remove the partial composite.
+            completed: list[Path] = []
+            try:
+                for name, component_task in components.items():
+                    sub_out = sub_outputs[name]
+                    console.print(
+                        f"\n[bold blue]Sub-model:[/bold blue] {name} (task={component_task})"
+                    )
+                    _run_component_export(component_task, sub_out)
+                    completed.append(sub_out)
```
DingmaomaoBJTU, line 534: `completed.append(sub_out)` only runs after `_run_component_export()` returns successfully, but the exporter can already have written or overwritten `sub_out` before a later step fails. For example, `HTPExporter.export()` writes the ONNX, then still verifies, reloads, tags, embeds metadata, saves, and cleans up sidecars. If one of those later steps raises, the partial-export warning omits a file touched by this run; if the first component fails after writing, no partial warning is emitted at all.

Can we track the current attempted/touched output as well, or include `sub_out` on exception when it exists or changed from its pre-export state before calling `_warn_partial_composite()`?

PR #1040, opened 2026-07-03
Warn users about silent dataset downloads and show live progress

### LINKED ISSUE #444: Warn before large downloads in long-running commands (build, quantize) + live elapsed-time counter
## Description

Long-running commands that trigger large downloads (calibration datasets, model weights) start the download silently with no upfront warning, no size estimate, and no time estimate. A developer on slow or metered connectivity has no chance to abort before the transfer is underway.

**Confirmed reproduction with `winml build` + calibration dataset:**

`winml build` downloads the entire `timm/mini-imagenet` dataset (~7 GB — 13 train + 3 validation + 2 test parquet files) even though the config specifies `"samples": 10`. Quantization took **896 s (~15 min)**, almost entirely spent downloading. The user sees no warning and no estimated time before this begins.

## Steps to Reproduce

```bash
winml build -c config.json -m ProsusAI/finbert -o output/
```

Where `config.json` specifies:
```json
{ "quant": { "dataset_name": "timm/mini-imagenet", "samples": 10 } }
```

## Expected Behavior

Before any large download begins, the CLI prints a warning with size and estimated time:

```
⚠  Downloading calibration dataset timm/mini-imagenet (~7.0 GB).
   Estimated time on 10 Mbps: ~95 min  |  100 Mbps: ~10 min
   Press Ctrl+C to cancel.
```

If the dataset size cannot be determined ahead of time, at minimum print the dataset name and that it may be large, before streaming begins.

The same pre-download warning should apply to:
- Model weight downloads (`winml build -m <huggingface_id>`)
- Any other command that triggers a network fetch > a configurable threshold (e.g., 500 MB)

## Actual Behavior

No warning is printed. The download starts immediately and silently inside the quantize `StageLive` block. The first visible signal is the spinner; the user has no indication of how long it will run or how much data will be transferred.

## Root Cause (initial analysis)

The `datasets` library streams or caches parquet shards for the full dataset split regardless of how many samples are consumed downstream. The `quantize_onnx` call in `_run_quantize_stage` ([build.py](src/winml/modelkit/commands/build.py)) does not query dataset size before fetching, and `StageLive` suppresses `datasets` progress bars to keep the display clean — removing the only secondary signal the user might have seen.

Two independent fixes are needed:

1. **Pre-download warning** (UX): Query the Hugging Face Hub API for dataset / model size before fetching and print a structured warning with size + time estimate. Block for 3 s (or until Ctrl+C) to give the user a chance to abort.
2. **Lazy / partial download** (efficiency): Investigate whether `datasets` streaming mode or shard-level access can be used to fetch only the N calibration samples without pulling all parquet files first. If feasible, this eliminates the problem for the `samples`-bounded case entirely.

## Environment

- OS: Windows 11
- ModelKit version: latest main (post #334)
- Network: WiFi (variable)

## Additional Context

This issue affects any command with a slow first-run experience:

| Command | Download trigger | Typical size |
|---------|-----------------|-------------|
| `winml build` (HF model) | Model weights | 0.1 – 10 GB |
| `winml build` (calibration) | Dataset parquet shards | 1 – 50 GB |
| `winml quantize` | Same as above | 1 – 50 GB |
| `winml eval` | Eval dataset | variable |

A developer running on coffee shop WiFi or a metered mobile hotspot will abandon the tool after one silent 15-minute hang. The pre-download warning is a low-cost, high-trust fix that should be prioritized independently of the lazy-download optimization.

zhenchaoni:
Fix #444

### Problem
Large HuggingFace dataset downloads are triggered silently inside blocking quantize/eval stages. The UI would freeze for minutes with only a static spinner while GB of data downloaded in the background — users couldn't tell whether the tool was working or hung (#444).

### Changes

**Surface dataset loads as warnings**
Dataset-loading logs are promoted from `info` → `warning` at every `load_dataset` entry point, so users get a visible heads-up before a potentially large fetch begins:
- `datasets/image.py`, `datasets/image_segmentation.py`, `datasets/text.py`
- `eval/base_evaluator.py`
- Added an equivalent warning to the Qwen3 GSM8K calibration loader (`quant/calibration/qwen3_transformer_only.py`)

**Live elapsed-time counter during blocking stages**
`StageLive` (`utils/console.py`) now renders a self-updating elapsed-seconds counter on Rich's refresh thread (15 fps), so the timer keeps ticking even while the main thread is blocked in native code (downloads, quantization). Added `_t0`/`_done` tracking so `__exit__` finalizes timing if a stage exits without an explicit `set_done`.

**Clearer eval messaging**
Split the eval console output — `Loading dataset...` is now distinct from `Evaluating...`, so the download phase is attributable instead of hidden behind a combined "Loading dataset and evaluating..." message.

**Suppress cosmetic noise**
Filter the scikit-learn "number of unique classes is greater than 50% of the number of samples" `UserWarning` (`_warnings.py`), which fires from HF `evaluate` metrics on small eval sample counts and would otherwise clutter the newly surfaced output.

### Result
Users now see an upfront warning when a dataset is about to load and a continuously updating timer during long stages, resolving the "frozen with no feedback" experience described in #444.

### DIFF
--- a/src/winml/modelkit/datasets/image.py
+++ b/src/winml/modelkit/datasets/image.py
@@ -74,7 +74,7 @@ class ImageDataset(BaseTaskDataset):
         # iterate the full remote stream into memory, which is worse than a
         # bulk download.
         streaming = self._config.get("streaming", False) and self._max_samples is not None
-        logger.info(f"Loading dataset: {self._dataset_name} with split: {self._data_split}")
+        logger.warning(f"Loading dataset: {self._dataset_name} with split: {self._data_split}")
         try:
             dataset = load_dataset(
                 self._dataset_name,
--- a/src/winml/modelkit/eval/evaluate.py
+++ b/src/winml/modelkit/eval/evaluate.py
@@ -326,7 +326,7 @@ def evaluate(config: WinMLEvaluationConfig) -> EvalResult:
 
     cls = get_evaluator_class(config)
     try:
-        console.print("[bold]Loading dataset and evaluating...[/bold]")
+        console.print("[bold]Evaluating...[/bold]")
         task_evaluator = cls(config, model)
         metrics = task_evaluator.compute()
     except DatasetValidationError as error:
--- a/src/winml/modelkit/utils/console.py
+++ b/src/winml/modelkit/utils/console.py
@@ -17,6 +17,7 @@ from __future__ import annotations
 
 import functools
 import logging
+import time
 from collections.abc import Callable
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, TypeVar
@@ -355,8 +356,11 @@ class StageLive:
         self._lines: list[RenderableType] = []
         self._live: _SafeLive | None = None
         self._status_idx: int = 0
+        self._t0: float = 0.0
+        self._done: bool = False
 
     def __enter__(self) -> StageLive:
+        self._t0 = time.monotonic()
         self._lines = [self._make_running_line()]
         self._status_idx = 0
         self._live = _SafeLive(
@@ -369,6 +373,8 @@ class StageLive:
         return self
 
     def __exit__(self, *_: object) -> None:
+        if not self._done:
+            self.set_done(time.monotonic() - self._t0)
         if self._live:
             self._live.update(self._render())
             self._live.stop()
@@ -383,13 +389,22 @@ class StageLive:
 
     # ── Status line management ────────────────────────────────────
 
-    def _make_running_line(self, detail: str = "") -> Text:
-        line = Text()
-        line.append(f"{ICON_RUNNING} ")
-        line.append(self._name.capitalize(), style="bold yellow")
-        if detail:
-            line.append(f"  {detail}", style="dim")
-        return line
+    def _make_running_line(self, detail: str = "") -> RenderableType:
+        # A self-rendering line so the elapsed counter ticks on Live's refresh
+        # thread (15fps) even while the main thread blocks in native code.
+        stage = self
+
+        class _RunningLine:
+            def __rich__(self) -> Text:
+                line = Text()
+                line.append(f"{ICON_RUNNING} ")
+                line.append(stage._name.capitalize(), style="bold yellow")
+                if detail:
+                    line.append(f"  {detail}", style="dim")
+                line.append(f"  {int(time.monotonic() - stage._t0)}s", style="dim")
+                return line
+
+        return _RunningLine()
 
     def set_status(self, detail: str) -> None:
         """Update the running status text."""
@@ -403,6 +418,7 @@ class StageLive:
         line.append(f"{self._name.capitalize():<48}", style="green")
         line.append(f"{elapsed:.1f}s", style="green")
         self._lines[self._status_idx] = line
+        self._done = True
         self._update()
 
     def set_error(self, error: str = "") -> None:
@@ -413,6 +429,7 @@ class StageLive:
         if error:
             line.append(f"  {error}", style="red")
         self._lines[self._status_idx] = line
+        self._done = True
         self._update()
 
     # ── Detail lines (indented under stage) ───────────────────────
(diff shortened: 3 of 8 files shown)


src/winml/modelkit/utils/console.py
```diff
@@ -369,6 +373,8 @@ def __enter__(self) -> StageLive:
         return self
 
     def __exit__(self, *_: object) -> None:
+        if not self._done:
+            self.set_done(time.monotonic() - self._t0)
```
DingmaomaoBJTU, line 377: This finalizer throws away the exception state (`*_`), so a stage that raises **without** first calling `set_error` now gets finalized as a green ✓ with an elapsed time — printed immediately before the traceback.

Concrete paths that hit this today:
- The entire `export` stage (`build.py` ~1398–1422): if `_load_model`/`export_onnx` raise, nothing calls `set_error`, so the failed stage renders as done ✓.
- `compile` stage (`build.py` ~1307): `raise RuntimeError("Compile reported success but output not found...")` runs with no `set_error`.
- `quantize` stage (`build.py` ~1213): if `quantize_onnx` itself raises (rather than returning `success=False`), no `set_error`.

Since every success path already calls `set_done`, this `if not self._done` branch in practice only fires on uncaught exceptions — i.e. it paints failures green. Suggest keying off the exception info instead:

```python
def __exit__(self, exc_type, exc_val, exc_tb) -> None:
    if not self._done:
        if exc_type is not None:
            self.set_error()
        else:
            self.set_done(time.monotonic() - self._t0)
    if self._live:
        ...
```

src/winml/modelkit/eval/evaluate.py
```diff
@@ -326,7 +326,7 @@ def evaluate(config: WinMLEvaluationConfig) -> EvalResult:
 
     cls = get_evaluator_class(config)
     try:
-        console.print("[bold]Loading dataset and evaluating...[/bold]")
+        console.print("[bold]Evaluating...[/bold]")
```
DingmaomaoBJTU, line 329: The ordering reads backwards now: this prints `Evaluating...`, but the dataset load actually happens on the next line inside `cls(config, model)` → `WinMLEvaluator.__init__` → `prepare_data()`. So the user sees `Evaluating...` first and then the `Loading dataset:` warning underneath it. Given the PR's goal is to make the download phase attributable, consider printing a `Loading dataset...` line before constructing the evaluator (or moving this `Evaluating...` print to just before `task_evaluator.compute()`).

src/winml/modelkit/datasets/image.py
```diff
@@ -74,7 +74,7 @@ def _load_and_sample(self, revision: str | None = None) -> Any:
         # iterate the full remote stream into memory, which is worse than a
         # bulk download.
         streaming = self._config.get("streaming", False) and self._max_samples is not None
-        logger.info(f"Loading dataset: {self._dataset_name} with split: {self._data_split}")
+        logger.warning(f"Loading dataset: {self._dataset_name} with split: {self._data_split}")
```
DingmaomaoBJTU, line 77: Promoting normal, successful status to `warning` is a bit of a log-level smell — these are informational messages, not warnings. It only surfaces because the default level is WARNING (INFO is hidden), but it means every successful run emits WARNING records that CI / log processing may flag, and it diverges from how user-facing status is surfaced elsewhere via `console.print`. Same applies to the other promoted sites (`image_segmentation.py`, `text.py`, `base_evaluator.py`, `qwen3_transformer_only.py`).

Since these loaders are deep library code with no console handle, `warning` is a pragmatic call — but worth a code comment or a follow-up toward a proper user-facing status channel so log semantics stay clean.

DingmaomaoBJTU (commented):
Nice improvement to the "frozen with no feedback" UX — the live elapsed timer and the upfront dataset heads-up are both welcome. A few things worth a look before merge: one likely regression in `StageLive.__exit__` (failures can render as a green ✓), the eval message ordering, and the log-level semantics of the info→warning promotion. Details inline.

PR #1042, opened 2026-07-03
feat(perf): add --apply-template flag to control genai prompt templating

### LINKED ISSUE #1030: feat: add a apply_template flag to control whether to apply template
So user could input prompt with template to test

As a follow up of https://github.com/microsoft/winml-cli/pull/1015

DingmaomaoBJTU:
Closes #1030 (follow-up to #1015).

## Problem

`winml perf --runtime winml-genai` always wraps `--prompt` in the bundle's own chat template before timing. There was no way to benchmark a prompt that is **already** formatted (or a raw completion prompt), so users couldn't test their own templated input.

## Change

Add `--apply-template/--no-apply-template` (default **on**, preserving current behavior). With `--no-apply-template`, the prompt is encoded and timed verbatim.

- `GenaiPerfConfig.apply_template` gates `GenaiPerfBenchmark._prompt_text`; when disabled, the prompt is used as-is with no `apply_chat_template` call.
- `apply_template` is recorded in the JSON report's `benchmark_info` for reproducibility.
- The flag is wired through the `perf` command and documented in `--prompt` / `--apply-template` help text.

## Usage

```bash
# Default: prompt wrapped in the bundle's chat template
winml perf -m path/to/bundle --runtime winml-genai --prompt "Hello"

# Benchmark a pre-formatted prompt verbatim
winml perf -m path/to/bundle --runtime winml-genai \
  --no-apply-template --prompt "<|user|>Hello<|end|><|assistant|>"
```


timenick:
I’m not sure why we need this user scenario, since onnxruntime-genai already includes an apply template API for it.

src/winml/modelkit/commands/_perf_genai.py
```diff
@@ -59,6 +59,12 @@
 # ``GenaiPerfConfig.prompt`` field default (a test asserts the two stay in sync).
 _DEFAULT_PROMPT = "Explain the theory of relativity in simple terms."
 
+# Unique placeholder used to recover a bundle's chat-template wrapper text (the
+# markers it adds around the user content) by rendering the template around it.
+# Distinctive enough that a real prompt never contains it, so partitioning the
+# rendered output on it reliably isolates the leading/trailing wrapper.
+_PROMPT_SENTINEL = "\ue000__winml_prompt_sentinel__\ue000"
```
xieofxie: this looks weird to me. Maybe the code dependency logic needs to be updated to avoid such mechanism.
We could either
```
prompt = apply_template or not in perf_genai
for i in n:
    session.run(prompt)
```
or
```
session.config(apply_template or not)
for i in n:
    session.run()
```

DingmaomaoBJTU: Removed the sentinel-based heuristic in 1b78ea2f. The remaining code follows your first suggestion: `_prompt_text()` resolves the prompt once (chat-templated when `--apply-template`, verbatim when `--no-apply-template`), `run()` encodes it once before the timed loop, and each iteration just calls `generate_timed` on the pre-encoded token ids — no per-iteration templating or marker detection left.

DingmaomaoBJTU:
This is the scenario from #1030 (the issue this PR closes): benchmarking a prompt that is *already* formatted, instead of having `perf` wrap `--prompt` in the single-turn chat template on every run.

The default (`--apply-template`) still goes through ort-genai's `apply_chat_template`, so that path is unchanged. `--no-apply-template` is the opposite direction: it feeds the prompt to the tokenizer verbatim so you can time the exact token sequence you care about — a multi-turn conversation, a RAG prompt with a large retrieved context, or a custom/non-default template. Since `perf` measures prefill (TTFT) and decode throughput, the prompt's real length and shape drive the numbers; the single-turn template wrapper doesn't represent those cases. The genai `apply_chat_template` API templates a prompt, but doesn't give `perf` a way to *skip* templating and benchmark a pre-formatted one, which is what this flag adds.

src/winml/modelkit/commands/_perf_genai.py
```diff
@@ -199,6 +200,7 @@ def to_dict(self) -> dict[str, Any]:
                 "device": self.config.device,
                 "compile": self.config.compile,
                 "compile_timeout": self.config.compile_timeout,
+                "apply_template": self.config.apply_template,
```
xieofxie: Reproducibility here is only partial: `benchmark_info` now records `apply_template` but never the `prompt` string itself. The PR's rationale is recording this "for reproducibility," yet a report saying `apply_template: false` doesn't capture *what verbatim prompt* was timed — which is exactly the case where the prompt content matters most, since the user supplied their own pre-formatted string. Consider also recording `prompt` here (or a hash/length if you'd rather not bloat the report). Pre-existing gap for `prompt`, but `--no-apply-template` makes it more visible.

xieofxie: Nit (cosmetic): `apply_template` sits between `compile_timeout` and `iterations`. Grouping it next to `prompt_tokens` / `max_new_tokens` would read more naturally since it's a prompt-shaping setting.

DingmaomaoBJTU: Done in 026d8a2f — `benchmark_info` now records the verbatim `prompt` string alongside `prompt_tokens`. Went with the full string rather than a hash/length so a report is directly reproducible (you can copy the exact prompt that was timed); it's the caller's own input, and benchmark prompts are bounded. Happy to switch to a hash+length if you'd prefer to keep the report leaner.

DingmaomaoBJTU: Moved in 026d8a2f — `apply_template` now sits with the prompt-shaping fields (`max_new_tokens`, `apply_template`, `prompt`, `prompt_tokens`) instead of next to `compile_timeout`.

xieofxie (approved):

PR #1046, opened 2026-07-03
feat(genai): honor --ep/--device as an EP override for GenaiSession (#1025)

### LINKED ISSUE #1025: GenaiSession: support --ep/--device to override genai_config.json EP routing
## Context

`GenaiSession` (added in #1015) decides whether to register WinML execution providers and whether to pre-compile pipeline stages by **reading the bundle's `genai_config.json`** (per-stage `session_options.provider_options`). This is the correct default: the bundle's config is the source of truth for EP routing.

As part of #1015 review (comment from @xieofxie), the hardcoded `_NEEDS_WINML_EPS = {mixed, qnn, dml}` gate was removed in favor of config detection (`_bundle_uses_hardware_ep`).

## Problem

The public `ep` constructor argument (and the `winml perf --device` flag that feeds it) is currently **informational only** — it is validated and reported in `benchmark_info`, but it no longer drives EP registration/compilation decisions (the config does).

## Proposal

Design and implement an explicit override path so a caller can force EP routing that differs from `genai_config.json`:

- `ep` (`GenaiSession`) / `--ep` + `--device` (`winml perf --runtime winml-genai`) should be able to **override** the config-derived decision (e.g. force CPU on a QNN bundle, or force a hardware EP for a bundle whose config is CPU-only).
- Define precedence clearly: explicit arg > config default.
- Decide whether an override should also rewrite the per-stage `session_options` that `og.Model` consumes (today those are respected verbatim), and how that interacts with EPContext pre-compilation.

## Acceptance criteria

- [ ] Precedence documented and tested (arg overrides config; absent arg respects config).
- [ ] `winml perf --runtime winml-genai --device <...>` exercises the override end-to-end.
- [ ] No hardcoded EP short-name → behavior mapping (stay generic).

DingmaomaoBJTU:
## Summary

Closes #1025.

Previously `GenaiSession`'s `ep` argument (and `winml perf --runtime winml-genai`'s `--ep`/`--device`) was **informational only**: EP registration and pre-compilation were decided purely from the bundle's `genai_config.json` per-stage `session_options`. This PR adds an explicit override path so callers can force EP routing that differs from the bundle config, with precedence **explicit arg > bundle config**.

### `GenaiSession`
- **`ep` default is now `None`** (was `"cpu"`), and the `"mixed"` sentinel is retired.
  - `ep=None` — **respect the bundle config** unchanged (registration + compile derive from `genai_config.json` exactly as before).
  - `ep=<concrete EP>` — **force the whole decoder pipeline onto that EP** by rewriting every stage's `provider_options`:
    - `cpu` strips hardware providers (CPU fallback; no WinML EP registration/compile).
    - a hardware EP routes every stage to it (registers the WinML EPs it needs; pre-compiles when `compile=True`). A stage's existing options are carried over only when it already targeted the same canonical EP.
- Registration and compile gates now derive from the **effective (post-override) config** via the existing generic `_bundle_uses_hardware_ep` — no hardcoded EP short-name → behavior map.
- Because `og.Config.clear_providers/append_provider` can only touch the top-level provider (not per-stage `session_options`), an override loads `og.Model` from a derived `_compiled/` bundle. `_prepare_compiled_bundle` now writes that derived bundle for override-only passes too (not just compilation).
- Invalid `ep` values raise `ValueError` (validated against the shared EP union), so an override can never silently become a no-op.

### `perf --runtime winml-genai`
- Precedence: **explicit `--ep` > concrete `--device` > respect config**. Removed `ep` from `_GENAI_IGNORED_FLAGS` (it is now honored).
- `device_to_genai_ep` returns `EPNameOrAlias | None` — `auto` (and any unrecognized device) maps to `None` (respect config), replacing the old `auto → "mixed"`.
- Reporting (`benchmark_info.ep` / device line) shows the resolved override alias, or `"config"` when respecting the bundle.

## Design notes / compatibility
- The rewrite and validation are **generic** — EPs are compared/aliased through the shared EP union (`normalize_ep_name` / `EP_NAME_TO_ALIAS` / `EP_NAMES`); the only literal provider name is the universal `cpu` fallback.
- The always-safe override direction is `cpu`. Forcing a hardware EP onto stages the bundle did not build for it (e.g. `embeddings`/`lm_head`) may fall back to JIT/CPU inside onnxruntime-genai, which is documented in the `ep` docstring.

### DIFF
--- a/src/winml/modelkit/commands/_perf_genai.py
+++ b/src/winml/modelkit/commands/_perf_genai.py
@@ -41,7 +41,7 @@ from ..session import (
     GenaiSessionError,
     GenerationConfig,
 )
-from ..utils.constants import EPNameOrAlias
+from ..utils.constants import EP_NAME_TO_ALIAS, EPNameOrAlias
 
 
 if TYPE_CHECKING:
@@ -60,30 +60,43 @@ RUNTIME_TYPE = "winml-genai"
 # ``GenaiPerfConfig.prompt`` field default (a test asserts the two stay in sync).
 _DEFAULT_PROMPT = "Explain the theory of relativity in simple terms."
 
-# --device -> GenaiSession EP override.  The default ("auto") resolves to
-# ``None`` = *respect the bundle config*: genai bundles route each stage via its
-# own session_options in genai_config.json (e.g. ctx/iter on the NPU/QNN,
-# embeddings/lm_head on CPU), and ``None`` honors that per-stage routing while
-# still registering the WinML EPs those stages need.  A concrete device forces
-# the whole decoder pipeline onto one EP override: "cpu" strips the hardware
-# providers (CPU fallback, no WinML EP registration), "qnn"/"dml" route every
-# stage to that accelerator.  An unknown device also falls back to ``None``.
-_DEVICE_TO_GENAI_EP: dict[str, EPNameOrAlias | None] = {
-    "cpu": "cpu",
-    "npu": "qnn",
-    "gpu": "dml",
-    "auto": None,
-}
+# Sentinel ``--device`` value meaning "respect the bundle's genai_config.json
+# routing" (no EP override).  It is the winml-genai default: a genai bundle is
+# mixed by design (e.g. ctx/iter on the NPU, embeddings/lm_head on CPU) and its
+# config already encodes that per-stage routing, so the common case leaves it
+# untouched.  A concrete ``--device`` (or ``--ep``) is an explicit override that
+# forces the *whole* decoder pipeline onto one EP.
+GENAI_CONFIG_DEVICE = "config"
 
 
-def device_to_genai_ep(device: str) -> EPNameOrAlias | None:
-    """Map a ``--device`` value to a :class:`GenaiSession` EP override.
+def resolve_genai_ep(device: str) -> EPNameOrAlias | None:
+    """Resolve a ``--device`` value to a :class:`GenaiSession` EP override.
 
-    Returns ``None`` for ``auto`` (and any unrecognized device), meaning the
-    bundle's ``genai_config.json`` routing is respected; otherwise a concrete
-    EP short name that forces the whole decoder pipeline onto that provider.
+    ``config`` -> ``None`` (respect ``genai_config.json`` as-is).  Any concrete
+    device (``auto``/``npu``/``gpu``/``cpu``) goes through the same
+    :func:`resolve_device` / :func:`resolve_eps` path the WinML ONNX runtime
+    uses, so it forces the whole pipeline onto the *best EP actually available
+    for that device on this machine* (e.g. an NPU that is VitisAI/OpenVINO
+    rather than QNN) instead of a static short-name guess.  Returns the EP's
+    canonical short alias, or ``None`` when the device resolves to no EP.
+
+    Raises:
+        ValueError: propagated from :func:`resolve_device` when the requested
+            device has no compatible EP available -- fail fast, like the ONNX
+            path, rather than silently falling back to CPU.
     """
-    return _DEVICE_TO_GENAI_EP.get(device.lower())
+    if device == GENAI_CONFIG_DEVICE:
+        return None
+
+    # Function-local import mirrors the ONNX path (perf.py) and avoids a
+    # module-level cycle: ``sysinfo`` pulls in heavier device-probing deps.
+    from ..sysinfo import resolve_device, resolve_eps
+
+    resolved_device, _ = resolve_device(device=device, ep=None)
+    eps = resolve_eps(resolved_device)
+    if not eps:
+        return None
+    return EP_NAME_TO_ALIAS[eps[0]]
 
 
 def genai_output_path(bundle_dir: str | Path) -> Path:
--- a/src/winml/modelkit/commands/_perf_genai.py
+++ b/src/winml/modelkit/commands/_perf_genai.py
@@ -328,6 +328,8 @@ class GenaiPerfBenchmark:
         if device and device not in ("config", "auto"):
             return device
         canonical = normalize_ep_name(ep)
+        if canonical is None:
+            return None
         devices = EP_SUPPORTED_DEVICES.get(canonical)
         return devices[0] if devices else None
 
--- a/src/winml/modelkit/commands/_perf_genai.py
+++ b/src/winml/modelkit/commands/_perf_genai.py
@@ -41,6 +41,7 @@ from ..session import (
     GenaiSessionError,
     GenerationConfig,
 )
+from ..utils.constants import EPNameOrAlias
 
 
 if TYPE_CHECKING:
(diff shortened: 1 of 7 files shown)


src/winml/modelkit/commands/_perf_genai.py
```diff
@@ -59,24 +60,30 @@
...
-_DEVICE_TO_GENAI_EP: dict[str, str] = {
+# --device -> GenaiSession EP override.  The default ("auto") resolves to
+# ``None`` = *respect the bundle config*: genai bundles route each stage via its
+# own session_options in genai_config.json (e.g. ctx/iter on the NPU/QNN,
+# embeddings/lm_head on CPU), and ``None`` honors that per-stage routing while
+# still registering the WinML EPs those stages need.  A concrete device forces
+# the whole decoder pipeline onto one EP override: "cpu" strips the hardware
+# providers (CPU fallback, no WinML EP registration), "qnn"/"dml" route every
+# stage to that accelerator.  An unknown device also falls back to ``None``.
+_DEVICE_TO_GENAI_EP: dict[str, EPNameOrAlias | None] = {
```
xieofxie: if we could update config, we don't need this mapping. We should always update to what user inputs and follow the device ep resolution.

But before this, I think we need a way to identify if it should be default.

Currently device=auto + ep=None means resolving the best capable EP + device for onnx files.

Need a discussion?

xieofxie: like device=follow + ep=None means follow

DingmaomaoBJTU: Good point — agree the hardcoded `device → ep` map (`npu→qnn`, `gpu→dml`) shouldn't live here. Since we now rewrite `genai_config.json`, `--device cpu/npu/gpu` should go through the same `resolve_device`/`resolve_eps` path the ONNX runtime uses, so it picks the **best available** EP for the device on this machine instead of a static guess (e.g. an NPU that's VitisAI/OpenVINO rather than QNN). I'll switch to that.

The genai-specific wrinkle behind the "what's default" question: a genai bundle is mixed by design — embeddings/lm_head on CPU, ctx/iter on NPU — and `genai_config.json` already encodes that per-stage routing. Any concrete `--device` forces the **whole** pipeline onto one EP, which is a deliberate override, not the common case (forcing everything onto one EP measured ~12x slower than respecting the config in my local run). So I think "respect the bundle config" has to stay the default.

Proposal, to make that explicit like you suggested:
- `--device config` (= default, `ep=None`) → respect `genai_config.json` as-is.
- `--device cpu|npu|gpu` → resolve the best available EP for that device via `resolve_eps`, rewrite every stage onto it.
- `--ep <name>` → explicit EP override, wins over `--device`.

Two calls I'd like from you:
1. Name for the "respect config" sentinel — `config` or `follow`?
2. What should `--device auto` mean for genai? I lean toward it aliasing `config` (respect the bundle), since auto-resolving one EP and forcing the whole pipeline is usually wrong for a mixed genai bundle. Or do you want `auto` = "resolve best single EP" to match the ONNX path, with `config`/`follow` as the default instead?

xieofxie: I don't know. look like default config is best. but it will diverge from onnx. Should we use different default for different runtime? look like acceptable

DingmaomaoBJTU: Thanks — went with **`config` as the default for winml-genai**, and yes it intentionally diverges from ONNX (which stays `auto`). A genai bundle is mixed per-stage by design, and forcing one EP across the whole pipeline measured ~12× slower than respecting the config in my local run, so "respect the bundle" has to be the genai default. Different default per runtime, as you said — acceptable.

Implemented in 07ef0ec5:

- Dropped the hardcoded `_DEVICE_TO_GENAI_EP` map. `--device auto|npu|gpu|cpu` now go through the same `resolve_device`/`resolve_eps` the ONNX path uses, so they pick the *best EP actually available* for that device (e.g. an NPU that's VitisAI/OpenVINO, not a static `npu→qnn`).
- `--device config` = respect `genai_config.json` as-is. It's the genai default (omitting `--device` maps to it), can be passed explicitly, and the `winml` runtime rejects it with a clear error.
- `--device auto` now matches ONNX: best device → best available EP, forced across the whole pipeline.
- Precedence unchanged: `--ep` > `--device` > `config`.

On your earlier point ("if we could update config, we don't need this mapping — always update the config"): that's exactly how the override already works — it rewrites `genai_config.json` in a derived `_compiled/` bundle rather than passing EP flags at load, so there's no separate mapping layer. `config` just means "don't rewrite". I kept the name `config` over `follow` since it reads as "respect the config file".

tests/e2e/test_perf_e2e.py
```diff
@@ -737,3 +739,185 @@
...
+                cmd,
+                capture_output=True,
+                text=True,
+                timeout=1500,
+                check=False,
+            )
+        except subprocess.TimeoutExpired:
+            pytest.skip("onnxruntime-genai model build timed out")
+
+        if proc.returncode != 0 or not (out / "genai_config.json").exists():
```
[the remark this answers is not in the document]
DingmaomaoBJTU: Fixed in a886b8d8: proc is now bound to None in the TimeoutExpired handler, so it's definitely-initialized on every path (timeout -> skip on None; otherwise check 
eturncode). Happy path unchanged.

src/winml/modelkit/session/genai_session.py
```diff
@@ -686,56 +753,181 @@ def _new_generator(self, cfg: GenerationConfig, prompt_len: int) -> Any:
...
+        behave exactly as before.
+
+        Otherwise every stage in ``model.decoder.pipeline`` has its
+        ``session_options.provider_options`` rewritten so the *whole* decoder
+        pipeline is forced onto the override EP:
+
+        * **CPU** — strip hardware ``provider_options`` (set to ``[]``) so
+          onnxruntime-genai falls back to the CPU provider.  Stages that were
+          already CPU are left untouched.
+        * **hardware EP** — route every stage to that EP as
```
xieofxie: need to set device for openvino, qnn. but could be next pr

src/winml/modelkit/session/genai_session.py
```diff
@@ -285,15 +307,18 @@ class GenaiSession:
     def __init__(
         self,
         bundle_dir: str | Path,
-        ep: str = "cpu",
+        ep: EPNameOrAlias | None = None,
```
xieofxie, line 334: we should also include device to match with _perf_genai.py ?

src/winml/modelkit/session/genai_session.py
```diff
@@ -686,56 +753,181 @@ def _new_generator(self, cfg: GenerationConfig, prompt_len: int) -> Any:
...
+        pipeline = effective.get("model", {}).get("decoder", {}).get("pipeline", [])
+        if isinstance(pipeline, list):
+            for stage_entry in pipeline:
+                if not isinstance(stage_entry, dict):
+                    continue
+                for stage_cfg in stage_entry.values():
+                    if isinstance(stage_cfg, dict):
+                        self._override_stage_provider(stage_cfg)
+
+        return effective, effective != cfg
```
xieofxie, line 866: Both this rewrite and `_bundle_uses_hardware_ep` only walk `model.decoder.pipeline`, so on a flat / empty-pipeline decoder an override rewrites nothing and registers no EP — yet the run still reports the requested EP (e.g. `--ep qnn` → `benchmark_info.ep: qnn`). The PR's "an override can never silently become a no-op" guarantee only covers *invalid* EP names, not this case (`test_force_ep_on_empty_pipeline_is_noop` and `test_ep_cpu_overrides_ep_only` both bake in the no-op). Consider warning — or reporting `config` — when an override was requested but zero stages were actually rewritten, so the report doesn't claim an EP that never took effect.

src/winml/modelkit/session/genai_session.py
DingmaomaoBJTU: Done in this PR (c803508f). When an override re-routes a stage to a device-parameterized EP (OpenVINO/VitisAI) that the bundle defines no options for, we now synthesize `{"device_type": <DEVICE>}` from `--device`, falling back to the EP's primary supported device from `EP_SUPPORTED_DEVICES` when `--device` is omitted.

QNN is handled without hardcoding a backend: if any pipeline stage already defines QNN options (`backend_path`/`soc_model`), a re-routed stage borrows them; otherwise we log a warning and write empty options (QNN falls back to its default backend) rather than guessing a `backend_path`.

Verified: `--ep openvino --device npu` → `[{"openvino": {"device_type": "NPU"}}]`; `--device npu` on a QNN bundle keeps the bundle's own `backend_path`/`soc_model` and reports `ep=qnn`.

src/winml/modelkit/session/genai_session.py
DingmaomaoBJTU: Yes — added a `device` param to `GenaiSession` in c803508f, and perf now threads `--device` through to it via `_session_device`, so the session and `_perf_genai.py` agree on the target device.

It's used when re-routing a stage to a device-parameterized EP (OpenVINO/VitisAI) that has no reusable options: the synthesized `device_type` comes from this `device` (falling back to the EP's primary supported device). QNN keeps reusing the bundle's own `backend_path`, so it doesn't depend on `device`.

src/winml/modelkit/session/genai_session.py
DingmaomaoBJTU: Fixed in c803508f. `effective_ep` is now computed from the *post-override* config via `_override_took_effect`: an override that rewrites/strips no stage (flat or empty pipeline, or an all-CPU bundle) yields `effective_ep = None`, so the report shows `config` instead of the requested EP, and `load()` logs a warning that the requested override did not take effect and the run follows `genai_config.json`.

Added an e2e case (`test_ep_qnn_on_flat_bundle_reports_config`) asserting that `--ep qnn` on a flat CPU bundle reports `ep: config` and never registers QNN. Verified on a real bundle: a QNN pipeline bundle correctly reports `ep=qnn`, while a flat/all-CPU pipeline reports `config`.

xieofxie (approved):

PR #1047, opened 2026-07-03
feat(quant): add dynamic quantization pass

DingmaomaoBJTU:
## Summary

Implements a real **dynamic quantization** pass in the `winml quantize` module. Previously `mode="dynamic"` existed in config but was a placeholder that silently routed to `StaticPass`.

Dynamic quantization quantizes **weights statically** (INT8) while **activation quantization parameters are computed at runtime** — so **no calibration data is required**. It emits QOperator-format ops (`DynamicQuantizeLinear` / `MatMulInteger` / `ConvInteger`) rather than QDQ nodes.

## Changes

- **New `DynamicPass`** (`quant/passes/dynamic.py`) wrapping ONNX Runtime's `quantize_dynamic`. Preserves model metadata, runs shape inference, and counts quantized nodes universally (no hardcoded/architecture-specific logic). Supports only 8-bit weights; 16-bit `weight_type` falls back to int8 with a warning.
- **Wired into the pass pipeline** (`quantizer.py`): `expand_precision` now maps `mode="dynamic"` → `DynamicPass` instead of `StaticPass`.
- **Config** (`config.py`): added `reduce_range` field with `to_dict`/`from_dict` round-trip (only serialized in dynamic mode).
- **CLI** (`commands/quantize.py`): new `winml quantize --precision dynamic` token. Honors `--weight-type` / `--per-channel` / `--symmetric`; warns that calibration-only options (`--samples`, `--method`, `--activation-type`, `--task`, `--model-id`) are ignored.
- **Exports** added in both `quant/__init__.py` and `quant/passes/__init__.py`.

## Behavior

| | Static | Dynamic |
|---|---|---|
| Weights | offline INT8 | offline INT8 |
| Activations | offline (needs calibration) | **runtime (no calibration)** |
| ONNX ops | QDQ | QOperator (DynamicQuantizeLinear / MatMulInteger / ConvInteger) |

Usage:
```
winml quantize -m model.onnx --precision dynamic
```

## Tests

- **Unit** (`tests/unit/test_quant_passes.py`): arg forwarding to `quantize_dynamic`, `WeightSymmetric` override, uint8/int8 mapping, 16-bit → int8 fallback warning, and `reduce_range` config serialization. Also updated the `expand_precision("dynamic")` routing test to assert `DynamicPass`.
- **E2E** (`tests/e2e/test_quantize_e2e.py`): `--precision dynamic` runs end-to-end, output contains `MatMulInteger`/`DynamicQuantizeLinear`, no QDQ nodes, passes `onnx.checker`, and runs under ORT-CPU.

Verification: unit quant passes + quantizer (31 passed), e2e `TestPrecision` (7 passed), precision + CLI-flags + integration (231 passed); `ruff check` clean.

### DIFF
--- a/src/winml/modelkit/commands/quantize.py
+++ b/src/winml/modelkit/commands/quantize.py
@@ -91,6 +91,13 @@ console = Console()
     help="Use symmetric quantization",
 )
 @click.option(
+    "--reduce-range/--no-reduce-range",
+    default=False,
+    show_default=True,
+    help="Quantize weights with 7 bits to reduce int8 saturation on pre-VNNI "
+    "CPUs (dynamic quantization only; ignored by other precisions).",
+)
+@click.option(
     "--task",
     type=str,
     default=None,
@@ -116,6 +123,7 @@ def quantize(
     activation_type: str | None,
     per_channel: bool,
     symmetric: bool,
+    reduce_range: bool,
     task: str | None,
     model_id: str | None,
     verbose: int,
@@ -145,6 +153,9 @@ def quantize(
 
         # Dynamic quantization (no calibration data needed)
         winml quantize -m model.onnx --precision dynamic
+
+        # Dynamic quantization with int8 weights + reduced range (pre-VNNI CPUs)
+        winml quantize -m model.onnx --precision dynamic --weight-type int8 --reduce-range
 
         # RTN int4 followed by FP16 conversion (two-pass pipeline)
         winml quantize -m model.onnx --precision int4 --precision fp16
@@ -182,6 +193,8 @@ def quantize(
             per_channel = qc["per_channel"]
         if not cli_utils.is_cli_provided(ctx, "symmetric") and "symmetric" in qc:
             symmetric = qc["symmetric"]
+        if not cli_utils.is_cli_provided(ctx, "reduce_range") and "reduce_range" in qc:
+            reduce_range = qc["reduce_range"]
         if not cli_utils.is_cli_provided(ctx, "task") and "task" in qc:
             task = qc["task"]
         if not cli_utils.is_cli_provided(ctx, "model_id") and "model_id" in qc:
@@ -204,6 +217,7 @@ def quantize(
             activation_type=activation_type,
             per_channel=per_channel,
             symmetric=symmetric,
+            reduce_range=reduce_range,
             task=task,
             model_id=model_id,
             console=console,
@@ -249,6 +263,7 @@ def quantize(
             weight_type=cast('Literal["uint8", "int8", "uint16", "int16"]', resolved_weight),
             per_channel=per_channel,
             symmetric=symmetric,
+            reduce_range=reduce_range,
         )
         label = "Dynamic quantization"
         console.print(f"[bold blue]Weight type:[/bold blue] {resolved_weight}")
@@ -366,6 +381,7 @@ def _run_multi_precision(
     activation_type: str | None,
     per_channel: bool,
     symmetric: bool,
+    reduce_range: bool,
     task: str | None,
     model_id: str | None,
     console: Console,
@@ -407,6 +423,7 @@ def _run_multi_precision(
         activation_type=cast('Literal["uint8", "int8", "uint16", "int16"]', resolved_activation),
         per_channel=per_channel,
         symmetric=symmetric,
+        reduce_range=reduce_range,
         task=task,
         model_id=model_id,
     )
--- a/src/winml/modelkit/commands/quantize.py
+++ b/src/winml/modelkit/commands/quantize.py
@@ -46,9 +46,10 @@ console = Console()
 @cli_utils.precision_option(
     default=(),
     multiple=True,
-    help_text="Quantization precision: fp16, int4, int8, int16, or w{x}a{y} where "
+    help_text="Quantization precision: fp16, int4, int8, int16, dynamic, or w{x}a{y} where "
     "x in {4,8,16}, y in {8,16} (e.g., w8a8, w8a16). "
     "int4 uses RTN weight-only quantization; "
+    "dynamic uses dynamic quantization (no calibration data); "
     "fp16 converts all FP32 tensors to FP16 (no QDQ). "
     "Repeat to chain passes in order (e.g. -p int4 -p fp16)",
     optional_message="Overridden by explicit --weight-type/--activation-type",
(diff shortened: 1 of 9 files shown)


timenick (approved):

src/winml/modelkit/commands/quantize.py
```diff
@@ -232,6 +237,23 @@ def quantize(
...
 
+    elif precision_lower == "dynamic":
+        # Dynamic quantization: weights quantized statically, activation
+        # quantization parameters computed at runtime (no calibration data).
+        _warn_ignored_dynamic_options(ctx, console)
+        if output is None:
+            output = model.parent / f"{model.stem}_dynamic.onnx"
+        resolved_weight = weight_type or "uint8"
+        config = WinMLQuantizationConfig(
+            mode="dynamic",
```
xieofxie, line 262: `reduce_range` looks half-wired for CLI users. The config field, its serialization, `DynamicPass` forwarding, and tests all exist, but this constructor never sets it, so it's permanently `False` from the CLI. This matters because the CLI defaults `weight_type` to `uint8` while ORT / the `DynamicPass` fallback is `int8`, and `reduce_range` is exactly the mitigation for int8 `u8*s8` saturation on pre-VNNI CPUs — which users currently can't reach. Either add a `--reduce-range` flag and pass it here, or document explicitly that it's config-file-only.

src/winml/modelkit/commands/quantize.py
```diff
@@ -302,11 +324,35 @@ def _cli_precision_to_mode(precision: str) -> str:
     p = precision.lower()
     if p == "fp16":
         return "fp16"
+    if p == "dynamic":
```
xieofxie, line 342: Can you confirm the upstream `--precision` token validator accepts `dynamic`? The named-precision set (e.g. `_NAMED_PRECISIONS` in `config/precision.py`) is `{auto, fp32, fp16, int4, int8, int16}` and this PR doesn't touch it. It likely works because `dynamic` predated this PR as a config mode and the e2e tests pass, but if there's a whitelist gating the token it could silently reject `--precision dynamic` before reaching this mapping — worth a quick check.

src/winml/modelkit/quant/passes/dynamic.py
```diff
@@ -0,0 +1,199 @@
...
+# QuantizeLinear is intentionally excluded: quantize_dynamic emits
+# DynamicQuantizeLinear (not QuantizeLinear) for activations and stores weights
+# as pre-quantized initializers, so a static QuantizeLinear never appears in its
+# output.
+_DYNAMIC_QUANT_OP_TYPES: frozenset[str] = frozenset(
+    {
+        "DynamicQuantizeLinear",
+        "MatMulInteger",
+        "ConvInteger",
+        "DequantizeLinear",
```
xieofxie, line 49: Including `DequantizeLinear` in the count set is well documented and justified for the embedding case, but note it's a heuristic: DQL nodes can originate from sources other than a quantized weight/embedding restore, so `nodes_quantized` could over-count. Since it's a reporting metric (not correctness-critical) this is fine — just flagging that it's fuzzier than the static QDQ-op count.

src/winml/modelkit/quant/passes/dynamic.py
```diff
@@ -0,0 +1,199 @@
...
+        logger.info(
+            "Running dynamic quantization (weight_type=%s, per_channel=%s, reduce_range=%s)...",
+            weight_type,
+            self._config.per_channel,
+            self._config.reduce_range,
+        )
+
+        input_model = load_onnx(model_path, validate=False)
+        metadata_snapshot = capture_metadata(input_model)
+        add_pre_process_metadata(input_model)
```
xieofxie, line 146: This adds pre-process *metadata* but never runs pre-quant shape inference / optimization. `StaticPass` routes through `get_qdq_config`, which handles the model; here `quantize_dynamic` is called directly and `infer_shapes` only runs *after* quantization (line 178). ORT recommends `quant_pre_process` (symbolic shape inference) before dynamic quantization for correctness/quality on non-trivial graphs. Worth confirming real transformer exports quantize correctly without a pre-quant shape-inference step.

src/winml/modelkit/commands/quantize.py
DingmaomaoBJTU: Confirmed there's no whitelist issue here. `precision_option` in `utils/cli.py` is deliberately `type=str` (not `click.Choice`) so mixed `w{x}a{y}` forms work — invalid values are rejected downstream rather than at the Click layer.

In quantize's single-precision path, `dynamic` is intercepted by the `elif precision_lower == "dynamic":` branch *before* `_resolve_quant_types` (the only validator) is reached. The `_NAMED_PRECISIONS` / `resolve_precision` machinery in `config/precision.py` is used by the build-path commands, not by quantize.

The e2e test `test_dynamic_precision_accepted` exercises this path. No code change needed.

src/winml/modelkit/quant/passes/dynamic.py
DingmaomaoBJTU: This is intentional — `DynamicPass` mirrors the established `StaticPass` pattern. Neither pass runs `quant_pre_process` / symbolic shape inference *before* quantizing: both call `add_pre_process_metadata` (metadata only) and run `infer_shapes` *after* quantization. In fact `quant_pre_process` isn't used anywhere in `src/`, so adding it to only `DynamicPass` would introduce an inconsistency between the two passes.

Empirically verified the pass handles a real transformer without pre-quant shape inference: exported `prajjwal1/bert-tiny` (132 fp32 nodes) then `--precision dynamic` produces `DynamicQuantizeLinear`×8, `MatMulInteger`×12, `DequantizeLinear`×2, and the output loads + runs correctly under ORT-CPU (output shape `(1, 512, 128)`).

If we do want pre-quant symbolic shape inference, I'd suggest a follow-up that adds it to both passes uniformly rather than diverging here.

src/winml/modelkit/commands/quantize.py
DingmaomaoBJTU: Good catch — `reduce_range` was only reachable via the config file and never from the CLI, so it was effectively `False` for CLI users.

Added a `--reduce-range/--no-reduce-range` flag (default `False`, mirroring `--per-channel`) and threaded it through the single-precision `dynamic` branch, the config-file default merge, and `_run_multi_precision`. It's dynamic-only (documented as "ignored by other precisions" in the help text, since only `DynamicPass` reads it).

Verified end-to-end on bert-tiny:

```
winml quantize -m bert_tiny.onnx --precision dynamic --weight-type int8 --reduce-range
```

Flag parsing + defaults are covered in `tests/unit/commands/test_boolean_flag_pairs.py`. (268871dd)

src/winml/modelkit/quant/passes/dynamic.py
DingmaomaoBJTU: Agreed — this is a reporting metric, not correctness-critical. I've expanded the comment on `_DYNAMIC_QUANT_OP_TYPES` to state explicitly that including `DequantizeLinear` makes `nodes_quantized` a heuristic that may slightly over-count (a DQL node could in principle come from something other than a quantized-weight restore), and that this is acceptable for a reporting metric — analogous to, if fuzzier than, the static pass's QDQ-op count. (268871dd)

xieofxie (approved):

PR #1054, opened 2026-07-06
Improve GenAI session: smarter max_length, DML skip, flat-decoder support, and response logging

xieofxie:
## Summary

Several improvements to GenAI session benchmarking and hardware EP detection.

### Smarter `max_length` allocation
`_new_generator` now computes `max_length` as `prompt_len + max_new_tokens`, capped at the bundle's `context_length`. This avoids pre-allocating the full KV cache (which can be 128K+ for DML bundles) when only a small generation is requested.

### Hardware EP detection improvements
- `_bundle_uses_hardware_ep` now returns the EP name (`str | None`) instead of `bool`, enabling better logging.
- DML is now treated like CPU — bundles using only DML no longer trigger WinML EP registration, since DML doesn't need it.
- Added support for **flat-decoder layout** (no `pipeline` wrapper), used by e.g. OpenVINO exports.

### Response text capture and logging
- Added `response_text` field to `GenerationTiming` to capture decoded model output.
- The first iteration's response is logged at INFO level; subsequent iterations at DEBUG.
- `generate_timed` now logs a summary line with input/output token counts and timing metrics.

### Tests
- Updated all `_bundle_uses_hardware_ep` tests to match the new return type.
- Added tests for flat-decoder layout (OpenVINO, CPU-only, no provider options).
- Added `test_max_length_is_prompt_plus_max_new_tokens` for the new max_length logic.
- Added DML-only test case (`test_dml_only_returns_none`).


xieofxie:
tested on
- [x] openvino npu
- [x] openvino gpu
- [x] dml
- [x] qnn

src/winml/modelkit/session/genai_session.py
```diff
@@ -489,33 +493,48 @@ def generate_timed(
...
         marks: list[float] = []
         generated = 0
+        output_token_ids: list[int] = []
         marks.append(clock())
         generator.append_tokens(tokens)
         marks.append(clock())
         while not generator.is_done():
             generator.generate_next_token()
             marks.append(clock())
+            output_token_ids.append(generator.get_next_tokens()[0])
```
DingmaomaoBJTU: This token fetch happens after the timestamp for the current `generate_next_token()`, so its cost is charged to the next token's decode interval (`marks[i + 1] - marks[i]`). On hardware EPs `get_next_tokens()` can force host/device synchronization or copies, which means `decode_s`/TPOT no longer measure only the `generate_next_token()` boundary described above. Can we either collect the token before taking the mark, or time each `generate_next_token()` with explicit start/end timestamps and fetch tokens outside those spans?

xieofxie: Good catch — moved token collection entirely out of the timing loop. Now using `generator.get_sequence(0)` after generation completes to fetch the full sequence in one call, then slicing off the prompt prefix. This keeps `decode_s`/TPOT measurements clean of any host/device sync overhead from token retrieval.

DingmaomaoBJTU (approved):
The latest commit (34eab68f) resolves my earlier concern: token fetching via get_sequence() is now done after the timing loop, so decode_s/TPOT measurements are no longer polluted by potential host/device sync costs from get_next_tokens(). CI is green. LGTM.

PR #1055, opened 2026-07-06
feat(perf): show dynamic input dims as dynamic(&lt;actual&gt;) in model info

### LINKED ISSUE #1052: bug: fix and verify batch for dynamic shape model, batch != 1 static shape model

xieofxie:
## Summary

The pre-benchmark model-info output previously printed dynamic input dimensions as `None`, which was ambiguous — it didn't convey that the axis was free, nor what size the generated benchmark data actually used.

This changes dynamic dims to render as `dynamic(<actual>)`:

**Before**
```
Inputs:      pixel_values         [None, 3, 64, 64]        float32
Outputs:     logits               [None, 1000]
```

**After**
```
Inputs:      pixel_values         [dynamic(1), 3, 64, 64]  float32
Outputs:     logits               [dynamic, 1000]
```

The value in parentheses is the concrete size the generated input tensor actually used for that axis (e.g. the resolved batch size), so both facts stay visible: the model left the axis free, and the benchmark ran it at size N.

## Changes

- Added `_format_input_shape(shape, actual)` — maps each `None` dim to `dynamic(<actual>)`, falling back to bare `dynamic` when no concrete shape is available; static dims are unchanged.
- `_print_model_info` gained an `actual_shapes` param, threaded from `self._inputs` (name → real `arr.shape`) at the call site.
- Added `TestFormatInputShape` covering single/multiple dynamic dims, all-static, and the no-actual fallback.

Close https://github.com/microsoft/winml-cli/issues/1052

### DIFF
--- a/src/winml/modelkit/commands/perf.py
+++ b/src/winml/modelkit/commands/perf.py
@@ -528,6 +528,11 @@ class PerfBenchmark:
             req_device=self.config.device,
             act_device=self._single.device,
             ep_name=self._single.ep_name,
+            actual_shapes=(
+                {name: tuple(arr.shape) for name, arr in self._inputs.items()}
+                if self._inputs
+                else None
+            ),
         )
 
         # [3] Run benchmark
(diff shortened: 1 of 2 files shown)


xieofxie:
tested on
- [x] : dynamic model
- [x] : static model: if not match, will output
[11:08:54 WARNING winml.modelkit.commands.perf] Requested --batch-size 10 could not be applied: the model's leading input dimension is statically 4. Throughput is scaled by the actual batch (4), not the requested value.

fenghuaxueyue98 (approved):

Issue #1038, opened 2026-07-03
feat: when exporting / building composite model as a folder output, we should have a file to know it is built by winml cli


xieofxie:
currently for build, we have build_manifest.json. 

may be we unify to winmlcli_manifest.json

To distinguish from hf folder, so we could load via WinmlAuto class instead of transformer's

Related pr https://github.com/microsoft/winml-cli/pull/1057

DingmaomaoBJTU:
**Triage Summary** (auto-generated 2026-07-27)

| Field | Value |
|-------|-------|
| **Priority** | P2 |
| **Type** | feature scale |
| **Component** | composite-model |
| **Owner** | @xieofxie |

_This issue has been triaged. Please update priority/assignment if needed._

PR #1057, opened 2026-07-06
refactor: introduce WinMLManifest dataclass as single source of truth for build_manifest.json

xieofxie:
## Summary

Introduces a `WinMLManifest` dataclass as the single source of truth for the `build_manifest.json` schema. This is a pure refactor — no new behavior, no filename changes.

### New: `WinMLManifest` / `ManifestStage` dataclasses (`utils/manifest.py`)
- Central schema for `build_manifest.json` — all producers construct a `WinMLManifest` and call `.save()`
- `to_dict()` drops `None` fields for compact JSON; `from_dict()` is forward-compatible (unknown keys preserved in `extras` at both top-level and stage-level)
- `_sanitize_value()` converts `Path` objects to strings and numpy scalars to native Python types, preserving numeric precision
- `find()`, `load()`, `manifest_path_for()` helpers for consumers to adopt incrementally

### Build producers updated to use `WinMLManifest`
- `build/hf.py`: raw dict → `WinMLManifest` + `ManifestStage`
- `build/onnx.py`: same refactor

### Tests
- 16 tests in `tests/unit/utils/test_manifest.py` (round-trip, None-omission, forward-compat extras at top and stage level, I/O, find, Path sanitization, numpy scalar sanitization)

### Cleanup
- Removed dead `_cleanup_partial_composite` function from `export.py` (defined but never called)


DingmaomaoBJTU (commented):
Solid, well-structured refactor — the `WinMLManifest`/`ManifestStage` dataclasses are clean, well-documented, forward-compatible (`extras`), and have a dedicated `test_manifest.py`. A few issues though, most notably the rename is **incomplete**.

## 🔴 Bugs / correctness

**1. Missed rename in `utils/cli.py` — the rename is not complete**

`classify_model_input` (`src/winml/modelkit/utils/cli.py:91-92`) still globs the old name:

```python
manifest_files = list(path.glob("build_manifest.json")) + list(path.glob("*_build_manifest.json"))
```

Producers now write `winml_manifest.json`, so a folder produced by the new `winml build`/`export` yields `is_winml_cli_folder=False`. It's currently *latent* (no command reads `is_winml_cli_folder`/`folder_has_onnx` — only `tests/unit/utils/test_cli.py` does), so the paired tests (~L447-474) still pass while validating **stale** behavior. This contradicts the PR's "all references updated (15 files)" claim. Please update the globs and those tests.

**2. Dead code: `_cleanup_partial_composite` (`commands/export.py:67`)**

Defined but never called. The composite-failure path (`export.py:595`) deliberately calls `_warn_partial_composite`, and its docstring states *"we do NOT delete anything."* So this helper is unreachable — either wire it in intentionally or remove it.

## 🟠 Coverage / design

**3. New "export writes manifest" behavior is untested**

`commands/export.py` now writes `winml_manifest.json` (single) and `<stem>_winml_manifest.json` (composite), but `tests/unit/commands/test_export.py` has no manifest assertions. The hf/onnx manifest-writing paths are tested; export's new path isn't. Please add coverage for both the single and composite-prefixed cases.

**4. No fallback for pre-existing `build_manifest.json` caches**

`inference/engine.py`, `inspect/resolver.py`, and `serve/app.py` now look *only* for `winml_manifest.json` (no fallback). Combined with #1, the codebase splits: `cli.py` recognizes only the **old** name while everything else recognizes only the **new** name — so neither pre-existing caches nor new outputs are recognized consistently across all components. If caches are considered disposable/pre-release this may be acceptable, but it deserves an explicit decision (or a one-time fallback read of the old filename).

## 🟡 Nits

- **`save()` uses `json.dumps(..., default=str)`** (`utils/manifest.py`), which the old writer didn't. If `export_stats`/`analyze_details` ever contain numpy scalars, numeric metrics silently serialize as strings (`"10"` rather than `10`). Consider sanitizing the payload instead of a blanket `default=str`.
- **Forward-compat asymmetry:** top-level unknown keys survive in `extras`, but unknown keys *inside a stage* are dropped by `from_dict` (filtered against `ManifestStage.__dataclass_fields__`). Minor, but the "forward-compatible" guarantee doesn't hold for stage entries.
- **Import convention:** producers use function-level `from ..utils.manifest import ...` (reaching into the submodule) even though both symbols are exported by `utils/__init__.py`; per the repo convention prefer `from ..utils import ManifestStage, WinMLManifest`.

src/winml/modelkit/inference/engine.py
```diff
@@ -161,7 +161,7 @@ def _find_build_artifacts(build_dir: Path, *, task: str | None = None) -> tuple[
     candidates: list[tuple[Path, dict | None]] = []
     for onnx_path in sorted(build_dir.glob("*_model.onnx")):
         prefix = onnx_path.name.rsplit("_model.onnx", 1)[0]
-        manifest_path = build_dir / f"{prefix}_build_manifest.json"
+        manifest_path = build_dir / f"{prefix}_winml_manifest.json"
```
timenick: The rename to `winml_manifest.json` is a clean break with no fallback to the old `build_manifest.json` name. For build/cache directories produced by a prior winml-cli version, `_find_build_artifacts` now gets `manifest=None` (the old-named file no longer matches).

In the prefixed layout that mostly just loses metadata — but when a cache dir holds multiple task variants, `if manifest is None or manifest.get("task") == task: return onnx_path` short-circuits on `manifest is None` and returns the first `*_model.onnx` **without** the task check — the exact "wrong output head → garbage results" case this function's own docstring warns about. On `main` the old-named manifest was still discovered, so this is a behavior change for pre-upgrade caches.

The PR's backward-compatibility note covers the JSON schema/field names, not the filename. Is the artifact cache (`~/.cache/winml/artifacts/...`) rebuilt/invalidated on upgrade so pre-existing `build_manifest.json` dirs never reach this path? If not, a one-release fallback glob for the old name (or a task guard that doesn't silently pass when the manifest is missing) would avoid the silent wrong-model selection.

xieofxie: I think we don't need fallback for the reasons
- inference engine is not public available
- winml cli is not GA
does it make sense to you ?

src/winml/modelkit/inspect/resolver.py
```diff
@@ -508,7 +508,7 @@ def resolve_cache(model_id: str) -> CacheInfo:
     # PRIMARY: Manifest-based resolution
     # -------------------------------------------------------------------------
     if model_dir.exists():
-        manifests = list(model_dir.glob("*build_manifest.json"))
+        manifests = list(model_dir.glob("*winml_manifest.json"))
```
DingmaomaoBJTU: This glob (and the matching ones in `inference/engine.py`, `serve/app.py`, `utils/cli.py`) now only match `winml_manifest.json`, so any model built with a prior version — whose artifacts still sit in the persistent cache dir as `build_manifest.json` — stops resolving here. The result is silent cache misses: `inspect`/`serve` lose the cached metadata and builds get redone even though the ONNX is right there, and there's no cache-version bump to invalidate cleanly.

Could we either also glob the legacy `*build_manifest.json` as a read fallback, or namespace/bump the cache dir so stale entries are handled deterministically? At minimum this rename probably warrants a migration note.

src/winml/modelkit/utils/manifest.py
```diff
@@ -0,0 +1,212 @@
...
+        d.update({k: _sanitize_value(v) for k, v in raw.items() if v is not None})
+        # Stage entries: also drop None fields inside each stage, merge extras.
+        if "stages" in d:
+            d["stages"] = [_sanitize_stage_dict(s) for s in d["stages"]]
+        d.update({k: _sanitize_value(v) for k, v in extras.items()})
+        return d
+
+    def save(self, path: Path) -> None:
+        """Write the manifest as indented JSON to *path*."""
+        path.write_text(json.dumps(self.to_dict(), indent=2, default=str))
```
DingmaomaoBJTU, line 157: With `default=str` kept here as the fallback, `_sanitize_value` only really handles `Path` — a `numpy` scalar in `export_stats`/`analyze_details` still ends up serialized as a string, which is exactly the "numeric metric accidentally serialised as a string" case the `_sanitize_value` docstring says it prevents. If those dicts can carry numpy types (export/quant stats often do), let's coerce `np.generic` via `.item()` inside `_sanitize_value`; otherwise the docstring overpromises what it guards against.

DingmaomaoBJTU (commented):
Nice refactor — WinMLManifest as a single source of truth is a clear improvement and the tests are thorough. Two things worth a look inline before merge.

xieofxie:
need to rethink about how to detect if the model is generated by winml cli

src/winml/modelkit/inspect/resolver.py
xieofxie: renamed back

src/winml/modelkit/utils/manifest.py
xieofxie: Addressed — _sanitize_value now also handles 
p.generic scalars via .item(), so numpy floats/ints/bools in xport_stats/nalyze_details serialize as native Python types instead of falling through to default=str. Added a test covering 
p.float32, 
p.int64, and 
p.bool_.

src/winml/modelkit/utils/manifest.py
```diff
@@ -0,0 +1,226 @@
...
+            stages.append(ManifestStage(**known_stage, extras=s))
+
+        known = set(cls.__dataclass_fields__) - {"stages", "extras"}
+        known_kwargs = {k: data.pop(k) for k in list(data) if k in known}
+        extras = data  # anything left is forward-compat extras
+
+        return cls(stages=stages, extras=extras, **known_kwargs)
+
+    @classmethod
+    def find(cls, directory: Path) -> list[WinMLManifest]:
```
DingmaomaoBJTU, line 189: `find()`, `load()`, and `from_dict()` have no production callers. `inspect/resolver.py`, `serve/app.py`, and `inference/engine.py` all still hand-parse manifests with raw `json.loads(...)` + manual dict access, and the build producers (`build/hf.py`, `build/onnx.py`) plus these consumers re-spell `"build_manifest.json"` / `"export_manifest.json"` as string literals instead of importing `MANIFEST_FILENAME` / `EXPORT_MANIFEST_FILENAME`.

So the read side of this "single source of truth" is effectively dead code (exercised only by tests), while the module docstring claims *all consumers should use load()/find()*. The filename stays duplicated in 5+ places and the hand-rolled parsers can silently drift from this dataclass on a schema change. Suggest either wiring the consumers onto `WinMLManifest.load()`/`find()` (and the constants), or trimming the unused surface and softening the docstring.

src/winml/modelkit/inspect/resolver.py
```diff
@@ -508,7 +508,9 @@ def resolve_cache(model_id: str) -> CacheInfo:
     # PRIMARY: Manifest-based resolution
     # -------------------------------------------------------------------------
     if model_dir.exists():
-        manifests = list(model_dir.glob("*build_manifest.json"))
+        manifests = list(model_dir.glob("*build_manifest.json")) + list(
```
DingmaomaoBJTU: Concrete cost of the read API above going unused: these glob literals plus the `json.loads` + manual `stages`/`filename` walk just below duplicate what `WinMLManifest.find()` / `load()` already do, so a schema change here drifts silently. `WinMLManifest.find(model_dir)` returning typed `ManifestStage`s would centralize it.

Separate question on the merge: build and export manifests are pooled and the newest-by-mtime wins. An `export_manifest.json` (single `export` stage) sitting next to build artifacts would then make `inspect` report optimize/quantize/compile as uncached. If build and export outputs never share a directory this is moot — is that guaranteed?

src/winml/modelkit/commands/export.py
```diff
@@ -59,7 +61,8 @@ def _delete_onnx_with_external_data(onnx_path: Path) -> None:
         logger.debug("Could not parse external data from %s", onnx_path, exc_info=True)
 
     if onnx_path.exists():
-        onnx_path.unlink()
+        with contextlib.suppress(OSError):
```
DingmaomaoBJTU: Nit: this now swallows a failed `unlink()` silently, unlike the sibling cleanup above (the `except Exception` block) which logs at debug on failure. A `logger.debug("Could not delete %s", onnx_path, exc_info=True)` inside the suppressed block would keep cleanup failures diagnosable and stay consistent with that pattern.

DingmaomaoBJTU (commented):
## Review — WinMLManifest refactor

Overall a clean, well-tested refactor: the naming split (`build_manifest.json` for build, `export_manifest.json` for export), the export-side manifest generation, and the composite prefix logic all look correct, and the earlier threads (numpy sanitization, filename revert, backward-compat glob) appear addressed.

Main structural feedback — the "single source of truth" is only half-wired: the **write** side goes through `WinMLManifest.save()`, but the **read** side (`find()`/`load()`/`from_dict()`) has no production callers and consumers still hand-parse JSON with duplicated filename literals. Details inline; nothing here is blocking.

🤖 Reviewed with GitHub Copilot CLI

src/winml/modelkit/inspect/resolver.py
xieofxie: good point! to keep it simple, let' remove the generation of export_manifest.json and let https://github.com/microsoft/winml-cli/issues/1038 fix it

src/winml/modelkit/commands/export.py
xieofxie: will revert changes in this file

src/winml/modelkit/inspect/resolver.py
xieofxie: Removed export manifest generation entirely — export.py is reverted to main. Export manifest support deferred to #1038. This keeps the PR as a pure refactor (WinMLManifest class for build_manifest.json only).

src/winml/modelkit/utils/manifest.py
xieofxie: Agreed — the read side (find/load/from_dict) is currently only exercised by tests. The consumers can adopt WinMLManifest.load()/find() incrementally in follow-up PRs. Softened the docstring to say 'should' rather than 'must' for consumers. The write side is fully wired (build/hf.py, build/onnx.py both use WinMLManifest.save()).

src/winml/modelkit/utils/manifest.py
```diff
@@ -0,0 +1,218 @@
...
+            if m is not None:
+                results.append(m)
+        return results
+
+    @classmethod
+    def _try_load(cls, path: Path) -> WinMLManifest | None:
+        """Load a manifest, returning ``None`` on read/parse failure."""
+        try:
+            return cls.load(path)
+        except (json.JSONDecodeError, OSError) as exc:
```
DingmaomaoBJTU: `_try_load` only catches `(json.JSONDecodeError, OSError)`, but `from_dict` raises **`TypeError`** for a file that is valid JSON yet missing a required field (`source` / `final_artifact`) — e.g. `WinMLManifest.__init__() missing 1 required positional argument: 'source'`. So a structurally-invalid manifest escapes here, contradicting the docstring ("returning None on read/parse failure").

This isn''t hypothetical for existing caches: the current `main` **hf** `build_manifest.json` writes no `source` key (only `onnx` builds did). After this PR, `WinMLManifest.load()`/`find()` on any pre-upgrade hf cache raises `TypeError`, and the consumers don''t catch it either — `resolve_cache` (`except (JSONDecodeError, KeyError, OSError)`) and serve''s `_manifest_from_engine` (`except (JSONDecodeError, OSError)`) both crash instead of degrading, and `engine._resolve_model_id_from_dir` (via `find()`) propagates it. Verified locally against the exact old-hf shape.

Note the asymmetry: unknown keys are absorbed into `extras` (forward-compat), but *missing* required keys hard-crash. Minimal fix: add `TypeError`/`ValueError` to this `except` (and/or default `source`/`final_artifact` to `None`) so old/partial manifests degrade instead of raising.

src/winml/modelkit/inspect/resolver.py
```diff
@@ -508,18 +509,20 @@ def resolve_cache(model_id: str) -> CacheInfo:
...
+        manifest_paths = sorted(
+            model_dir.glob(f"*{MANIFEST_FILENAME}"),
+            key=lambda p: p.stat().st_mtime,
+            reverse=True,
+        )
+        if manifest_paths:
             try:
-                manifest = json.loads(manifest_path.read_text())
-                manifest_stages = {s["name"]: s for s in manifest.get("stages", [])}
+                manifest = WinMLManifest.load(manifest_paths[0])
```
DingmaomaoBJTU, line 519: Follow-on from the `_try_load` note: this `WinMLManifest.load()` can raise `TypeError` (valid JSON, missing a required field — e.g. old hf manifests without `source`), but the surrounding `except (json.JSONDecodeError, KeyError, OSError)` on line 549 doesn''t catch it. The documented "fall through to filename scanning" path is then skipped and `resolve_cache` crashes — a regression vs. the previous `json.loads(...) + .get()`, which tolerated missing keys. Adding `TypeError`/`ValueError` here (and in serve''s `_manifest_from_engine`, same gap) restores the graceful degradation.

src/winml/modelkit/build/hf.py
```diff
@@ -168,13 +168,15 @@ def build_hf_model(
     def _name(base: str) -> str:
         return f"{cache_key}_{base}" if cache_key else base
 
+    from ..utils.manifest import MANIFEST_FILENAME
```
DingmaomaoBJTU: Nit (consistency): `MANIFEST_FILENAME` is imported via the deep `from ..utils.manifest import ...` while the very next manifest line uses `from ..utils import ManifestStage, WinMLManifest` through the package API — and both symbols are exported from `utils/__init__`, so `MANIFEST_FILENAME` could ride that same `from ..utils import` line. Same in `onnx.py`. The PR description mentions standardizing on `from ..utils import ...`, so this just keeps that consistent. Non-blocking.

DingmaomaoBJTU (commented):
## Follow-up review — update since `d2221e56`

The two new commits cleanly address my earlier feedback:
- **Export manifest generation removed** → PR is now a focused, pure refactor. This also drops the build/export manifest pooling + mtime-merge ambiguity I flagged, and reverts the `contextlib.suppress(OSError)` unlink from my third note.
- **Consumers wired** to `WinMLManifest` + `MANIFEST_FILENAME` (`resolver`, `serve`, `engine._resolve_model_id_from_dir`), so the read API is no longer test-only dead code and the filename literal is centralized.

One verified item remains — a robustness / backward-compat regression: `WinMLManifest.load()`/`find()` raise an **uncaught `TypeError`** on a valid-JSON manifest missing a required field, which is exactly the shape of pre-upgrade **hf** `build_manifest.json` files (no `source` key). Consumers don''t catch it, so `inspect`/`serve` crash instead of degrading. Details inline; the fix is ~1 line. Holding approval on that one point.

🤖 Reviewed with GitHub Copilot CLI

src/winml/modelkit/utils/manifest.py
xieofxie: yes, let's default source/final_artifact to None

src/winml/modelkit/build/hf.py
xieofxie: Fixed — consolidated into a single \rom ..utils import MANIFEST_FILENAME, ManifestStage, WinMLManifest\ line in both hf.py and onnx.py.

src/winml/modelkit/inspect/resolver.py
xieofxie: Added \TypeError\ to the except clause here and in serve's \_manifest_from_engine\. Graceful degradation is restored.

src/winml/modelkit/utils/manifest.py
xieofxie: Fixed — \source\ and \inal_artifact\ now default to \None\, and \_try_load\ catches \TypeError\/\ValueError\ so old/partial manifests gracefully degrade instead of crashing.

DingmaomaoBJTU (commented):
Nice refactor — centralizing the schema in `WinMLManifest` reads well, and the test coverage (round-trip, None-omission, forward-compat extras, sanitization, find) is solid. A few things worth a look before merge:

- **Robustness:** the two direct `WinMLManifest.load()` callers (`resolver.py`, `serve/app.py`) catch a narrower exception set than `_try_load`, so a valid-JSON-but-non-object manifest can raise an uncaught `ValueError` and crash them. I confirmed this locally.
- **Behavioral drift in a "pure refactor":** `serve` now returns `to_dict()` instead of the raw file, which reshapes the response and forces `schema_version: 1`; and `from_dict` silently downgrades a future-versioned manifest on re-serialise.
- **Single-source-of-truth gap:** the new `manifest_path_for()` helper has no production callers — the producers still build the path themselves.

Rest are minor style/scope notes inline. Leaving as COMMENT — not blocking.

src/winml/modelkit/inspect/resolver.py
```diff
@@ -543,8 +546,8 @@ def resolve_cache(model_id: str) -> CacheInfo:
                     total_cached=total_cached,
                     total_size_mb=round(total_size_mb, 2),
                 )
-            except (json.JSONDecodeError, KeyError, OSError) as exc:
-                logger.debug("Failed to read manifest %s: %s", manifest_path, exc)
+            except (json.JSONDecodeError, KeyError, TypeError, OSError) as exc:
```
DingmaomaoBJTU: `WinMLManifest.load()` can raise a bare `ValueError` that this tuple doesn't catch, so it would escape and crash `resolve_cache`. A manifest file that's valid JSON but not an object (e.g. a bare string) makes `from_dict` run `dict(data)`, which raises `ValueError` — not `JSONDecodeError` or `TypeError`. I confirmed this locally. `_try_load` already guards against it by catching `ValueError`; this direct caller doesn't. Suggest catching `ValueError` here (it also subsumes `json.JSONDecodeError`), or routing both direct-`load()` sites through one shared safe-loader so the caught set can't drift.

src/winml/modelkit/serve/app.py
```diff
@@ -741,11 +741,13 @@ async def cli_command(command: str, request: CliRequest) -> CliResponse:
...
-        manifest_file = Path(engine.model_path) / "build_manifest.json"
+        from ..utils.manifest import MANIFEST_FILENAME, WinMLManifest
+
+        manifest_file = Path(engine.model_path) / MANIFEST_FILENAME
         if manifest_file.exists():
             try:
-                return cast("dict[str, Any]", json.loads(manifest_file.read_text()))
-            except (json.JSONDecodeError, OSError) as e:
+                return WinMLManifest.load(manifest_file).to_dict()
+            except (json.JSONDecodeError, TypeError, OSError) as e:
```
DingmaomaoBJTU: Same `ValueError` gap as in `resolver.py`: on a valid-JSON-but-non-object manifest `load()` raises a plain `ValueError`, which isn't in this tuple, so the endpoint would 500 instead of falling back to the default. Please add `ValueError` (it also covers `json.JSONDecodeError`).

src/winml/modelkit/serve/app.py
```diff
@@ -741,11 +741,13 @@ async def cli_command(command: str, request: CliRequest) -> CliResponse:
...
     if engine.model_path:
-        manifest_file = Path(engine.model_path) / "build_manifest.json"
+        from ..utils.manifest import MANIFEST_FILENAME, WinMLManifest
+
+        manifest_file = Path(engine.model_path) / MANIFEST_FILENAME
         if manifest_file.exists():
             try:
-                return cast("dict[str, Any]", json.loads(manifest_file.read_text()))
-            except (json.JSONDecodeError, OSError) as e:
+                return WinMLManifest.load(manifest_file).to_dict()
```
DingmaomaoBJTU, line 749: Heads up: this quietly changes the served manifest shape vs. the previous raw `json.loads`. `to_dict()` drops `None` fields (so skipped stages lose `filename`/`elapsed_seconds`) and hard-codes `schema_version: 1` even if the on-disk file has a different version. For a pure refactor the `/manifest` response now differs. If unintended, return the raw dict; if intended, worth calling out in the PR.

src/winml/modelkit/utils/manifest.py
```diff
@@ -0,0 +1,218 @@
...
+    @classmethod
+    def load(cls, path: Path) -> WinMLManifest:
+        """Deserialise from a JSON file on disk."""
+        return cls.from_dict(json.loads(path.read_text()))
+
+    @classmethod
+    def from_dict(cls, data: dict[str, Any]) -> WinMLManifest:
+        """Construct from a raw JSON dict (as stored on disk)."""
+        data = dict(data)  # shallow copy — don't mutate caller's dict
+        data.pop("schema_version", None)
```
DingmaomaoBJTU: `from_dict` pops and discards `schema_version`, while `to_dict` always writes `SCHEMA_VERSION` (1). So loading a future v2 manifest and re-serialising silently rewrites it to v1 while keeping the v2 fields in `extras` — the version label then contradicts the content. Since forward-compat is an explicit goal here, consider preserving the original version (stash it in `extras`, or add a field) instead of dropping it.

src/winml/modelkit/utils/manifest.py
```diff
@@ -0,0 +1,218 @@
...
+    def save(self, path: Path) -> None:
+        """Write the manifest as indented JSON to *path*."""
+        path.write_text(json.dumps(self.to_dict(), indent=2, default=str))
+        logger.debug("Manifest persisted: %s", path)
+
+    # ------------------------------------------------------------------
+    # Factory helpers
+    # ------------------------------------------------------------------
+
+    @staticmethod
```
DingmaomaoBJTU: This helper encodes the `{prefix}_{filename}` convention, but the producers (`build/hf.py`, `build/onnx.py`) build the path via their own `_name()` closure + `MANIFEST_FILENAME` and never call it — so the path convention still lives in two places and this method has no production callers (only tests). Either wire it into the producers (`manifest_path_for(output_dir, prefix=cache_key)`) or drop it; as-is it undercuts the "single source of truth" goal.

src/winml/modelkit/utils/manifest.py
```diff
@@ -0,0 +1,218 @@
...
+    from pathlib import PurePath
+
+    if isinstance(value, PurePath):
+        return str(value)
+    if isinstance(value, dict):
+        return {k: _sanitize_value(v) for k, v in value.items()}
+    if isinstance(value, list):
+        return [_sanitize_value(v) for v in value]
+    # numpy scalars → native Python types (int, float, bool, etc.)
+    import numpy as np
```
DingmaomaoBJTU: `import numpy as np` runs on every (recursive) call of `_sanitize_value`. numpy is a required dependency now, so hoisting this to a module-level import avoids the repeated import lookups per value.

src/winml/modelkit/build/hf.py
```diff
@@ -168,13 +168,15 @@ def build_hf_model(
     def _name(base: str) -> str:
         return f"{cache_key}_{base}" if cache_key else base
 
+    from ..utils import MANIFEST_FILENAME, ManifestStage, WinMLManifest
```
DingmaomaoBJTU: Any reason this import is function-local? Nothing under `utils/` imports from `build/`, so there's no circular-import risk, and the rest of this module imports its `..`-siblings at module level. Module-level would be more consistent here (same applies to the matching line in `build/onnx.py`).

tests/unit/utils/test_manifest.py
```diff
@@ -0,0 +1,226 @@
...
+        )
+        path = tmp_path / MANIFEST_FILENAME
+        m.save(path)
+        data = json.loads(path.read_text())
+        assert isinstance(data["export_stats"]["output_path"], str)
+        assert data["export_stats"]["count"] == 5  # numeric stays numeric
+
+    def test_save_sanitizes_numpy_scalars(self, tmp_path: Path) -> None:
+        """Numpy scalars in export_stats are coerced to native Python types."""
+        np = pytest.importorskip("numpy")
```
DingmaomaoBJTU: numpy is a required dependency (this PR even removed the try/except around the numpy import for that reason), so `importorskip` would silently skip this test in a broken env rather than fail. Per the repo convention skips are for hardware/EP only — a plain `import numpy as np` is more appropriate.

tests/unit/commands/test_export.py
```diff
@@ -1118,6 +1118,9 @@ def fake_export_onnx(**kwargs):
             out.parent.mkdir(parents=True, exist_ok=True)
             out.write_bytes(b"onnx")
 
+        default_export_cfg = WinMLExportConfig()
```
DingmaomaoBJTU: This looks unrelated to the manifest refactor — it just hoists two configs into locals with no behavior change, and `commands/export.py` was reverted to main in this PR. Looks like leftover churn from the abandoned export-manifest work; consider dropping it to keep the diff focused.

src/winml/modelkit/serve/app.py
xieofxie: yes, if local has schema_version, we should keep it. else if it is none, we should use 1 as default

src/winml/modelkit/utils/manifest.py
xieofxie: let's drop it

src/winml/modelkit/utils/manifest.py
xieofxie: yes in deed

tests/unit/commands/test_export.py
xieofxie: Reverted — test_export.py now has zero diff vs main.

src/winml/modelkit/build/hf.py
xieofxie: Good catch — moved to module-level in both hf.py and onnx.py.

tests/unit/utils/test_manifest.py
xieofxie: Fixed — replaced \pytest.importorskip\ with a plain \import numpy as np\.

src/winml/modelkit/utils/manifest.py
xieofxie: Dropped the method. Producers continue using their own \_name()\ closure — single convention lives there.

src/winml/modelkit/utils/manifest.py
xieofxie: Hoisted to module-level.

src/winml/modelkit/utils/manifest.py
xieofxie: Fixed — \schema_version\ is now a dataclass field (default=1). \rom_dict\ preserves the on-disk value; \	o_dict\ writes \self.schema_version\. A v2 manifest round-trips without version downgrade.

src/winml/modelkit/serve/app.py
xieofxie: Addressed — \	o_dict()\ now uses the preserved \schema_version\ from disk (defaults to 1 if absent). The None-dropping behavior is intentional for a compact response.

src/winml/modelkit/serve/app.py
xieofxie: Added \ValueError\ to the except tuple.

src/winml/modelkit/inspect/resolver.py
xieofxie: Added \ValueError\ to the except tuple here as well.

src/winml/modelkit/utils/manifest.py
```diff
@@ -140,9 +140,10 @@ class WinMLManifest:
 
     def to_dict(self) -> dict[str, Any]:
         """Serialise to a JSON-safe dict (``schema_version`` injected)."""
-        d: dict[str, Any] = {"schema_version": SCHEMA_VERSION}
+        d: dict[str, Any] = {"schema_version": self.schema_version}
         raw = asdict(self)
         extras = raw.pop("extras", {})
+        raw.pop("schema_version", None)
```
xieofxie: why we pop this?

xieofxie: Good question — the pop was to avoid duplicating \schema_version\ in the output (it's already written as the first key on line 143). Clarified: now we \pop\ it from \
aw\ explicitly so it's clear it's consumed, not silently discarded.

DingmaomaoBJTU (approved):
Thanks for the quick turnaround — all the points from my previous review are addressed, and I verified the behavior locally:

- Both direct `WinMLManifest.load()` callers now catch `ValueError`, so a valid-JSON-but-non-object manifest degrades gracefully instead of crashing `resolve_cache` / the serve endpoint.
- `schema_version` is now a preserved dataclass field — confirmed a v2 manifest round-trips as `2` (no more silent downgrade), unversioned files still default to `1`, and top-level/stage `extras` are preserved.
- numpy import hoisted to module level, `..utils` imports moved to module scope in both build producers, unused `manifest_path_for` dropped, `importorskip` swapped for a plain import, and the unrelated `test_export.py` churn reverted.

Clean, well-tested refactor. LGTM 👍

PR #1058, opened 2026-07-06
feat: Add composite model support to build command

xieofxie:
## Summary

Adds composite (seq2seq) model detection and fan-out to `winml build`, mirroring the pattern already implemented in `winml export`. Previously, running `winml build` on a composite model like Qwen3 would attempt a single-model build and fail. Now, composite models are automatically detected and each sub-component is built independently into its own subdirectory.

### `src/winml/modelkit/commands/build.py`
- Added `_warn_partial_composite_build()` helper that warns when a composite build fails mid-run, listing completed components without deleting them (same safety semantics as export's `_warn_partial_composite`).
- In the single-model branch, added composite detection via `resolve_composite_components` with the same error handling as export:
  - `ValueError` → `UsageError` (e.g. ambiguous composite needing explicit `--task`)
  - `RuntimeError` re-raised (empty registry guard)
  - `OSError` suppressed (model ref not resolvable as HF config)
  - Unexpected errors → `ClickException`
- When composite is detected:
  - Rejects `--use-cache` (cache keys are ambiguous for multi-component builds)
  - Generates a config per component via `generate_build_config(task=component_task)`
  - Carries over `quant`/`compile` from the outer config (already patched by CLI overrides)
  - Runs `_run_single_build` per component into `output_dir/<component_name>/`
  - Tracks completed components for partial-failure warnings

### `tests/unit/commands/test_build.py`
- Added `TestBuildComposite` class with 6 tests:
  - Fan-out runs one `_run_single_build` per component into correct subdirectories
  - `--use-cache` rejected for composite models
  - `ValueError` during composite resolution surfaces as usage error
  - Unexpected errors during detection are surfaced, not masked
  - Partial build failure warns and keeps completed directories
  - Sub-model configs inherit `quant`/`compile` from outer config


DingmaomaoBJTU (commented):
I left a few inline comments around composite detection, component config metadata, optimization default behavior, and pattern removability.

src/winml/modelkit/commands/build.py
```diff
@@ -792,19 +816,106 @@ def _patch_device(cfg: WinMLBuildConfig) -> None:
...
+            # Detect composite pipeline (registry-driven, same pattern as
+            # export command). A composite fans out into one build per
+            # sub-component; a plain model builds to the single output dir.
+            components = None
+            if model and not model_is_onnx:
+                try:
+                    from ..loader.resolution import resolve_composite_components
+
+                    task_hint = config.loader.task if config.loader else None
+                    components = resolve_composite_components(
```
DingmaomaoBJTU, line 865: Passing `config.loader.task` here misses the auto-detected seq2seq composite path. `resolve_composite_components()` only bridges `text2text-generation` to composite components when `task=None`; with an auto-generated config, `loader.task` can be `text2text-generation`, so BART/T5/Marian builds can fall through to the single-model path. This also ignores `config.loader.model_type`, so explicit model-type overrides will not be honored. Can we call this with `task=None` for auto-detected configs and pass `model_type=config.loader.model_type` when present?

src/winml/modelkit/commands/build.py
```diff
@@ -792,19 +816,106 @@ def _patch_device(cfg: WinMLBuildConfig) -> None:
...
+                            model,
+                            task=component_task,
+                            trust_remote_code=trust_remote_code,
+                            device=device,
+                            precision=precision,
+                            ep=ep,
+                        )
+                        # Carry over quant/compile from the outer config (already
+                        # patched by CLI overrides like --no-quant / --no-compile).
+                        component_config.quant = config.quant
```
DingmaomaoBJTU: Overwriting the generated component config's `quant` object with the outer config loses the component-specific `task`, `model_id`, and `model_type` populated by `generate_build_config()`. That means encoder/decoder calibration can run with the outer task metadata instead of the sub-model metadata, and the same mutable quant object is shared across sub-builds. Please deep-copy the outer settings that need to carry over and preserve or restore the component metadata.

src/winml/modelkit/optim/api.py
```diff
@@ -165,6 +165,7 @@ def optimize_onnx(
     output: str | Path | None = None,
     *,
     config: str | Path | dict[str, Any] | None = None,
+    passes: int = 1,
```
DingmaomaoBJTU: This changes the public `optimize_onnx()` default from two optimizer runs to one. Existing callers such as `build/common.py` still call it without `passes`, so normal builds will now get less optimization than before. If a narrower path needs one pass, can we keep the default behavior at 2 and pass `passes=1` only at that call site?

src/winml/modelkit/pattern/base.py
```diff
@@ -1760,6 +1758,9 @@ def _match_single_skeleton(
                             # src_node_matched = False
                             continue
                         edge_info = self._get_registered_edge_info(input_edge, node_name)
+                        if edge_info is None:
```
DingmaomaoBJTU, line 1771: Now that missing edge registrations are treated as an ordinary non-match, removability still relies on `edge_info_by_name` being complete. If an intermediate output has an unregistered external consumer, `_compute_removable()` sees no consumer and can mark the match removable, allowing the rewriter to delete nodes that still feed outside the match. Please derive consumers from `graph.node` inputs or conservatively mark matches non-removable when any registration is incomplete.

src/winml/modelkit/commands/build.py
xieofxie: yes, look like we should apply it following the same logic for non composite model.
- no quant / no compile - clear it
- if ctx set device, precision, overwrite else keep

src/winml/modelkit/optim/api.py
xieofxie: I will change it is default to 2

src/winml/modelkit/commands/build.py
xieofxie: Fixed — now passes model_type=config.loader.model_type to resolve_composite_components(). The task hint is still passed from the config (needed when the user explicitly sets a task), but model_type is forwarded so explicit overrides are honored.

src/winml/modelkit/commands/build.py
xieofxie: Fixed — now deep-copies the outer quant/compile settings and restores the component-specific task, model_id, and model_type that generate_build_config() populated. When the outer config has quant=None or compile=None (from --no-quant/--no-compile), the component config is cleared entirely.

src/winml/modelkit/optim/api.py
xieofxie: Fixed — default is now passes=2. The build pipeline (build/common.py) explicitly passes passes=1 since the autoconf loop already handles iterative re-optimization. The pytorch normalizer keeps the default passes=2.

src/winml/modelkit/pattern/base.py
xieofxie: Fixed — _compute_removable() now does a fallback scan over graph.node inputs to detect unregistered external consumers of intermediate tensors. If any graph node outside the match consumes an intermediate tensor, the match is conservatively marked non-removable.

src/winml/modelkit/commands/build.py
```diff
@@ -792,19 +816,124 @@ def _patch_device(cfg: WinMLBuildConfig) -> None:
...
+            components = None
+            if model and not model_is_onnx:
+                try:
+                    from ..loader.resolution import resolve_composite_components
+
+                    task_hint = config.loader.task if config.loader else None
+                    model_type_hint = config.loader.model_type if config.loader else None
+                    components = resolve_composite_components(
+                        model,
+                        task=task_hint,
```
DingmaomaoBJTU, line 867: This still passes the auto-generated loader task back into `resolve_composite_components()`. In the no-`-c` build path there is no explicit `--task`, so `task_hint` is the auto-detected task; for seq2seq models that can be `text2text-generation`, which takes the resolver's explicit-task path and skips the seq2seq composite bridge. Can we only pass `task` for config-file/explicit-task configs, and pass `task=None` for auto-generated build configs while still forwarding `model_type`?

src/winml/modelkit/commands/build.py
```diff
@@ -792,19 +816,124 @@ def _patch_device(cfg: WinMLBuildConfig) -> None:
...
+                        )
+                        # Carry over quant/compile settings from the outer config
+                        # (already patched by CLI overrides like --no-quant /
+                        # --no-compile). Deep-copy to avoid sharing mutable state
+                        # across sub-builds, and preserve component-specific
+                        # metadata (task, model_id, model_type) that
+                        # generate_build_config populated.
+                        if config.quant is None:
+                            component_config.quant = None
+                        elif component_config.quant is not None:
```
DingmaomaoBJTU: This still drops outer quant settings when the generated component config has `quant=None`. For example, a config-file build with static quant settings can enter this branch with `config.quant` populated while `generate_build_config()` returns a component config without quant for the current device/precision policy; the outer quant config is then silently ignored. Please deep-copy the outer quant settings whenever `config.quant` is not `None`, and then restore component `task`, `model_id`, and `model_type` metadata.

src/winml/modelkit/commands/build.py
```diff
@@ -792,19 +816,124 @@ def _patch_device(cfg: WinMLBuildConfig) -> None:
...
+                            saved_model_id = component_config.quant.model_id
+                            saved_model_type = component_config.quant.model_type
+                            import copy
+
+                            component_config.quant = copy.deepcopy(config.quant)
+                            component_config.quant.task = saved_task
+                            component_config.quant.model_id = saved_model_id
+                            component_config.quant.model_type = saved_model_type
+
+                        if config.compile is None:
```
DingmaomaoBJTU, line 949: This only propagates the `--no-compile`/`compile=None` case. If the outer config has a non-default compile section, the component keeps whatever `generate_build_config()` produced instead of inheriting the user's compile settings. Please deep-copy `config.compile` when it is not `None` too, so composite builds preserve config-file compile options.

src/winml/modelkit/build/common.py
```diff
@@ -87,6 +87,7 @@ def run_optimize_analyze_loop(
     optimize_onnx(
         model=model_path,
         output=optimized_path,
+        passes=1,
```
DingmaomaoBJTU: This still changes build-pipeline behavior from the previous two optimizer runs to one. The autoconf loop does not cover all callers (`auto=False`, `max_optim_iterations=0`, or no discovered patterns), so normal builds can still get less optimization than before. If the intent is only to avoid extra passes in a specific loop iteration, can we keep the initial optimize call at the API default and use `passes=1` only for the explicit re-optimize loop where needed?

src/winml/modelkit/pattern/base.py
```diff
@@ -1624,6 +1626,20 @@ def _compute_removable(self, matched_nodes: list[str], skeleton_output: str) ->
 
                 # Check all consumers of this tensor
                 consumers = self.edge_info_by_name.get(output_tensor, {})
+
+                # If there's a producer registered but no consumers, yet the
+                # tensor name appears as an input to some graph node, edge
+                # registration is incomplete — conservatively not removable.
+                if not consumers and output_tensor in self.producer_lookup:
```
DingmaomaoBJTU: This fallback only runs when there are no registered consumers. If the edge map is partially incomplete (one consumer registered and another external consumer missing), `_compute_removable()` will trust the partial map and can still mark the match removable. Please compare against the actual graph-node inputs for every intermediate tensor, or at least mark the match non-removable whenever actual graph consumers differ from `edge_info_by_name`.

DingmaomaoBJTU (commented):
Thanks for the update. I re-reviewed the latest head and am not approving yet because a few of the earlier concerns still look partially unresolved.

src/winml/modelkit/build/common.py
xieofxie: reverted. This is for qwen3 as two optimizations will crash. Will fix in later pr

src/winml/modelkit/commands/build.py
xieofxie: Fixed in e8226ab. We now only forward config.loader.task when it came from an explicit -c config file. For auto-generated configs (no -c), we pass 	ask=None so 
esolve_composite_components uses the 
esolve_task().composite path that applies the seq2seq bridge. Verified winml build -m google-t5/t5-small -o temp/t5-build -v now detects the encoder/decoder composite and builds to completion. model_type is still forwarded for explicit overrides.

src/winml/modelkit/commands/build.py
xieofxie: Fixed in e8226ab. The carry-over now deep-copies the outer quant whenever config.quant is not None, regardless of whether the component config had a quant object. When the component did have one, we preserve its 	ask/model_id/model_type metadata and overlay the outer settings; otherwise we just take the deep-copied outer quant.

src/winml/modelkit/commands/build.py
xieofxie: Fixed in e8226ab. We now deep-copy config.compile into each component when it is not None, in addition to clearing it when the outer compile is None.

src/winml/modelkit/pattern/base.py
xieofxie: Fixed in e8226ab. _compute_removable no longer relies on dge_info_by_name for the consumer check. It collects the match's intermediate tensors (bailing early if any is a graph output) and then does a single authoritative scan over graph.node inputs to find any external consumer. This is independent of edge-registration completeness, so an unregistered external consumer can't wrongly mark the match removable.

src/winml/modelkit/optim/api.py
```diff
@@ -179,6 +180,9 @@ def optimize_onnx(
         config: Configuration source - JSON file path or capabilities dict.
             Values from config are overridden by **capabilities kwargs.
             JSON keys use kebab-case (e.g., "gelu-fusion").
+        passes: Number of optimization passes to run. Each pass runs the full
+            pipeline (shape inference + optimization pipes). A second pass can
+            stabilize the graph after initial rewrites. Defaults to 1.
```
DingmaomaoBJTU: The docstring says `Defaults to 1`, but the parameter is declared `passes: int = 2` (line 168) and the updated `test_accepts_string_path` asserts the default runs `optimizer.optimize` twice. Please change this to `Defaults to 2`.

src/winml/modelkit/optim/api.py
```diff
@@ -165,6 +165,7 @@ def optimize_onnx(
     output: str | Path | None = None,
     *,
     config: str | Path | dict[str, Any] | None = None,
+    passes: int = 2,
```
DingmaomaoBJTU: With the default at `2`, every current caller runs two passes: the build pipeline (`build/common.py:151`) calls `optimize_onnx(...)` without `passes`, and `export/pytorch.py:148` explicitly passes `passes=2` (equal to the default). So no caller ever uses a non-default value, and the build path still runs the same 2 optimize passes as before this PR — i.e. the `passes=1` mitigation for the Qwen3 `SymbolicShapeInference` segfault mentioned in the commit history isn't actually in effect.

If the pattern-matcher `assert` -> `None` fix is what really resolved that crash, is the `passes` knob still needed at all? If it is, did the build pipeline intend to pass `passes=1`? Either way it'd be good to make the intent explicit so this doesn't read as dead configuration (and the redundant `passes=2` in `pytorch.py` could then be dropped).

src/winml/modelkit/commands/build.py
```diff
@@ -792,19 +817,139 @@ def _patch_device(cfg: WinMLBuildConfig) -> None:
...
+                            # Preserve component metadata (if the component had a
+                            # quant config), then overlay the outer quant settings.
+                            saved_meta: tuple[Any, Any, Any] | None = None
+                            if component_config.quant is not None:
+                                saved_meta = (
+                                    component_config.quant.task,
+                                    component_config.quant.model_id,
+                                    component_config.quant.model_type,
+                                )
+                            component_config.quant = copy.deepcopy(config.quant)
```
DingmaomaoBJTU, line 942: Edge case in the quant carry-over: when the outer `config.quant` is set but the component's generated config has `quant is None`, `saved_meta` stays `None`, so `component_config.quant = deepcopy(config.quant)` leaves the **outer** model's `task` / `model_id` / `model_type` on the sub-model's quant config. Calibration for that sub-model would then run against the outer task/model instead of itself.

`test_composite_carries_outer_quant_when_component_quant_none` actually locks this in: the component's loader task is `feature-extraction`, yet the inherited quant keeps `task='text-generation'` and `model_id='outer/model'`. In the `else`/`saved_meta is None` branch, consider setting `task` / `model_id` / `model_type` from `component_config.loader` (and the `model` arg) so the sub-model calibrates against itself. Since `generate_build_config` uses the same device/precision/ep for every component this may be rare in practice, but the silent mislabel is worth guarding.

DingmaomaoBJTU (commented):
Reviewed the composite-build fan-out together with the optimize / pattern-matcher changes.

The composite detection, the flat `<component>_model.onnx` layout (which lines up with `InferenceEngine._find_build_artifacts`' `*_model.onnx` glob in `inference/engine.py`), the per-component `dict(extra_kwargs)` copy (avoids the `.pop()` in `_build_hf_pipeline` bleeding across sub-builds), and the `_warn_partial_composite_build` semantics all look correct. The pattern-matcher assert -> `None` change and the graph-authoritative `_compute_removable` are also sound.

Three points below: one doc fix, one design question about the new `passes` parameter, and one latent quant-metadata gap for composite sub-models.

src/winml/modelkit/optim/api.py
xieofxie: good point, reverted and tracked via https://github.com/microsoft/winml-cli/issues/1077

src/winml/modelkit/commands/build.py
xieofxie: Good catch. Fixed in ecc8142. In the saved_meta is None branch (component produced no quant of its own) we now derive the calibration identity from the sub-model itself — 	ask from the component task, model_id from the model arg, and model_type from the component loader — instead of leaving the outer model's identity on the deep-copied quant. Updated the test to assert the sub-model's own task/model_id/model_type are used.

DingmaomaoBJTU (approved):
Re-reviewed the latest updates — all three points from my previous review are resolved:

1. **`passes` parameter (optim/api.py + export/pytorch.py)** — reverted entirely. `optimize_onnx` is back to main's unconditional double-optimize, and the Qwen3 segfault is handled by the pattern-matcher `assert`→`None` / graph-authoritative `_compute_removable` changes (which remain in place). This cleanly removes the dead config and the `Defaults to 1` docstring mismatch.

2. **Composite sub-model quant calibration identity (build.py)** — the `component_config.quant is None` branch now derives `task` / `model_id` / `model_type` from the sub-model's own `component_task` / `model` / `component_loader`, so calibration runs against the sub-model instead of inheriting the outer model's identity. `test_composite_carries_outer_quant_when_component_quant_none` asserts the corrected identity.

The composite fan-out, flat `<component>_model.onnx` layout (matches `InferenceEngine._find_build_artifacts`' `*_model.onnx` glob), per-component `dict(extra_kwargs)` isolation, and partial-failure warning all look correct. CI is green (lint, analyze, commands, models, optim, remaining). LGTM 👍

PR #1059, opened 2026-07-06
feat(perf): support composite model generation benchmarking

### LINKED ISSUE #1028: feat: perf supports ry style LLM
after https://github.com/microsoft/winml-cli/pull/1015

xieofxie:
Close https://github.com/microsoft/winml-cli/issues/1028

## Summary

Updates `winml perf` to support composite models (e.g. Qwen decoder-only) loaded via `WinMLCompositeModel.from_onnx`, benchmarked with `model.generate()` using text prompts — matching `eval.py`'s composite model CLI pattern.

### CLI
- `--model` now accepts `multiple=True` with `role=path` syntax for composite models  
  (e.g. `-m decoder_prefill=prefill.onnx -m decoder_gen=gen.onnx`)
- Added `--model-id` for HF config/tokenizer resolution (ported from `eval.py`)
- `--prompt` and `--max-new-tokens` now work for WinML composite generation (previously genai-only)

### Engine
- `_resolve_model_path()` — ported from `eval.py` for consistent composite `-m role=path` parsing
- `_load_model()` — handles dict `model_path` → `WinMLAutoModel.from_onnx(dict, hf_config=...)`
- `_is_generative_composite` — detects `WinMLDecoderOnlyModel` (e.g. Qwen)
- `_run_generation()` — benchmarks via `model.generate()` with tokenized prompt, reports TTFT, decode tokens/sec, TPOT

### Data classes
- `BenchmarkConfig` — added `model_path`, `hf_model_id`, `prompt`, `max_new_tokens`
- `GenerationBenchmarkResult` — LLM-style metrics matching `GenaiBenchmarkResult`'s structure
- `display_generation_report()` — console output for generation results

## Usage

```bash
winml perf \
  -m decoder_prefill=output/prefill/model.onnx \
  -m decoder_gen=output/gen/model.onnx \
  --model-id Qwen/Qwen3-0.6B \
  --task text-generation \
  --prompt "Hello, how are you?" \
  --max-new-tokens 50 \
  --iterations 10 --warmup 2
```

## Testing

All existing tests pass (49 CLI tests, 12 composite tests, 41 genai perf tests, 15 auto_onnx tests, 5 qwen3 composite tests).

### DIFF
--- a/src/winml/modelkit/commands/perf.py
+++ b/src/winml/modelkit/commands/perf.py
@@ -2343,7 +2343,7 @@ def perf(
                 )
 
             profiler = tracer_cls(
-                Path(onnx_for_trace),
+                Path(cast("str", onnx_for_trace)),
                 output_dir=output_dir,
                 level=op_tracing,
             )
--- a/src/winml/modelkit/commands/perf.py
+++ b/src/winml/modelkit/commands/perf.py
@@ -32,6 +32,10 @@ from ..utils import cli as cli_utils
 from ..utils.constants import EPName, EPNameOrAlias
 from ..utils.logging import configure_logging
 from ._live_chart import LiveMonitorDisplay
+from ._perf_generation import (
+    GenerationBenchmarkResult,
+    display_generation_report,
+)
 
 
 if TYPE_CHECKING:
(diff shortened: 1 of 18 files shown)


src/winml/modelkit/commands/perf.py
```diff
@@ -101,6 +101,15 @@ class BenchmarkConfig:
     ep_options: dict[str, str] | None = None
     shape_config: dict | None = None
 
+    # Composite model support: dict of {role: onnx_path} for from_onnx dispatch
+    model_path: str | dict[str, str] | None = None
+    # HF model ID for composite models loaded from ONNX (needed for hf_config)
+    hf_model_id: str | None = None
+
+    # Generation benchmark (decoder-only composite models)
+    prompt: str = "Explain the theory of relativity in simple terms."
```
xieofxie: use a same var to share with click option or use as str | None = None?

src/winml/modelkit/commands/perf.py
```diff
@@ -101,6 +101,15 @@ class BenchmarkConfig:
...
+    max_new_tokens: int = 128
```
xieofxie, line 118: do we also support apply_chat_template ?

src/winml/modelkit/commands/perf.py
```diff
@@ -211,6 +220,89 @@ def to_dict(self) -> dict[str, Any]:
         return result
 
 
+@dataclass
+class GenerationBenchmarkResult:
```
xieofxie: could it be shared with winml-genai code ?

src/winml/modelkit/commands/perf.py
```diff
@@ -421,6 +513,17 @@ def _is_composite(self) -> bool:
...
+    @property
+    def _is_generative_composite(self) -> bool:
+        """True when the loaded model is a decoder-only composite (e.g. Qwen).
+
+        These models support ``model.generate()`` via ``GenerationMixin`` and
+        should be benchmarked with text prompts rather than random inputs.
+        """
+        from ..models.winml.decoder_only import WinMLDecoderOnlyModel
+
+        return isinstance(self._model, WinMLDecoderOnlyModel)
```
xieofxie: should we use class or task? which is better

src/winml/modelkit/commands/perf.py
```diff
@@ -1298,7 +1550,40 @@ def display_console_report(result: BenchmarkResult, console: Console) -> None:
     console.print()
 
 
-def write_json_report(result: BenchmarkResult, output_path: Path) -> None:
+def display_generation_report(result: GenerationBenchmarkResult, console: Console) -> None:
```
xieofxie: could it be shared with winml-genai code path?

src/winml/modelkit/commands/perf.py
```diff
@@ -1623,8 +1914,96 @@ def _run_genai_runtime(ctx: click.Context, *, console: Console, json_mode: bool)
     run_genai_perf(config, console=console, json_mode=json_mode)
 
 
+def _resolve_model_path(
```
xieofxie, line 1866: could we share it with eval command?

src/winml/modelkit/commands/perf.py
```diff
@@ -1874,34 +2266,34 @@
...
+    is_composite = isinstance(model_path, dict)
+    is_onnx = False
+    if not is_composite and model_path is not None:
+        _mi = cli_utils.classify_model_input(model_path)
+        is_onnx = _mi.kind is cli_utils.ModelInputKind.ONNX_FILE
+    elif not is_composite and hf_model_id is not None and model_path is None:
+        try:
+            _mi = cli_utils.classify_model_input(hf_model)
+            is_onnx = _mi.kind is cli_utils.ModelInputKind.ONNX_FILE
+        except click.UsageError:
```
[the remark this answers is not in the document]
xieofxie: Fixed — added an explanatory comment for why we pass.

src/winml/modelkit/commands/perf.py
xieofxie: Changed to \str | None = None\ in the dataclass. Extracted a shared \_DEFAULT_PROMPT\ constant used by both the dataclass fallback in \_run_generation()\ and the click option default.

src/winml/modelkit/commands/perf.py
xieofxie: Added \pply_template: bool = True\ to \BenchmarkConfig\. In \_run_generation()\, when enabled, the prompt is wrapped via \	okenizer.apply_chat_template()\ (matching the genai path's behavior). The \--apply-template/--no-apply-template\ CLI flag now works for both composite models and winml-genai bundles.

src/winml/modelkit/commands/perf.py
xieofxie: The result classes share the same metric fields but use different config types (BenchmarkConfig vs GenaiPerfConfig) and genai has extra fields (prefill_mean_ms, context_length, avg_token_latency_ms). Aligned the display function to match display_genai_report's TTFT table format for visual consistency. A shared GenerationMetrics protocol could unify them further in a follow-up.

src/winml/modelkit/commands/perf.py
xieofxie: Switched from isinstance(WinMLDecoderOnlyModel) to capability-based: \self._is_composite and hasattr(self._model, 'generate')\. This checks the actual capability (GenerationMixin) without importing the class, and is more future-proof — any composite that supports generate() gets the generation benchmark path.

src/winml/modelkit/commands/perf.py
xieofxie: Aligned the display format to match display_genai_report (TTFT percentile table + decode/total summary lines). The function signatures differ because they take different result types with different config objects. A shared display helper accepting a protocol could unify them in a follow-up.

src/winml/modelkit/commands/perf.py
xieofxie: Done — moved to \cli_utils.resolve_model_path()\ in utils/cli.py. Both eval.py and perf.py now delegate to the shared implementation.

src/winml/modelkit/commands/perf.py
xieofxie: yes, we should have a shared base. is it possible to add prefill_mean_ms, context_length, avg_token_latency_ms ? maybe even 0 if not valid

src/winml/modelkit/commands/perf.py
xieofxie: is there any shared part? you could adjust code in winml-genai to align them

src/winml/modelkit/commands/perf.py
xieofxie: Done — created \_perf_generation.py\ with a shared \GenerationBenchmarkResult\ that includes all fields from both runtimes: \prefill_mean_ms\, \context_length\, \vg_token_latency_ms\ (default 0 when unused). Both \_perf_genai.py\ and \perf.py\ now import from it. Runtime-specific JSON fields (compile, bundle_dir, etc.) go into \xtra_info\ which merges into \enchmark_info\ in \	o_dict()\.

src/winml/modelkit/commands/perf.py
xieofxie: Done — both runtimes now share \display_generation_report()\ from \_perf_generation.py\. The layout adapts based on which optional fields are populated (e.g. prefill line only shows when \prefill_mean_ms > 0\). \display_genai_report\ in \_perf_genai.py\ is now just an alias to the shared function.

xieofxie:
it is too slow, low pri

we should focus on direct model performance

PR #1066, opened 2026-07-07
feat(perf): support --input-data npz for real benchmark inputs

### LINKED ISSUE #1065: feat: profiling support npz, npy file as model input

xieofxie:
Closes #1065

## Summary

Adds a `--input-data` option to `winml perf` so users can benchmark with **real input tensors** loaded from a `.npz` file instead of the randomly generated inputs the command produces by default. This makes latency/throughput numbers reflect representative data and the correct dynamic shapes.

## Behavior

```bash
winml perf -m model.onnx --input-data inputs.npz
```

- **`.npz` only.** A single-array `.npy` carries no input names to bind against, so it's rejected with a message pointing users to `.npz`.
- **Strict validation:**
  - the archive's keys must exactly match the model's input names — any missing or unexpected key is an error (an unexpected key is usually a typo that would otherwise leave a required input unset);
  - each array's dtype must equal the model's expected input dtype.
- **Shapes come from the data.** `--batch-size` and `--shape-config` are ignored (with a warning) when `--input-data` is set. Throughput scaling stays correct because `effective_batch_size` reads the leading dim from the provided arrays.
- **Not supported** with `--module` (submodules have their own internal inputs) or `--runtime winml-genai` (driven by `--prompt`) — both raise a clear `UsageError`.

## Changes

- `load_input_data(path, io_config)` helper in `perf.py` (raises `click.UsageError` with actionable messages).
- `input_data` field on `BenchmarkConfig`; branch in `PerfBenchmark._generate_inputs` to load real inputs.
- CLI option wiring with the `--module` / `winml-genai` guards and batch/shape warnings.
- `docs/commands/perf.md`: flag table, example, and pitfall note.
- Tests in `tests/unit/commands/test_perf_cli.py`: helper happy path + missing/unexpected/dtype/.npy/non-npz errors, `_generate_inputs` using provided data (and effective batch derived from it), and the `--input-data` + `--module` CLI guard.

## Notes

- Shape correctness beyond dtype is left to ORT at runtime — a bad shape surfaces as a normal inference error rather than a re-implemented shape check.


src/winml/modelkit/commands/perf.py
```diff
@@ -297,6 +300,84 @@ def _resolve_shape(
...
+            f"Expected exactly: {expected_names}."
+        )
+
+    for name, expected_dtype in zip(expected_names, expected_types, strict=True):
+        got = provided[name].dtype
+        want = np.dtype(expected_dtype)
+        if got != want:
+            raise click.UsageError(
+                f"--input-data dtype mismatch for '{name}': got {got}, expected {want}."
+            )
```
DingmaomaoBJTU, line 384: `load_input_data` is stricter than the actual inference path, which may surprise users. `WinMLSession.run()` -> `_prepare_inputs()` silently casts inputs to the expected dtype (`arr.astype(expected_type)`), and `_validate_inputs()` only *warns* on unexpected keys — so inputs that run fine normally become a hard error here.

The dtype check in particular bites a common case: numpy defaults integer literals to int64 and floats to float64, so `np.savez('in.npz', input_ids=np.array([[101, 2054]]))` yields int64 and fails against an int32 `input_ids`, even though a normal run would have cast it. Consider casting-with-a-warning to match `_prepare_inputs`, or make the error tell the user to pre-cast (`.astype(np.int32)`). (The unexpected-key hard error is more defensible as typo detection, but it's still stricter than the session's warn-only behavior.)

tests/unit/commands/test_perf_cli.py
```diff
@@ -824,6 +826,125 @@ def test_to_dict_emits_effective_batch_size(self) -> None:
         assert info["effective_batch_size"] == 1
 
 
+# =============================================================================
+# --INPUT-DATA TESTS
+# =============================================================================
+
+
+class TestLoadInputData:
```
DingmaomaoBJTU, line 834: Nice coverage of the error paths. One gap: the `except` around `np.load` in `load_input_data` (the "Could not read --input-data file" branch) isn't exercised. Worth a quick test that writes a `.npz`-suffixed file with garbage bytes and asserts the `UsageError`:

```python
path = tmp_path / "corrupt.npz"
path.write_bytes(b"not an archive")
with pytest.raises(click.UsageError, match="Could not read"):
    load_input_data(path, self._IO)
```

DingmaomaoBJTU (commented):
Reviewed the `--input-data` feature — overall clean and well-tested. Strict validation with actionable messages, correct `--module` / `--runtime winml-genai` guards, `effective_batch_size` correctly derives the batch from the provided tensors, and the docs update is thorough.

Two inline notes below (dtype strictness + a small test-coverage gap), plus two cross-cutting observations that don't map onto a changed line:

1. **`--op-tracing` ignores `--input-data`.** The op-trace path builds its own session and generates random inputs (`QNNProfiler.run()` -> `_generate_inputs`), so `--input-data ... --op-tracing basic` measures latency on the real tensors but traces ops on random ones. Worth a warning when both are set, or a note in the docs.

2. **HuggingFace model IDs export with default shapes.** Because `--input-data` nulls `--shape-config` and ignores `--batch-size`, an HF model is exported with default shapes; real data only binds if the export kept the relevant axes dynamic. Fine for the common case (and already "left to ORT at runtime"), just worth being explicit about.

Not blocking — nice work.

src/winml/modelkit/commands/perf.py
xieofxie: Good catch, we should follow casting-with-a-warning to match _prepare_inputs

xieofxie: Done in c4c2d549. load_input_data now casts each array to the model's expected dtype with a warning (matching _prepare_inputs's silent cast) instead of hard-erroring. Missing/unexpected-key errors are kept as typo protection. Covered by the new 	est_dtype_cast_with_warning.

tests/unit/commands/test_perf_cli.py
xieofxie: Added test_corrupt_npz_errors, which writes a bogus .npz and asserts the Could-not-read UsageError is raised, covering that except branch.

xieofxie:
Addressed the review-body notes in c4c2d549:

- **op-tracing now honors `--input-data`.** `OpTracer`/`QNNProfiler` take an `input_data` arg; `QNNProfiler.run()` uses a new `_resolve_inputs()` that traces on the same real tensors as the benchmark when the keys match the traced session, and falls back to random inputs with a warning otherwise (e.g. a compiled context model with different input names). The perf op-tracing call site only passes the loaded inputs when `--input-data` was set, so plain op-tracing still uses fresh random data. Covered by three new `_resolve_inputs` tests.
- **HF default-shapes caveat** documented in `docs/commands/perf.md` (real data only binds if the export kept axes dynamic), alongside the dtype-casting behavior and the op-tracing note.

src/winml/modelkit/commands/perf.py
```diff
@@ -1895,6 +2017,11 @@ def perf(
                 f"Submodule benchmarking requires a HuggingFace model ID "
                 f"(e.g., 'microsoft/resnet-50'), not '{hf_model}'."
             )
+        if input_data is not None:
```
DingmaomaoBJTU, line 2032: `--input-data` is guarded for `--module` and `--runtime winml-genai`, but composite dual-encoder models (CLIP/SigLIP) are a third unsupported case that slips through.

Composite-ness is only known after `_load_model`, so this reaches `run()` → `_run_sub_models`, which spawns a child `PerfBenchmark(self.config)` per sub-model. Each child then hits `load_input_data(self.config.input_data, <sub-model io_config>)`, and a single `.npz` can't match two different encoders' input names — so at least one sub-model raises `UsageError`, which `_run_sub_models` re-wraps as `RuntimeError: Sub-model '…' failed: …` (a traceback rather than the clean up-front error the other two cases give).

Suggest guarding in `run()` before dispatching to `_run_sub_models`, e.g.:

```python
if self._is_composite and self.config.input_data is not None:
    raise click.UsageError(
        "--input-data is not supported for composite (dual-encoder) models; "
        "each sub-model has its own inputs a single .npz cannot address."
    )
```

src/winml/modelkit/commands/perf.py
```diff
@@ -1948,6 +2075,22 @@ def perf(
...
+    # (--batch-size, --shape-config) don't apply. Warn instead of silently
+    # ignoring so the user isn't surprised by the reported batch.
+    if input_data is not None:
+        if cli_utils.is_cli_provided(ctx, "batch_size"):
+            console.print(
+                "[yellow]Warning:[/yellow] --batch-size is ignored when "
+                "--input-data is set; the batch comes from the provided tensors."
+            )
+        if shape_config is not None:
+            console.print(
```
DingmaomaoBJTU: Minor: when both `--shape-config` and `--input-data` are passed, the block above already printed `Shape overrides: {…}` before this warning fires — so the output announces the overrides and then immediately says they're ignored. Consider skipping the shape-config parse/print (or moving this `input_data` check above it) when `--input-data` is set, so the user only sees the ignore-warning.

DingmaomaoBJTU (commented):
Reviewed the `--input-data` feature — overall solid. The dtype-cast-with-warning fix (mirroring `WinMLSession._prepare_inputs`) and the QNN op-tracing reuse with a key-mismatch fallback both read well, and the test coverage (helper happy/error paths, `_generate_inputs`, and the profiler `_resolve_inputs` cases) is thorough. Two comments inline: one functional gap for composite dual-encoder models, and one minor output nit.

src/winml/modelkit/commands/perf.py
xieofxie: Fixed in 56e9a489. Added a guard in `PerfBenchmark.run()` (after `_load_model`, before dispatching to `_run_sub_models`) that raises a clean `UsageError` for composite models when `input_data` is set, so the user gets the up-front message instead of the re-wrapped `RuntimeError`. Covered by `test_input_data_composite_model_rejected` and documented in the perf docs.

src/winml/modelkit/commands/perf.py
xieofxie: Fixed in 56e9a489. The shape-config parse/print is now skipped entirely when `--input-data` is set (the branch only emits the ignore-warning), so the output no longer announces `Shape overrides: {…}` and then contradicts it. Covered by `test_shape_config_ignored_suppresses_overrides_print`.

DingmaomaoBJTU (approved):
Both review comments addressed cleanly:

- **Composite models** — `run()` now rejects `--input-data` before dispatching to `_run_sub_models`, giving a clean `UsageError` instead of the re-wrapped `Sub-model '…' failed` RuntimeError. The comment explains why the guard can't live with the up-front `--module`/`--runtime` checks, and `test_input_data_composite_model_rejected` covers it.
- **Shape-config print nit** — the parse/print is now skipped entirely when `--input-data` is set, so the user sees a single ignore-warning rather than `Shape overrides: {…}` followed by an immediate contradiction. `test_shape_config_ignored_suppresses_overrides_print` covers it.

Docs updated to list composite (dual-encoder) models alongside `--module`/`winml-genai` as unsupported. Verified locally: `tests/unit/commands/test_perf_cli.py` + `tests/unit/optracing/test_qnn_profiler.py` — 89 passed, ruff clean. LGTM ✅

Issue #1069, opened 2026-07-07
--task rejects composite pipeline tasks (summarization/translation) that --list-tasks advertises


timenick:
## Problem

`winml inspect --list-tasks -m t5-small` lists `summarization` and `translation`, but `winml inspect -m t5-small --task summarization` is **rejected**:

```
Error: Invalid task 'summarization'. Valid: any-to-any, ... (40 total).
See 'winml inspect --list-tasks' for the full list.
```

The error message even points users back to `--list-tasks` — the exact list that contains the rejected value. Directly contradictory.

## Root cause

A validator/registry split:
- `_validate_task` (`src/winml/modelkit/commands/inspect.py:46`) gates `--task` on `KNOWN_TASKS` only.
- Composite pipeline tasks (`summarization`, `translation`, `table-question-answering`, …) are deliberately **excluded** from `KNOWN_TASKS` (`src/winml/modelkit/loader/task.py:104-110`) and live in a separate registry (`composite_pipeline_tasks`).

So the composite tasks — the *primary* thing a user wants from T5/BART — can be listed but not passed.

## Additional subtlety (needs investigation)

If validation is bypassed, `resolve_loader_config('t5-small', task='summarization')` resolves to `T5DecoderWrapper` with `composite=None` — i.e. it routes to the **decoder half**, not the full composite pipeline. So "just widen the validator" may be insufficient; composite resolution on the explicit-`--task` path needs to be verified/wired so `--task summarization` builds the whole pipeline.

## Scope

- `_validate_task` in `src/winml/modelkit/commands/inspect.py` (Click callback — note: no access to `-m`/`--model-type` at validation time, so it can only validate against a global set, not per-model).
- `_validate_task_supported_for_model` in `src/winml/modelkit/commands/build.py:235`.
- Verify `winml config` / `build` accept the composite task and build the full pipeline.

## Expected

`--task <composite-task>` (e.g. `summarization`, `translation`) is accepted wherever `--list-tasks` advertises it, and builds the composite pipeline. `--list-tasks` and valid `--task` inputs stay in sync.

Related: #1067 (the `--list-tasks` model-specific listing fix).

DingmaomaoBJTU:
## Triage Summary

| Field | Value |
|-------|-------|
| **Priority** | P1 — major feature broken |
| **Component** | commands/inspect, loader/task |
| **Owner** | @timenick |
| **Quality** | Yes |

[...]

PR #1070, opened 2026-07-07
fix(inspect): make --list-tasks -m <model> show model-specific tasks

### LINKED ISSUE #1067: winml inspect --list-tasks -m <model> ignores the model and dumps all tasks
## Problem

`winml inspect -m openai/clip-vit-base-patch16 --list-tasks` prints the entire global task taxonomy (~50 tasks) instead of the tasks that model actually supports.

## Cause

The `--list-tasks` branch in `src/winml/modelkit/commands/inspect.py` short-circuits before any model resolution and always dumps `KNOWN_TASKS`, ignoring `-m`/`--model-type`.

## Expected

When combined with `-m`/`--model-type`, `--list-tasks` should list only the tasks the model supports: Optimum-exportable tasks (via `get_supported_tasks`) plus registered composite pipeline tasks (via `composite_pipeline_tasks`).

Example — `winml inspect -m openai/clip-vit-base-patch16 --list-tasks` should show:

```
feature-extraction
image-classification
zero-shot-image-classification
```

Plain `winml inspect --list-tasks` (no model) keeps dumping the full taxonomy.

timenick:
## Summary

`winml inspect --list-tasks` previously ignored `-m`/`--model-type` and always dumped the full `KNOWN_TASKS` taxonomy (~40 tasks). When a model target is given, it now lists only the tasks that model actually supports.

- **Model-specific listing**: with `-m`/`--model-type`, list Optimum-exportable tasks intersected with `KNOWN_TASKS` (dropping `-with-past` KV-cache export flavors that aren't user-facing) unioned with registered composite pipeline tasks (via `composite_pipeline_tasks`).
- **Registration fix**: the listing now imports `winml.modelkit.models.hf` so ModelKit's `register_onnx_overwrite` tasks surface (e.g. `sam` → `mask-generation`). Importing only optimum's `model_configs` silently dropped them and was flaky depending on prior import order.
- **Fast path unchanged**: plain `winml inspect --list-tasks` (no model) still dumps the full taxonomy without importing transformers.

### Examples

```
$ winml inspect -m t5-small --list-tasks
feature-extraction
summarization
text2text-generation
translation

$ winml inspect --model-type sam --list-tasks
feature-extraction
mask-generation
```

Closes #1067

## Notes

- Until #1069 lands, `--list-tasks` still shows composite tasks like `summarization`/`translation` that `--task` currently rejects (validator/registry split). The listing is correct — accepting those in `--task` is tracked separately in #1069.


src/winml/modelkit/commands/inspect.py
```diff
@@ -57,6 +57,43 @@ def _validate_task(ctx: click.Context, param: click.Parameter, value: str | None
...
+    )
+    from ..loader.task import KNOWN_TASKS
+
+    # Intersect with KNOWN_TASKS to drop Optimum's `-with-past` export flavors,
+    # which are not user-facing WinML tasks. Composite tasks are added separately
+    # (they intentionally live outside KNOWN_TASKS), so union after the filter.
+    optimum_tasks = get_supported_tasks(model_type, resolve_optimum_library(model_type))
+    tasks: set[str] = {t for t in optimum_tasks if t in KNOWN_TASKS}
+    # Composite registrations key on the hyphenated, lowercased model_type.
+    tasks.update(composite_pipeline_tasks(model_type.lower().replace("_", "-")))
```
DingmaomaoBJTU, line 104: Normalization is asymmetric here: `composite_pipeline_tasks` gets `model_type.lower().replace("_", "-")`, but `get_supported_tasks` / `resolve_optimum_library` on line 90 get the raw `model_type`. Optimum's lookup is an exact dict match with no normalization, and its keys are lowercase but a *mix* of underscore (`gpt_neo`, and our own `depth_anything`) and hyphen (`vision-encoder-decoder`, `megatron-bert`). So a non-canonical `--model-type` silently drops the Optimum half: `--model-type depth-anything` returns nothing while `depth_anything` returns `depth-estimation`, and `--model-type BERT` comes back empty. Via `-m` it's fine since `hf_config.model_type` is already canonical. Can we normalize once and feed both calls the same value (and document that `--model-type` must be the canonical config form)?

src/winml/modelkit/commands/inspect.py
```diff
@@ -135,14 +172,39 @@ def inspect(
...
+            # hand-coded set directly from loader.task to keep this branch fast —
+            # going through _list_tasks_for_model pulls in transformers (~10s warm).
+            from ..loader.task import KNOWN_TASKS
+
+            for t in sorted(KNOWN_TASKS):
+                click.echo(t)
+            return
 
-        for t in sorted(KNOWN_TASKS):
+        for t in _list_tasks_for_model(resolved_type):
```
DingmaomaoBJTU: When `_list_tasks_for_model` returns `[]` we print nothing and exit 0, so a typo or wrong-separator `--model-type` is indistinguishable from success (`winml inspect --model-type asdf --list-tasks` -> no output, rc 0). `--task` fails loudly with suggestions; could we at least warn on stderr (or exit non-zero) when the resolved list is empty? The docstring's "(it genuinely has no exportable tasks)" conflates unknown/typo with genuinely-none.

tests/cli/test_inspect_cli.py
```diff
@@ -115,6 +115,59 @@ def test_list_tasks_is_sorted(self, extra_args: list[str]) -> None:
...
+    def test_list_tasks_model_type_is_strict_subset(self) -> None:
+        """--list-tasks --model-type clip must be a strict subset of the full taxonomy."""
+        full = set(self._task_lines(_run("--list-tasks").output))
+        result = _run("--list-tasks", "--model-type", "clip")
+        assert result.exit_code == 0
+        clip_tasks = self._task_lines(result.output)
+        assert clip_tasks, "Expected clip to have at least one task"
+        assert clip_tasks == sorted(clip_tasks)
+        clip_set = set(clip_tasks)
+        assert clip_set < full, "Model-specific tasks must be a strict subset of all tasks"
```
DingmaomaoBJTU: This strict-subset property only holds because clip's composite task (`zero-shot-image-classification`) happens to be in `KNOWN_TASKS`. Only the Optimum half is intersected with `KNOWN_TASKS`; composite tasks are unioned unfiltered -- so for `t5` the list includes `summarization`/`translation`, which are deliberately *outside* `KNOWN_TASKS` (and `test_list_tasks_excludes_with_past_flavors` asserts exactly that). The same `clip_set < full` assertion would fail there. Fine to keep for clip, but the message "Model-specific tasks must be a strict subset of all tasks" reads as a universal guarantee the feature doesn't make.

DingmaomaoBJTU (commented):
Took a close look -- the core change is solid and the fast-path split (plain `--list-tasks` still avoids transformers) is preserved cleanly. Nice catch on the `winml.modelkit.models.hf` import being required for the Optimum overwrites, and the tests pin the important behaviors. Three non-blocking notes inline: one robustness nit on `--model-type` normalization, one UX nit on silent empty output, and one test-scoping observation.

src/winml/modelkit/commands/inspect.py
timenick: Good catch — I dug into this and the asymmetry is real, but "normalize once and feed both" turns out to be infeasible: Optimum's `_SUPPORTED_MODEL_TYPE` is an exact dict lookup with **mixed** separator conventions — `gpt_neo`/`visual_bert`/`m2m_100` are underscored while `megatron-bert`/`xlm-roberta`/`vision-encoder-decoder` are hyphenated. So no single normalization can reconstruct the canonical key from an arbitrary variant. The only reliable key is the raw `config.model_type`, which the `-m` path already yields.

Resolved in `cc1c48a6`: documented that `--model-type` must be the canonical `config.model_type` (help text + docstring), and — since a non-canonical form now can't be silently absorbed — made the empty case fail loudly (see the L207 thread). `resolve_optimum_library` keeps the raw value too, since `WRAPPED_LIBRARY_MODEL_TYPES` is likewise keyed on the canonical form (`timm_wrapper`).

Generated with Claude Code

src/winml/modelkit/commands/inspect.py
timenick: Fixed in `cc1c48a6`. Empty results now raise a `ClickException` (rc 1). When the user passed `--model-type`, the message points at the canonical form (`gpt_neo`, not `gpt-neo`; `bert`, not `BERT`) since a typo/separator mismatch is the likeliest cause; when the type was resolved from `-m`, it names the source model. Added `test_list_tasks_unknown_model_type_fails_loudly` and `test_list_tasks_non_canonical_separator_fails_loudly`, and dropped the misleading "(it genuinely has no exportable tasks)" from the docstring.

Generated with Claude Code

tests/cli/test_inspect_cli.py
timenick: Agreed — corrected in `cc1c48a6`. The assertion message now reads "clip's tasks should be a strict subset of the full taxonomy" and the docstring spells out that this holds only because clip's composite task (`zero-shot-image-classification`) lives inside `KNOWN_TASKS`, explicitly noting it's not a general guarantee and that t5's `summarization`/`translation` would break it.

Generated with Claude Code

DingmaomaoBJTU (approved):

PR #1071, opened 2026-07-07
fix(tasks): accept composite pipeline tasks as explicit --task

### LINKED ISSUE #1069: --task rejects composite pipeline tasks (summarization/translation) that --list-tasks advertises

timenick:
## Summary

`winml inspect --list-tasks -m t5-small` advertises composite pipeline tasks like `summarization` / `translation`, but `--task summarization` was **rejected** — the error even pointed back to `--list-tasks`, which listed the value it just refused. This makes `--list-tasks` (advertised) and `--task` (accepted) consistent, across inspect / config / export / build / perf.

- **Validation** now accepts the union of `KNOWN_TASKS` and a new hand-coded `COMPOSITE_TASKS` constant (`loader/task.py`), locked in sync with the composite registry by a test. Composite tasks stay OUT of `KNOWN_TASKS` (cache-key/alias rules unchanged). In practice only `summarization`/`translation` were actually blocked — the other 4 composite tasks were already in `KNOWN_TASKS`.
- **`resolve_task`'s explicit-task branch** now tags the composite (via `resolve_composite`, exact-key) so `winml inspect --task summarization` renders the Composite Pipeline panel. `--task text2text-generation` stays `composite=None` (single-decoder export) — the granular-vs-composite distinction is preserved.
- **Pure composites** (`table-question-answering`, which has no single model class) now resolve a representative display class instead of raising `ValueError`.
- **build** validation (`_validate_task_supported_for_model`) accepts composite tasks registered for the architecture; genuinely unsupported combinations (e.g. `t5 --task image-classification`) are still rejected.
- config / export / perf already accepted composite `--task` (they route through `resolve_composite_components`) — no change needed there.

## Behavior

| command | before | after |
|---|---|---|
| `inspect -m t5-small --task summarization` | "Invalid task" | resolves, shows composite panel |
| `inspect -m bart --task table-question-answering` | `ValueError` (crash) | resolves, shows composite panel |
| `inspect -m t5-small --task text2text-generation` | single decoder | **unchanged** (single decoder, no composite) |
| `build -m t5-small --task summarization` | rejected | accepted |

Closes #1069


DingmaomaoBJTU (approved):
Reviewed — LGTM. The two-layer approach is right: widen validation (`KNOWN_TASKS | COMPOSITE_TASKS`) and tag the composite on the explicit `--task` resolve path so inspect renders the pipeline panel.

Verified:
- `COMPOSITE_TASKS` exactly matches the 6 distinct registered composite tasks, and `test_composite_tasks.py::test_matches_composite_registry` locks it in sync with the live registry — good mitigation for the hand-coded constant vs. keeping the `--list-tasks`/validation fast path transformers-free.
- Exact-key lookup on the original user string preserves the granular-vs-composite distinction (`summarization` -> composite, `text2text-generation` -> single decoder). Both cases are tested.
- Pure composite (`table-question-answering`) resolves a representative display class instead of raising ValueError.
- No regression risk on the build/export/config/perf paths: they never read `TaskResolution.composite`; composite fan-out is driven independently (registry membership in `models/auto.py` / `resolve_composite_components` in export+config). The build validation gate normalizes model_type to match registry keys.
- Both validators stay in sync (`get_known_tasks()` union updated for the resolver path too).
- CI green across lint + all test shards.

Nice, well-commented and thoroughly tested fix.

PR #1074, opened 2026-07-08
feat: Add dynamic axes export support

### LINKED ISSUE #929: feat: support dynamic axes / dynamic dimensions in `winml export`
## Summary

The `winml export` command (`src/winml/modelkit/commands/export.py`) has no first-class support for **dynamic axes / dynamic dimensions** in the exported ONNX. Everything the CLI surfaces is aimed at producing **static** shapes. Dynamic axes are only reachable by smuggling a raw `dynamic_axes` dict through an `--export-config` JSON, and the toolchain actively warns against it. We should decide on, and implement, proper dynamic-dimension support (e.g. dynamic batch / sequence length).

### The CLI exposes only static-shape controls
- No `--dynamic-axes` (or equivalent) flag exists on the `export` command.
- `--shape-config` (`export.py:126`) takes **concrete integer** overrides like `{"sequence_length": 2048, "height": 640}` — it *pins* symbolic dims to fixed values (the opposite of dynamic).
- `--input-specs` shapes are parsed as a fixed `tuple(spec["shape"])` (`export.py:334`) — concrete integers only, no symbolic axis names.

### The capability exists underneath but is not wired to the CLI
`WinMLExportConfig` already has a `dynamic_axes` field that is forwarded to `torch.onnx.export`:
- `src/winml/modelkit/export/config.py:83` — `dynamic_axes: dict[str, dict[int, str]] | None = None`
- `src/winml/modelkit/export/config.py:301` — `from_dict` reads `data.get("dynamic_axes")`
- `src/winml/modelkit/export/htp/exporter.py:442` — `if export_config.dynamic_axes: onnx_kwargs["dynamic_axes"] = export_config.dynamic_axes`

Because `export.py` loads the `--export-config` JSON into `config_kwargs` and builds the config via `WinMLExportConfig.from_dict(config_kwargs)` (`export.py:353`, `export.py:369`), a user **can** currently enable dynamic dims only by hand-writing a `"dynamic_axes": {...}` key in an `--export-config` JSON file. There is no dedicated, validated, discoverable option.

### The HTP/QNN path deliberately discourages dynamic dims
The default export path is built around static dimensions on purpose:
- `exporter.py:76-77` — *"dynamic_axes: Not set (defaults to None = static dimensions). This prevents dynamic batch which causes MatMulAddFusion failure."*
- `exporter.py:369-384` — after export it inspects the graph and **warns** if it finds a symbolic (`dim_param`) batch dimension, recommending removal of `dynamic_axes`.
- `config.py:144-148` — logs a warning if `dynamic_axes` marks axis 0 (batch) dynamic.

## Why this matters
Some downstream scenarios (variable batch size, variable sequence length, variable image resolution) require dynamic dimensions in the exported ONNX. Today users must reverse-engineer the `dynamic_axes` schema and bypass the intended CLI surface, with no validation and conflicting warnings.

## Proposed scope (to refine)
- [ ] Decide the supported surface: a dedicated `--dynamic-axes` option and/or a `dynamic` field per input in `--input-specs` / `--shape-config`.
- [ ] Allow symbolic dimension names (strings) in `InputTensorSpec.shape` instead of integers-only.
- [ ] Reconcile with the HTP/QNN static-shape requirement: when does dynamic break fusions (e.g. MatMulAddFusion), and should it be blocked / gated per EP?
- [ ] Update the warnings in `exporter.py` / `config.py` so intentional dynamic export isn't flagged as an error.
- [ ] Add tests covering a dynamic-axes export (none currently exercise this path through the CLI).
- [ ] Document the feature and its compatibility caveats.

## Open questions
- Should dynamic axes be opt-in per-EP (e.g. allowed for CPU/CUDA, blocked/warned for QNN)?
- What is the desired UX — explicit `dynamic_axes` dict, or higher-level flags like `--dynamic-batch` / `--dynamic-seq-len`?

xieofxie:
## Summary

- Add a first-class `--dynamic-axes` option for `winml export`.
- Allow symbolic string dimensions in `--input-specs` and infer matching dynamic ONNX input axes.
- Normalize/validate dynamic axes mappings, including JSON string axis keys.
- Keep static shapes as the QNN-safe default while making requested dynamic batch informational instead of warning-shaped.
- Document dynamic axes, symbolic input specs, and QNN compatibility caveats.

Fixes #929

## Validation

- `uv run ruff check --fix src\winml\modelkit\onnx\io.py src\winml\modelkit\export\config.py src\winml\modelkit\commands\export.py src\winml\modelkit\export\htp\exporter.py tests\unit\export\test_config_validation.py tests\unit\export\test_pytorch_export.py tests\unit\commands\test_export.py`
- `uv run pytest tests\unit\export\test_config_validation.py tests\unit\export\test_pytorch_export.py tests\unit\commands\test_export.py`

### DIFF
--- a/src/winml/modelkit/commands/export.py
+++ b/src/winml/modelkit/commands/export.py
@@ -85,6 +85,23 @@ def _warn_partial_composite(completed: list[Path]) -> None:
     )
 
 
+def _load_json_object(path: Path, option_name: str) -> dict:
+    """Load a JSON object from a CLI option path."""
+    try:
+        with path.open() as f:
+            data = json.load(f)
+    except json.JSONDecodeError as e:
+        raise click.ClickException(f"Invalid JSON in {option_name}: {path}: {e}") from e
+    except Exception as e:
+        raise click.ClickException(f"Failed to load {option_name}: {e}") from e
+
+    if not isinstance(data, dict):
+        raise click.ClickException(
+            f"{option_name} must contain a JSON object, got {type(data).__name__}"
+        )
+    return data
+
+
 @click.command()
 @cli_utils.model_option(
     required=True,
@@ -151,6 +168,15 @@ def _warn_partial_composite(completed: list[Path]) -> None:
     default=None,
     help='JSON with shape overrides (e.g., {"sequence_length": 2048, "height": 640}).',
 )
+@click.option(
+    "--dynamic-axes",
+    type=click.Path(exists=True, path_type=Path),
+    default=None,
+    help=(
+        "JSON dynamic axes mapping for ONNX export "
+        '(e.g., {"input_ids": {"0": "batch", "1": "sequence"}}).'
+    ),
+)
 @cli_utils.build_config_option()
 @cli_utils.verbosity_options()
 @cli_utils.no_color_option()
@@ -171,6 +197,7 @@ def export(
     input_specs: Path | None,
     export_config: Path | None,
     shape_config: Path | None,
+    dynamic_axes: Path | None,
     config_file: Path | None,
 ) -> None:
     r"""Export HuggingFace model to ONNX format with HTP.
@@ -213,6 +240,9 @@ def export(
 
         # Custom ONNX export configuration
         winml export -m bert-base-uncased -o bert.onnx --export-config config.json
+
+        # Dynamic dimensions from a dedicated JSON file
+        winml export -m bert-base-uncased -o bert.onnx --dynamic-axes dynamic_axes.json
     """
     # Classify the -m value once (existence-first). Export only works with
     # HuggingFace model IDs — reject ONNX files and folders early.
@@ -275,6 +305,8 @@ def export(
         console.print(f"[bold blue]Input specs:[/bold blue] {input_specs}")
     if export_config:
         console.print(f"[bold blue]Export config:[/bold blue] {export_config}")
+    if dynamic_axes:
+        console.print(f"[bold blue]Dynamic axes:[/bold blue] {dynamic_axes}")
 
     output_path = Path(output)
 
@@ -282,8 +314,7 @@ def export(
     export_config_dict: dict = {}
     if export_config:
         try:
-            with export_config.open() as f:
-                export_config_dict = json.load(f)
+            export_config_dict = _load_json_object(export_config, "--export-config")
             console.print(f"[dim]Loaded export config: {list(export_config_dict.keys())}[/dim]")
         except Exception as e:
             console.print(f"[bold red]Failed to load export config:[/bold red] {e}")
@@ -292,19 +323,13 @@ def export(
     # Load shape overrides from JSON (task-independent).
     shape_overrides = None
     if shape_config:
-        try:
-            with shape_config.open() as f:
-                shape_overrides = json.load(f)
-            if not isinstance(shape_overrides, dict):
-                raise click.ClickException(
-                    f"--shape-config must contain a JSON object, "
-                    f"got {type(shape_overrides).__name__}"
-                )
-        except json.JSONDecodeError as e:
-            raise click.ClickException(
-                f"Invalid JSON in --shape-config: {shape_config}: {e}"
-            ) from e
+        shape_overrides = _load_json_object(shape_config, "--shape-config")
         console.print(f"[dim]Shape overrides: {shape_overrides}[/dim]")
+
+    dynamic_axes_dict = None
+    if dynamic_axes:
+        dynamic_axes_dict = _load_json_object(dynamic_axes, "--dynamic-axes")
+        console.print(f"[dim]Dynamic axes: {dynamic_axes_dict}[/dim]")
 
     # One-time warnings (apply to every sub-model in a composite export).
     if torch_module:
@@ -423,6 +448,8 @@ def export(
             config_kwargs["verbose"] = bool(verbose)
         if cli_utils.is_cli_provided(ctx, "dynamo"):
             config_kwargs["dynamo"] = dynamo
+        if dynamic_axes_dict is not None:
+            config_kwargs["dynamic_axes"] = dynamic_axes_dict
 
         # Add input/output tensors if we resolved them
         if input_tensors:
--- a/src/winml/modelkit/commands/export.py
+++ b/src/winml/modelkit/commands/export.py
@@ -313,12 +313,8 @@ def export(
     # Load export configuration from JSON file if provided (task-independent).
     export_config_dict: dict = {}
     if export_config:
-        try:
-            export_config_dict = _load_json_object(export_config, "--export-config")
-            console.print(f"[dim]Loaded export config: {list(export_config_dict.keys())}[/dim]")
-        except Exception as e:
-            console.print(f"[bold red]Failed to load export config:[/bold red] {e}")
-            raise click.ClickException(f"Failed to load export config: {e}") from e
+        export_config_dict = _load_json_object(export_config, "--export-config")
+        console.print(f"[dim]Loaded export config: {list(export_config_dict.keys())}[/dim]")
 
     # Load shape overrides from JSON (task-independent).
     shape_overrides = None
--- a/src/winml/modelkit/config/build.py
+++ b/src/winml/modelkit/config/build.py
@@ -756,9 +756,7 @@ def generate_hf_build_config(
 
         # Extract input shapes and dtypes from export_config -- NO HARDCODED VALUES
         input_tensors = [t for t in (export_config.input_tensors or []) if t.shape is not None]
-        input_shapes = [
-            _concrete_input_shape(t.shape) for t in input_tensors if t.shape is not None
-        ]
+        input_shapes = [t._concrete_shape() for t in input_tensors]
         input_dtypes = [t.dtype for t in input_tensors]
         if not input_shapes:
             raise ValueError(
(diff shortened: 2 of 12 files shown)


src/winml/modelkit/export/config.py
```diff
@@ -146,7 +246,7 @@ def __post_init__(
                 if 0 in axes:  # Batch dimension is dynamic
                     logger.warning(
                         "Dynamic batch detected for input '%s'. "
-                        "This prevents MatMulAddFusion and causes BiasGelu!",
+                        "QNN optimizations such as MatMulAddFusion may require static batch.",
```
DingmaomaoBJTU, line 250: After `_merge_dynamic_axes`, `self.dynamic_axes` only ever contains axes the user explicitly opted into — either via `--dynamic-axes` or inferred from symbolic `--input-specs` shapes. So this `logger.warning` now fires on every export that uses the new feature as documented.

That's inconsistent with the exporter change in `_verify_onnx_export`, which emits an *info* (`"Dynamic batch confirmed ... as requested"`) for exactly this case. Net effect: for the same axis the user sees a WARNING at config time and a reassuring INFO at verify time.

Since there's no longer an "unrequested dynamic batch" path reaching this point, consider downgrading to `info`/`debug` (or dropping it) so the opt-in feature doesn't warn about itself.

src/winml/modelkit/export/config.py
```diff
@@ -28,6 +28,96 @@
...
+        return None
+
+    dynamic_axes: DynamicAxes = {}
+    for spec in input_tensors:
+        if not spec.name or not spec.shape:
+            continue
+        axes = {
+            axis: dim_name
+            for axis, dim_name in enumerate(spec.shape)
+            if isinstance(dim_name, str)
```
DingmaomaoBJTU: `_normalize_dynamic_axes` rejects blank axis names (`if not isinstance(dim_name, str) or not dim_name`), but this inference path only checks `isinstance(dim_name, str)`. An input spec like `{"x": {"shape": ["", 128]}}` would produce an unnamed dynamic axis `{0: ""}` and only fail later in `to_tensor()` -> `_concrete_shape()` with a less obvious "empty symbolic dim" error.

Suggest `if isinstance(dim_name, str) and dim_name:` here so validation stays consistent with the explicit path and fails early.

src/winml/modelkit/config/build.py
```diff
@@ -1258,6 +1260,11 @@ def _find_submodules_by_class(
     return results
 
 
+def _concrete_input_shape(shape: tuple[int | str, ...]) -> tuple[int, ...]:
```
DingmaomaoBJTU: This duplicates `InputTensorSpec._concrete_shape` (onnx/io.py). At the call site the argument is `t.shape` where `t` is already an `InputTensorSpec`, so you could call `t._concrete_shape()` and drop this helper.

Bonus: `_concrete_shape` also validates empty/invalid dims, whereas this version silently passes anything non-`str` through — so reusing it tightens validation too.

DingmaomaoBJTU (commented):
Nice, well-scoped addition — the docs, config validation, and the end-to-end symbolic-shape export test all look solid, and the layered `--dynamic-axes` > `--export-config` precedence is clear.

A few consistency nits below around the dynamic-batch messaging and symbolic-dim validation. Nothing blocking.

src/winml/modelkit/export/config.py
xieofxie: yes, we should align with _verify_onnx_export

src/winml/modelkit/export/config.py
xieofxie: updated

src/winml/modelkit/config/build.py
xieofxie: removed

src/winml/modelkit/commands/export.py
```diff
@@ -275,15 +305,16 @@ def export(
...
 
     output_path = Path(output)
 
     # Load export configuration from JSON file if provided (task-independent).
     export_config_dict: dict = {}
     if export_config:
         try:
-            with export_config.open() as f:
-                export_config_dict = json.load(f)
+            export_config_dict = _load_json_object(export_config, "--export-config")
```
DingmaomaoBJTU: Now that `_load_json_object` raises its own `ClickException` with a descriptive message, this surrounding try/except re-wraps it: a bad file here surfaces as `Failed to load export config: Invalid JSON in --export-config: ...` (prefix doubled), and the error prints twice — once via the red `console.print` and again from Click's default handler. `--shape-config` and `--dynamic-axes` just let `_load_json_object` propagate and read cleaner. Suggest dropping this try/except so all three options behave consistently.

src/winml/modelkit/config/build.py
```diff
@@ -756,7 +756,7 @@ class name. Uses torchinfo to discover submodules and infer
 
         # Extract input shapes and dtypes from export_config -- NO HARDCODED VALUES
         input_tensors = [t for t in (export_config.input_tensors or []) if t.shape is not None]
-        input_shapes = [t.shape for t in input_tensors if t.shape is not None]
+        input_shapes = [t._concrete_shape() for t in input_tensors]
```
DingmaomaoBJTU: This calls `InputTensorSpec._concrete_shape()` — a private method — from another module. It's now used both here and internally in `io.py`, so consider promoting it to a public method (e.g. `concrete_shape()`) instead of reaching into a `_`-prefixed API across module boundaries.

src/winml/modelkit/export/config.py
```diff
@@ -28,6 +28,97 @@
...
+    if not inferred:
+        return explicit
+
+    merged: DynamicAxes = {name: dict(axes) for name, axes in explicit.items()}
+    for tensor_name, axes in inferred.items():
+        merged_axes = merged.setdefault(tensor_name, {})
+        for axis, dim_name in axes.items():
+            existing = merged_axes.get(axis)
+            if existing is not None and existing != dim_name:
+                raise ValueError(
```
DingmaomaoBJTU, line 114: Quick check on the intended UX: if a symbolic dim from `--input-specs` and an explicit `dynamic_axes` (e.g. from `--export-config`) name the same axis differently, this raises and the user gets a hard `Configuration error`. Since both refer to the same axis, letting the explicit value win (or documenting the conflict rule) might be friendlier than failing. It's covered by a test so likely intentional — just flagging the behavior.

DingmaomaoBJTU (commented):
Took a pass over the dynamic axes support. The normalization / merge / inference split in `export/config.py` is clean, symbolic dims resolving to size 1 for the dummy tensor is a sensible default, and the test coverage is solid. A few non-blocking notes inline: an error-wrapping inconsistency in the `--export-config` path, a cross-module private-method call, and a small question about the conflict-on-merge UX. Nothing blocking from my side.

src/winml/modelkit/export/config.py
xieofxie: we should raise for conflict

src/winml/modelkit/commands/export.py
xieofxie: Fixed by letting _load_json_object propagate directly for --export-config, matching --shape-config and --dynamic-axes error handling.

src/winml/modelkit/config/build.py
xieofxie: Fixed by promoting the helper to InputTensorSpec.concrete_shape() and using that from both to_tensor() and the build config path.

src/winml/modelkit/commands/export.py
DingmaomaoBJTU: Confirmed fixed in c32287a — the try/except is gone and --export-config now propagates the ClickException from _load_json_object directly, consistent with --shape-config/--dynamic-axes. Test updated to assert the new message. Thanks!

src/winml/modelkit/config/build.py
DingmaomaoBJTU: Confirmed fixed in c32287a — concrete_shape() is now public and both call sites (io.py internal + build.py cross-module) use it. Clean. Thanks!

src/winml/modelkit/export/config.py
DingmaomaoBJTU: Sounds good — raising on conflict is a defensible choice, and the CLI surfaces it clearly as \Configuration error: Conflicting dynamic axis for '...' axis N: 'a' vs 'b'\. Agreed we can keep it. Thanks for confirming!

DingmaomaoBJTU (approved):
Did a full holistic pass over the complete change set (not just the incremental delta) and everything checks out. Approving. ✅

**What I verified this round:**
- All three of my previous comments were addressed cleanly — the `--export-config` error-handling is now consistent with `--shape-config`/`--dynamic-axes`, `concrete_shape()` was promoted to a public method used by both call sites, and the conflict-on-merge behavior is intentional with a clear CLI error message.
- Symbolic (`str`) dims don't leak into numeric code paths: the `batch_size` comparison in `__post_init__` guards with `isinstance(spec.shape[0], int)`, dummy generation goes through `concrete_shape()`, and the metadata embedding (`to_dict()` + `json.dumps`) round-trips symbolic shapes correctly.
- `dynamic_axes` int-key ↔ JSON string-key normalization round-trips through `to_dict`/`from_dict`.
- `_verify_onnx_export` and `__post_init__` are aligned: requested dynamic batch logs INFO, unexpected dynamic batch still WARNs.
- Explicit `--dynamic-axes` supports output tensors too (covered by tests).

**Checks:** `ruff` clean on all changed files; full CI green (lint, CodeQL, test commands/models/optim/remaining).

Nice, well-tested feature. Thanks for iterating on the feedback!

PR #1076, opened 2026-07-08
fix(sysinfo): filter device-qualified EP aliases from available EPs

### LINKED ISSUE #1075: bug: 'Available EPs' error lists device-qualified .AUTO alias (e.g. OpenVINOExecutionProvider.AUTO)
### Description
When a requested EP is unavailable, the error message lists ORT's
device-qualified aliases (e.g. `OpenVINOExecutionProvider.AUTO`) as if they
were real execution providers.

### Repro
Request an unavailable EP on a system with OpenVINO:

```
Requested EP 'qnn' is not available on this system. Available EPs:
['CPUExecutionProvider', 'DmlExecutionProvider', 'OpenVINOExecutionProvider',
 'OpenVINOExecutionProvider.AUTO']
```

`OpenVINOExecutionProvider.AUTO` is not a canonical EP name and should not appear.

### Root cause
`ort.get_ep_devices()` reports device-qualified aliases as distinct `ep_name`
values. `_get_device_ep_map_from_ort` (src/winml/modelkit/sysinfo/device.py)
consumed them raw, so they leaked into `_get_available_eps()` and the error
messages. The parallel path in commands/analyze.py already filtered these.

### Expected
```
Available EPs: ['CPUExecutionProvider', 'DmlExecutionProvider', 'OpenVINOExecutionProvider']
```

### Notes
Fix ready locally: normalize each `ep_name` via `normalize_ep_name` and drop
anything not in `SUPPORTED_EPS`, at the single source of truth.

timenick:
## Summary
- `ort.get_ep_devices()` reports device-qualified aliases (e.g. `OpenVINOExecutionProvider.AUTO`) as distinct `ep_name` values, which leaked into `_get_available_eps()` and the `"Available EPs: [...]"` error messages in `resolve_device`.
- `_get_device_ep_map_from_ort` now normalizes each `ep_name` and drops anything not in the canonical `SUPPORTED_EPS` set, matching the filtering already done in `commands/analyze.py`. Fixing at this single source of truth cleans up both the error message and device resolution.

Before:
```
Available EPs: ['CPUExecutionProvider', 'DmlExecutionProvider', 'OpenVINOExecutionProvider', 'OpenVINOExecutionProvider.AUTO']
```
After:
```
Available EPs: ['CPUExecutionProvider', 'DmlExecutionProvider', 'OpenVINOExecutionProvider']
```

Fixes #1075


src/winml/modelkit/sysinfo/device.py
```diff
@@ -119,8 +125,16 @@ def _get_device_ep_map_from_ort() -> dict[str, tuple[EPName, ...]]:
...
             device_name = DEVICE_TYPE_TO_DEVICE.get(ep_device.device.type)
-            if device_name is not None:
-                result.setdefault(device_name.lower(), []).append(ep_device.ep_name)
+            if device_name is None:
+                continue
+            # ORT reports device-qualified aliases (e.g. "OpenVINOExecutionProvider.AUTO")
+            # as distinct ep_name values. Normalize and drop anything that isn't a
+            # canonical EPName so downstream availability sets and error messages
+            # only list real providers.
+            ep_name = normalize_ep_name(ep_device.ep_name)
```
xieofxie: do we need normalize ?

timenick: See comments above. To drop the .auto EP

xieofxie: I mean we could assume ep_device.ep_name is also normalized, so we could just check against SUPPORTED_EPS

xieofxie (approved):

PR #1079, opened 2026-07-09
refactor: Merge cli_utils.classify_model_input into utils.model_input (#1078)

### LINKED ISSUE #1078: refactor: merge cli_utils.classify_model_input and src/winml/modelkit/utils/model_input.py
Introduced in https://github.com/microsoft/winml-cli/pull/582

xieofxie:
## Summary

Fixes #1078.

The CLI carried **two parallel `-m/--model` classifiers**:

- `utils/cli.py` — enum `ModelInputKind`, rich `ModelInput` (folder metadata), and a `classify_model_input` that **raises `click.UsageError`** on bad input. Used by the command modules.
- `utils/model_input.py` — string-literal kinds, leaner `ModelInput`, a **pure** `classify_model_input` (returns `invalid`, no raising) plus `resolve_model_input` (Hub download). Used by the library layer (`auto.py`, `engine.py`) and inspect's hub check.

`inspect` even ran **both** on the same value. This merges them into a single, canonical classifier.

## Changes

- **`utils/model_input.py`** is now the single classifier + resolver:
  - `ModelInputKind` is a `str`-based `Enum`, so both `kind is ModelInputKind.ONNX_FILE` and `kind == "hub_onnx"` keep working from one type.
  - `ModelInput` gains folder metadata (`is_hf_folder`, `folder_has_onnx`, `is_winml_cli_folder`) and an `error` message.
  - `classify_model_input` is **pure** (no I/O, never raises): it folds in `hub_onnx` detection, folder metadata, and the strict path / HF-id validation with friendly messages, returning `kind=INVALID` + `error` for bad input.
- **`utils/cli.py`** drops its duplicate enum/dataclass/classifier and re-exports the canonical ones. `is_onnx_file_path` / `normalize_model_arg` are unchanged.
- **Command callers** (`build`, `config`, `export`, `inspect`, `perf`) now raise `click.UsageError` on `kind=INVALID` using the classifier's `error` message. `inspect` collapses its double-classify into a single call and reports a missing local `.onnx` as `ONNX file not found`.

## Notes on reconciled semantics

- A non-`.onnx` path-shaped value with ≥2 `/` segments (e.g. `org/repo/path`) is now `INVALID` ("path does not exist") rather than being forwarded to the Hub as a doomed `hf_id`. Library pass-through (`resolve_model_input(x).local_path or x`) is unaffected for these inputs.
- The classifier stays existence-independent for the `.onnx` kind; `inspect` performs the existence check to surface the friendly "ONNX file not found" message.

All `tests/unit/utils` and `tests/unit/commands` suites pass (1140 tests).


src/winml/modelkit/commands/build.py
```diff
@@ -633,7 +634,7 @@ def build(
             # ``normalize_model_arg``), route to the ONNX config generator instead
             # of treating the path as a HuggingFace repo id (which would try to
             # load the .onnx file as a JSON config and crash).
-            if cli_utils.is_onnx_file_path(model):
+            if classify_model_input(model).kind is ModelInputKind.ONNX_FILE:
```
DingmaomaoBJTU: When `-c` isn't supplied, INVALID inputs never reach the guard added below at lines 722-725. `classify_model_input("nonexistent.onnx")` and `classify_model_input("has spaces")` return `kind=INVALID` (not `ONNX_FILE`), so they fall into the `else` branch and get handed to `generate_build_config()` -> `generate_hf_build_config()` as an HF id, which fails with a confusing config / "Build failed:" error instead of the friendly "ONNX file not found" / "not a valid HuggingFace identifier". That reintroduces the #553 behavior for the common config-less path, and the new regression test (`test_build_onnx_suffix_but_not_exists_raises`) only exercises the `-c` path, so it doesn't catch this.

Suggest classifying once here and rejecting INVALID before routing:

```python
model_input = classify_model_input(model)
if model_input.kind is ModelInputKind.INVALID:
    raise click.UsageError(model_input.error or f"Invalid model input: {model}")
if model_input.kind is ModelInputKind.ONNX_FILE:
    ...
```

Then the re-classification at lines 722-725 becomes redundant.

src/winml/modelkit/commands/inspect.py
```diff
@@ -246,30 +247,31 @@ def inspect(
...
+            # error) from an existing one (inspection simply isn't wired up yet).
+            from pathlib import Path
+
+            if model_input.local_path and not Path(model_input.local_path).exists():
+                raise click.ClickException(f"ONNX file not found: {model}")
+            raise click.ClickException(
+                "ONNX file inspection is not yet supported. "
+                "Use 'winml config -m model.onnx' for ONNX build config."
+            )
+        if model_input.kind is ModelInputKind.HUB_ONNX:
```
DingmaomaoBJTU: Minor: this `HUB_ONNX` branch raises the identical message as the existing-`ONNX_FILE` branch just above (lines 266-269). Could collapse the two into a single branch to drop the duplicated string. Non-blocking.

src/winml/modelkit/commands/perf.py
```diff
@@ -2092,8 +2093,10 @@ def perf(
     # Classify the -m value once (existence-first) so module mode and the
     # single-model path share one source of truth. Raises cleanly on a missing
     # .onnx or an invalid id instead of a confusing downstream config error.
-    model_input = cli_utils.classify_model_input(hf_model)
-    is_onnx = model_input.kind is cli_utils.ModelInputKind.ONNX_FILE
+    model_input = classify_model_input(hf_model)
+    if model_input.kind is ModelInputKind.INVALID:
```
DingmaomaoBJTU, line 2125: Only a **bare** `missing.onnx` reaches this `INVALID` guard. A path-shaped missing file — `./missing.onnx`, `/abs/missing.onnx`, `C:\...\missing.onnx` — classifies as `ONNX_FILE` (the pure classifier no longer requires existence, and `_is_local_path()` is true), so it skips this check and proceeds into the benchmark with a non-existent path, deferring to a downstream loader error.

Two things worth reconciling:
- The comment just above ("Raises cleanly on a missing .onnx ... instead of a confusing downstream config error") no longer holds for those path-shaped inputs.
- `inspect` re-checks existence to keep the friendly "ONNX file not found", but `perf`/`build`/`config` don't — so the error differs by command for the same input.

Either re-check existence here like `inspect` does, or soften the comment to reflect the intentional deferred-error design.

src/winml/modelkit/utils/model_input.py
```diff
@@ -9,104 +9,203 @@
...
+            and a ModelKit build manifest (a ``winml build`` output folder).
+        error: Human-readable reason the value is ``INVALID``. ``None`` for
+            every valid kind.
     """
 
     kind: ModelInputKind
     raw: str
     local_path: str | None = None
     hf_id: str | None = None
+    is_hf_folder: bool = False
```
DingmaomaoBJTU: `is_hf_folder`, `folder_has_onnx`, and `is_winml_cli_folder` are computed here and unit-tested, but no caller in `src/` actually reads them (already true on `main`). Since this PR's goal is to unify the classifiers, it may be a good moment to decide: keep them as intentional back-compat/public surface (worth a one-line note), or drop the carried-over dead classification work. Not blocking — just flagging it.

DingmaomaoBJTU (approved):
Approving — the classifier consolidation is a clear improvement and the core logic checks out (pure `classify_model_input`, str-based `ModelInputKind`, and the `resolve_model_input(x).local_path or x` library pass-through is genuinely unaffected by the `org/repo/path -> INVALID` change). Left a few inline notes as non-blocking follow-ups: the INVALID-guard ordering in `build`/`perf`/`config` (path-shaped or `-c`-less invalid inputs bypass the friendly guard), an `inspect` DRY nit, and some computed-but-unused folder-metadata fields.

tests/unit/utils/test_cli.py
```diff
@@ -28,6 +26,7 @@
     precision_option,
     quant_option,
 )
+from winml.modelkit.utils.model_input import ModelInputKind, classify_model_input
```
xieofxie: these tests should be moved into test_model_input.py

src/winml/modelkit/utils/model_input.py
xieofxie: yes, let's clean them up

src/winml/modelkit/commands/build.py
xieofxie: Fixed. Moved the classify + INVALID guard up to right after `normalize_model_arg`, so INVALID inputs are rejected before either the `-c` or config-less routing. The later re-classification is now removed. Added config-less regression tests (`test_build_configless_missing_onnx_raises`, `test_build_configless_invalid_id_raises`).

src/winml/modelkit/commands/inspect.py
xieofxie: Collapsed the `HUB_ONNX` and existing-`ONNX_FILE` branches into a single `kind in (ONNX_FILE, HUB_ONNX)` branch so the "not yet supported" message lives in one place.

src/winml/modelkit/commands/perf.py
xieofxie: Reconciled by re-checking existence here like `inspect` does: a path-shaped missing `.onnx` now raises a friendly "ONNX file not found" instead of deferring to a downstream loader error. Updated the comment to reflect that the classifier stays existence-agnostic and this check owns the friendly error.

src/winml/modelkit/utils/model_input.py
xieofxie: Cleaned up as agreed: dropped `is_hf_folder`, `folder_has_onnx`, and `is_winml_cli_folder` (plus the glob/manifest work in `_classify_local_dir`) since no `src/` caller reads them. Removed the corresponding folder-metadata tests.

tests/unit/utils/test_cli.py
xieofxie: Done. Removed the `TestClassifyModelInput` class from test_cli.py; classifier coverage now lives entirely in test_model_input.py (added the dots-in-name and dir-wins-over-hf-shape cases there).

DingmaomaoBJTU (approved):
Re-reviewed after the fixes in `928f2195e` (+ the `main` merge and export-test hardening in `23167ab8a`). Every thread from the previous pass is properly addressed and CI is green — approving.

Verified:
- **build.py (was blocking):** classify + INVALID guard moved up right after `normalize_model_arg`, so INVALID inputs are rejected before *both* the `-c` and config-less routing; the late re-classification is gone and routing reuses the up-front `model_input`. The config-less `#553` regression is closed, and the two new tests (`test_build_configless_missing_onnx_raises`, `test_build_configless_invalid_id_raises`) exercise exactly that path.
- **perf.py:** path-shaped missing `.onnx` now gets the friendly "ONNX file not found" via an existence re-check (`Path` is imported at module level); comment updated to match the existence-agnostic classifier.
- **inspect.py:** `ONNX_FILE`/`HUB_ONNX` collapse is behavior-preserving (existing file still falls through to "not yet supported"; missing file still raises "not found").
- **model_input.py:** `is_hf_folder`/`folder_has_onnx`/`is_winml_cli_folder` dropped cleanly — no dangling references in `src/` or `tests/`.
- **tests:** classifier coverage consolidated into `test_model_input.py`; all CI shards (commands/analyze/models/optim/remaining), lint, and CodeQL pass.

PR #1081, opened 2026-07-09
feat(build): one-command Qwen3 genai bundle via winml build (npu + qnn)

DingmaomaoBJTU:
## Why

Exporting Qwen3 as an onnxruntime-genai bundle for the NPU previously required a bespoke `scripts/qwen3.py` run that emitted raw ONNX graphs and left the genai assembly to manual steps. This makes `winml build` produce the full, runnable genai bundle in one command, while keeping the existing per-model output unchanged for every other target.

## What

For registered decoder-LLM families, `winml build -m <model> -o <dir> --ep qnn` targeted at the NPU now assembles a complete onnxruntime-genai bundle:

- `ctx.onnx` / `iter.onnx` - transformer prefill + decode on the NPU (QNN, w8a16)
- `embeddings.onnx` / `lm_head.onnx` - companions on CPU
- `genai_config.json` + tokenizer

The routing is **opt-in and non-destructive**: it triggers only when `--ep qnn` is passed explicitly together with an NPU target — either `--device npu`, or a `--device auto` that resolves to the NPU on this machine. Any other target (CPU, GPU, or an auto-detected NPU without `--ep qnn`) keeps today's stock composite build.

All model-specific detail lives in a data-only **genai-bundle recipe** registered by the model package (`GenaiBundleRecipe` + `build_genai_bundle`), so the `winml build` routing itself stays architecture-agnostic. `scripts/qwen3.py` is thinned to delegate to the same orchestrator, and registering a recipe for another decoder family is all that is needed to give it the same one-command build.

## Reproduce (NPU + CPU hybrid)

The bundle runs as an **NPU + CPU hybrid**: the transformer stages (`ctx` / `iter`) execute on the QNN HTP while `embeddings` and `lm_head` run on CPU. The split is encoded per stage in the bundle's `genai_config.json`, so `--device auto` ("mixed") honors it without any extra flags.

```bash
# 1. Build the bundle in one command. --device auto resolves to the NPU here;
#    --device npu is equivalent.
winml build -m Qwen/Qwen3-0.6B -o out/qwen3-bundle --device auto --ep qnn

# 2. Pre-compile each QNN stage to an EPContext binary under _compiled/.
#    This may exit with a native teardown crash *after* the artifacts are
#    written to disk; the salvage path recovers them (see Notable).
winml perf -m out/qwen3-bundle --runtime winml-genai --device auto --compile

# 3. Benchmark the pre-compiled bundle in a fresh process (no --compile).
#    Loading _compiled/ directly is the stable path — it skips the fragile
#    in-process compile-then-load.
winml perf -m out/qwen3-bundle/_compiled --runtime winml-genai --device auto \
    --iterations 10 --warmup 2 --max-new-tokens 64
```

Observed on a Snapdragon X Elite NPU (step 3, hybrid, HTP burst mode): **TTFT ~136 ms, decode ~21.6 tok/s (~46 ms/token)**, stable across repeated runs.

## Notable

- **Routing bug fix (found during e2e).** `WinMLAutoModel.from_pretrained` computed an explicit `model_type` override but then delegated to the base composite `from_pretrained`, which re-derived the *native* model_type and built the stock graph, breaking quantization/calibration. It now resolves the concrete composite class for the overridden type. This is the fix that actually lets the surgical bundle build succeed.
- **Compile once, then run directly.** onnxruntime-genai compiles the QNN context in-memory at model creation, which faults before the first token when done in the same process that also just prepared the bundle. The reliable pattern is the two-step workflow above: `--compile` once to produce `_compiled/`, then benchmark `-m <bundle>/_compiled` in a fresh process. Documented in the sample.
- **Teardown crash is now salvaged.** The `--compile` step may still exit with a native `0xC0000374` / `0xC0000005` during QNN-EP teardown, *after* each stage's EPContext is written. The salvage path recovers the valid `_compiled/` artifacts from that crash, so the subsequent direct run succeeds. The crash originates in the native runtime below winml-cli and does not affect the generated tokens or the saved perf metrics.

## Verification

- Scoped unit tests green across the genai-bundle registry/orchestrator, build routing (including `--device auto`), the script `device`→EP mapping, and the `--no-quant` rejection guard; ruff clean.
- End to end on a Snapdragon X Elite NPU via the two-step workflow above: `winml build ... --ep qnn` produced the bundle, `--compile` wrote `_compiled/`, and the fresh-process run generated tokens on the QNN HTP as an NPU+CPU hybrid (**TTFT ~136 ms, decode ~21.6 tok/s**).


src/winml/modelkit/commands/build.py
```diff
@@ -441,6 +441,83 @@ def _validate_loader_tasks_for_model(
     return hf_config
 
 
+def _maybe_build_genai_bundle(
```
xieofxie, line 444: a little too specific.. but have the best flexibility

xieofxie: maybe use a runtime flag instead (serve as --target)? 
`--runtime winml / winml-genai`
So in the future if we have more recipes for each, user could easily switch instead of automatically switch and hide another

scripts/qwen3.py
```diff
@@ -159,87 +104,35 @@ def _add_export_parser(sub: argparse._SubParsersAction) -> None:  # type: ignore
...
     t0 = time.monotonic()
-
-    # --- Transformer (ctx + iter) ---
-    print(f"\n=== building transformer stages (device={args.device}) ===")
-    transformer = WinMLQwen3TransformerOnlyModel.from_pretrained(
+    config_path = build_genai_bundle(
         args.model_id,
+        args.output,
+        QWEN3_GENAI_BUNDLE_RECIPE,
+        ep="qnn" if args.device == "npu" else args.device,
```
timenick: Before this refactor, `_DEVICE_TO_EP` mapped `gpu` to `DmlExecutionProvider`. This branch now passes the device token itself (`ep="gpu"`) into `build_genai_bundle`. `normalize_ep_name()` has no `gpu` EP alias and intentionally returns unknown values unchanged, so the downstream precision resolver rejects this path with `Unknown EP 'gpu'`. As a result, `scripts/qwen3.py export --device gpu` regresses while the CPU and NPU choices still work.

Could we preserve device-to-EP resolution here (or accept an explicit EP) and add coverage for the GPU branch?

src/winml/modelkit/commands/build.py
```diff
@@ -441,6 +441,83 @@ def _validate_loader_tasks_for_model(
...
+        raise click.UsageError(
+            "genai bundle output is a directory; pass --output-dir, not --use-cache."
+        )
+    if not output_dir:
+        raise click.UsageError("--output-dir is required for a genai bundle build.")
+
+    bundle_dir = Path(output_dir)
+    override_precision = precision if cli_utils.is_cli_provided(ctx, "precision") else None
+    console.print(f"\n[bold blue]Genai bundle[/bold blue] ({model_type}): {model} -> {bundle_dir}")
+    config_path = build_genai_bundle(
```
timenick, line 556: This fast path runs after `config_or_configs` has been loaded and patched, but it does not forward that config or the pipeline controls derived from it. For example, `--no-quant` sets `config.quant = None` before this point, yet `override_precision` remains `None`; `build_genai_bundle` then falls back to the recipe's `w8a16` precision and quantizes anyway. A config file with `"quant": null` has the same problem (the new routing fixture uses exactly that config but only asserts that routing occurred).

Because this branch returns before the normal pipeline, the existing optimize/analyze/compile controls are bypassed as well. Could we either plumb the supported settings through, explicitly reject unsupported combinations, or restrict this fast path to the auto/default pipeline? Please also add a regression case for explicit `--no-quant` so a user-supplied control cannot become a silent no-op.

src/winml/modelkit/models/winml/genai_bundle.py
```diff
@@ -0,0 +1,333 @@
...
+# =========================================================================
+
+
+def _node_summary(path: str | Path) -> str:
+    """One-line op-type summary of an ONNX graph (loads shape metadata only)."""
+    model = onnx.load(str(path), load_external_data=False)
+    counts = collections.Counter(n.op_type for n in model.graph.node)
+    gqa_io: set[str] = set()
+    for node in model.graph.node:
+        if node.op_type == "GroupQueryAttention":
```
timenick: `_node_summary()` makes this otherwise generic orchestrator depend on specific Ops and topology: `GroupQueryAttention`, `QuantizeLinear` / `DequantizeLinear`, `Gather`, `MatMulNBits`, and QDQ connectivity to GQA. On `main`, this diagnostic lived only in the Qwen3 developer script; moving it into the generic source module crosses the recipe boundary and conflicts with the project's no-hardcoded-Op rule.

Could the Qwen3 recipe own an optional diagnostic callback instead, or could this become a genuinely generic Op-count summary?

src/winml/modelkit/commands/build.py
DingmaomaoBJTU: Good point on flexibility. For this PR the trigger is intentionally the `device`+`ep` pair (locked with the product owner): it's opt-in (`--ep qnn` must be passed explicitly) and non-destructive — every other target keeps the stock composite build, so nothing is silently swapped out from under the user.

I agree an explicit `--runtime` / `--target` selector is the better long-term shape once a single family has multiple recipes (where a device+ep pair can no longer name the intended output unambiguously). That's a broader CLI-surface change beyond this PR, so I filed #1090 to track the design discussion and kept the current implicit trigger here. Thanks!

src/winml/modelkit/commands/build.py
timenick: Thanks for rejecting explicitly supplied pipeline flags instead of silently ignoring them. The config-file half of this issue still remains, though: `-c config.json` is accepted (and the positive routing tests use it), but after `single_config.loader.model_type` is read, `build_genai_bundle()` receives none of that config. A user-supplied `"quant": null`, export shapes, optimization settings, or compile settings are therefore still silently discarded.

This also conflicts with the command documentation describing the config as a portable, self-contained pipeline specification. Could this guard reject an explicitly supplied `config_file`, or could the supported config fields be plumbed into the recipe build?

src/winml/modelkit/config/build.py
```diff
@@ -714,21 +714,28 @@ class name. Uses torchinfo to discover submodules and infer
...
         if parent_config.quant is None:
             parent_config.quant = WinMLQuantizationConfig()
         parent_config.quant.weight_type = policy.weight_type
         parent_config.quant.activation_type = policy.activation_type
     elif policy.precision == "fp16":
         # Pure FP16: no QDQ, only FP16 conversion via quantize stage
-        parent_config.quant = WinMLQuantizationConfig(mode="fp16")
+        if parent_config.quant is None:
+            parent_config.quant = WinMLQuantizationConfig()
+        parent_config.quant.mode = "fp16"
```
timenick, line 732: Preserving `model_type` in the FP16 and RTN branches makes the registered model-type finalizer run later in `quantize_onnx()`. For `qwen3_transformer_only`, that finalizer unconditionally rewrites the config to `mode="static"`, `weight_type="int8"`, and `activation_type="uint16"`. Consequently, an explicit bundle override such as `--precision fp16` or `--precision w4a16` is forwarded to the transformer builder but is ultimately changed back to the fixed w8a16 scheme.

The docs and `GenaiBundleRecipe` API describe this as a transformer precision override, while the new tests only verify that the argument and `model_type` survive—not which quantization pass actually runs. If this transformer supports only w8a16, could unsupported overrides be rejected? Otherwise, the finalizer needs to respect the explicitly selected mode.

src/winml/modelkit/session/genai_session.py
```diff
@@ -960,13 +960,85 @@ def _compile_stage(
...
+
+        Returns ``True`` when a valid EPContext is now present at *ctx_out*.
+        """
+        from ..onnx import is_compiled_onnx
+
+        # ORT emits the auto-named EPContext next to the source model; the
+        # compiler then copies it into ctx_out's directory. Search both, plus
+        # ctx_out itself (a crash after the copy). ``dict.fromkeys`` dedups the
+        # directories while preserving order in case they coincide.
+        candidates: list[Path] = [ctx_out]
```
timenick: This can salvage an artifact that predates the compile attempt. `_compile_stage()` is reached when the cached `ctx_out` is not fresh, but it does not remove or quarantine that stale file before spawning the compiler. If the subprocess exits non-zero before replacing it, `_salvage_epcontext()` checks the old `ctx_out` first, accepts it because it merely contains an `EPContext` node, and the caller then writes a new marker with the current provider options. A context compiled for old options can therefore be mislabeled as the new one and reused; it also wins over any newly written auto-named candidate.

The structural check also does not verify that an `embed_mode=0` context's referenced `.bin` exists, so a partially flushed graph can be accepted as complete. Could salvage be limited to candidates created or updated by this subprocess (for example by snapshotting/quarantining pre-existing files), and validate the referenced external context before marking the stage successful?

src/winml/modelkit/commands/build.py
DingmaomaoBJTU: Addressed in 702f7064 — the fast path now rejects unsupported controls instead of letting them silently no-op:

- CLI pipeline flags (`--quant/--no-quant`, `--optimize/--no-optimize`, `--analyze/--no-analyze`, `--compile/--no-compile`, `--max-optim-iterations`, `--allow-unsupported-nodes`) were already rejected via the `_unsupported_controls` guard when explicitly passed. Added `test_explicit_no_quant_is_rejected_not_silently_ignored` to pin the `--no-quant` case you called out.
- The remaining gap was a config *file* (`-c`), whose fields (e.g. `"quant": null`) bypass `is_cli_provided`. Once a recipe matches, the fast path now rejects `-c/--config` outright, so a supplied config can no longer be silently discarded. The routing fixtures no longer inject `model_type` via `-c` — they mock `generate_build_config`. New regression: `test_explicit_config_file_is_rejected_for_bundle`.

src/winml/modelkit/config/build.py
DingmaomaoBJTU: Addressed in 702f7064 by rejecting the unsupported override rather than silently reverting it. `build_genai_bundle` now raises when a `--precision` override differs from the transformer precision **and** the transformer's `model_type` has a registered quant finalizer (`has_quant_finalizer`) — exactly the case where the finalizer would rewrite it back to the pinned scheme (w8a16 for `qwen3_transformer_only`).

The gate is keyed on the finalizer registry, not any model name, so a transformer *without* a finalizer still honors overrides. It lives in the orchestrator so it covers both the CLI fast path and `scripts/qwen3.py`. Docs updated to state the transformer precision is recipe-fixed. Tests: `test_precision_override_rejected_when_transformer_finalizer_pinned`, `test_matching_precision_override_allowed_when_finalizer_pinned`, `test_no_precision_override_uses_recipe_default_when_finalizer_pinned`.

src/winml/modelkit/session/genai_session.py
DingmaomaoBJTU: Addressed in 702f7064.

- **Stale artifact:** `_compile_stage` now snapshots the mtimes of every EPContext candidate path *before* spawning the compiler. `_salvage_epcontext` rejects any pre-existing candidate whose mtime did not advance, so a stale `ctx_out` from an earlier compile (e.g. different provider options) is never salvaged or promoted — only files this subprocess actually created or rewrote qualify. A newly written auto-named candidate therefore wins over an untouched stale `ctx_out`.
- **Missing external context:** added `_epcontext_is_valid`, which beyond the structural EPContext check resolves an `embed_mode=0` context's `ep_cache_context` next to the graph and requires the `.bin` to exist and be non-empty before promoting (a partially flushed graph is rejected). `embed_mode=1`/absent embeds the blob inline, so only the structural check applies there.

New tests: `test_stale_pre_existing_epcontext_is_not_salvaged`, `test_embed_mode_zero_missing_bin_is_not_salvaged`. The existing salvage tests were reworked so the mock subprocess writes the artifact *after* the snapshot (modeling the real teardown-crash ordering).

src/winml/modelkit/config/build.py
timenick: Thanks for adding the pinned-precision guard. It still rejects valid equivalent inputs because the comparison uses the raw string before normalization/resolution. An explicitly supplied `--precision auto` resolves to `w8a16` for NPU, but `"auto" != "w8a16"` here, so a generic invocation fails even though its effective precision matches the recipe. The same happens with a case variant such as `W8A16`, which downstream precision validation otherwise accepts. Could we treat `auto` as no conflicting override and normalize explicit values before comparing, with regression coverage for both cases?

src/winml/modelkit/session/genai_session.py
timenick: Thanks for strengthening the EPContext validation. This loop also visits secondary EPContext nodes, which are allowed to omit `ep_cache_context`. The in-repo EPContext schema specifically documents the QNN representation: one `main_context=1` node carries the context, while the remaining `main_context=0` nodes skip it; `CompileStage._finalize_output()` already ignores those secondary nodes for the same reason. If they use `embed_mode=0`, the current empty `cache_ref` check rejects an otherwise valid multi-partition graph, so teardown salvage fails and falls back to the original JIT/crash path. Could we parse `main_context` (defaulting to 1), validate the external file only for the main node, and add a multi-node regression test?

DingmaomaoBJTU: Good catch — fixed in 7de447b3. `_epcontext_is_valid` now reads `main_context` and skips secondary (`main_context=0`) nodes, requiring the external `.bin` only for the main context node, mirroring `CompileStage._finalize_output()`'s handling. A multi-partition QNN graph (one `main_context=1` node + secondary `main_context=0` nodes that omit `ep_cache_context`) is therefore salvaged instead of falling back to the JIT/crash path. New regression tests: `test_salvages_multipartition_qnn_epcontext_with_secondary_nodes` and `test_multipartition_epcontext_missing_main_bin_is_not_salvaged`.

src/winml/modelkit/config/build.py
DingmaomaoBJTU: Good point — fixed in 93cbbc76. The guard no longer compares the raw override string:

- `auto` is treated as deferring to the recipe scheme (it resolves to the device default `w8a16` on NPU = the pinned precision), so a generic `--precision auto` invocation is no longer rejected.
- The comparison is now case-insensitive, so an equivalent variant like `W8A16` is accepted (matching the case-insensitive downstream resolver).

A finalizer-pinned transformer's precision also collapses to the canonical recipe value for the downstream builder/logs. Regression coverage: `test_auto_precision_override_allowed_when_finalizer_pinned` and `test_case_variant_precision_override_allowed_when_finalizer_pinned` (both assert the effective precision resolves to `w8a16`), alongside the existing conflicting-override rejection test.

timenick: Thanks for handling the NPU `auto` case, but this exemption is broader than the condition it relies on. The existing `resolve_precision()` policy maps `auto` to `w8a16` only for NPU; GPU and CPU both resolve it to `fp32`. `scripts/qwen3.py` supports all three devices and forwards the selected device here, while this pinned-finalizer branch now always collapses `auto` to the recipe's `w8a16`. As a result, `scripts/qwen3.py export --device gpu --precision auto` is accepted but silently builds with `w8a16` instead of the effective `fp32`, reintroducing the ignored-explicit-precision behavior. Could we resolve `auto` using the supplied device/EP before comparing, allow it only when the effective precision matches the pinned recipe, and add GPU/CPU coverage?

timenick: Confirmed that this recipe's supported contract is NPU-only. Under that constraint, `auto` resolves to `w8a16` for every supported invocation, so the CPU/GPU scenario I raised is outside the recipe's valid scope and does not make this change incorrect. The broader device choices exposed by the standalone script are a separate contract question and should not block this fix. Withdrawing my previous concern.

timenick (approved):

PR #1083, opened 2026-07-10
feat: add dynamic export controls to build and perf

xieofxie:
## Summary
- Add shared CLI option helpers for export-related JSON controls.
- Wire `--shape-config`, `--input-specs`, `--export-config`, and `--dynamic-axes` into `winml build` where applicable.
- Wire `--input-specs`, `--export-config`, and `--dynamic-axes` into `winml perf` Hugging Face build flow.
- Update build/perf docs and command tests for the new flags.

## Validation
- `uv run ruff check --fix src\winml\modelkit\utils\cli.py src\winml\modelkit\commands\export.py src\winml\modelkit\commands\build.py src\winml\modelkit\commands\perf.py`
- `uv run pytest tests\unit\commands\test_build.py tests\unit\commands\test_perf_cli.py tests\unit\commands\test_export.py`
- `uv run mypy -m winml.modelkit.utils.cli -m winml.modelkit.commands.export -m winml.modelkit.commands.build -m winml.modelkit.commands.perf`


src/winml/modelkit/utils/cli.py
```diff
@@ -1121,6 +1180,76 @@ def load_build_config(config_path: Path) -> tuple[WinMLBuildConfig, dict]:
...
+        if shape is not None:
+            if not isinstance(shape, list):
+                raise click.ClickException(f"{option_name}['{name}'].shape must be a list")
+            shape = tuple(shape)
+
+        dtype = spec.get("dtype")
+        if dtype is not None and not isinstance(dtype, str):
+            raise click.ClickException(f"{option_name}['{name}'].dtype must be a string")
+
+        specs.append(InputTensorSpec(name=name, dtype=dtype or "float32", shape=shape))
```
DingmaomaoBJTU: For `build`/`perf` this builds a fresh `InputTensorSpec` list that later **fully replaces** the auto-resolved `input_tensors` (`merge_config` replaces lists, it doesn't patch them). That differs from `export`, which patches fields onto the auto-resolved specs. Two footguns fall out of the replace + `dtype or "float32"` default:

- Omitting `dtype` forces `float32`. For a text model, `{"input_ids": {"shape": ["batch","seq"]}}` produces a float32 `input_ids`, and `generate_dummy_inputs()` then feeds a float tensor into the embedding during export -> export fails.
- Any auto-resolved inputs you don't list (e.g. `attention_mask`) and their `value_range`s are dropped, so a partial file silently breaks the export.

The docs/help say these specs "merge into" the export config, which doesn't match the replace behavior. Can we patch the auto-resolved specs the way `export` does (or at least preserve `dtype`/`value_range` for inputs the user doesn't fully specify)?

src/winml/modelkit/utils/cli.py
```diff
@@ -1121,6 +1180,76 @@ def load_build_config(config_path: Path) -> tuple[WinMLBuildConfig, dict]:
...
+        overrides.update(load_json_object(export_config, "--export-config"))
+    if input_specs is not None:
+        overrides["input_tensors"] = load_input_tensor_specs(input_specs)
+    if dynamic_axes is not None:
+        overrides["dynamic_axes"] = load_json_object(dynamic_axes, "--dynamic-axes")
+
+    if overrides:
+        # Validate field names/types and dynamic-axis conflicts early, but return
+        # the sparse mapping so build config merges do not clobber unrelated fields.
+        WinMLExportConfig.from_dict(overrides)
```
DingmaomaoBJTU: `load_json_object` raises a clean `ClickException`, but this `from_dict` validation can raise a raw `ValueError`/`TypeError` for realistic input -- a non-integer axis key in `--dynamic-axes`, conflicting explicit-vs-symbolic axes, or `batch_size <= 0` from `--export-config`. In `build`/`perf` this runs outside any try, so the user gets a traceback instead of a friendly message. Wrap it and re-raise as `click.ClickException`/`UsageError` for consistency.

src/winml/modelkit/commands/build.py
```diff
@@ -634,6 +686,12 @@ def build(
             # of treating the path as a HuggingFace repo id (which would try to
             # load the .onnx file as a JSON config and crash).
             if cli_utils.is_onnx_file_path(model):
+                if export_overrides:
```
DingmaomaoBJTU, line 825: `--shape-config` isn't covered by this ONNX guard, so `winml build -m model.onnx --shape-config shapes.json` silently ignores it (`shape_overrides` is never passed to the ONNX `generate_build_config` below). That's inconsistent with the sibling flags rejected here, and the build docs state "Export-shape flags ... are rejected for pre-exported ONNX inputs." Consider rejecting (or at least warning on) `shape_config` here too.

xieofxie:
Tested
- winml perf -m microsoft/resnet-50 -v --dynamic-axes .\temp\resnet-dynamic.json --batch-size 2 --ep cpu
-  winml build -m microsoft/resnet-50 -v --dynamic-axes .\temp\resnet-dynamic.json  --ep cpu --output-dir temp/resnet_dyn

content: `{"pixel_values": {"0": "batch"}}`

DingmaomaoBJTU (commented):
Overall the refactor reads well — the shared CLI option helpers and de-duplicating `_load_json_object` into `utils/cli.py` are a nice cleanup. A few concerns inline, the main one being that in `build`/`perf` `--input-specs` **replaces** the auto-resolved `input_tensors` wholesale (via `merge_config`) rather than patching them by name the way `winml export` does, which combined with the `dtype or "float32"` default can silently drop inputs / force the wrong dtype. Also flagged the unwrapped `from_dict` validation (raw tracebacks on bad `--dynamic-axes`/`--export-config`) and the `--shape-config` gap in the ONNX guard.

src/winml/modelkit/utils/cli.py
xieofxie: Good catch. Fixed in efe36f37: build/perf now patch --input-specs onto the auto-resolved input_tensors by name (mirroring export) instead of replacing the list. Unlisted inputs like attention_mask and their value_ranges are preserved, and omitted dtype stays None (no more forced float32 that broke text models). Added unit tests for the patch helper and the dtype/value_range preservation.

src/winml/modelkit/utils/cli.py
xieofxie: Fixed in efe36f37: load_export_overrides now wraps WinMLExportConfig.from_dict and re-raises ValueError/TypeError as click.UsageError, so bad --dynamic-axes/--export-config give a friendly message instead of a traceback. Added a test for a non-integer axis key.

src/winml/modelkit/commands/build.py
xieofxie: Fixed in efe36f37: --shape-config is now rejected for pre-exported ONNX inputs with a UsageError instead of being silently ignored, consistent with the sibling export-shape flags and the docs. Added a test.

src/winml/modelkit/config/build.py
```diff
@@ -683,8 +726,30 @@ class name. Uses torchinfo to discover submodules and infer
...
+                # Shallow-copy so the caller's dict isn't mutated when we drop the key.
+                override = {
+                    **override,
+                    "export": {k: v for k, v in export_override.items() if k != "input_tensors"},
+                }
+
         parent_config = merge_config(parent_config, override)
 
+        if input_spec_patches is not None and parent_config.export is not None:
+            parent_config.export.input_tensors = _patch_input_tensors(
```
DingmaomaoBJTU: Setting `input_tensors` after `merge_config` bypasses `WinMLExportConfig.__post_init__`, which is the only place `_dynamic_axes_from_input_tensors` runs (export/config.py:225). The auto-resolved specs carry concrete int shapes (export/io.py:487 uses the dummy tensors’ shapes), so `dynamic_axes` is derived as empty during the merge, and the symbolic dims from `--input-specs` patched in here are never converted afterwards. Since the exporter reads the stored `export_config.dynamic_axes` (htp/exporter.py:484), a run like `--input-specs '{"input_ids":{"shape":["batch","seq"]}}'` (without `--dynamic-axes`) now produces a **static** model.

This is a regression vs. the previous wholesale-replace path (which reconstructed the config and therefore ran `__post_init__` on the user specs), and it diverges from `winml export` and from the documented “Symbolic string dimensions infer dynamic axes”. `TestPatchInputTensors` only asserts on the returned list, so it doesn’t cover this.

Suggestion: re-derive dynamic axes after patching — e.g. reconstruct the export config so `__post_init__` runs, or explicitly
```python
parent_config.export.dynamic_axes = _merge_dynamic_axes(
    _normalize_dynamic_axes(parent_config.export.dynamic_axes),
    _dynamic_axes_from_input_tensors(parent_config.export.input_tensors),
)
```

src/winml/modelkit/commands/build.py
```diff
@@ -757,6 +788,27 @@ def build(
...
+                for cfg in config_list:
+                    if cfg.export is None:
+                        raise click.UsageError(
+                            "--input-specs, --export-config, and --dynamic-axes require "
+                            "a HuggingFace export config; they are not supported when "
+                            "the build config has export=null."
+                        )
+                if isinstance(config_or_configs, list):
+                    config_or_configs = [
+                        merge_config(cfg, {"export": export_overrides}) for cfg in config_or_configs
```
DingmaomaoBJTU: The auto path now patches `--input-specs` onto the resolved specs by name (`_patch_input_tensors`), but this `-c` branch still merges wholesale. Since `merge_config` replaces lists entirely, `winml build -c config.json --input-specs '{"input_ids":{...}}'` drops every other input defined in the config file; and because `load_input_tensor_specs` now leaves `dtype=None`, those specs export as float32 — the same footgun `_patch_input_tensors` was written to avoid. Consider routing this branch through the same by-name patch (against each config’s `export.input_tensors`) so `-c` and the auto path behave consistently.

DingmaomaoBJTU (commented):
Follow-up on the fix commit (efe36f37): the three earlier comments are well addressed and nicely tested — thanks! Two issues remain from this change: a functional regression around dynamic-axes inference, and a consistency gap on the `-c` path. Details inline.

src/winml/modelkit/config/build.py
xieofxie: Great catch, that was a real regression. Fixed in aa7b4278: instead of assigning input_tensors after merge_config (which bypassed __post_init__), the patched list is now routed back through merge_config, which reconstructs WinMLExportConfig and re-runs __post_init__ -> dynamic_axes are re-derived from symbolic --input-specs dims and merged with any explicit --dynamic-axes (with the conflict check intact). Added tests asserting dynamic_axes derivation, explicit+symbolic merge, and conflict detection.

src/winml/modelkit/commands/build.py
xieofxie: Fixed in aa7b4278: the -c branch now routes through the same by-name patch via a shared merge_export_overrides helper, so config-file inputs (and their dtypes) are preserved instead of being replaced wholesale. Added a -c integration test asserting attention_mask survives, input_ids keeps dtype int64 (not float32), and symbolic dims derive dynamic_axes.

xieofxie:
also need to add to config command.. next pr

DingmaomaoBJTU (approved):
Approving — the follow-up commit (aa7b4278) fully resolves both remaining concerns:

1. **Dynamic-axes regression fixed.** `merge_export_overrides` now routes the patched `input_tensors` list back through `merge_config`, so `WinMLExportConfig.__post_init__` re-runs and re-derives `dynamic_axes` from symbolic `--input-specs` dims. Covered by `TestMergeExportOverrides::test_symbolic_dims_derive_dynamic_axes` and the `-c` integration test.
2. **`-c` path consistency fixed.** The `-c` branch now goes through `merge_export_overrides` instead of a wholesale `merge_config`, so `--input-specs` patches each config's `export.input_tensors` by name (preserving unlisted inputs and their dtype/value_range). Covered by `test_input_specs_patches_config_file_inputs`.

Nice touches: the shared helper is exported in `config/__init__.py`, no-mutation is asserted, explicit-vs-inferred axis merging and conflict detection are tested, and the now-unused `merge_config` import was removed. All CI checks are green.

(Note: I couldn’t run pytest locally — win_arm64 has no torch 2.13.0 wheel — so this relies on code review plus the passing CI test jobs.)

PR #1089, opened 2026-07-10
feat: add --submodel to perf and build commands

xieofxie:
## Summary

Expands the `--submodel` capability (added to `winml export` in #1031) to the `winml perf` and `winml build` commands, letting users select a single sub-model of a composite model instead of processing all of them.

## Changes

**`winml build`** (`build.py`)
- Adds `--submodel <name>` option.
- After the existing `resolve_composite_components()` detection, validates and filters the components dict, mirroring `export`:
  - `--submodel` on a non-composite model → `BadParameter` ("not a composite model").
  - Unknown name → `BadParameter` listing available sub-models.
  - Otherwise narrows the fan-out to just that one sub-model build.

**`winml perf`** (`perf.py`)
- Adds `--submodel <name>` option and `BenchmarkConfig.submodel` field.
- perf detects composites *post-load*, so `_run_sub_models()` filters the loaded `sub_models` dict to the requested name (unknown → `UsageError`), and `run()` errors when `--submodel` is passed for a non-composite model.

## Tests
- `TestBuildSubmodel`: filters to one component, rejects unknown name, rejects non-composite.
- `TestPerfSubmodelFilter`: benchmarks only the requested sub-model, runs only that session, rejects unknown name, rejects non-composite.

Verification: ruff clean, mypy clean on both modules; build/perf-composite (119) and perf-cli (68) tests pass.


xieofxie:
no need to add to config as it fast and it generates config per submodule

src/winml/modelkit/commands/build.py
```diff
@@ -1247,6 +1257,22 @@ def _patch_device(cfg: WinMLBuildConfig) -> None:
                         f"Composite model detection failed unexpectedly: {e}"
                     ) from e
 
+            # ── --submodel validation ──────────────────────────────────────
+            if submodel is not None:
```
timenick, line 1284: This validation runs too late to enforce the new option across all build modes. `_maybe_build_genai_bundle()` can fully handle the request and return before reaching this block, so both `--export-type optimized` and the implicit QNN/NPU shortcut build the complete bundle while silently ignoring `--submodel`. The array-config module branch above likewise builds every module without consulting it. Please validate or reject `--submodel` before those dispatches (or explicitly support it there), and add coverage for both early paths.

src/winml/modelkit/commands/perf.py
```diff
@@ -2305,6 +2348,46 @@ def perf(
...
+    # --submodel: narrow a composite model to one sub-component, benchmarked as
+    # a standalone single-session model. The composite is detected the same way
+    # `winml export` / `winml build` / `winml inspect` do (registry-driven, via
+    # the seq2seq bridge), so it works even when --task is omitted. The selected
+    # component is then loaded through the normal single-model path using its own
+    # component task — exactly how the composite builds that sub-model — which
+    # sidesteps the config-ambiguous pipeline task (e.g. t5 translation vs
+    # summarization) and avoids building the other sub-models just to discard.
+    # =========================================================================
+    if submodel is not None:
```
timenick, line 2368: This validation is also bypassed by the earlier `--runtime winml-genai` return. `winml perf --runtime winml-genai --submodel ...` runs the GenAI benchmark while silently ignoring the requested sub-model, and `submodel` is not included in `_GENAI_IGNORED_FLAGS`, so the user receives no warning either. Please reject this combination alongside `--module`/`--input-data`, or explicitly define and implement its GenAI behavior.

xieofxie: Fixed. _run_genai_runtime() now rejects --submodel right alongside --module, so \winml perf --runtime winml-genai --submodel ...\ errors instead of silently ignoring the sub-model. Since it's a hard rejection I left it out of \_GENAI_IGNORED_FLAGS\ (same as \--module\). Added a test in test_perf_genai.py.

src/winml/modelkit/commands/build.py
xieofxie: Fixed. --submodel is now validated before both early dispatches. _maybe_build_genai_bundle() raises a BadParameter once it commits to a bundle build (both explicit --export-type optimized and the implicit QNN/NPU shortcut), and the array-config module-mode branch rejects it as well. Added coverage for the genai-bundle paths in test_build_genai_bundle.py and for module mode in test_build.py (TestBuildSubmodel).

timenick (approved):

PR #1094, opened 2026-07-12
Add audio-classification fp16 recipe for audeering/wav2vec2-large-robust-24-ft-age-gender

ssss141414:
## Add audio-classification fp16 recipe — audeering/wav2vec2-large-robust-24-ft-age-gender

Recipe-only PR. Adds `examples/recipes/audeering_wav2vec2-large-robust-24-ft-age-gender/cpu/cpu/audio-classification_fp16_config.json`.

### Environment (this run)
- winml-cli **v0.2.0** (editable install)
- Base: rebased onto `origin/main` — branch HEAD `b2e88b86`
- Host EPs (`onnxruntime.get_available_providers()`): `CPUExecutionProvider`, `AzureExecutionProvider` → only **CPU** is runtime-verified this run. Accelerator buckets, if any, are carried-over / prior-verified on capable hardware.

### Why the recipe is needed (baseline gap — proven)
Clean `origin/main`, no recipe:
```
winml config -m audeering/wav2vec2-large-robust-24-ft-age-gender    # and winml build
→ FAIL: "wav2vec2 doesn't support task text-classification.
         Supported: feature-extraction, automatic-speech-recognition,
         audio-classification, audio-frame-classification, audio-xvector."
```
Task is auto-mis-detected as `text-classification`. The recipe's `loader.task = audio-classification` fixes detection so the model builds.

### fp16 caveat (documented)
Recipe `quant.mode = "fp16"` **alone is not sufficient** on CPU: the auto-precision policy (cpu→fp32) overrides the recipe, producing an fp32 graph (`model.onnx.data` ≈ 1204 MB). To actually get fp16 you must pass the CLI flag explicitly:
```
winml build -m <model> -c <recipe> --precision fp16
```
→ correct fp16 (`model.onnx.data` ≈ 602 MB; `winml perf` reports `Model Precision: fp16`).

### Goal ladder (CPU, this run)
| Level | Check | Result |
|---|---|---|
| L0 | build (CPU, recipe) | **PASS** — fp32 and fp16 (`--precision fp16`) both build |
| L1 | perf (CPU, fp16) | **PASS** — P50 132.33 ms, tput 7.59/s, RAM load +1225.5 MB; input `input_values[1,16000]` f32, output `logits[1,3]` |
| L2 | numeric parity vs PyTorch | **FAIL (blocked upstream — see below)** |
| L3 | accuracy eval | Not meaningful (see L2) |
| analyze | op-level `--ep all` | OpenVINO(CPU) **773/0/0/0** (all supported); CPU EP no rule data |

### L2 numeric parity — genuine FAIL, root cause is upstream model packaging (not a winml bug)
The checkpoint declares `config.architectures = ['Model']` (audeering's custom `AgeGenderModel` head; `id2label = {0: female, 1: male, 2: child}`), **but the repo ships no `trust_remote_code` modeling file**. As a result `AutoModelForAudioClassification.from_pretrained(..., trust_remote_code=True)` silently falls back to the standard `Wav2Vec2ForSequenceClassification` and **randomly re-initializes the head**:
```
Some weights of Wav2Vec2ForSequenceClassification were not initialized from the
model checkpoint ... newly initialized: ['classifier.bias','classifier.weight',
'projector.bias','projector.weight']
```
So both the PyTorch reference **and** winml's export load a randomly-initialized head (different seed on each load). The logits are meaningless (~0.03 random), and parity can never converge:
- fp32: cosine ≈ 0.67, max_abs ≈ 3.4e-2 (realistic normalized input)
- fp16: cosine ≈ 0.30

This is a real defect the recipe cannot fix: the model's custom age/gender head weights do not map onto any standard HF class, and there is no custom modeling code to load them. The build/perf paths still work, but the exported ONNX is numerically invalid.

**Disposition:** needs-work. Correctly exporting this model requires a Lane-A custom loader/OnnxConfig that maps the audeering `Model` head weights (or the model is out of scope until upstream ships modeling code). The recipe here is retained because it fixes task detection and enables the build/perf pipeline; the L2 gate correctly flags the parity blocker.


xieofxie:
if without config, could perf run it directly?

ssss141414:
> if without config, could perf run it directly?

As PR's description, it will failed on build stage without config. And the following command will not be run (perf/ eval)...

```
Because of the custom head, winml build auto-config mis-resolves the task to text-classification, and the build fails:

Error: wav2vec2 doesn't support task text-classification for the onnx backend.
Supported tasks are: feature-extraction, automatic-speech-recognition,
audio-classification, audio-frame-classification, audio-xvector.
```

xieofxie:
> > if without config, could perf run it directly?
> 
> As PR's description, it will failed on build stage without config. And the following command will not be run (perf/ eval)...
> 
> ```
> Because of the custom head, winml build auto-config mis-resolves the task to text-classification, and the build fails:
> 
> Error: wav2vec2 doesn't support task text-classification for the onnx backend.
> Supported tasks are: feature-extraction, automatic-speech-recognition,
> audio-classification, audio-frame-classification, audio-xvector.
> ```

same as another pr

xieofxie (approved):

PR #1102, opened 2026-07-13
feat: Add ONNX data to QNN profiler metrics

xieofxie:
## Summary
- Add env-gated ONNX metadata enrichment for QNN basic profiler metrics via WINMLCLI_OP_ADD_DATA
- Include ONNX node type, attributes, and input tensor metadata keyed by input name
- Add focused unit coverage for default and env-enabled output shapes

## Verification
- uv run ruff check --fix src\winml\modelkit\optracing\result.py src\winml\modelkit\optracing\qnn\profiler.py tests\unit\optracing\test_qnn_profiler.py
- uv run pytest tests\unit\optracing\test_qnn_profiler.py

### DIFF
--- a/src/winml/modelkit/optracing/qnn/profiler.py
+++ b/src/winml/modelkit/optracing/qnn/profiler.py
@@ -105,6 +105,7 @@ def _csv_operator_metrics(
                 onnx_op_type=op_data.get("onnx_op_type"),
                 onnx_attributes=op_data.get("onnx_attributes"),
                 onnx_inputs=op_data.get("onnx_inputs"),
+                onnx_outputs=op_data.get("onnx_outputs"),
             )
         )
     # duration is the headline metric; percent breaks ties when US timing is
@@ -136,6 +137,7 @@ def _load_onnx_operator_data(onnx_path: Path) -> dict[str, dict[str, Any]]:
             "onnx_op_type": node.op_type,
             "onnx_attributes": {attr.name: _serialize_attribute(attr) for attr in node.attribute},
             "onnx_inputs": _node_input_metadata(node, opset_versions, value_info, initializers),
+            "onnx_outputs": _node_output_metadata(node, opset_versions, value_info, initializers),
         }
     return operator_data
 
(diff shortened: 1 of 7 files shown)


xieofxie:
Sample result
```
    {
      "name": "/resnet/encoder/stages.1/layers.0/layer/layer.1/convolution/Conv",
      "op_path": "/resnet/encoder/stages.1/layers.0/layer/layer.1/convolution/Conv",
      "op_id": 588,
      "duration_us": 45.286626051604955,
      "percent_of_total": 4.619123992288399,
      "onnx_op_type": "Conv",
      "onnx_attributes": {
        "dilations": [
          1,
          1
        ],
        "group": 1,
        "kernel_shape": [
          3,
          3
        ],
        "pads": [
          1,
          1,
          1,
          1
        ],
        "strides": [
          2,
          2
        ]
      },
      "onnx_inputs": {
        "X": {
          "name": "/resnet/encoder/stages.1/layers.0/layer/layer.0/activation/Relu_output_0_DequantizeLinear_Output",
          "data_type": "FLOAT",
          "dims": [
            1,
            128,
            56,
            56
          ]
        },
        "W": {
          "name": "onnx::Conv_533_DequantizeLinear_Output",
          "data_type": "FLOAT",
          "dims": [
            128,
            128,
            3,
            3
          ]
        },
        "B": {
          "name": "onnx::Conv_534",
          "data_type": "FLOAT",
          "dims": [
            128
          ]
        }
      },
      "onnx_outputs": {
        "Y": {
          "name": "/resnet/encoder/stages.1/layers.0/layer/layer.1/activation/Relu_output_0",
          "data_type": "FLOAT",
          "dims": [
            1,
            128,
            28,
            28
          ]
        }
      }
    },
```

src/winml/modelkit/optracing/qnn/profiler.py
```diff
@@ -422,7 +628,10 @@ def _from_csv(
                 f"warmup, got {len(measured)} from {len(samples)} total."
             )
 
-        operators = _csv_operator_metrics(measured)
+        onnx_operator_data = (
+            _load_onnx_operator_data(self.onnx_path) if _is_op_add_data_enabled() else None
```
DingmaomaoBJTU: When `WINMLCLI_OP_ADD_DATA` is set, any exception raised inside `_load_onnx_operator_data` propagates out of `_from_csv` and discards profiling results that were already computed successfully. Several paths here can throw: `load_onnx` (missing/corrupt model, or a `*_ctx.onnx` context binary), `TensorProto.DataType.Name()` on an unknown enum value, and `_serialize_attribute` on unsupported attribute types.

Since this is opt-in enrichment, a metadata failure shouldn't be able to void an otherwise-valid trace. Suggest wrapping the call in try/except, logging a warning, and falling back to `None`.

Verified locally: with the env var set and an unreadable model path, `_from_csv` raises `FileNotFoundError` instead of returning the (valid) operator metrics.

src/winml/modelkit/optracing/qnn/profiler.py
```diff
@@ -86,22 +91,223 @@ def _csv_operator_metrics(samples: list[dict[str, Any]]) -> list[OperatorMetrics
...
+        return list(value)
+    if attr_type == AttributeProto.GRAPH:
+        return _graph_attribute_metadata(value)
+    if attr_type == AttributeProto.GRAPHS:
+        return [_graph_attribute_metadata(graph) for graph in value]
+    if attr_type == AttributeProto.SPARSE_TENSOR:
+        return _sparse_tensor_attribute_metadata(value)
+    if attr_type == AttributeProto.SPARSE_TENSORS:
+        return [_sparse_tensor_attribute_metadata(tensor) for tensor in value]
+    return value
```
DingmaomaoBJTU: This fallthrough returns the raw protobuf value for any attribute type not handled above — notably `TYPE_PROTO` / `TYPE_PROTOS` (and `UNDEFINED`). Those are not JSON-serializable, so `OpTraceResult.to_json()` fails with `TypeError: Object of type TypeProto is not JSON serializable`.

Verified locally with a `TYPE_PROTO` attribute. Suggest handling those cases explicitly, or using a safe fallback such as `str(value)`, so serialization can't break.

src/winml/modelkit/optracing/qnn/profiler.py
```diff
@@ -86,22 +91,223 @@ def _csv_operator_metrics(samples: list[dict[str, Any]]) -> list[OperatorMetrics
...
+    opset_versions = _collect_opset_versions(model)
+
+    operator_data: dict[str, dict[str, Any]] = {}
+    for node in model.graph.node:
+        if not node.name:
+            continue
+        operator_data[node.name] = {
+            "onnx_op_type": node.op_type,
+            "onnx_attributes": {attr.name: _serialize_attribute(attr) for attr in node.attribute},
+            "onnx_inputs": _node_input_metadata(node, opset_versions, value_info, initializers),
```
DingmaomaoBJTU, line 139: This captures input tensor shapes/types but not the node's output tensor(s). Given the branch name (add_op_size), the output shape is often the more useful "size" signal for a profiler — and an op's inputs are usually another op's outputs anyway. Was omitting `onnx_outputs` intentional, or worth capturing alongside inputs? Non-blocking.

DingmaomaoBJTU (commented):
Reviewed the ONNX metadata enrichment for QNN basic profiler metrics. The schema-name input resolution, the env gating, and the on/off test coverage look solid. Left a few inline notes — the main one is that an enrichment error can currently void an otherwise-successful trace.

DingmaomaoBJTU (approved):

src/winml/modelkit/optracing/qnn/profiler.py
xieofxie: good point! wrapping the call in try/except, logging a warning, and falling back to None

src/winml/modelkit/optracing/qnn/profiler.py
xieofxie: yes, we should add it

src/winml/modelkit/optracing/qnn/profiler.py
xieofxie: Fixed in fba7ad85: ONNX metadata enrichment now fails open with a warning and falls back to plain profiling metrics when model metadata cannot be loaded or serialized.

src/winml/modelkit/optracing/qnn/profiler.py
xieofxie: Fixed in fba7ad85: TYPE_PROTO / TYPE_PROTOS are converted through protobuf JSON conversion, and unsupported attribute payloads now use a JSON-safe fallback so result serialization cannot fail.

src/winml/modelkit/optracing/qnn/profiler.py
xieofxie: Fixed in fba7ad85: added onnx_outputs keyed by ONNX schema output names, with tensor name, data_type, and dims included just like onnx_inputs.

DingmaomaoBJTU (approved):

PR #1103, opened 2026-07-13
fix(genai): isolate EPContext compile from og.Model load (#1087)

### LINKED ISSUE #1087: genai --compile: model load can native-crash when compile and load share one process; isolate og.Model load into its own subprocess
## Summary

`winml perf -m <bundle> --runtime winml-genai --device npu --compile` can silently native-crash at **model load** (before the first token) even when all EPContext stages are already compiled and cached. The failure is not a wedged HTP and not the model itself — it is the `--compile` code path running the crash-prone stage compilation and the `og.Model()` load in the **same parent process**.

Reproduced independently on two setups (Windows ARM64 / QNN HTP).

## Root cause

In `GenaiSession.load()` (`src/winml/modelkit/session/genai_session.py`), `--compile` (`compile=True`) does both of these in one process:

1. `_prepare_compiled_bundle()` — compiles each EPContext-capable stage in a spawned subprocess. QNN can native-crash during interpreter/driver **teardown** of those subprocesses (PR #1081 adds an EPContext salvage that recovers the compiled artifact, but the native fault still happened in-session).
2. Immediately afterward, `og.Model(og.Config(_compiled/))` is created in the **same** parent process to run generation.

The compile orchestration appears to leave the parent process / native QNN state fragile, so the subsequent `og.Model()` load faults (`python.exe` faulting, `0xC0000005` / `0xC0000374`) with no Python traceback.

## Evidence it is process-isolation, not the bundle

Loading the already-compiled, self-contained `_compiled/` directory directly in a **fresh** process (no `--compile`, so no `_prepare_compiled_bundle()`) generates cleanly:

```
winml perf -m out/qwen3-bundle/_compiled --runtime winml-genai --device npu \
  --max-new-tokens 20 --prompt "What is the capital of France?"
```

Result: winml-genai / qnn / npu, 15-token prompt -> 20 tokens, TTFT ~537 ms (P50 531), decode ~5.0 tok/s, 10 iterations, results JSON saved. The same `_compiled/` bundle that faults under `--compile` runs fine when loaded on its own.

`_mirror_non_onnx_files()` already skips files that exist (`if dst.exists(): continue`), so this is **not** re-mirror memory pressure — it is specifically compile + load sharing a process.

## Workaround (already documented)

`docs/samples/qwen3-genai-bundle.md` now describes the two-step flow: run `--compile` once to produce `<bundle>/_compiled/`, then run `winml perf -m <bundle>/_compiled` (no `--compile`) in a fresh process. The second command loads the compiled EPContext graphs directly, so it does not JIT-compile anything.

## Proposed fix (root cause)

In `--compile` mode, after `_prepare_compiled_bundle()` returns, load and run `og.Model()` in its **own isolated subprocess** (spawn / re-exec) so the crash-prone compile orchestration cannot poison the process that performs generation. Equivalent options to weigh:

- Have `winml perf --compile` internally delegate generation to a fresh child process pointed at the prepared `_compiled/` directory.
- Or split the public flow so compilation and generation are never done in one process.

Acceptance: `winml perf -m <bundle> --compile --runtime winml-genai --device npu` produces tokens end-to-end in a single command, without a manual second invocation, even when a stage compile teardown-crashes.

## Related

- PR #1081 (adds the Qwen3 genai bundle build + the EPContext salvage that makes each `--compile` run recover completed stages; commit `cfc3e015`).

DingmaomaoBJTU:
## Problem

`winml perf -m <bundle> --compile --runtime winml-genai --device npu` can native-crash (`0xC0000005` / `0xC0000374`) at `og.Model()` load, before the first token — even when every stage is already compiled and cached.

Root cause: `GenaiSession.load()` runs the crash-prone EPContext **compile orchestration** (`_prepare_compiled_bundle`, which spawns and reaps per-stage QNN compile workers that native-fault during interpreter/driver teardown) and then loads `og.Model` in the **same process**. Orchestrating those crash-prone subprocesses leaves the orchestrating process's native accelerator state fragile, so the subsequent model load faults.

The proven workaround was to compile in one process (`--compile`) and then load `_compiled/` in a **fresh** process (no `--compile`).

## Fix

Run the compile orchestration in a throwaway **spawned subprocess**: it writes `_compiled/` to disk and exits, keeping the parent process — the one that loads and runs `og.Model` — pristine. This mirrors the two-step workaround (the load process never runs the poison-prone orchestration) while keeping the in-process streaming/timed generation API intact for every `GenaiSession` consumer, with no cross-process result serialization.

The parent only spawns/reaps **one clean orchestrator child** (which exits 0); it never directly reaps the crashing per-stage QNN workers, so it stays clean for the model load. The child writing `_compiled/` to disk is the durable hand-off — the queue only carries the resolved load-dir path as a fast-path signal.

- Add module-level `_prepare_compiled_bundle_worker` (spawn target): reconstructs a compile-only `GenaiSession`, runs `_prepare_compiled_bundle`, posts `("ok", <load_dir>)` / `("error", <detail>)`.
- Add `GenaiSession._prepare_compiled_bundle_isolated`: spawns the worker, drains its single result, and on any child failure falls back to the on-disk `_compiled/` bundle if present — **never** an in-process compile (which would reintroduce the poisoning), else the original bundle dir.
- `load()` routes the `--compile` path through the isolated variant; the override-only path (JSON rewrite + file mirror, no native accelerator work) stays in-process.
- Docs: `winml perf --compile` now runs compile + generation end-to-end in one command; the manual two-step is kept as an optional way to reuse a prebuilt `_compiled/` bundle.

## Verification

- `tests/unit/session/test_genai_session.py` — 142 pass (12 new: worker posts ok/error, isolated wrapper returns path + drains-on-exit + fallbacks, `load()` routing).
- `tests/unit/commands/test_perf_genai.py` — 54 pass (the only `compile=True` consumer, unchanged by design).
- `ruff check` clean on both changed Python files.

Note: the actual native-crash fix is validated by design (it mirrors the proven cross-process workaround) plus unit tests with mocked subprocess machinery — I don't have NPU hardware to reproduce end-to-end.

Fixes #1087.

### DIFF
--- a/docs/samples/qwen3-genai-bundle.md
+++ b/docs/samples/qwen3-genai-bundle.md
@@ -117,20 +117,26 @@ time-to-first-token (prefill) and decode throughput, and writes a results JSON u
     generated tokens and the saved perf metrics are unaffected — and originates in the
     native runtime below winml-cli, not in the bundle or the build.
 
-!!! tip "More reliable on the NPU: compile once, then run the compiled bundle directly"
-    Doing the EPContext compilation and the generation load in the **same** `--compile`
-    process can fault natively at model load (before the first token) on some setups — the
-    stage compilation can leave the process in a fragile native state. The `--compile` run
-    still writes the compiled bundle to `out/qwen3-bundle/_compiled/` (its `genai_config.json`
-    points at the compiled `context_ctx.onnx` / `iterator_ctx.onnx`), so run generation as a
-    **second, fresh** command pointed straight at that directory — without `--compile`:
+!!! tip "`--compile` compiles and generates in one command"
+    Earlier this could fault at model load: the EPContext stage compilation and the
+    `og.Model` generation load ran in the **same** process, and the compile's crash-prone
+    QNN teardown left that process in a fragile native state. `winml perf --compile` now runs
+    the stage compilation in an **isolated subprocess**, so the process that loads the model
+    for generation stays pristine — the single command above compiles every stage and then
+    generates tokens end-to-end. (This fixes
+    [issue #1087](https://github.com/microsoft/winml-cli/issues/1087).)
+
+    The `--compile` run still writes the compiled bundle to `out/qwen3-bundle/_compiled/`
+    (its `genai_config.json` points at the compiled `context_ctx.onnx` / `iterator_ctx.onnx`),
+    so you can optionally skip recompilation on later runs by pointing `-m` straight at that
+    directory — without `--compile`:
 
     ```bash
-    # 1. Compile once (produces out/qwen3-bundle/_compiled/; re-run if a stage is still missing)
+    # 1. Compile once (produces out/qwen3-bundle/_compiled/)
     winml perf -m out/qwen3-bundle --runtime winml-genai --device npu --compile \
       --compile-timeout 600 --max-new-tokens 20 --prompt "What is the capital of France?"
 
-    # 2. Run the compiled bundle in a fresh process (loads EPContext directly, no re-compile)
+    # 2. Re-run against the compiled bundle (loads EPContext directly, no re-compile)
     winml perf -m out/qwen3-bundle/_compiled --runtime winml-genai --device npu \
       --max-new-tokens 20 --prompt "What is the capital of France?"
     ```
@@ -138,9 +144,7 @@ time-to-first-token (prefill) and decode throughput, and writes a results JSON u
     Dropping `--compile` is safe **only** because `-m` points at the already-compiled
     `_compiled/` directory, whose stages are EPContext graphs loaded as-is — nothing is
     JIT-compiled. (Dropping `--compile` on the original bundle would JIT-compile the source
-    ONNX and fault, per the warning above.) A root-cause fix that isolates model load into its
-    own process is tracked in
-    [issue #1087](https://github.com/microsoft/winml-cli/issues/1087).
+    ONNX and fault, per the warning above.)
 
 ## How it maps to the composite system
 
--- a/docs/samples/qwen3-genai-bundle.md
+++ b/docs/samples/qwen3-genai-bundle.md
@@ -117,35 +117,6 @@ time-to-first-token (prefill) and decode throughput, and writes a results JSON u
     generated tokens and the saved perf metrics are unaffected — and originates in the
     native runtime below winml-cli, not in the bundle or the build.
 
-!!! tip "`--compile` compiles and generates in one command"
-    Earlier this could fault at model load: the EPContext stage compilation and the
-    `og.Model` generation load ran in the **same** process, and the compile's crash-prone
-    QNN teardown left that process in a fragile native state. `winml perf --compile` now runs
-    the stage compilation in an **isolated subprocess**, so the process that loads the model
-    for generation stays pristine — the single command above compiles every stage and then
-    generates tokens end-to-end. (This fixes
-    [issue #1087](https://github.com/microsoft/winml-cli/issues/1087).)
-
-    The `--compile` run still writes the compiled bundle to `out/qwen3-bundle/_compiled/`
-    (its `genai_config.json` points at the compiled `context_ctx.onnx` / `iterator_ctx.onnx`),
-    so you can optionally skip recompilation on later runs by pointing `-m` straight at that
-    directory — without `--compile`:
-
-    ```bash
-    # 1. Compile once (produces out/qwen3-bundle/_compiled/)
-    winml perf -m out/qwen3-bundle --runtime winml-genai --device npu --compile \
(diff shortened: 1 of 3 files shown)


docs/samples/qwen3-genai-bundle.md
```diff
@@ -117,30 +117,34 @@ time-to-first-token (prefill) and decode throughput, and writes a results JSON u
...
     native runtime below winml-cli, not in the bundle or the build.
 
-!!! tip "More reliable on the NPU: compile once, then run the compiled bundle directly"
-    Doing the EPContext compilation and the generation load in the **same** `--compile`
-    process can fault natively at model load (before the first token) on some setups — the
-    stage compilation can leave the process in a fragile native state. The `--compile` run
-    still writes the compiled bundle to `out/qwen3-bundle/_compiled/` (its `genai_config.json`
-    points at the compiled `context_ctx.onnx` / `iterator_ctx.onnx`), so run generation as a
-    **second, fresh** command pointed straight at that directory — without `--compile`:
+!!! tip "`--compile` compiles and generates in one command"
```
DingmaomaoBJTU: Please don't include this

DingmaomaoBJTU: Removed the entire `--compile` one-command tip block in bf4f04ec — the sample now goes straight from the teardown caveat to the `## How it maps` section.

src/winml/modelkit/session/genai_session.py
```diff
@@ -995,6 +1047,114 @@ def _existing_opts_for_ep(provider_options: Any, target: EPName) -> dict:
...
+        # exists to prevent): prefer a fully written _compiled/ bundle if present,
+        # otherwise load the original bundle and let og.Model surface any error.
+        logger.warning(
+            "Isolated compile subprocess did not report success (status=%s, detail=%s); "
+            "falling back to on-disk state",
+            status,
+            payload,
+        )
+        compiled_dir = self._bundle_dir / self._COMPILED_SUBDIR
+        if (compiled_dir / "genai_config.json").exists():
```
xieofxie: Can we avoid loading `_compiled/` after a failed/silent worker unless we can prove it was produced for the current effective config and stage inputs? Right now any existing `_compiled/genai_config.json` from a previous run satisfies this fallback, so if the isolated child fails before rewriting the derived bundle after the source config, stage files, EP override, or provider options changed, the parent can silently load stale EPContext artifacts/config instead of surfacing the failure. A success marker from the child, or a freshness check that covers the current config and stage inputs/options, would make this fallback safe.

src/winml/modelkit/session/genai_session.py
```diff
@@ -458,8 +502,16 @@ def load(self) -> None:
...
+        # subprocess so its crash-prone accelerator teardown cannot poison THIS
+        # process, which must stay pristine to load og.Model (issue #1087).  An
+        # override-only pass (no compilation) merely rewrites JSON + mirrors
+        # files, touches no native accelerator state, and stays in-process.
         load_dir = self._bundle_dir
-        if self._compile or overridden:
+        if self._compile:
+            load_dir = self._prepare_compiled_bundle_isolated(effective_cfg, overridden=overridden)
+        elif overridden:
             load_dir = self._prepare_compiled_bundle(effective_cfg, overridden=overridden)
```
xieofxie: in this case, it will not compile, correct? the name _prepare_compiled_bundle is a little misleading

src/winml/modelkit/session/genai_session.py
DingmaomaoBJTU: Good catch — fixed in d7f99234. The child now writes a completion marker (`.winml_bundle_complete`) as the final step of `_prepare_derived_bundle`, after the config + mirrored files are on disk. The isolated parent removes any stale marker before spawning the child, and the post-failure fallback reuses an existing `_compiled/` **only** when that marker is present — proving it was produced by this run for the current effective config + stage inputs. Without the marker it now loads the original bundle and lets `og.Model` surface the failure instead of silently serving stale EPContext artifacts. Added tests for the fresh-marker reuse, the no-marker refusal, and the stale-marker-cleared-before-spawn cases.

src/winml/modelkit/session/genai_session.py
DingmaomaoBJTU: Agreed — in the override-only path (`compile=False`, the `elif overridden:` branch) it rewrites stage routing without compiling anything, so `compiled` was misleading. Renamed `_prepare_compiled_bundle` → `_prepare_derived_bundle` (and its `_isolated`/`_worker` variants) in d7f99234. "Derived bundle" matches the term already used in the surrounding comments and is accurate whether or not a stage is compiled.

xieofxie (approved):

PR #1104, opened 2026-07-13
Add `--export-type` (generic | optimized) selector to `winml build`

### LINKED ISSUE #1090: Add an explicit `--export-type` (GENERIC | SPECIALIZED) selector for `winml build` instead of device+ep auto-routing
## Context

PR #1081 makes `winml build` emit an onnxruntime-genai **bundle** for a registered
decoder-LLM family (currently Qwen3). The trigger is intentionally implicit: a registered
family **plus** an explicit `--ep qnn` **plus** an NPU target (`--device npu`, or a
`--device auto` that resolves to the NPU). Every other target keeps the stock
single/composite build.

So a single family already has **two possible outputs**, selected today only by the
`device`+`ep` pair:

- a **single-model / composite ONNX** build (today's stock output), and
- a **runtime-specialized multi-component bundle** — the onnxruntime-genai layout
  (`ctx` / `iter` / `embeddings` / `lm_head` + `genai_config.json` + tokenizer), produced
  from a registered recipe.

Note both outputs are already target-aware (the stock build does device-aware quantization
and can compile an EPContext for a specific EP). The real distinction is **output topology
and consuming runtime** — one ONNX model for ORT single-shot inference vs. a bundle for the
onnxruntime-genai generation loop — *not* "portable vs target-specific."

## Suggestion (from review of #1081)

@xieofxie suggested (review threads on `src/winml/modelkit/commands/build.py`:
[3555879362](https://github.com/microsoft/winml-cli/pull/1081#discussion_r3555879362),
[3555913471](https://github.com/microsoft/winml-cli/pull/1081#discussion_r3555913471))
that instead of auto-routing on `device`+`ep`, we expose an **explicit** selector for the
kind of output, rather than hiding one output behind a device/ep combination:

> So in the future if we have more recipes for each, user could easily switch instead of
> automatically switch and hide another.

## Proposal: an explicit `--export-type` selector

Add an explicit **export-type** selector on `winml build`:

```
--export-type GENERIC | SPECIALIZED
```

- **GENERIC** *(default)* — the stock single-model / composite ONNX build. This is today's
  output for every target, so the default preserves current behavior.
- **SPECIALIZED** — a runtime-specialized export produced from a registered recipe (today:
  the onnxruntime-genai NPU bundle).

`--device` / `--ep` / `--precision` remain **target descriptors** *within* the chosen
export-type; they no longer silently decide *which kind* of artifact is produced.

The point: an explicit selector scales better than a device+ep pair once a family has more
than one output, and gives users the choice without hidden routing.

> **Naming — open.** `--export-type` overloads "export": there is already a `winml export`
> command and an export *stage* inside the build pipeline, so `--export-type` may read as
> "the export stage's type." Alternatives to weigh: `--output-kind`, `--artifact`,
> `--profile`, `--build-type`, or a registry-keyed `--recipe` (see *Binary vs. extensible*
> below). `GENERIC`/`SPECIALIZED` value casing / case-insensitivity is an implementation
> detail.

## Key decision: does `SPECIALIZED` *infer* the target, or *require* it?

The recipe already knows the target it needs (today: QNN NPU). Two options:

1. **Infer (recommended).** `winml build -m <model> -o out --export-type SPECIALIZED`
   selects the recipe's target automatically. `--device`/`--ep` are only needed to pick
   among a recipe's supported targets, and an explicit *contradiction*
   (e.g. `--device cpu`) or unavailable hardware is an error. This makes the explicit
   selector genuinely more ergonomic than the device+ep trigger it replaces.
2. **Require.** The user must still pass the recipe's `--device`/`--ep` (e.g. `--ep qnn`
   on an NPU); a mismatch is an error. Closest to today's behavior but offers little over
   the implicit trigger.

This is the central UX decision and is currently unspecified. `--device auto` must be
defined under either option: it should resolve to the recipe's supported target if one is
available on the host, else error.

## Validation: `SPECIALIZED` with an unmet target fails fast (hard error)

When `--export-type SPECIALIZED` is passed but the target cannot serve the selected recipe
— today the only recipe is the QNN NPU bundle, so an explicit non-QNN `--ep`, a
non-NPU `--device`, or an NPU that is unavailable on the host — the build must **abort with
a clear error and a non-zero exit**.

- This is a **fail-fast error** (a `UsageError`, like the existing `-c`/`--use-cache`
  rejections), **not** a soft warning that continues.
- It must **not** silently fall back to `GENERIC` or build a mismatched artifact.

The same fail-fast applies when the family has **no** specialized recipe at all. Because the
user asked for `SPECIALIZED` explicitly, an unmet target is a user error to surface, not
something to auto-correct.

## Prerequisite: recipes must declare their supported targets

The validation above is **not expressible today**: `GenaiBundleRecipe`
(`models/winml/genai_bundle.py`) has no ep/device field — QNN+NPU is hardcoded at the call
site (`build_genai_bundle(..., ep="qnn", device="npu")`) and as orchestrator defaults. This
work must first add a declarative *supported-targets* field to the recipe (e.g.
`supported_targets: tuple[(ep, device), ...]`) so `build` can validate `--export-type
SPECIALIZED` against it and drive target inference, keeping the routing architecture-agnostic.

## Inherited constraints under `SPECIALIZED`

`SPECIALIZED` keeps the recipe-driven pipeline, so the controls the fast path already
rejects still apply and must error consistently: `-c/--config`, `--use-cache`,
`--quant/--no-quant`, `--optimize/--no-optimize`, `--analyze/--no-analyze`,
`--compile/--no-compile`, `--max-optim-iterations`, `--allow-unsupported-nodes`.
`--precision` remains the one supported transformer knob.

## Binary vs. extensible axis

`GENERIC`/`SPECIALIZED` is a **binary**, but the stated motivation is scaling to *multiple*
recipes per family. Two specialized recipes for one family would make `SPECIALIZED`
ambiguous again — the same problem, one level up. Decide whether the axis is a fixed 2-value
enum plus a later recipe sub-selector, or an **extensible, registry-keyed value** from the
start (e.g. `--export-type <recipe-name>` / a dedicated `--recipe`), so the primary flag
names the concrete recipe when a family has more than one.

## Interaction with the current device+ep trigger

`--export-type` is the explicit selector and **overrides** routing when passed. When
**omitted**, the existing `--ep qnn` + NPU trigger from #1081 is kept as a **convenience
shortcut** for `SPECIALIZED`, so today's opt-in, non-destructive behavior is unchanged and
no existing command breaks. `--export-type GENERIC` forces the stock build even for a
registered family on the NPU.

## Reconcile with `perf` (two vocabularies)

`perf` already selects the engine with `--runtime winml | winml-genai`
(`commands/perf.py`). Under this proposal a user **builds** `--export-type SPECIALIZED` then
**benchmarks** `--runtime winml-genai` — two different vocabularies for the same bundle.
Options: (a) share one vocabulary across `build`/`perf`/`export`; (b) keep `--export-type`
(build *artifact*) and `--runtime` (inference *engine*) as deliberately distinct axes and
document the mapping; or (c) have `perf` auto-detect a genai bundle from its directory
layout so `--runtime` is rarely needed. This is a decision, not just a note — the original
issue proposed `winml | winml-genai` precisely for this symmetry.

## Discoverability

There is no way today to learn that a model supports `SPECIALIZED`. Consider surfacing the
supported export-types (and their targets) in `winml inspect` / `analyze`, and/or having the
`GENERIC` build emit a hint when the family also has a specialized recipe.

## Why it's not in #1081

The device+ep trigger was locked with the product owner for that PR: it is opt-in
(`--ep qnn` must be explicit), non-destructive (every other target is unchanged), and needs
no new surface. Adding an explicit `--export-type` flag is a broader CLI-surface design
change, so it's better evaluated on its own.

## Proposed follow-up

- Add `--export-type GENERIC | SPECIALIZED` as the explicit output selector for `build`,
  defaulting to `GENERIC`; settle the flag name (see *Naming*).
- Decide **infer vs. require** for `SPECIALIZED` targets, including `--device auto`.
- Add a declarative *supported-targets* field to `GenaiBundleRecipe` so target validation
  and inference are data-driven (prerequisite).
- Fail-fast (non-zero exit, `UsageError`) when `SPECIALIZED` can't serve the target or the
  family has no recipe — never silently fall back to `GENERIC`.
- Keep the `--ep qnn` + NPU trigger as a shortcut for `SPECIALIZED` when `--export-type` is
  omitted; `--export-type` overrides it when passed.
- Preserve the existing `SPECIALIZED` control rejections (config/cache/quant/optimize/etc.).
- Decide **binary enum vs. registry-keyed recipe value** for multi-recipe families.
- Reconcile with `perf`'s `--runtime winml | winml-genai` (shared vocab, documented mapping,
  or perf auto-detect).
- Surface supported export-types for discoverability (`inspect`/`analyze`, or a build hint).

Tracking the design discussion here so #1081 can merge with the current implicit trigger.

DingmaomaoBJTU:
Closes #1090.

## Summary

Replaces the implicit device+ep auto-routing to the onnxruntime-genai bundle with an explicit `--export-type` selector on `winml build`, so choosing between a stock ONNX export and a runtime-optimized bundle is a deliberate, discoverable decision instead of a side effect of `--ep qnn` + NPU.

| `--export-type` | Result |
|---|---|
| `generic` (default) | Stock single/composite ONNX build (unchanged). |
| `optimized` | The family's registered runtime-optimized recipe — today the onnxruntime-genai NPU bundle. |

## Behavior

- **`optimized` resolves the target, then builds for it.** `--ep`/`--device` are resolved exactly like a generic build — an explicit value is honored, otherwise the host is hardware-probed — and the bundle is built for that resolved `(ep, device)`. What you see is what you get: on an NPU host `--export-type optimized` alone builds the NPU bundle; pin `--ep qnn --device npu` to build it on any host (e.g. CI).
- **Fail-fast, no silent fallback.** `optimized` errors when the family has no recipe, the input is a pre-exported `.onnx`, it is module mode, or the resolved `(ep, device)` is not one the recipe's `supported_targets` declares (the error names the ep/device).
- **Backward compatible.** When `--export-type` is omitted, the existing shortcut still applies: a registered family with an explicit `--ep qnn` on an NPU target (`--device npu`, or `--device auto` resolving to the NPU) routes to its bundle. `--export-type generic` always forces the stock build, even on the NPU.

## Changes

- **Recipe data model** (`genai_bundle.py`): new `GenaiTarget` frozen dataclass and a required `supported_targets` field on `GenaiBundleRecipe`, validated (non-empty) at registration. New symbols exported from the package `__init__`.
- **Qwen3 recipe**: declares `supported_targets=(GenaiTarget(ep="qnn", device="npu"),)`.
- **`winml build`** (`build.py`): adds the `--export-type` choice option and refactors the routing into small helpers (`_resolve_genai_recipe`, `_resolve_optimized_target`, `_maybe_build_genai_bundle`) that keep the switch data-driven and architecture-agnostic. `optimized` resolves `--ep`/`--device` like the generic path and matches the resolved target against the recipe's `supported_targets`.
- **Docs**: `docs/commands/build.md` (flag row + reframed genai-bundle section) and `docs/samples/qwen3-genai-bundle.md` now lead with `--export-type optimized` and document the resolve-first-then-error behavior plus the backward-compatible shortcut.

## Sample

Build the optimized onnxruntime-genai NPU bundle for a decoder LLM. On an NPU host no target flags are needed; pin `--ep qnn --device npu` to build the bundle on a host without the accelerator (e.g. CI):

```bash
winml build -m Qwen/Qwen3-0.6B -o out/ --export-type optimized
# host-independent (explicit target skips host detection):
winml build -m Qwen/Qwen3-0.6B -o out/ --export-type optimized --ep qnn --device npu
```

Force a stock ONNX build instead:

```bash
winml build -m Qwen/Qwen3-0.6B -o out/ --export-type generic
```

The selector is discoverable in `--help`:

```
--export-type [generic|optimized]
                                Output selector. 'generic' builds the stock
                                single/composite ONNX model. 'optimized'
                                builds the registered runtime-optimized recipe
                                for the family (today: the onnxruntime-genai
                                NPU bundle) for the resolved --ep/--device; it
                                errors if the family has no recipe or the
                                resolved target is not one the recipe supports.
                                ...
```

Fail-fast validation (verified in terminal). No recipe for the model type:

```
$ winml build -m microsoft/resnet-50 -o out --export-type optimized
Error: --export-type optimized: no optimized recipe is registered for model type 'resnet'.
```

Resolved target the recipe doesn't support (here forced with `--device cpu`):

```
$ winml build -m Qwen/Qwen3-0.6B -o out --export-type optimized --device cpu
Error: --export-type optimized is not supported for ep=CPUExecutionProvider, device=cpu (supported: --ep qnn --device npu).
```

## Tests

New CLI routing tests (resolve-and-build, `generic` forces stock, unregistered-family error, unsupported resolved target errors, explicit target builds on a non-NPU host) plus focused unit tests for target matching/rejection and the `optimized` preconditions; registry tests for the empty-`supported_targets` guard. The genai build/routing suites (`tests/unit/commands/test_build_genai_bundle.py` and `tests/unit/models/winml/`) pass; `ruff check` and strict `mypy` clean.


src/winml/modelkit/commands/build.py
```diff
@@ -629,6 +746,18 @@ def _maybe_build_genai_bundle(
 @cli_utils.precision_option(
     optional_message="With -c, applied only when --device or --precision is passed.",
 )
+@click.option(
+    "--export-type",
+    type=click.Choice(["generic", "specialized"], case_sensitive=False),
```
xieofxie: specialized is a strange name, not meaningful when set it - what should I expect?

DingmaomaoBJTU: How about optimized?

xieofxie: no idea now.. need time to think

xieofxie: optimized looks good

src/winml/modelkit/commands/build.py
```diff
@@ -942,20 +1076,21 @@ def _patch_device(cfg: WinMLBuildConfig) -> None:
...
+        # ctx/iter/embeddings/lm_head + genai_config.json + tokenizer), inferring
+        # the recipe's target. When ``--export-type`` is omitted, a registered
+        # family targeted at the NPU HTP via an explicit ``--ep qnn`` (with
+        # ``--device npu`` or a ``--device auto`` resolving to the NPU) still
+        # routes here as a backward-compatible shortcut. The switch is data-driven
+        # (the genai-bundle recipe registry) and architecture-agnostic -- every
+        # model-specific value lives in the recipe. ``--export-type generic`` (or
+        # any other device/ep combination without the flag) keeps the stock
+        # single/composite build.
         if _maybe_build_genai_bundle(
```
xieofxie, line 1097: This optimized path is still reached only after the earlier `if ep is None: resolve_check_device_ep(...)` block has run. That means `winml build -m Qwen/Qwen3-0.6B -o out --export-type optimized` still probes local hardware and can fail before the recipe target is inferred, which contradicts the new contract/docs that optimized builds are hardware-independent and don't require detecting the NPU locally. Can we route explicit `--export-type optimized` before auto-resolving device/EP, or skip that resolution when the optimized target will be inferred from the recipe?

xieofxie:
For your description

`Build the optimized onnxruntime-genai NPU bundle for a decoder LLM (the target is inferred from the recipe, so --ep/--device are optional):`

If it is the case, I suggest we resolve first and then error that the --export-type is not supported for ep=xx, device=xx

Infer target from recipe is bad due to
1. conflict with ep device flags resolving
2. if we have more optimized recipes, it is hard to resolve

src/winml/modelkit/commands/build.py
DingmaomaoBJTU: Good catch — fixed in ffc3af92. `--export-type optimized` now skips the `resolve_check_device_ep` auto-EP probe entirely: the optimized branch resolves (ep, device) from the recipe via `_resolve_optimized_target`, so the build no longer depends on detecting the accelerator locally. Added a regression test that makes `resolve_check_device_ep` raise and asserts the bundle still builds (and the probe is not called).

DingmaomaoBJTU: Good catch — and per your follow-up design comment we deliberately went the other way. The contract is now **resolve-first-then-error** (WYSIWYG) rather than hardware-independent.

`--export-type optimized` now resolves `--ep`/`--device` exactly like a generic build — so yes, the `if ep is None: resolve_check_device_ep(...)` probe still runs when you don't pin flags — and then matches that resolved `(ep, device)` against the recipe's `supported_targets`, failing fast when it isn't supported:

```
$ winml build -m Qwen/Qwen3-0.6B -o out --export-type optimized --device cpu
Error: --export-type optimized is not supported for ep=CPUExecutionProvider, device=cpu (supported: --ep qnn --device npu).
```

So on a non-NPU host with no flags it now errors (naming the ep/device) instead of silently producing a bundle, and `--ep qnn --device npu` builds it on any host. The contract/docs were updated to match. Pushed in 19520139.

DingmaomaoBJTU:
Adopted — thanks for the push. Inferring the target was replaced with **resolve-first-then-error**, exactly as you suggested.

`--export-type optimized` now resolves `--ep`/`--device` the same way a generic build does (explicit value honored, otherwise hardware-probed) and then matches that resolved `(ep, device)` against the recipe's `supported_targets`. No match → fail fast, naming the target:

```
$ winml build -m Qwen/Qwen3-0.6B -o out --export-type optimized --device cpu
Error: --export-type optimized is not supported for ep=CPUExecutionProvider, device=cpu (supported: --ep qnn --device npu).
```

This removes the conflict with ep/device resolution and scales cleanly as more optimized recipes are added — the resolved target is simply looked up. `--ep qnn --device npu` builds the bundle on any host. Pushed in 19520139; PR description + docs updated.

src/winml/modelkit/models/winml/genai_bundle.py
```diff
@@ -99,6 +118,10 @@ class GenaiBundleRecipe:
         assemble: Callable assembling the copied/converted components into a
             bundle directory and writing ``genai_config.json``; must accept the
             keywords the orchestrator passes (see :func:`build_genai_bundle`).
+        supported_targets: The ``(ep, device)`` pairs this recipe can build for.
+            The first entry is the preferred/default target inferred when the
```
xieofxie: This still describes the previous "infer the recipe's first target" behavior, but the latest `build.py` flow now resolves `--ep`/`--device` first and only uses `supported_targets` to match or reject that resolved target. Could you update this to say the entries are the targets the recipe can serve (and maybe that explicit `--ep qnn --device npu` is how to select this one on any host)? There's a similar stale example comment in `build.py` saying the target is inferred.

DingmaomaoBJTU: Done — both stale spots updated in 8d70ae75.

- `genai_bundle.py` `supported_targets` docstring now says the entries are the targets the recipe can build for, that `--export-type optimized` resolves `--ep`/`--device` first and matches against them (rejecting an unlisted target), and that `--ep qnn --device npu` selects this recipe's target on any host.
- `build.py` example comment no longer says "infers the recipe's NPU/QNN target"; it now notes optimized resolves `--ep`/`--device` like a generic build (no flags needed on an NPU host, or pin `--ep qnn --device npu` to build anywhere).

xieofxie (approved):

PR #1105, opened 2026-07-14
build: track uv lockfile

DingmaomaoBJTU:
## Summary
- Track the resolved uv dependency graph in uv.lock.
- Make CI install dependencies with uv sync --locked.
- Prevent unreviewed upstream dependency drift from changing development or CI environments.

## Validation
- uv lock --check
- uv sync --locked --all-extras --all-groups

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -1,6 +1,14 @@
 version = 1
 revision = 3
 requires-python = "==3.11.*"
+resolution-markers = [
+    "platform_machine == 'x86_64' and sys_platform == 'linux'",
+    "platform_machine == 'AMD64' and sys_platform == 'win32'",
+]
+supported-markers = [
+    "platform_machine == 'x86_64' and sys_platform == 'linux'",
+    "platform_machine == 'AMD64' and sys_platform == 'win32'",
+]
 
 [[package]]
 name = "aiohappyeyeballs"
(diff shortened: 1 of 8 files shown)


uv.lock
```diff
@@ -0,0 +1,4132 @@
+version = 1
```
xieofxie, line 1: should we check in this?  I thought we should update pyproject.toml to support only x64

DingmaomaoBJTU: Update: based on the latest decision, I reverted the x64-only metadata change and restored the lockfile approach. `uv.lock` is checked in again, and the docs, lint, and modelkit CI workflows use `uv sync --locked`. I verified the restored lock with `uv lock --check` and `uv sync --locked --all-extras --all-groups`.

DingmaomaoBJTU: Went with a combined approach that follows your suggestion:

- Added `[tool.uv].environments` restricting resolution to `linux/x86_64` and `win32/AMD64` (the `linux/x86_64` entry is needed for the docs CI on ubuntu-latest). This declares x64-only support and makes `uv sync` fail fast with a clear "unsupported environment" error on native Arm64 Python instead of an opaque missing-wheel error.
- Kept `uv.lock` checked in for reproducibility; it is now x64-only (re-resolved under the new environments).
- Documented the Windows-on-Arm requirement in README and installation docs: Arm64 devices must use an x64 Python 3.11 build (runs under emulation), since PyTorch and the Windows ML runtime packages ship no `win_arm64` wheels.

Note: metadata alone cannot auto-switch a user's native Arm64 interpreter to x64, so the docs cover that final step. Verified with `uv lock --check` and `uv sync --locked --all-extras --all-groups` on x64 Python.

xieofxie (approved):

PR #1106, opened 2026-07-14
feat(config): add dynamic export controls to config command

xieofxie:
## Summary

Adds the dynamic export controls (`--input-specs`, `--export-config`, `--dynamic-axes`) to `winml config`, bringing it to parity with `winml build` / `winml perf` / `winml export`. This is a follow-up to #1083, which introduced these controls for build/perf and the shared `cli_utils` option/loader helpers.

## What changed

- **`commands/config.py`**
  - Adds the three shared options (`input_specs_option`, `export_config_option`, `dynamic_axes_option`) and their params.
  - Loads a sparse override mapping via `cli_utils.load_export_overrides` (validates early, doesn't clobber auto-detected fields).
  - New `_merge_export_overrides` helper applies them via the shared `merge_export_overrides`, so `--input-specs` patches the auto-resolved `input_tensors` **by name** (unlisted inputs and their `dtype`/`value_range` preserved) and symbolic dims re-derive dynamic axes through `WinMLExportConfig.__post_init__`.
  - Wired into the single-config, `--module` list, and composite-pipeline paths.
  - Rejected for pre-exported ONNX inputs and for configs with `export=null` (clear `UsageError`).
  - Shows the provided files in the console summary; adds docstring examples.
- **`docs/commands/config.md`**: three new flag rows, an example, and a pitfall note.
- **`tests/unit/commands/test_config_cli.py`**: `TestConfigExportControls` — help visibility, override forwarding/parsing, no-op when absent, and ONNX rejection.

## Verification

- `ruff check` + `ruff format` clean.
- `tests/unit/commands/test_config_cli.py` (55) and `test_config_composite_resolution.py` + `test_config_value_priority.py` (63) pass.
- Doc flag-alignment check passes.


src/winml/modelkit/commands/config.py
```diff
@@ -542,6 +610,7 @@ def _generate_pipeline_configs(
             trust_remote_code=trust_remote_code,
             ep=ep,
         )
+        cfg = _merge_export_overrides(cfg, export_overrides)
```
DingmaomaoBJTU: For composite pipelines this applies the same `export_overrides` to every sub-component. Since `_patch_input_tensors` appends any name that doesn't match an existing input, a shared `--input-specs`/`--dynamic-axes` (e.g. `{"input_ids": ...}`) gets patched onto the component that has that input but silently appended as a bogus extra input on the ones that don't (and a `dynamic_axes` entry for a missing input just persists). Unlike `build`, which never fans overrides out across heterogeneous components, this can emit invalid per-component configs. Can we scope the overrides to components whose inputs actually match, or at least document the broadcast behavior?

tests/unit/commands/test_config_cli.py
```diff
@@ -693,3 +693,92 @@ def test_non_qdq_onnx_has_default_quant(self, runner: CliRunner, tmp_path: Path)
...
+        input_specs.write_text(
+            json.dumps({"pixel_values": {"dtype": "float32", "shape": ["batch", 3, 224, 224]}})
+        )
+        export_config = tmp_path / "export.json"
+        export_config.write_text(json.dumps({"opset_version": 18}))
+        dynamic_axes = tmp_path / "dynamic_axes.json"
+        dynamic_axes.write_text(json.dumps({"pixel_values": {"0": "batch"}}))
+
+        with patch(
+            "winml.modelkit.config.merge_export_overrides",
```
DingmaomaoBJTU: This patches out `merge_export_overrides`, so the test only proves the sparse override dict is forwarded — the actual by-name patching and symbolic-dim -> dynamic-axes re-derivation (the behavior this PR really adds) is never exercised for `config`. Could we add one integration-style test that lets `merge_export_overrides` run and asserts on `to_dict()` (unlisted input preserved; a `["batch", ...]` shape produces `dynamic_axes`)? Also nothing currently covers the `export=null` rejection in `_merge_export_overrides`, the `--module` list path, or the composite path.

src/winml/modelkit/commands/config.py
```diff
@@ -266,6 +302,15 @@ def config(
                 ) from e
             _shape_config_file = shape_config_path.name
 
+        # Load export CLI overrides (--input-specs/--export-config/--dynamic-axes).
+        # Returned sparse so unspecified fields don't clobber auto-detected values;
+        # applied per generated config via merge_export_overrides below.
+        export_overrides = cli_utils.load_export_overrides(
```
DingmaomaoBJTU: Minor: this loads and validates the override JSON before the ONNX-input rejection below (line 324). So `winml config -m model.onnx --dynamic-axes bad.json` surfaces "Invalid export configuration" instead of the clearer "not supported for pre-exported ONNX". Moving the ONNX guard ahead of the load would give the more relevant error and skip the file I/O.

DingmaomaoBJTU (commented):
A few notes on the new export controls — mainly the composite-pipeline fan-out and test coverage. Otherwise nice work bringing this to parity with build/perf.

src/winml/modelkit/commands/config.py
xieofxie: it should not support composite pipeline, we should raise

tests/unit/commands/test_config_cli.py
xieofxie: Rewrote these into integration-style tests: they build a real `WinMLBuildConfig` and let `merge_export_overrides` actually run, then assert on `to_dict()` output — verifying by-name `input_tensors` patching and symbolic-dim to `dynamic_axes` derivation. Also added coverage for `export=null` rejection, the `--module` list path, the composite rejection, and the ONNX-input rejection ordering.

src/winml/modelkit/commands/config.py
xieofxie: Moved the ONNX-input rejection ahead of the JSON load (config.py:319): it now checks raw flag presence and raises the ONNX-specific error before `load_export_overrides` parses/validates the JSON. Bad JSON on an ONNX input now surfaces the ONNX problem instead of 'Invalid export configuration', and we skip the needless file I/O.

src/winml/modelkit/commands/config.py
xieofxie: Done — config now raises a `UsageError` when export overrides (`--input-specs`/`--export-config`/`--dynamic-axes`) are passed for a composite (multi-component) model, instead of fanning them out across sub-components. Each sub-component has its own export inputs, so a shared spec would append bogus inputs on components lacking the named input. See config.py:371.

src/winml/modelkit/commands/config.py
```diff
@@ -345,15 +415,15 @@ def config(
                 ep=ep,
             )
             if isinstance(result, list):
-                configs = result
+                configs = [_merge_export_overrides(cfg, export_overrides) for cfg in result]
```
DingmaomaoBJTU: The `--module` path fans the same `--input-specs`/`--dynamic-axes` onto every matched submodule config by name. Since `_patch_input_tensors` *appends* any spec whose name doesn't match an existing input, running this against submodules whose real input names differ from the spec would append bogus inputs to each config — the same failure mode the composite branch rejects just above. For same-class `--module` matches the inputs are homogeneous so it's usually fine, but two things are worth a look:

1. The composite-block comment says "config never fans these overrides out", which reads as contradicting this line — it does fan them across submodule configs here.
2. `config.md`'s pitfall note only calls out the composite rejection. A short caveat that `--module` applies the specs to *every* submodule (and unmatched names get appended) would help users avoid surprise inputs.

src/winml/modelkit/commands/config.py
```diff
@@ -306,6 +362,20 @@ def config(
                 hf_model, model_type, task, trust_remote_code=trust_remote_code
             )
             if pipeline_components:
+                # Export controls target a single export graph; a composite model
+                # has one export per sub-component with distinct inputs. Applying a
+                # shared --input-specs/--dynamic-axes across them would append bogus
+                # inputs on components that lack the named input. Reject instead of
+                # emitting invalid per-component configs (unlike build, config never
+                # fans these overrides out).
+                if export_overrides:
```
DingmaomaoBJTU: Minor UX asymmetry: `load_export_overrides` (≈line 329) validates the override JSON *before* this composite rejection runs, so a composite model + a malformed `--dynamic-axes`/`--input-specs` file surfaces `Invalid export configuration` instead of this composite-specific message. That's the opposite of the ONNX path, which you deliberately reject before loading/validating (and cover with `test_onnx_rejection_precedes_json_validation`). Composite detection needs the model resolved, so you can't reject purely on flag presence as cheaply — but if you want the symmetry, moving the `load_export_overrides` call below `_resolve_composite_model_components` and rejecting here on `(input_specs or export_config or dynamic_axes)` would let the composite message win first.

src/winml/modelkit/commands/config.py
```diff
@@ -407,6 +477,21 @@ def config(
                 f"   \U0001f4c1 [bold]Shape config:[/bold] "
                 f"{_shape_config_file}  [green]\u2713[/green]"
             )
+        if input_specs:
+            console.print(
+                f"   \U0001f4c1 [bold]Input specs:[/bold]  "
+                f"{Path(input_specs).name}  [green]\u2713[/green]"
```
DingmaomaoBJTU: Trivial: `input_specs`/`export_config`/`dynamic_axes` are already `Path` (declared `Path | None`), so `Path(input_specs).name` can just be `input_specs.name` (same for `export_config`/`dynamic_axes` below).

DingmaomaoBJTU (commented):
Reviewed the export-controls addition to `winml config`. Overall this is clean and mirrors `winml build` faithfully — the shared `merge_export_overrides` / `_patch_input_tensors` by-name patching, the `export=null` guard, and the up-front ONNX rejection (with its ordering test) are all solid, and the integration-style tests that let `merge_export_overrides` actually run are a nice improvement over pure mocks. No correctness bugs found. A few non-blocking notes below: two minor UX/consistency points and one trivial nit.

src/winml/modelkit/commands/config.py
xieofxie: yes, it should also not be supported by --module

xieofxie: Done — `--module` now raises a `UsageError` on `--input-specs`/`--export-config`/`--dynamic-axes` up front (before any config generation), matching the composite rejection, so it no longer fans a shared spec across submodule configs. Removed the per-config merge from the `--module` list branch and updated both the composite-block comment and `config.md` (flag rows + pitfall) to note the `--module` rejection.

src/winml/modelkit/commands/config.py
xieofxie: Done — simplified `Path(input_specs).name` to `input_specs.name` for all three (they're already `Path | None`).

src/winml/modelkit/commands/config.py
xieofxie: Good catch — restored the symmetry. `load_export_overrides` now runs *after* `_resolve_composite_model_components`, and the composite branch rejects on raw flag presence (`input_specs or export_config or dynamic_axes`) before the JSON is loaded/validated. A composite model + malformed `--dynamic-axes` now surfaces the composite-specific message instead of 'Invalid export configuration'. Added `test_composite_rejection_precedes_json_validation` to lock it in.

tests/unit/commands/test_config_cli.py
```diff
@@ -693,3 +693,283 @@ def test_non_qdq_onnx_has_default_quant(self, runner: CliRunner, tmp_path: Path)
...
+        ):
+            result = runner.invoke(
+                config, ["-m", "some/seq2seq", "--dynamic-axes", str(dynamic_axes)]
+            )
+
+        assert result.exit_code != 0
+        assert "composite" in result.output
+        assert "Invalid export configuration" not in result.output
+        mock_pipeline.assert_not_called()
+        """_merge_export_overrides raises when the generated config has export=null."""
```
DingmaomaoBJTU, line 921: This refactor accidentally fused two tests. The old `def test_merge_export_overrides_rejects_export_null` line was replaced by `def test_composite_rejection_precedes_json_validation`, but the old method's body was left behind — so lines 919–931 (the stray `"""...export=null"""` statement, `import click`, and the `pytest.raises(..., match="export=null")` block) now execute at the tail of the composite test.

Two effects:
1. `test_merge_export_overrides_rejects_export_null` no longer exists as a discrete test — that coverage silently vanished from the suite.
2. `test_composite_rejection_precedes_json_validation` now also asserts the export=null behavior under the wrong name. Both assertions pass, so CI stays green and this is easy to miss.

Fix: pull lines 919–931 back out into their own `def test_merge_export_overrides_rejects_export_null(self) -> None:` method (with its docstring).

DingmaomaoBJTU (commented):
Thanks for the follow-up — this cleanly resolves all three earlier notes: `--module` and composite now both reject on raw flag presence *before* `load_export_overrides`, so the path-specific error wins over `Invalid export configuration`; `load_export_overrides` is scoped to the single-HF path (verified `export_overrides` is still defined before `_merge_export_overrides` on that branch); the `Path(...).name` nit is gone; and the docs read accurately. One new issue from the test refactor — see inline.

tests/unit/commands/test_config_cli.py
xieofxie: Good catch — exactly right, the refactor left the `export=null` body orphaned at the tail of `test_composite_rejection_precedes_json_validation`. Split it back into its own `test_merge_export_overrides_rejects_export_null` method. Collection now reports 62 items (was 61), confirming the discrete test is restored; full file passes.

DingmaomaoBJTU (approved):
Approving. All review feedback is addressed:

- `--input-specs`/`--export-config`/`--dynamic-axes` are rejected up front (on raw flag presence, before loading the JSON) for the ONNX, `--module`, and composite paths, so the path-specific error wins over a downstream `Invalid export configuration`; `load_export_overrides` is scoped to the single-HF path and `_merge_export_overrides` still applies there.
- `ae412b3c` splits the accidentally-fused test back out — `test_merge_export_overrides_rejects_export_null` is a discrete method again and the composite test ends cleanly.
- `d7aeaa8a` (build-test autouse stub for `resolve_composite_components`) is a sound, well-targeted flake fix that matches the existing autouse pattern; composite-specific tests still override it.

Docs read accurately. Nice work. (Note: I reviewed by reading the diff/branch; the suite couldn't be run on my ARM64 host, so a green CI run is worth a final glance before merge.)

PR #1108, opened 2026-07-14
feat(telemetry): record model_id on WinMLCLIAction

### LINKED ISSUE #1107: Record model_id on WinMLCLIAction telemetry
## Summary

CLI telemetry (`WinMLCLIAction`) currently records `action_name`, `device`, `ep`, `duration_ms`, and `success`, but not which model each command ran against. We want visibility into which models people run — especially public HuggingFace models — without leaking PII.

## Problem

The `-m/--model` value can be a HuggingFace ID (`microsoft/resnet-50`), a local `.onnx` file path, or a build output directory. Local paths may embed a username, so recording the raw value would reintroduce exactly the path/username leakage the existing exception-scrubbing pipeline exists to prevent.

## Proposal

Add a `model_id` attribute to `WinMLCLIAction`:

- Clean HF ids (`org/name`, not present on disk) → recorded as-is.
- Local paths / directories / bare names → anonymized to `<local:.onnx>` / `<local:dir>` (no path fragment ever emitted).
- Commands exposing `--model-id` (eval/quantize) record that clean HF id directly.

timenick:
## Summary

- Add a `model_id` attribute to the `WinMLCLIAction` telemetry event so we can see which model each CLI command ran against.
- Anonymize local model references: `<local:.onnx>` / `<local:dir>` — no path fragment (basename, dir names, username) is ever emitted. Clean HuggingFace ids (`org/name`) pass through as-is.
- For commands exposing `--model-id` (eval/quantize), the clean HF id is recorded directly, bypassing the scrubbed `-m` value.

## Changes

- `telemetry/utils.py`: new `_scrub_model_ref` classifier + `<local:...>` marker.
- `telemetry/telemetry.py`: `log_action` gains a `model_id` param; `model_id` added to the `WinMLCLIAction` allowlist.
- `telemetry/click_group.py`: wires `--model-id`-preferred, otherwise-scrubbed-`-m` into emission.

Closes #1107

### DIFF
--- a/src/winml/modelkit/telemetry/click_group.py
+++ b/src/winml/modelkit/telemetry/click_group.py
@@ -30,6 +30,7 @@ from typing import Any
 import click
 
 from .telemetry import Telemetry
+from .utils import _scrub_model_ref
 
 
 _INSTRUMENTED_ATTR = "_winmlcli_instrumented"
@@ -100,8 +101,12 @@ def _instrument(cmd: click.Command) -> click.Command:
             raise
         finally:
             duration_ms = int((time.perf_counter() - start) * 1000)
+            model_id = _param(ctx, "model_id")
+            if model_id is None:
+                model_id = _scrub_model_ref(_param(ctx, "model"))
             telemetry.log_action(
                 action_name=cmd.name or "",
+                model_id=model_id,
                 device=_param(ctx, "device"),
                 ep=_param(ctx, "ep"),
                 duration_ms=duration_ms,
--- a/src/winml/modelkit/telemetry/click_group.py
+++ b/src/winml/modelkit/telemetry/click_group.py
@@ -101,9 +101,13 @@ def _instrument(cmd: click.Command) -> click.Command:
             raise
         finally:
             duration_ms = int((time.perf_counter() - start) * 1000)
-            model_id = _param(ctx, "model_id")
-            if model_id is None:
-                model_id = _scrub_model_ref(_param(ctx, "model"))
+            # Both inputs go through the scrubber: --model-id is trusted to be
+            # a clean HF id but isn't validated, so a stray path there is
+            # anonymized too. --model-id wins when present (eval/quantize);
+            # otherwise fall back to the scrubbed -m value.
+            model_id = _scrub_model_ref(_param(ctx, "model_id")) or _scrub_model_ref(
+                _param(ctx, "model")
+            )
             telemetry.log_action(
                 action_name=cmd.name or "",
                 model_id=model_id,
--- a/src/winml/modelkit/telemetry/utils.py
+++ b/src/winml/modelkit/telemetry/utils.py
@@ -151,6 +151,40 @@ def _looks_like_path(token: str) -> bool:
     return "\\" in token and _PACKAGE_ROOT in token.replace("\\", "/")
 
 
+def _model_ref_local_marker(value: str) -> str:
+    """Anonymized marker for a local model reference.
+
+    Emits only the file extension or the literal ``dir`` — never any path
+    fragment (no basename, no directory names, no username).
+    """
+    tail = value.replace("\\", "/").rsplit("/", 1)[-1]
+    ext = Path(tail).suffix
+    return f"<local:{ext}>" if ext else "<local:dir>"
+
+
+def _scrub_model_ref(value: str | tuple[str, ...] | None) -> str | None:
+    """Classify a ``-m/--model`` reference for telemetry.
+
+    Clean HuggingFace ids (``org/name``, single slash, not present on
+    disk) pass through verbatim. Everything else — drive-letter paths,
+    absolute paths, on-disk paths, and bare names — collapses to a
+    ``<local:...>`` marker that carries no path content.
+    """
+    if isinstance(value, tuple):
+        value = value[0] if value else None
+    if not value:
+        return None
+    normalized = value.replace("\\", "/")
+    if re.match(r"^[A-Za-z]:[\\/]", value) or normalized.startswith("/"):
+        return _model_ref_local_marker(value)
+    if Path(value).exists():
+        return _model_ref_local_marker(value)
+    parts = normalized.split("/")
+    if len(parts) == 2 and parts[0] and parts[1]:
+        return value
+    return _model_ref_local_marker(value)
+
+
 def _encode_cache_entry(entry: dict[str, Any]) -> str:
     """Encode a cache entry to a single-line string: ``base64(json(dict))``."""
     raw = json.dumps(entry, ensure_ascii=False).encode("utf-8")
--- a/src/winml/modelkit/telemetry/utils.py
+++ b/src/winml/modelkit/telemetry/utils.py
@@ -151,6 +151,14 @@ def _looks_like_path(token: str) -> bool:
     return "\\" in token and _PACKAGE_ROOT in token.replace("\\", "/")
 
 
+# A clean HuggingFace Hub id: one or two segments of the Hub charset, with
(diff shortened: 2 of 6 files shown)


src/winml/modelkit/telemetry/utils.py
```diff
@@ -151,6 +151,40 @@ def _looks_like_path(token: str) -> bool:
...
+        return None
+    normalized = value.replace("\\", "/")
+    if re.match(r"^[A-Za-z]:[\\/]", value) or normalized.startswith("/"):
+        return _model_ref_local_marker(value)
+    if Path(value).exists():
+        return _model_ref_local_marker(value)
+    parts = normalized.split("/")
+    if len(parts) == 2 and parts[0] and parts[1]:
+        return value
+    return _model_ref_local_marker(value)
```
xieofxie, line 201: **Single-segment Hub ids collapse to `<local:dir>`, defeating the feature for legacy/canonical models.**

Only `org/name` (exactly one slash) passes through; a bare Hub id with no org prefix falls through to the local marker. These are all valid, commonly-used hub ids:

```
bert-base-uncased  -> <local:dir>
gpt2               -> <local:dir>
resnet-50          -> <local:dir>
distilbert-base-uncased-finetuned-sst-2-english -> <local:dir>
```

(verified by running the PR's `_scrub_model_ref` directly.) Notably `winml perf -m bert-base-uncased` is a documented example in `perf.py`'s docstring, and `perf` has no `--model-id`, so this exact invocation records `<local:dir>` instead of `bert-base-uncased`. That contradicts the PR goal ("see which model each CLI command ran against").

It's the safe direction privacy-wise, but it silently drops signal for every org-less model. Suggestion: when the value has no path separator, doesn't exist on disk, and matches the Hub-id charset (`[A-Za-z0-9._-]+`, optionally one `/`), treat it as a clean id and pass it through. That keeps drive/absolute/on-disk paths collapsing to the marker while preserving canonical ids.

src/winml/modelkit/telemetry/utils.py
```diff
@@ -151,6 +151,40 @@ def _looks_like_path(token: str) -> bool:
...
+    if not value:
+        return None
+    normalized = value.replace("\\", "/")
+    if re.match(r"^[A-Za-z]:[\\/]", value) or normalized.startswith("/"):
+        return _model_ref_local_marker(value)
+    if Path(value).exists():
+        return _model_ref_local_marker(value)
+    parts = normalized.split("/")
+    if len(parts) == 2 and parts[0] and parts[1]:
+        return value
```
xieofxie, line 200: Minor edge: the classifier isn't aware of the eval `-m role=path` form. `_scrub_model_ref("encoder=sub/model.onnx")` splits to `['encoder=sub', 'model.onnx']` (len 2) and returns the value **verbatim**, leaking the `sub/model.onnx` fragment — which the PR summary says is never emitted. In practice composite `role=path` always requires `--model-id`, so the scrubbed-`-m` branch isn't hit today; this is latent rather than live. Still worth a guard (e.g. treat a token containing `=` as non-Hub) so the function is safe standalone.

src/winml/modelkit/telemetry/click_group.py
```diff
@@ -100,8 +101,12 @@ def wrapped_invoke(ctx: click.Context) -> Any:
             raise
         finally:
             duration_ms = int((time.perf_counter() - start) * 1000)
+            model_id = _param(ctx, "model_id")
+            if model_id is None:
+                model_id = _scrub_model_ref(_param(ctx, "model"))
```
xieofxie: Defense-in-depth: the `--model-id` value is recorded verbatim (the `if model_id is None:` guard means it never goes through `_scrub_model_ref`). Nothing validates that `--model-id` is a clean HF id — a user could pass a local path there. Running it through `_scrub_model_ref` too would be a no-op for legitimate `org/name` ids and would scrub any path that slips in:

```python
model_id = _scrub_model_ref(_param(ctx, "model_id")) or _scrub_model_ref(_param(ctx, "model"))
```

Given the whole point of this change is to guarantee no path content reaches telemetry, closing this trusted-input gap seems worthwhile.

xieofxie (commented):
Solid, well-tested change — the `<local:...>` marker approach is a clean way to keep model signal without leaking path fragments, and the `--model-id`-preferred / scrubbed-`-m` wiring plus the existence-wins tie-break are nicely covered by tests.

One concern I'd like to resolve before merge: single-segment canonical Hub ids get anonymized as `<local:dir>`, which defeats the feature's goal for a real class of models (details inline). Two smaller notes on `--model-id` scrubbing and the `role=path` edge.

src/winml/modelkit/telemetry/utils.py
timenick: Fixed in e07fb13a. Single-segment canonical Hub ids now pass through: classification is based on a Hub-id charset regex (`[A-Za-z0-9._-]+(?:/[A-Za-z0-9._-]+)?`) rather than requiring exactly one slash. So `bert-base-uncased`, `gpt2`, `resnet-50` etc. are recorded verbatim, and the documented `perf -m bert-base-uncased` example now yields `bert-base-uncased` instead of `<local:dir>`. Drive/absolute/on-disk paths still collapse to the marker. One deliberate refinement: a single-segment name that carries a file extension (e.g. `model.onnx`) is still treated as a local file ref, since Hub ids don't carry file extensions. Added regression cases in `test_utils_scrubbing.py`.

Generated with Claude Code

src/winml/modelkit/telemetry/utils.py
timenick: Fixed in e07fb13a. The Hub-id charset regex excludes `=`, so `encoder=sub/model.onnx` no longer matches and now collapses to `<local:.onnx>` — the `sub/model.onnx` fragment is never emitted verbatim, so the function is safe standalone as you suggested. Added a regression case for the `role=path` form in `test_utils_scrubbing.py`.

Generated with Claude Code

src/winml/modelkit/telemetry/click_group.py
timenick: Fixed in e07fb13a. `--model-id` now goes through `_scrub_model_ref` too:

```python
model_id = _scrub_model_ref(_param(ctx, "model_id")) or _scrub_model_ref(_param(ctx, "model"))
```

This is a no-op for legitimate `org/name` ids and anonymizes any path that slips into `--model-id`, closing the trusted-input gap — no path content can reach telemetry via that field. Added a regression test (`test_action_scrubs_path_in_model_id_param`) asserting a `C:\...\secret.onnx` in `--model-id` records `<local:.onnx>`.

Generated with Claude Code

xieofxie (approved):
Re-reviewed at `e07fb13`. All three points from my previous review are addressed cleanly, each with a targeted regression test:

1. **Single-segment Hub ids pass through** — the new `_HUB_ID_RE` (one/two Hub-charset segments, ≤1 slash) means `bert-base-uncased`, `gpt2`, `resnet-50` now record verbatim, so the documented `perf -m bert-base-uncased` no longer collapses to `<local:dir>`. The extension carve-out (`model.onnx` → local) keeps bare file paths anonymized while `org/model.v2` stays an id.
2. **`role=path` guarded** — the `=` falls outside the Hub charset, so `encoder=sub/model.onnx` → `<local:.onnx>` instead of leaking the fragment.
3. **`--model-id` scrubbed** — both inputs now flow through `_scrub_model_ref`, so a stray path in the trusted `--model-id` slot is anonymized too.

I ran the telemetry suite locally: `65 passed`. The residual `reldir/name`-vs-`org/name` ambiguity is inherent and already handled by the existence-wins tie-break. LGTM.

PR #1109, opened 2026-07-14
feat(perf): auto-build a genai bundle from a model id for --runtime winml-genai

DingmaomaoBJTU:
## What

`winml perf -m <hf-model-id> --runtime winml-genai` now auto-builds a genai bundle from a HuggingFace model id instead of requiring a prebuilt bundle directory.

## Why

The `winml` runtime already auto-builds an ONNX from a model id on demand. The `winml-genai` runtime forced users to first run `winml build ... --device npu --ep qnn` and then point `perf` at the output directory. This mirrors the winml behavior so a bare model id "just works".

## How

- New `_autobuild_genai_bundle` helper: resolves the model family, looks up its genai-bundle recipe, and builds into the managed cache dir (`get_model_dir(model)/genai-bundle`), pinning `ep=qnn` / `device=npu` (genai bundles target the NPU HTP via QNN). A prior build is reused unless `--rebuild` / `--ignore-cache` forces a fresh one.
- `_run_genai_runtime` now distinguishes inputs: a directory is used as-is (must contain `genai_config.json`), an `.onnx` path is rejected, anything else is treated as a model id and auto-built on demand.
- When auto-built, the report path is derived from the model id (the cache dir name is not a meaningful slug).
- Docs: `perf.md` and the `qwen3-genai-bundle.md` sample updated with the one-command flow.

Unblocks `winml perf -m Qwen/Qwen3-0.6B --runtime winml-genai`.

### DIFF
--- a/docs/commands/perf.md
+++ b/docs/commands/perf.md
@@ -16,7 +16,8 @@ $ winml perf [options]
 
 | Flag | Short | Type | Default | Description |
 |---|---|---|---|---|
-| `--model` | `-m` | `TEXT` | — | HuggingFace model ID or path to a local `.onnx` file. Required. |
+| `--model` | `-m` | `TEXT` | — | HuggingFace model ID or path to a local `.onnx` file. Required. With `--runtime winml-genai`, also accepts a prebuilt genai **bundle directory**, or a HuggingFace model ID that is auto-built into a bundle on demand. |
+| `--runtime` | | `winml\|winml-genai` | `winml` | Inference runtime. `winml` benchmarks single-shot ONNX inference; `winml-genai` benchmarks an onnxruntime-genai bundle (LLM generation: time-to-first-token + decode tokens/sec). With `winml-genai`, a model ID that is not a bundle directory is auto-built into one (cached under `~/.cache/winml/`, targeting the NPU HTP via QNN) before benchmarking; pass `--rebuild` to force a fresh build. |
 | `--task` | | `TEXT` | auto-detected | Explicit task override (e.g., `image-classification`). Inferred from the model if omitted. |
 | `--iterations` | | `INTEGER` | `100` | Number of timed inference iterations used to compute statistics. |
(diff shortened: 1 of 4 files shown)


xieofxie (approved):
## Review Summary

Well-structured PR with clean logic, comprehensive test coverage, and good documentation. The auto-build pattern mirrors the existing winml runtime's on-the-fly behavior nicely. A few observations below (none blocking).

### Key observations

1. **Input routing heuristic** (perf.py:1912-1925): When a model string looks like a filesystem path (e.g. user typo'd a directory) but doesn't exist, `is_dir()` returns False and it falls into the auto-build branch. The resulting error is serviceable but could confuse users who clearly intended a local path. A cheap heuristic (check for path separators) could surface a friendlier "path does not exist" before attempting the HF lookup.

2. **Cache-hit integrity** (perf.py:1836): The cache check keys solely on `genai_config.json` existing. A prior build interrupted mid-write could leave that file but an incomplete bundle. Either a lightweight sanity check (valid JSON + at least one `.onnx`) or a comment documenting why this is acceptable (GenaiSession will fail fast) would be helpful.

3. **`--ep` and `_GENAI_AUTOBUILD_FLAGS`** (perf.py:1760): The auto-build pins `ep=qnn`, but `"ep"` is not in `_GENAI_AUTOBUILD_FLAGS`. If a user passes `--ep qnn` alongside a model id, it gets warned as "ignored" even though it's consistent. Not a bug (correct from the benchmark-routing angle) but could cause minor user confusion. Worth considering.

4. **Dense table row** (perf.md:11): The `--runtime` description is quite long for a markdown table cell. Consider moving the detailed explanation below the table or to a subsection.

### Positives

- ExitStack pattern for temp-dir lifetime management is elegant
- Test matrix is thorough: autobuild, cache-hit, rebuild, ignore-cache, warning suppression, recipe rejection
- Good error messages with actionable hints (pass a prebuilt bundle directory)
- Documentation tip box in the sample page is well-placed

**Verdict: Approve** — the concerns above are all minor/non-blocking.

chinazhangchao (approved):

Issue #1110, opened 2026-07-14
WinMLCLIError telemetry loses root cause — ClickException wrappers make errors unanalyzable


timenick:
## Problem

`WinMLCLIError` records `exception_type`, a scrubbed+capped `exception_message`, and a structured `exception_stack`. In practice, commands wrap the real failure in a `click.ClickException` before it reaches telemetry, so the recorded event loses the information needed to analyze the failure.

Two failure shapes observed in production telemetry:

**1. Root cause buried in a truncated message**

```
Command:      inspect
ErrorType:    ClickException
ErrorMessage: Inspection error: The checkpoint you are trying to load has model
              type `qwen3_5` but Transformers does not recognize this
              architecture. This could be because of an issue with the
              checkpoint, or becau…      <- truncated at the 200-char cap
Count:        7
```

The real exception (a Transformers `ValueError`) is stringified into the message at `commands/inspect.py:328` (`raise click.ClickException(f"Inspection error: {e}") from e`). `ErrorType` collapses to `ClickException`, so distinct root causes can't be grouped, and the useful text is cut off by the 200-char cap.

**2. No information at all**

```
Command:      quantize
ErrorType:    ClickException
ErrorMessage: Quantization failed
Count:        3
```

`commands/quantize.py:327` raises `click.ClickException(f"{label} failed")` with no `e` and no `from e`. The root cause is dropped at the source — telemetry has nothing to record.

## Impact

- `ErrorType` is almost always `ClickException`, so errors can't be grouped or triaged by real cause.
- The most diagnostic signal is forced into the free-text `exception_message`, which is the least reliable field (PII-scrubbed and capped at 200 chars).

## Proposed direction (two parts)

**1. Telemetry layer (broad, command-agnostic)**

In `Telemetry.log_error` (`telemetry/telemetry.py:212`), walk the `__cause__` / `__context__` chain and record the root cause alongside the outer exception, e.g. new `root_cause_type` (and optionally a scrubbed `root_cause_message`) attributes. This immediately recovers shape #1 for every command without touching command code, and stays architecture-agnostic (no hardcoded command/exception names).

**2. Command layer (root fix, per-command)**

[...]

PR #1111, opened 2026-07-14
feat(telemetry): record root cause on WinMLCLIError

timenick:
## Summary

- Record the innermost root cause of an exception chain on `WinMLCLIError`, so wrapped errors (e.g. `ClickException` around a `ValueError`) become analyzable instead of collapsing to `ErrorType=ClickException`.
- New attributes `root_cause_type` (for grouping) and `root_cause_message` (scrubbed, capped at 500 — larger than the outer 200 cap since the root cause is the diagnostic payload). Both are `None` when the outer exception has no deeper cause, so "no chain" is visually distinct from a real root cause.

## Changes

- `telemetry/utils.py`: new `_root_cause` chain-walker (prefers `__cause__` over `__context__`, walks to innermost, cycle-safe via `id()`); `_format_exception_message` gains a `cap` param; new `_ROOT_CAUSE_MESSAGE_CAP = 500`.
- `telemetry/telemetry.py`: `log_error` resolves the root cause and emits the two new attributes; both added to the `WinMLCLIError` allowlist.

## Notes

- This is **Part 1 (telemetry layer)** of #1110. Part 2 (fixing command-layer sites that drop the cause, e.g. `commands/quantize.py:327`) is intentionally out of scope — the plan is to use the new telemetry data to prioritize the highest-frequency offenders.
- The 500-char cap is an initial value; it can be tuned once real telemetry shows how often root-cause messages are truncated.

Part of #1110

### DIFF
--- a/src/winml/modelkit/telemetry/utils.py
+++ b/src/winml/modelkit/telemetry/utils.py
@@ -300,14 +300,19 @@ def _root_cause(exc: BaseException) -> BaseException:
 
     Follows ``__cause__`` (explicit ``raise ... from e``) in preference to
     ``__context__`` (implicit, set when raising inside an ``except`` block),
-    repeatedly, until neither is set. Returns ``exc`` itself when there is
-    no chain. Cycle-safe: a chain that loops back on itself terminates
-    rather than spinning forever.
+    repeatedly, until neither is set. A ``__context__`` explicitly suppressed
+    by ``raise ... from None`` (``__suppress_context__``) is honored — the
+    walk stops there, matching Python's own traceback printing and respecting
+    the developer's intent to hide that inner error. Returns ``exc`` itself
+    when there is no chain. Cycle-safe: a chain that loops back on itself
+    terminates rather than spinning forever.
     """
     seen: set[int] = {id(exc)}
     current = exc
     while True:
-        nxt = current.__cause__ or current.__context__
+        nxt = current.__cause__
+        if nxt is None and not current.__suppress_context__:
+            nxt = current.__context__
         if nxt is None or id(nxt) in seen:
             return current
         seen.add(id(nxt))
--- a/src/winml/modelkit/telemetry/utils.py
+++ b/src/winml/modelkit/telemetry/utils.py
@@ -242,3 +242,22 @@ def _extract_exception_stack(tb: Any) -> list[dict[str, Any]]:
         }
         for frame in frames
     ]
+
+
+def _root_cause(exc: BaseException) -> BaseException:
+    """Return the innermost cause of an exception chain.
+
+    Follows ``__cause__`` (explicit ``raise ... from e``) in preference to
+    ``__context__`` (implicit, set when raising inside an ``except`` block),
+    repeatedly, until neither is set. Returns ``exc`` itself when there is
+    no chain. Cycle-safe: a chain that loops back on itself terminates
+    rather than spinning forever.
+    """
+    seen: set[int] = {id(exc)}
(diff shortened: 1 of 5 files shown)


src/winml/modelkit/telemetry/utils.py
```diff
@@ -292,3 +293,22 @@ def _extract_exception_stack(tb: Any) -> list[dict[str, Any]]:
...
+    Follows ``__cause__`` (explicit ``raise ... from e``) in preference to
+    ``__context__`` (implicit, set when raising inside an ``except`` block),
+    repeatedly, until neither is set. Returns ``exc`` itself when there is
+    no chain. Cycle-safe: a chain that loops back on itself terminates
+    rather than spinning forever.
+    """
+    seen: set[int] = {id(exc)}
+    current = exc
+    while True:
+        nxt = current.__cause__ or current.__context__
```
xieofxie: `_root_cause` ignores `__suppress_context__`, so it walks into a context the developer explicitly suppressed with `raise ... from None`. Verified against this PR's code:

```python
try:
    try:
        raise ValueError("SECRET inner detail")
    except ValueError:
        raise RuntimeError("clean outer") from None
except RuntimeError as outer:
    _root_cause(outer)  # -> ValueError('SECRET inner detail')
```

`from None` sets `__cause__=None` and `__suppress_context__=True`; since `__cause__ or __context__` falls through to the still-populated `__context__`, the suppressed inner exception gets recorded as `root_cause_type`/`root_cause_message`. Python's own traceback printing honors the flag and hides it, so telemetry diverges from what the developer intended to expose — and `from None` is commonly used precisely to hide a noisy or sensitive inner error. Low severity (message is still scrubbed, telemetry-internal), but easy to align:

```python
nxt = current.__cause__
if nxt is None and not current.__suppress_context__:
    nxt = current.__context__
```

Worth a small test covering the `from None` case alongside the existing `__context__` one.

xieofxie (commented):
Nice, well-scoped Part-1 change. The chain-walker is clean — `__cause__`-over-`__context__` preference, innermost walk, and `id()`-based cycle guard are all correct and thoroughly tested (I ran the full telemetry suite locally: `224 passed`). The `cap` parameterization preserves the scrub-before-cap ordering, and `has_root = root is not exc` gives a clean null distinction for "no chain". Splitting the command-layer `from`-dropping fixes into Part 2 is the right call.

One behavioral nuance on `_root_cause` worth considering before merge (inline) — otherwise LGTM.

src/winml/modelkit/telemetry/utils.py
timenick: Good catch, fixed in eae6f92a. Verified your repro — `from None` did surface the suppressed inner exception. `_root_cause` now honors `__suppress_context__`: it reads `__cause__` first, and only falls back to `__context__` when the cause is absent *and* context isn't suppressed — matching Python's own traceback printing. Added a `test_root_cause_honors_suppressed_context` regression alongside the existing implicit-`__context__` case.

Generated with Claude Code

xieofxie (approved):
Re-reviewed at `eae6f92a`. The `__suppress_context__` point is fixed exactly as suggested: `_root_cause` now only falls through to `__context__` when it isn't suppressed, so `raise ... from None` correctly surfaces no chain — matching Python's own traceback behavior and the developer's intent to hide the inner error. The docstring is updated and `test_root_cause_honors_suppressed_context` locks it in. Ran the affected telemetry tests locally: `66 passed`. LGTM.

PR #1113, opened 2026-07-15
feat(loader): add pipeline_tag fallback to task detection

xieofxie:
## Summary

When architecture-based task detection (Stage 1c) fails because `config.architectures` contains a generic name like `'Model'` that doesn't map to any real transformers class, task resolution falls through to the last-resort `feature-extraction` default. However, HuggingFace Hub knows the correct task via `pipeline_tag` (e.g. `audio-classification` for `audeering/wav2vec2-large-robust-24-ft-age-gender`).

This PR adds a **Stage 1e** between Stage 1c (TasksManager architecture detection) and Stage 1d (last-resort default) that queries the Hub's `pipeline_tag` as a fallback.

### `src/winml/modelkit/loader/resolution.py`
- Added `TaskSource.PIPELINE_TAG = "pipeline-tag"` enum value for provenance tracking
- Added `KNOWN_TASKS` to imports from `.task`
- Added Stage 1e block in `resolve_task()`:
  - Only runs when `opt_task is None` (architecture detection failed) and `model_id` exists
  - Checks `model_id` is not a local path via `_is_local_path` from `hub_utils`
  - Queries `HfApi().model_info(model_id).pipeline_tag`
  - Normalizes the tag and validates against `KNOWN_TASKS`
  - Wrapped in `try/except Exception` for graceful fallthrough on network errors or invalid model IDs

### `tests/unit/loader/test_detect_task.py`
Added 4 tests:
- Valid pipeline_tag is used when architecture detection fails
- Invalid/unknown pipeline_tag falls through to default
- API failure (network error) falls through gracefully
- Local path model_id skips the Hub API call entirely

### DIFF
--- a/src/winml/modelkit/loader/resolution.py
+++ b/src/winml/modelkit/loader/resolution.py
@@ -24,7 +24,6 @@ from typing import TYPE_CHECKING, cast
 
 from .task import (
     HF_TASK_DEFAULTS,
-    KNOWN_TASKS,
     get_default_task_for_model_id,
     get_supported_tasks,
     normalize_task,
@@ -556,7 +555,7 @@ def resolve_task(
     if opt_task is None and not getattr(config, "architectures", None) and model_type:
         # Populate Optimum's ONNX export-config registry before querying it;
         # get_supported_tasks returns [] if this hasn't been imported.
-        import optimum.exporters.onnx.model_configs  # noqa: F401
+        import optimum.exporters.onnx.model_configs
 
         supported = get_supported_tasks(model_type, resolve_optimum_library(model_type))
         if supported:
@@ -574,18 +573,29 @@ def resolve_task(
         except ValueError:
             opt_task = None
 
-    # 1e. Hub pipeline_tag fallback
-    if opt_task is None and model_id:
+    # 1d. Hub pipeline_tag fallback
+    if opt_task is None and model_id and model_type:
         from ..utils.hub_utils import get_pipeline_tag
 
         tag = get_pipeline_tag(model_id)
         if tag:
             normalized_tag = normalize_task(tag)
-            if normalized_tag in KNOWN_TASKS:
+            # Gate on the model-type's ONNX-exportable set, NOT the full KNOWN_TASKS
+            # display taxonomy. A Hub pipeline_tag is a HuggingFace *pipeline* label and
+            # may name a task with no export path (e.g. text-to-image,
+            # reinforcement-learning, time-series-forecasting). Admitting one would flow a
+            # non-exportable task into Stage 2 (model-class) / Stage 3 instead of degrading
+            # to the last-resort default. Populate Optimum's ONNX export-config registry
+            # first (as Stage 1b does) so get_supported_tasks doesn't return [].
+            import optimum.exporters.onnx.model_configs  # noqa: F401
+
+            if normalized_tag in get_supported_tasks(
+                model_type, resolve_optimum_library(model_type)
+            ):
                 opt_task = normalized_tag
                 source = TaskSource.PIPELINE_TAG
 
-    # 1d. last-resort default
+    # 1e. last-resort default
     if opt_task is None:
         opt_task = next(iter(HF_TASK_DEFAULTS))
         source = TaskSource.HF_TASK_DEFAULT
--- a/src/winml/modelkit/loader/resolution.py
+++ b/src/winml/modelkit/loader/resolution.py
@@ -442,8 +442,8 @@ def resolve_task(
     """Resolve a single model's task + class from an HF config.
 
     Stages: 0 user override -> 1 detect (override / no-architectures /
-    TasksManager / default) -> 2 model class -> 3 modality upgrade
-    (detection path only) -> 4 composite tag.
+    TasksManager / pipeline-tag / default) -> 2 model class -> 3 modality
+    upgrade (detection path only) -> 4 composite tag.
 
     ``model_type_override`` lets a caller drive resolution with a build variant
     (e.g. ``qwen3_transformer_only``) without mutating the loaded HF config; when
@@ -576,20 +576,14 @@ def resolve_task(
 
     # 1e. Hub pipeline_tag fallback
     if opt_task is None and model_id:
-        try:
-            from ..utils.hub_utils import _is_local_path
+        from ..utils.hub_utils import get_pipeline_tag
 
-            if not _is_local_path(model_id):
-                from huggingface_hub import HfApi
-
-                tag = HfApi().model_info(model_id).pipeline_tag
-                if tag:
-                    normalized_tag = normalize_task(tag)
-                    if normalized_tag in KNOWN_TASKS:
-                        opt_task = normalized_tag
-                        source = TaskSource.PIPELINE_TAG
-        except Exception:  # graceful fallthrough: network errors, invalid model_id, etc.
-            pass
+        tag = get_pipeline_tag(model_id)
+        if tag:
+            normalized_tag = normalize_task(tag)
+            if normalized_tag in KNOWN_TASKS:
+                opt_task = normalized_tag
+                source = TaskSource.PIPELINE_TAG
 
     # 1d. last-resort default
     if opt_task is None:
--- a/src/winml/modelkit/loader/resolution.py
+++ b/src/winml/modelkit/loader/resolution.py
@@ -588,7 +588,7 @@ def resolve_task(
                     if normalized_tag in KNOWN_TASKS:
                         opt_task = normalized_tag
                         source = TaskSource.PIPELINE_TAG
-        except Exception:
+        except Exception:  # graceful fallthrough: network errors, invalid model_id, etc.
             pass
 
     # 1d. last-resort default
--- a/src/winml/modelkit/loader/resolution.py
+++ b/src/winml/modelkit/loader/resolution.py
@@ -24,6 +24,7 @@ from typing import TYPE_CHECKING, cast
 
 from .task import (
     HF_TASK_DEFAULTS,
+    KNOWN_TASKS,
     get_default_task_for_model_id,
     get_supported_tasks,
     normalize_task,
@@ -253,6 +254,7 @@ class TaskSource(str, Enum):
     SENTINEL_DEFAULT = "sentinel-default"  # (model_type, None) sentinel
     TASKS_MANAGER = "tasks-manager"  # Optimum inference (incl. fill-mask upgrade)
     WRAPPED_LIBRARY = "wrapped-library"  # no architectures -> first supported task
+    PIPELINE_TAG = "pipeline-tag"  # Hub pipeline_tag fallback
     HF_TASK_DEFAULT = "hf-task-default"  # last-resort default
 
 
@@ -571,6 +573,23 @@ def resolve_task(
             source = TaskSource.TASKS_MANAGER
         except ValueError:
             opt_task = None
+
+    # 1e. Hub pipeline_tag fallback
+    if opt_task is None and model_id:
+        try:
+            from ..utils.hub_utils import _is_local_path
+
+            if not _is_local_path(model_id):
+                from huggingface_hub import HfApi
+
+                tag = HfApi().model_info(model_id).pipeline_tag
+                if tag:
+                    normalized_tag = normalize_task(tag)
+                    if normalized_tag in KNOWN_TASKS:
+                        opt_task = normalized_tag
+                        source = TaskSource.PIPELINE_TAG
+        except Exception:
+            pass
 
     # 1d. last-resort default
     if opt_task is None:
--- a/src/winml/modelkit/utils/hub_utils.py
+++ b/src/winml/modelkit/utils/hub_utils.py
@@ -144,6 +144,28 @@ def is_hub_model(model_name_or_path: str) -> tuple[bool, dict]:
         return False, {"type": "local", "path": model_name_or_path}
 
 
+_PIPELINE_TAG_TIMEOUT = 10  # seconds
+
+
+def get_pipeline_tag(model_id: str) -> str | None:
+    """Return the Hub ``pipeline_tag`` for *model_id*, or ``None``.
+
+    Lightweight helper that skips the full metadata extraction of
+    ``is_hub_model``. Returns ``None`` (never raises) when *model_id* is a
+    local path, the Hub is unreachable, or the model has no tag.
+    """
+    if _is_local_path(model_id):
+        return None
+    try:
+        from huggingface_hub import HfApi
+
+        info = HfApi().model_info(model_id, timeout=_PIPELINE_TAG_TIMEOUT)
+        return getattr(info, "pipeline_tag", None)
+    except Exception:
+        logger.debug("pipeline_tag lookup failed for '%s'", model_id, exc_info=True)
+        return None
+
+
 def inject_hub_metadata(onnx_model: Any, model_name_or_path: str, metadata: dict) -> None:
     """Inject HuggingFace Hub metadata into ONNX model.
 
--- a/src/winml/modelkit/utils/hub_utils.py
+++ b/src/winml/modelkit/utils/hub_utils.py
@@ -144,7 +144,7 @@ def is_hub_model(model_name_or_path: str) -> tuple[bool, dict]:
         return False, {"type": "local", "path": model_name_or_path}
 
 
-_PIPELINE_TAG_TIMEOUT = 10  # seconds
+_PIPELINE_TAG_TIMEOUT = 5  # seconds
 
 
 def get_pipeline_tag(model_id: str) -> str | None:
--- a/tests/unit/loader/test_detect_task.py
+++ b/tests/unit/loader/test_detect_task.py
@@ -231,31 +231,37 @@ def test_resolve_task_case1_surfaces_modality_aware_task() -> None:
 
 
 # =============================================================================
-# Stage 1e — Hub pipeline_tag fallback
+# Stage 1d — Hub pipeline_tag fallback
 # =============================================================================
 
 _GET_PIPELINE_TAG = "winml.modelkit.utils.hub_utils.get_pipeline_tag"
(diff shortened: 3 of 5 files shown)


src/winml/modelkit/loader/resolution.py
```diff
@@ -572,6 +574,23 @@
...
+            if not _is_local_path(model_id):
+                from huggingface_hub import HfApi
+
+                tag = HfApi().model_info(model_id).pipeline_tag
+                if tag:
+                    normalized_tag = normalize_task(tag)
+                    if normalized_tag in KNOWN_TASKS:
+                        opt_task = normalized_tag
+                        source = TaskSource.PIPELINE_TAG
+        except Exception:
```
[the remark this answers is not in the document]
xieofxie: Fixed — added an explanatory comment documenting that the bare \xcept\ is intentional: network errors, invalid model IDs, and missing pipeline_tag all gracefully fall through to Stage 1d (last-resort default).

xieofxie:
a follow up of https://github.com/microsoft/winml-cli/pull/1094

tested via winml export -m audeering/wav2vec2-large-robust-24-ft-age-gender -o temp/wav2vec2.onnx

src/winml/modelkit/loader/resolution.py
```diff
@@ -572,6 +574,23 @@ def resolve_task(
         except ValueError:
             opt_task = None
 
+    # 1e. Hub pipeline_tag fallback
+    if opt_task is None and model_id:
```
DingmaomaoBJTU: `resolve_task` is the single entry point for inspect / config / build / eval / inference (per this module's docstring), and until now the Stage-1 detection path was purely local. This adds a blocking Hub call on the last-resort path, which is a meaningful behavior change:

- The function docstring's `Stages:` line still reads `1 detect (override / no-architectures / TasksManager / default)` and doesn't mention this new 1e stage — worth updating so the enumeration stays accurate.
- There's no caching, so every `resolve_task` on such a model re-hits the Hub. For something like `winml inspect` run repeatedly this could be noticeable. Worth confirming that's acceptable, or memoizing per model_id.

src/winml/modelkit/loader/resolution.py
```diff
@@ -572,6 +574,23 @@ def resolve_task(
...
+        try:
+            from ..utils.hub_utils import _is_local_path
+
+            if not _is_local_path(model_id):
+                from huggingface_hub import HfApi
+
+                tag = HfApi().model_info(model_id).pipeline_tag
```
DingmaomaoBJTU: `HfApi().model_info()` is called without a timeout. `requests` has no default timeout, so on a stalled/black-holed connection this can block `resolve_task` indefinitely. Consider passing a short `timeout=`.

Also — `hub_utils.is_hub_model()` already calls `model_info()` and returns `pipeline_tag` in its metadata dict. Could this reuse that (or a small shared helper like `get_pipeline_tag(model_id)`) instead of re-implementing the local-path check + `model_info` call and reaching into the private `_is_local_path`? Centralizing Hub access keeps token/error handling in one place.

src/winml/modelkit/loader/resolution.py
```diff
@@ -572,6 +574,23 @@ def resolve_task(
...
+                if tag:
+                    normalized_tag = normalize_task(tag)
+                    if normalized_tag in KNOWN_TASKS:
+                        opt_task = normalized_tag
+                        source = TaskSource.PIPELINE_TAG
+        except Exception:  # graceful fallthrough: network errors, invalid model_id, etc.
```
DingmaomaoBJTU: Swallowing every `Exception` with a bare `pass` drops all diagnostics — when the fallback silently fails, there's no way to tell why a model still resolved to the `feature-extraction` default. Elsewhere in this codebase the same situation leaves a breadcrumb (e.g. `hub_utils.is_hub_model`'s ModelCard block does `logger.debug(..., exc_info=True)`). A `logger.debug` here would match that convention and help troubleshooting without changing control flow. This module already has a `logger`.

tests/unit/loader/test_detect_task.py
```diff
@@ -228,3 +228,81 @@ def test_resolve_task_case1_surfaces_modality_aware_task() -> None:
...
+    assert r.source == TaskSource.HF_TASK_DEFAULT
+
+
+def test_resolve_task_pipeline_tag_skips_local_path() -> None:
+    """When model_id is a local path, the Hub API is not called."""
+    cfg = _FakeConfig("faketype", name_or_path="./local-model")
+    with (
+        patch(_INFER, side_effect=ValueError("unknown arch")),
+        patch(_IS_LOCAL, return_value=True),
+        patch(_HF_API, side_effect=AssertionError("must not call HfApi for local paths")),
```
DingmaomaoBJTU: This negative guard doesn't actually protect against a regression. `AssertionError` is a subclass of `Exception`, so if the `if not _is_local_path(...)` guard were ever removed, `HfApi()` would raise this AssertionError — but the production `except Exception: pass` would swallow it, `opt_task` would stay `None`, and the test would still fall through to `HF_TASK_DEFAULT` and pass.

Consider patching `HfApi` with a `MagicMock` and asserting `mock_hfapi.assert_not_called()` (or `mock_hfapi.return_value.model_info.assert_not_called()`) so the test fails loudly if the Hub is ever hit for a local path.

DingmaomaoBJTU (commented):
Thanks for the pipeline_tag fallback — the overall shape is right: it only runs on the last-resort path (opt_task is None), guards against local paths, and falls through gracefully. A few things worth addressing before merge, left inline:

1. resolve_task now does blocking network I/O on a path that used to be purely local — no timeout, no caching, and the docstring's stage list wasn't updated.
2. The bare `except Exception: pass` drops all diagnostics.
3. The local-path negative test can't actually fail, because the AssertionError it raises is swallowed by that same `except Exception`.

src/winml/modelkit/loader/resolution.py
xieofxie: if we don't know it, we should try our best to get it, so a network call is acceptable

xieofxie: Fixed:
- Updated the docstring's Stages line to include \pipeline-tag\ in the enumeration.
- On caching: since this only fires when architecture detection *fails* (generic \config.architectures\), and the caller (xieofxie) confirmed a network call is acceptable here, I've kept it uncached for now. Can add \@lru_cache\ on \get_pipeline_tag\ later if repeated calls become a concern.

src/winml/modelkit/loader/resolution.py
xieofxie: Fixed both:
- Extracted a \get_pipeline_tag(model_id)\ helper in \hub_utils.py\ that centralizes the local-path check, \HfApi().model_info()\ call with a **10-second timeout**, and error logging. Stage 1e now just calls this helper.
- No longer reaches into private \_is_local_path\ from resolution.py.

src/winml/modelkit/loader/resolution.py
xieofxie: Fixed — the new get_pipeline_tag() helper now does logger.debug with exc_info=True on any exception, matching the existing hub_utils convention. The bare except: pass in resolution.py is gone entirely.

tests/unit/loader/test_detect_task.py
xieofxie: Fixed — the test now mocks get_pipeline_tag at source (winml.modelkit.utils.hub_utils.get_pipeline_tag) with return_value=None and asserts mock_tag.assert_called_once_with('./local-model'). The local-path guard is encapsulated inside get_pipeline_tag, so the test verifies the contract without relying on AssertionError being swallowed.

src/winml/modelkit/loader/resolution.py
```diff
@@ -572,6 +574,17 @@ def resolve_task(
...
             opt_task = None
 
+    # 1e. Hub pipeline_tag fallback
+    if opt_task is None and model_id:
+        from ..utils.hub_utils import get_pipeline_tag
+
+        tag = get_pipeline_tag(model_id)
+        if tag:
+            normalized_tag = normalize_task(tag)
+            if normalized_tag in KNOWN_TASKS:
```
DingmaomaoBJTU: `KNOWN_TASKS` is the display/validation taxonomy and includes tasks that have no ONNX export path (`text-to-image`, `reinforcement-learning`, `time-series-forecasting`, `any-to-any`, `inpainting`, ...). I checked, and `sentence-similarity` is safe because optimum's `map_from_synonym` collapses it to `feature-extraction` inside `normalize_task` — but those others pass through unchanged. So for a model whose `pipeline_tag` is one of them and whose Stage 1c failed, we'd now set `opt_task` to a non-exportable task instead of the old `feature-extraction` default. For `inspect` that's arguably more accurate, but on the build/eval path it turns a usable default into a hard failure at export / model-class resolution. Consider gating on the exportable set (e.g. intersect with `get_supported_tasks` / what TasksManager can resolve) rather than the full `KNOWN_TASKS`.

src/winml/modelkit/loader/resolution.py
```diff
@@ -572,6 +574,17 @@ def resolve_task(
         except ValueError:
             opt_task = None
 
+    # 1e. Hub pipeline_tag fallback
+    if opt_task is None and model_id:
```
DingmaomaoBJTU: This puts a Hub round-trip (`model_info`) on `resolve_task`, which is shared by build / eval / inference too, not just `inspect` — those callers reach here whenever `--task` isn't passed and arch detection fails. Task resolution has been deliberately offline/local up to now, so worth confirming we want the network reach on all of those paths vs. scoping it to `inspect`. Offline mode degrades fine (the `except` returns `None`) and it's guarded to the fallback, so this is more "is this intended" than a bug. Minor: `model_info` retries transient errors, so worst-case latency can exceed the 10s timeout.

src/winml/modelkit/loader/resolution.py
```diff
@@ -572,6 +574,17 @@ def resolve_task(
         except ValueError:
             opt_task = None
 
+    # 1e. Hub pipeline_tag fallback
```
DingmaomaoBJTU: Nit: this block is labeled `1e` but executes *before* `1d` (last-resort default), so the code reads 1c -> 1e -> 1d. The docstring already lists them in execution order ("TasksManager / pipeline-tag / default"); renaming pipeline-tag to `1d` and the default to `1e` would make the inline labels match.

tests/unit/loader/test_detect_task.py
```diff
@@ -228,3 +228,58 @@ def test_resolve_task_case1_surfaces_modality_aware_task() -> None:
...
+    cfg = _FakeConfig("faketype", name_or_path="someone/some-model")
+    with (
+        patch(_INFER, side_effect=ValueError("unknown arch")),
+        patch(_GET_PIPELINE_TAG, return_value=None),
+    ):
+        r = resolve_task(cfg)
+    assert r.source == TaskSource.HF_TASK_DEFAULT
+
+
+def test_resolve_task_pipeline_tag_skips_local_path() -> None:
```
DingmaomaoBJTU, line 281: This test mocks `get_pipeline_tag`, so it never actually exercises the local-path skip — that logic lives *inside* `get_pipeline_tag`, which is patched out here. The assertion only confirms the helper is called with `"./local-model"` and returns the mocked `None`, so the name/docstring overclaim what's covered: the real `_is_local_path` rejection and the `except Exception` fallthrough have no direct test. Suggest a focused unit test on `get_pipeline_tag` itself (local path -> `None` without hitting the network; `model_info` raising -> `None`).

DingmaomaoBJTU (commented):
Reviewed the `pipeline_tag` fallback — the core idea is sound and the `resolve_task` dispatch is well covered. A few things inline worth a look: the `KNOWN_TASKS` gate admits display-only tasks that aren't ONNX-exportable; this adds a Hub network call to the shared `resolve_task` path (build/eval/inference, not just inspect); the 1e/1d labels read out of execution order; and the local-path test mocks the function under test so it doesn't cover the skip.

One more, not inline: the PR description looks slightly stale — it says the `HfApi().model_info(...)` call is inlined in `resolution.py` and lists 2 changed files, but the implementation factors it into a `get_pipeline_tag` helper in `utils/hub_utils.py` (3 files). Worth updating so the description matches.

src/winml/modelkit/utils/hub_utils.py
```diff
@@ -144,6 +144,28 @@ def is_hub_model(model_name_or_path: str) -> tuple[bool, dict]:
         return False, {"type": "local", "path": model_name_or_path}
 
 
+_PIPELINE_TAG_TIMEOUT = 10  # seconds
```
xieofxie: make it smaller ?

timenick: **Stage 1e latency on unreachable network**

I measured this on the PR head (`591e02ee`) with `audeering/wav2vec2-large-robust-24-ft-age-gender` (architectures=`['Model']`, which forces Stage 1e):

| Scenario | Cost |
|---|---|
| Online, cold, **with** Stage 1e | resolve_task 5.64s |
| Online, cold, **without** Stage 1e | resolve_task 5.50s |
| **Net Stage 1e cost (online)** | **~0.14s** (negligible) |
| `HF_HUB_OFFLINE=1` + no network | 0.58s (huggingface_hub short-circuits — no issue) |
| **Network unreachable, not offline** | **blocks 10.3s** (= `_PIPELINE_TAG_TIMEOUT`) |

So the fallback is essentially free when online, and explicit-offline users are already handled by the Hub client. The one real cost is users who are **not** explicitly offline but whose network is down/slow (corporate proxy, VPN): every `inspect`/`build`/`inference`/`eval` that reaches Stage 1e blocks for the full 10s. For a best-effort fallback, 10s is high — suggest lowering to 3–5s.

🤖 Generated with Claude Code

src/winml/modelkit/loader/resolution.py
```diff
@@ -572,6 +574,17 @@ def resolve_task(
...
+    if opt_task is None and model_id:
+        from ..utils.hub_utils import get_pipeline_tag
+
+        tag = get_pipeline_tag(model_id)
+        if tag:
+            normalized_tag = normalize_task(tag)
```
timenick, line 582: **`KNOWN_TASKS` membership doesn't guarantee an exportable task**

A Hub `pipeline_tag` is a HuggingFace *pipeline* label, while every other Stage 1 branch produces an Optimum-resolvable task that Stage 2 (`TasksManager.get_model_class_for_task`) and Stage 3 (modality upgrade) rely on. `KNOWN_TASKS` includes names that aren't Optimum-exportable (e.g. `sentence-similarity`, abbrev `None`). If a `pipeline_tag` normalizes to one of those, it passes this guard and flows into Stage 2, where `get_model_class_for_task` will fail and fall back to `_resolve_model_class_from_config` — potentially surfacing a task with no export path. Consider intersecting the accepted set with the exportable tasks (e.g. `get_supported_tasks` / `HF_TASK_DEFAULTS`) rather than `KNOWN_TASKS` alone.

🤖 Generated with Claude Code

src/winml/modelkit/utils/hub_utils.py
xieofxie: let's half it to 5

src/winml/modelkit/loader/resolution.py
xieofxie: sharing by build / eval / inference is by design

src/winml/modelkit/loader/resolution.py
xieofxie: Fixed in bb2b3944. Stage 1d now gates on the model-type's ONNX-exportable set via get_supported_tasks(model_type, resolve_optimum_library(model_type)) instead of the full KNOWN_TASKS display taxonomy. So a pipeline_tag that normalizes to a non-exportable task (text-to-image, reinforcement-learning, time-series-forecasting, ...) is rejected and resolution degrades to the last-resort default rather than surfacing a task Stage 2/3 can't build. (Duplicate of the same concern in the later CodeQL/Claude review — addressed by one change.)

src/winml/modelkit/loader/resolution.py
xieofxie: Fixed in bb2b3944 (same change as the earlier thread on this file). The accepted set is now the model-type's ONNX-exportable tasks (get_supported_tasks) rather than KNOWN_TASKS, so a pipeline_tag normalizing to a non-Optimum-resolvable task like sentence-similarity no longer flows into Stage 2 — it's rejected at the gate and falls through to the default.

src/winml/modelkit/utils/hub_utils.py
xieofxie: Lowered _PIPELINE_TAG_TIMEOUT from 10s to 5s in bb2b3944 (matches your 3-5s suggestion and xieofxie's 'half it to 5'). Thanks for the detailed measurements — they confirm the online cost is negligible and only the not-explicitly-offline-but-unreachable case pays the timeout, so 5s is the right ceiling there.

xieofxie: Done — halved to 5s in bb2b3944.

src/winml/modelkit/loader/resolution.py
xieofxie: Fixed in bb2b3944 — renamed the inline labels so they match execution order: pipeline_tag is now 1d and last-resort default is 1e (code reads 1c -> 1d -> 1e), consistent with the docstring's 'TasksManager / pipeline-tag / default'.

tests/unit/loader/test_detect_task.py
xieofxie: Fixed in bb2b3944 — added tests/unit/utils/test_hub_utils.py with focused tests on get_pipeline_tag itself: local path returns None without constructing HfApi (asserts assert_not_called), a reachable model returns its tag, a model with no tag returns None, and a model_info exception is swallowed to None. The resolver test's docstring was also corrected to stop overclaiming.

timenick (approved):

PR #1126, opened 2026-07-16
feat(optim): add algebraic graph rewrites

xieofxie:
## Summary

- add an opt-in algebraic rewrite pipe after ORT constant folding
- replace statically bounded `Split` nodes with `Slice` nodes
- fold safe channel-wise affine branches into producing `Conv` weights and bias
- fold inference `Conv + static Add + BatchNormalization` patterns
- guard graph outputs, nested subgraph captures, shared consumers, dynamic shapes, and repeated optimizer runs
- add generated ONNX numerical-equivalence and graph-safety coverage

## QNN result

On the evaluated fixed-batch U-Net, the combined rewrites plus `htp_graph_finalization_optimization_mode=3` reduced p50 latency from 31.125 ms to 22.729 ms and reduced QNN DDR traffic by roughly 30%.

### DIFF
--- a/src/winml/modelkit/optim/pipes/algebraic.py
+++ b/src/winml/modelkit/optim/pipes/algebraic.py
@@ -1,0 +1,1202 @@
+# -------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+# --------------------------------------------------------------------------
+"""Conservative, opt-in algebraic ONNX graph rewrites.
+
+The transformations in this pipe intentionally match topology, static tensor
+metadata, attributes, and constant values.  They do not rely on exporter names
+or on a particular model architecture.
+"""
+
+from __future__ import annotations
+
+import logging
+from dataclasses import dataclass
+from typing import Any, ClassVar
+
+import numpy as np
+import onnx
+
+from ..capabilities import algebraic
+from .base import BasePipe, PipeConfig, caps_dict
+
+
+logger = logging.getLogger(__name__)
+
+
+ALGEBRAIC_CAPABILITIES: dict[str, Any] = caps_dict(
+    algebraic.STATIC_SPLIT_TO_SLICE,
+    algebraic.CONV_CHANNEL_AFFINE_FOLDING,
+    algebraic.CONV_ADD_BATCH_NORMALIZATION_FOLDING,
+)
+
+
+@dataclass
+class AlgebraicRewritePipeConfig(PipeConfig):
+    """Independent switches for the exact algebraic rewrites."""
+
+    static_split_to_slice: bool = False
+    conv_channel_affine_folding: bool = False
+    conv_add_batch_normalization_folding: bool = False
+
+
+@dataclass
+class _GraphIndex:
+    """Small graph index rebuilt after each mutation."""
+
+    producers: dict[str, onnx.NodeProto]
+    consumers: dict[str, list[onnx.NodeProto]]
+    initializers: dict[str, onnx.TensorProto]
+    shapes: dict[str, tuple[int | None, ...]]
+    graph_outputs: set[str]
+
+    @classmethod
+    def build(cls, model: onnx.ModelProto) -> _GraphIndex:
+        from onnx import numpy_helper
+
+        graph = model.graph
+        producers: dict[str, onnx.NodeProto] = {}
+        consumers: dict[str, list[onnx.NodeProto]] = {}
+        for node in graph.node:
+            for output in node.output:
+                if output:
+                    producers[output] = node
+            consumed_names = {input_name for input_name in node.input if input_name}
+            for attribute in node.attribute:
+                if attribute.type == onnx.AttributeProto.GRAPH:
+                    consumed_names.update(_captured_tensor_names(attribute.g))
+                elif attribute.type == onnx.AttributeProto.GRAPHS:
+                    for nested_graph in attribute.graphs:
+                        consumed_names.update(_captured_tensor_names(nested_graph))
+            for input_name in consumed_names:
+                consumers.setdefault(input_name, []).append(node)
+
+        initializers = {initializer.name: initializer for initializer in graph.initializer}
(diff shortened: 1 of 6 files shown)


src/winml/modelkit/optim/pipes/algebraic.py
```diff
@@ -0,0 +1,1202 @@
...
+            original_conv.op_type != "Conv"
+            or len(original_conv.output) != 1
+            or not original_conv.output[0]
+        ):
+            continue
+        conv_output = original_conv.output[0]
+        conv = index.producers.get(conv_output)
+        if conv is None or conv.op_type != "Conv":
+            continue
+        conv_shape = _static_shape(index, conv_output)
```
xieofxie, line 763: `_static_shape` requires *every* dimension (including batch) to be static, so the whole affine fold bails on any model with a dynamic batch dim — which is extremely common. Channel-wise folding only needs the channel axis (`conv_shape[1]`) to be known; the direct `Conv → Mul/Add(channel-const)` path is mathematically valid for any N/H/W. Was requiring full-static shapes intentional, or could the direct path relax to only needing a static channel dim? As written the optimization silently no-ops for most non-fixed-batch models.

Verified empirically: `Conv → Mul([1,C,1,1])` with input `[None,1,3,3]` is left unfolded, while the identical fully-static graph folds.

src/winml/modelkit/optim/pipes/algebraic.py
```diff
@@ -0,0 +1,1202 @@
...
+            continue
+        epsilon = _attribute(batch_norm, "epsilon", 1e-5)
+        try:
+            epsilon = float(epsilon)
+        except (TypeError, ValueError):
+            continue
+        denominator = variance + epsilon
+        if epsilon < 0 or np.any(denominator <= 0) or not np.all(np.isfinite(denominator)):
+            continue
+        gamma = scale / np.sqrt(denominator)
```
xieofxie, line 1046: The BN math runs entirely in the weight dtype. Because of the guards above, `scale`/`variance`/`mean`/`beta` are all `weight_dtype`, so for an fp16 model `denominator`, `np.sqrt`, `gamma`, and the `new_bias = bias*gamma + beta - gamma*mean` in `_copy_conv_parameters_for_bn` all execute in fp16. That contrasts with the affine fold, which deliberately upcasts to `result_type(weight_dtype, float32)` (line 857) before folding. ONNX BatchNormalization is typically evaluated with an fp32 accumulator, so folding at fp16 can drift beyond tolerance on the low-precision models this PR targets for QNN. Consider computing `gamma`/`new_bias` in `result_type(weight_dtype, float32)` and casting back to match the affine path.

src/winml/modelkit/optim/pipes/algebraic.py
```diff
@@ -0,0 +1,1202 @@
...
+    index = _GraphIndex.build(model)
+    opset = next(
+        (int(opset.version) for opset in model.opset_import if opset.domain in ("", "ai.onnx")),
+        0,
+    )
+    if opset and opset < 10:
+        return
+    replacements: dict[int, list[onnx.NodeProto]] = {}
+    for split in list(model.graph.node):
+        if split.op_type != "Split" or len(split.input) < 1 or not split.input[0]:
```
xieofxie, line 1088: There's no domain guard here (same for the `Conv`/`Add`/`BatchNormalization` checks in the fold functions). An op is `(domain, op_type)` — a custom-domain node literally named `Split` (or `Conv`/`BatchNormalization`) would be rewritten with standard-op semantics. ONNX/QNN models can carry non-`""` domain nodes. Cheap to add `node.domain in ("", "ai.onnx")` to each pattern match for safety.

src/winml/modelkit/optim/pipes/algebraic.py
```diff
@@ -0,0 +1,1202 @@
...
+        """Apply enabled rewrites to a copied model and return the copy."""
+        if not self.should_process(config):
+            return model
+
+        import onnx
+
+        result = onnx.ModelProto()
+        result.CopyFrom(model)
+        allocator = _NameAllocator(result)
+        introduced: set[str] = set()
```
xieofxie, line 1188: `introduced` is populated (every `_new_initializer` does `introduced.add(name)`) but never read — only `introduced_nodes` is consumed, by `_prune_introduced_nodes`, and initializer cleanup goes through `_prune_unused_initializers` (reference counting) instead. So this set is dead state threaded through all three folds. Either drop it or wire it into the pruning if it was meant to bound initializer removal.

src/winml/modelkit/optim/pipes/algebraic.py
```diff
@@ -0,0 +1,1202 @@
...
+    index: _GraphIndex,
+    node: onnx.NodeProto,
+    input_name: str,
+) -> tuple[int, list[tuple[int, int]]] | None:
+    """Return a static Split axis and output channel-independent boundaries."""
+    input_shape = index.shapes.get(input_name)
+    if input_shape is None or len(node.output) == 0:
+        return None
+
+    axis_input_index = 2 if len(node.input) > 2 else None
```
xieofxie, line 350: Minor: `Split` has no `axis` input in any opset — `axis` is always an attribute, and the only optional input is `split` (index 1). Treating input index 2 as a possible axis source is dead/defensive code that misreads the schema and could mislead a future reader. Reading `axis` purely from the attribute would be clearer.

xieofxie (commented):
Deep-dived the three rewrites (math + topology guards), ran the new suite (82 passed) and ruff (clean), and probed edge cases empirically. The core algebra is correct and the guards (graph outputs, subgraph captures, shared consumers, overlapping/routed channel intervals, static-vs-dynamic split sizes, repeated runs) are impressively thorough. The channel-affine and Conv+Add+BN folds are numerically sound, and idempotency holds.

Main thing worth discussing is the full-static-shape requirement, which quietly disables the affine fold on any dynamic-batch model. A few smaller robustness/cleanliness notes inline. None are blocking correctness bugs — they're about applicability, precision on low-precision models, and code hygiene.

PR #1130, opened 2026-07-17
ModelKit: unify information rules into pattern rules (single source of truth)

fangyangci:
## Summary
This PR removes duplicated recommendation configuration between:
- `src/winml/modelkit/analyze/rules/information_rules/*_information.json`
- `src/winml/modelkit/pattern/rules/*.json`

and consolidates recommendation metadata into `SkeletonPatternRules` as a single source of truth.

## Motivation
Previously, recommendation switches/naming/details were maintained in multiple places, which could diverge and become hard to reason about.
A concrete example was QNN attention mapping mismatch (`pattern_to_id`) between pattern rules and information rules.

## What changed
- Removed legacy information rule files:
  - `default_information.json`
  - `qc_information.json`
  - `ov_information.json`
- Extended pattern config/schema to carry recommendation metadata directly in skeleton alternatives:
  - `explanation` on skeleton pattern
  - `enabled` / `details` / `action_items` on alternatives
- Updated runtime and information pipeline to consume recommendation metadata from pattern rules:
  - `runtime_checker` propagates explanation/action metadata
  - `information_engine` derives actions from skeleton alternatives
- Added metadata reuse for repeated `from -> to` pairs in pattern config loading.
- Unified QNN attention target to:
  - `SUBGRAPH/TransposeAttentionPattern -> SUBGRAPH/ExpandedAttentionPattern`
- Updated unit tests to validate skeleton-driven behavior and removed dependency on legacy information rules.

## Outcome
- One authoritative location for recommendation config
- Lower risk of drift between duplicated files
- More understandable mapping between detected patterns and suggested actions

## Validation
- Updated unit tests in:
  - `tests/unit/analyze/core/test_information_engine.py`
  - `tests/unit/analyze/models/test_information.py`


src/winml/modelkit/analyze/core/information_engine.py
```diff
@@ -116,35 +116,9 @@ def __init__(
...
-        self._predefined_info = self._rule_loader.load_information_rules(ihv_type=ihv_type)
-        load_predefined_ms = int((time.perf_counter() - load_predefined_start) * 1000)
-
-        # Build pattern_id -> Information lookup for quick matching
-        self._info_by_pattern: dict[str, Information] = {
-            info.pattern_id: info for info in self._predefined_info if info.pattern_id
-        }
+        # Keep API compatibility for callers that still pass custom rules_dir.
+        # Pattern-level recommendations are now sourced from skeleton alternatives.
+        _ = rules_dir
```
DingmaomaoBJTU: [功能不等价 - 静默回归] rules_dir 被直接丢弃，没有任何警告日志。老代码会真正用 RuleLoader(rules_dir=rules_dir) 加载对应目录的 information rules；现在如果调用方传入非 None 的 rules_dir，会静默失效，实际走硬编码 pattern rules 路径。建议加: if rules_dir is not None: logger.warning('rules_dir %s is ignored; recommendations now sourced from skeleton alternatives', rules_dir)

src/winml/modelkit/analyze/core/information_engine.py
```diff
@@ -1009,7 +930,12 @@ def _check_patterns(self) -> list[Information]:
                 continue
 
             # Generate explanation with count
-            if count == 1:
+            if representative.explanation:
+                if count == 1:
+                    explanation = representative.explanation
+                else:
+                    explanation = f"{count} instances of pattern '{pattern_id}' detected. {representative.explanation}"
```
DingmaomaoBJTU: [行为差异] count > 1 且有自定义 explanation 时，新格式 '{count} instances of pattern detected. {explanation}' 丢失了分类状态（are not supported / have partial support）。建议拼接 classification.value，如: f'{count} instances of pattern are {classification.value}: {representative.explanation}'

src/winml/modelkit/analyze/core/information_engine.py
```diff
@@ -1286,15 +1096,29 @@ def _extract_actions(self, pattern_runtime: PatternRuntime) -> list[Action]:
...
             level, status = self._determine_action_level_and_status(
                 classification, alt_classification
             )
 
-            # Skip if no improvement (None level)
+            # Fallback: keep optimization suggestions that carry action_items
+            # even when support-level transition has no improvement.
             if level is None:
-                continue
+                if not alternative.action_items:
```
DingmaomaoBJTU, line 1121: [新增行为，建议明确] 老代码对 level is None (SUPPORTED->SUPPORTED 无改善) 是直接 continue，不产生任何 action。新代码改为：只要 alternative 携带 action_items 就强制生成 ActionLevel.OPTIONAL action。这是新行为，PR description 没有提到。当前 JSON 只有 PARTIAL/UNSUPPORTED 源，暂不触发——但若未来在 SUPPORTED 源图案上加了 action_items，会静默生出推荐。建议加 logger.debug 说明 fallback 触发了。

src/winml/modelkit/pattern/config.py
```diff
@@ -316,18 +331,48 @@ def _load_config(self) -> None:
...
-                    reason=alt_data.get("reason"),
-                    pattern_class=alt_data.get("pattern_class"),
-                    module=alt_data.get("module"),
+            # Parse alternatives (metadata only, not loaded as Pattern instances).
+            # For repeated from->to pairs across different pattern classes, allow
+            # metadata reuse by inheriting the first explicit action payload.
+            alternatives: list[PatternAlternative] = []
+            pattern_id = pattern_data["pattern_id"]
+            for alt_data in pattern_data.get("alternatives", []):
+                pattern_to_id = alt_data["pattern_to_id"]
```
DingmaomaoBJTU: [脆弱的顺序依赖] pair_metadata_defaults 是先出现者胜：第一个显式定义 details/action_items 的 skeleton 条目会把元数据渗透给同 (pattern_id, pattern_to_id) 对的后续条目。Gelu2/3/4Pattern、LayerNormalizationMulPattern 全依赖各自 Variant 1 / Pow 变体排在 JSON 最前面。若有人在前面插入新变体但没带元数据，所有变体的 action_items 会静默丢失，没有报错。建议：(a) 所有变体显式写元数据（最安全）；或 (b) 在 schema 和 _load_config 里显式文档化 JSON 顺序有语义，并加 warning 检测遗漏。

DingmaomaoBJTU (commented):
对此 PR 做了功能等价性分析，整体方向正确（单一数据源替换双维护，修复了 QNN attention target mismatch）。发现以下几处需要关注的问题，请参见 inline comments。

DingmaomaoBJTU (commented):
Reviewed for functional equivalence. Overall approach is sound (single source of truth, fixes QNN attention target). Found 4 areas requiring attention—see inline comments below.

src/winml/modelkit/analyze/core/information_engine.py
DingmaomaoBJTU: [Functional regression - silent failure] The rules_dir parameter is discarded without warning. Old code used RuleLoader(rules_dir) to load from provided directory; now callers passing custom rules_dir (integration tests, plugins) silently fail. Recommend: if rules_dir is not None: logger.warning(...)

src/winml/modelkit/analyze/core/information_engine.py
DingmaomaoBJTU: [Behavior change - lost context] When count>1 with custom explanation, new format drops classification context (are not supported/have partial support). Reduces user actionability. Consider: f'{count} instances of pattern are {classification.value}: {explanation}'

src/winml/modelkit/analyze/core/information_engine.py
DingmaomaoBJTU, line 1121: [New behavior - undocumented fallback] Old code would skip (continue) when level is None. New code generates ActionLevel.OPTIONAL if action_items exist, even without support improvement. This is semantic expansion not mentioned in PR description. Currently harmless (only PARTIAL/UNSUPPORTED sources), but future SUPPORTED patterns with action_items will silently generate actions. Recommend: add logger.debug to make this explicit.

src/winml/modelkit/pattern/config.py
DingmaomaoBJTU: [Fragile ordering dependency] pair_metadata_defaults implements first-wins inheritance: first rule defining details/action_items for (pattern_id, pattern_to_id) propagates to subsequent rules. Gelu2/3/4 and LayerNormMul rely on Variant 1/Pow appearing first in JSON. If new variant inserted before current first variant without metadata, all variants silently lose action_items with no error. Recommend: (a) explicitly define metadata on all variants (safest); or (b) document JSON order semantics in schema + add warning when metadata appears missing.

src/winml/modelkit/analyze/core/information_engine.py
fangyangci: Addressed in 4a55624c. I removed the obsolete `rules_dir` parameter from InformationEngine instead of keeping a no-op argument. Since recommendations are now sourced from skeleton alternatives in pattern rules (single source of truth), keeping `rules_dir` in this API path would be misleading and still effectively ignored. I also verified there are no active InformationEngine call sites passing `rules_dir` in the current code path.

fangyangci: Addressed in 4a55624c. I removed the obsolete `rules_dir` parameter from InformationEngine rather than keeping a deprecated-but-ignored argument. This avoids silent failure semantics in this code path and aligns with the PR goal: recommendations now come from pattern-rule skeleton alternatives as the single source of truth. I also checked that no current InformationEngine caller passes `rules_dir`.

src/winml/modelkit/analyze/core/information_engine.py
fangyangci: Addressed in ab290fe6. For aggregated subgraph results (count > 1) with custom explanation, we now preserve support-level context in the prefix before appending the custom explanation (for example: are not supported / have partial support / are fully supported). Added regression test test_aggregated_custom_explanation_keeps_support_context in tests/unit/analyze/core/test_information_engine.py.

fangyangci: Addressed in ab290fe6. The multi-instance custom explanation format now keeps classification context in the message prefix instead of only saying detected. We now emit support-aware prefixes (for example: are not supported / have partial support / are fully supported) and then append the custom explanation. Also added regression test test_aggregated_custom_explanation_keeps_support_context.

src/winml/modelkit/analyze/core/information_engine.py
fangyangci: Thanks for calling this out. We reviewed this path and confirmed this fallback is intentional and preserves prior behavior from the pre-merge information-action flow: when a rule carries explicit action_items, we keep the migration/optimization suggestion even if the support-level transition itself is not an improvement. In the old Information-based path, these rule-authored actions were still surfaced as migration guidance; this PR keeps that semantic in the unified pattern-rule path to avoid losing rule-defined rewrite guidance.

fangyangci: Confirmed this is not an accidental behavior expansion for this PR, but a compatibility carry-over of the previous information-rule fallback semantics. If action_items are explicitly authored by the rule, we still emit an optimization/migration recommendation even when support-level delta is neutral, so rule-defined rewrite guidance is not dropped after unifying to a single source of truth.

src/winml/modelkit/pattern/config.py
fangyangci: Addressed in ca728a55. We removed pair-level metadata inheritance from UnifiedPatternConfig, so alternative metadata is now explicit per entry (no cross-variant reuse and no ordering semantics). To preserve behavior, we explicitly filled metadata for the variants that previously depended on reuse (Gelu2/3/4 and LayerNormalizationMul in default.json), and added regression coverage in test_variants_define_explicit_alternative_action_items.

fangyangci: Implemented in ca728a55. We dropped the inherited pair_metadata_defaults behavior entirely and require explicit metadata on each alternative record. This removes fragile first/ordering dependency. We also made variant entries explicit in default.json (Gelu2/3/4 and LayerNormalizationMul) and added a regression test to ensure these variants carry action_items/details without inheritance.

DingmaomaoBJTU:
## Summary: All functional equivalence concerns have been addressed in this update

This update resolves all four functional equivalence issues raised in the previous review:

### 1. **Silent parameter regression (RESOLVED)**
- The `rules_dir` parameter has been completely removed from `InformationEngine.__init__()` 
- This eliminates the risk of silent failures for callers expecting custom rules_dir support
- **Resolution method**: Complete removal is cleaner than silent ignoring

### 2. **Lost classification context in multi-instance explanations (RESOLVED)**
- Multi-instance explanations now preserve classification information
- Example: `"2 instances of pattern 'X' are not supported."` (lines 200-215 in updated information_engine.py)
- Support phrase is dynamically generated based on `SupportLevel` classification
- **Test verification**: `test_skeleton_metadata_drives_explanation_and_action_items` validates this behavior

### 3. **Undocumented action fallback semantics (RESOLVED)**  
- Fallback for non-improvement alternatives with `action_items` is now explicit and well-handled
- When `level is None` but alternative has `action_items`, generates `ActionLevel.OPTIONAL` action
- Includes explanatory details: `"Pattern 'X' is already {classification}. Alternative 'Y' can still provide graph optimization benefits."`
- **Test verification**: `test_extract_actions_keeps_action_items_without_support_improvement` validates this behavior
- **Result**: Better user messaging than debug logging

### 4. **Fragile ordering dependency on pair-level metadata (RESOLVED)**
- Eliminated the "first-wins" inheritance pattern completely
- Every alternative now explicitly carries its own metadata from JSON: `enabled`, `details`, `action_items`
- Config loading comment now explicitly states: "Metadata must be explicit per alternative entry."
- New method `get_pattern_config()` provides direct access to skeleton-level explanation
- **Test verification**: `test_variants_define_explicit_alternative_action_items` validates each variant has explicit action_items
- **Result**: Robust, order-independent metadata management

### Additional improvements:
- Added deduplication logic for repeated from→to alternative pairs (lines 397, 413-420)
- Added `enabled` field filtering for alternatives (lines 401-402)
- Expanded `RuntimeTestResult` model to carry `enabled`, `details`, `action_items` (runtime_checks.py)
- Updated schema and JSON rules to support new fields

The refactoring now achieves true functional equivalence with a cleaner, more explicit architecture.

DingmaomaoBJTU (approved):

PR #1139, opened 2026-07-20
Add two-ONNX comparison to winml eval --mode compare

xieofxie:
## Summary

Extends `winml eval --mode compare` so the reference can be a second **ONNX file** instead of only the HuggingFace PyTorch model derived from `--model-id`.

A new `--reference <path>` option points at a reference `.onnx`. When set, both the candidate (`-m`) and the reference run as **raw ORT sessions** (`WinMLSession`) on identical inputs — random by default, or the tensors from `--input-data` — and the existing tensor-similarity metrics (cosine, SQNR, PSNR, MSE, max-abs-diff) are computed over their aligned outputs. Because both are raw ONNX graphs, no HF config/task is loaded, so `--model-id` and `--task` are not required in this path. Both sessions honor `--device` / `--ep`.

## Usage

```bash
winml eval -m candidate.onnx --mode compare --reference reference.onnx
```

## Changes

- **`eval/config.py`**: add `reference_path` field (`cli_name="reference"`) with `to_dict` / `from_dict` support.
- **`commands/eval.py`**: add `--reference` option; `_resolve_reference()` normalizes hub refs and validates `.onnx` existence; `allow_missing_model_id` lets a bare ONNX candidate skip the `--model-id` requirement; guard so `--reference` only works with `--mode compare`; print the reference path.
- **`eval/evaluate.py`**: skip task resolution and load no torch model for the two-ONNX path; guard `None` model_id/task and print the reference in the config summary.
- **`eval/tensor_similarity_evaluator.py`**: add `_ONNXSessionModel` wrapper (exposes `onnx_path`, `io_config`, `__call__` returning torch tensors); build both raw sessions when `reference_path` is set, reusing the existing similarity metrics.
- **`docs/commands/eval.md`**: document `--reference`, clarify `--mode`, add a two-ONNX example.
- **Tests**: cover config round-trip, `_resolve_reference` validation, the compare-mode guard, bare-ONNX candidate handling, the two-ONNX evaluate flow, and the `_ONNXSessionModel` wrapper.

## Scope

Phase 1 supports an ONNX-file reference only. HF-local-folder and composite-model references are deferred. The branch is rebuilt on current `main`, so the diff contains only the `--reference` feature (the earlier `--input-data` work landed separately in #1221).

### DIFF
--- a/src/winml/modelkit/commands/eval.py
+++ b/src/winml/modelkit/commands/eval.py
@@ -749,11 +749,20 @@ def display_eval_report(result: EvalResult, console: Console) -> None:
     # archive (via EvalResult.num_samples), not the unused config default.
     samples = result.num_samples if result.num_samples is not None else ds.samples
 
-    # Header
+    # Header — model_id when building from HF, otherwise the ONNX path(s). A
+    # composite model_path is a {role: path} dict; join its paths so the title
+    # stays a readable string instead of a raw dict repr.
+    if cfg.model_id:
+        eval_name = cfg.model_id
+    elif isinstance(cfg.model_path, dict):
+        eval_name = ", ".join(str(path) for path in cfg.model_path.values())
+    else:
+        eval_name = str(cfg.model_path)
+
     console.print()
     console.print(
         Panel.fit(
-            f"[bold]Evaluation: {cfg.model_id or cfg.model_path}[/bold]",
+            f"[bold]Evaluation: {eval_name}[/bold]",
             border_style="blue",
         )
     )
@@ -767,7 +776,10 @@ def display_eval_report(result: EvalResult, console: Console) -> None:
     elif ds.path:
         console.print(f"[dim]Dataset:[/dim]    {ds.path}")
     console.print(f"[dim]Samples:[/dim]    {samples}")
-    if cfg.model_path:
+    if isinstance(cfg.model_path, dict):
+        for role, path in cfg.model_path.items():
+            console.print(f"[dim]ONNX ({role}):[/dim] {path}")
+    elif cfg.model_path:
         console.print(f"[dim]ONNX:[/dim]       {cfg.model_path}")
     if cfg.reference_path:
         console.print(f"[dim]Reference:[/dim]  {cfg.reference_path}")
--- a/src/winml/modelkit/commands/eval.py
+++ b/src/winml/modelkit/commands/eval.py
@@ -198,6 +198,17 @@ logger = logging.getLogger(__name__)
         "sample axis (N samples), and all inputs must share the same N."
     ),
 )
+@click.option(
+    "--reference",
+    "reference",
+    type=str,
+    default=None,
+    help=(
+        "Reference ONNX file to compare the candidate against (use with "
+        "--mode compare). Compares two ONNX models on identical random inputs; "
+        "--model-id / --task are not required in this mode."
+    ),
+)
 @cli_utils.skip_build_option()
 @cli_utils.format_option()
 @cli_utils.build_config_option()
@@ -240,6 +251,7 @@ def eval(
     show_schema: bool,
     mode: EvalMode,
     input_data: str | None,
+    reference: str | None,
     config_file: Path | None,
     skip_build: bool,
 ) -> None:
@@ -251,6 +263,8 @@ def eval(
         winml eval -m model.onnx --model-id microsoft/resnet-50
 
         winml eval --mode compare -m cand.onnx --model-id microsoft/resnet-50 --input-data data.npz
+
+        winml eval --mode compare -m cand.onnx --reference baseline.onnx
 
     Run `winml eval --schema --task <task>` to see the dataset columns
     and options expected by each task.
@@ -287,8 +301,12 @@ def eval(
     if cfg.input_data is not None and cfg.mode != "compare":
         raise click.UsageError("--input-data is only valid with --mode compare.")
 
+    if cfg.reference_path is not None and cfg.mode != "compare":
+        raise click.UsageError("--reference is only valid with --mode compare.")
+
     # ── 2. Resolve in place ──
-    _resolve_model(cfg, model, model_id)
+    _resolve_model(cfg, model, model_id, allow_missing_model_id=cfg.reference_path is not None)
+    _resolve_reference(cfg)
     _apply_export_overrides(cfg, shape_config_path, input_specs, export_config, dynamic_axes)
     _resolve_device(cfg)
     _resolve_label_mapping(cfg)
@@ -420,11 +438,52 @@ def _resolve_model(
     cfg: WinMLEvaluationConfig,
     model: tuple[str, ...],
     model_id: str | None,
+    *,
+    allow_missing_model_id: bool = False,
 ) -> None:
     """Resolve ``-m`` / ``--model-id`` into ``cfg.model_path`` / ``cfg.model_id``."""
-    model_path, resolved_id = _resolve_model_path(model=model, model_id=model_id)
+    model_path, resolved_id = _resolve_model_path(
+        model=model, model_id=model_id, allow_missing_model_id=allow_missing_model_id
+    )
     cfg.model_path = model_path
     cfg.model_id = resolved_id
+
+
+def _resolve_reference(cfg: WinMLEvaluationConfig) -> None:
+    """Validate and normalize ``cfg.reference_path`` for two-ONNX compare.
+
+    Requires the candidate (``-m``) to be a single ONNX file (composite
+    ``role=path`` candidates and build-from-id are not supported with
+    ``--reference`` yet). Resolves Hub-hosted ONNX refs to local paths.
+    """
+    if cfg.reference_path is None:
+        return
+
+    if not isinstance(cfg.model_path, str):
+        raise click.UsageError(
+            "--reference requires the candidate (-m) to be a single ONNX file. "
+            "Composite (role=path) candidates and build-from-id are not "
+            "supported with --reference."
+        )
+
+    ref = cfg.reference_path
+    if Path(ref).suffix.lower() != ".onnx":
+        raise click.BadParameter(
+            f"--reference must be an .onnx file, got: {ref}",
+            param_hint="--reference",
+        )
+    try:
+        ref = cli_utils.normalize_model_arg(ref) or ref
+    except Exception as e:
+        raise click.ClickException(
+            f"Failed to resolve Hub-hosted reference ONNX path {ref!r}: {e}"
+        ) from e
+    if not Path(ref).exists():
+        raise click.BadParameter(
+            f"Reference ONNX file not found: {ref}",
+            param_hint="--reference",
+        )
+    cfg.reference_path = ref
 
 
 def _apply_export_overrides(
@@ -567,8 +626,14 @@ def _resolve_model_path(
     *,
     model: tuple[str, ...],
     model_id: str | None,
+    allow_missing_model_id: bool = False,
 ) -> tuple[str | dict[str, str] | None, str | None]:
-    """Turn repeated -m values + --model-id into (model_path, model_id)."""
+    """Turn repeated -m values + --model-id into (model_path, model_id).
+
+    When ``allow_missing_model_id`` is set (two-ONNX ``--mode compare``), a
+    plain ``-m <file>.onnx`` is accepted without ``--model-id`` because the
+    candidate runs as a raw ORT session with no HF config resolution.
+    """
     if not model:
         if model_id is not None:
             return None, model_id
@@ -644,6 +709,8 @@ def _resolve_model_path(
                 param_hint="-m/--model",
             )
         if model_id is None:
+            if allow_missing_model_id:
+                return value, None
             raise click.UsageError(
                 "When using an ONNX file, --model-id is required "
                 "for preprocessor and config resolution."
@@ -686,7 +753,7 @@ def display_eval_report(result: EvalResult, console: Console) -> None:
     console.print()
     console.print(
         Panel.fit(
-            f"[bold]Evaluation: {cfg.model_id}[/bold]",
+            f"[bold]Evaluation: {cfg.model_id or cfg.model_path}[/bold]",
             border_style="blue",
         )
     )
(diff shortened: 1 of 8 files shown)


DingmaomaoBJTU (commented):
## Review Comments

### ⚠️ Unrelated e2e test removals (~200 lines)

This commit removes `TestProcessExitCleanup`, `TestInspectWarningNoise`, and `test_benchmark_qnn_process_exit_releases_native_session` plus subprocess helpers from 3 e2e files. These are unrelated to the `--reference` feature — were they moved elsewhere or intentionally dropped? Should be a separate commit if intentional.

### ⚠️ Potential bug: torch→numpy in `_ONNXSessionModel.__call__`

`_inference_model` passes **torch tensors** via `model(**inputs)`. `_ONNXSessionModel.__call__` forwards them directly to `self._session.run(inputs)`. Does `WinMLSession.run()` handle torch tensor inputs, or does it expect numpy arrays? If the latter, a `.numpy()` conversion is needed inside `__call__` before calling `self._session.run()`.

### Minor: No `--input-data` + `--reference` combined test

The code supports combining `--input-data` with `--reference` (real tensors fed to two-ONNX compare), but no unit or e2e test covers this interaction path.

### 👍 Overall

Feature is well-designed — clean `_ONNXSessionModel` wrapper, proper validation in `_resolve_reference`, good docs/tests coverage. The concerns above are the main items.

xieofxie:
Thanks for the thorough review @DingmaomaoBJTU. Went through all three:

**① "Unrelated e2e test removals" — not an actual removal; the branch was just behind `main`.**

The branch was rebuilt on `main` at `deadd822` (#1221), and `main` moved 2 commits ahead since (#1239, #1240), which is where `TestProcessExitCleanup`, `TestInspectWarningNoise`, `test_benchmark_qnn_process_exit_releases_native_session`, `ep_path.py`, `test_ensure_provider_ready.py`, etc. landed. A two-dot `git diff main HEAD` on an out-of-date branch renders those additions-on-main as deletions-on-branch, but nothing was actually removed here. I've merged current `main` into the branch, so those files/tests are all present and the diff now shows only the 8 `--reference` files.

**② `torch → numpy` in `_ONNXSessionModel.__call__` — not a bug.**

`WinMLSession.run()` already accepts torch tensors; it converts them to numpy in `_prepare_inputs` before running:

```python
# session.py
if hasattr(value, "numpy"):  # torch.Tensor
    arr = value.cpu().numpy()
```

Its docstring even states inputs may be "torch.Tensor or numpy arrays". So forwarding the tensors directly is correct and no `.numpy()` is needed in `__call__`. The new `test_compute_runs_real_inputs_through_both_sessions` exercises this end-to-end (torch tensors in, numpy out) as a guard.

**③ No `--input-data` + `--reference` combined test — fixed.**

Added `TestONNXReferenceWithInputData` with two cases:
- `test_prepare_data_uses_input_data_over_candidate_session` — confirms `prepare_data` builds an `InputDataDataset` over the candidate `_ONNXSessionModel`'s `io_config` (not a `RandomDataset`) when both options are set, proving they compose.
- `test_compute_runs_real_inputs_through_both_sessions` — runs the compare loop end-to-end on real `.npz` samples through both raw ORT sessions.

All eval/command unit tests pass and ruff is clean.

tests/e2e/test_inspect_e2e.py
```diff
@@ -245,51 +197,6 @@ def test_hierarchy_flag_accepted_without_model(self):
         assert data["hierarchy"] is None  # hierarchy requires -m
```
DingmaomaoBJTU, line 245: This commit removes ~200 lines of unrelated e2e tests (`TestProcessExitCleanup`, `TestInspectWarningNoise`, subprocess helpers across 3 e2e files). These are orthogonal to the `--reference` feature — should they be in a separate commit/PR to keep this change focused?

src/winml/modelkit/eval/tensor_similarity_evaluator.py
```diff
@@ -32,16 +36,65 @@
...
+
+        self.onnx_path = Path(onnx_path)
+        self._session = WinMLSession(onnx_path=self.onnx_path, device=device, ep=ep)
+
+    @property
+    def io_config(self) -> dict:
+        """ONNX I/O metadata (delegated to the session)."""
+        return self._session.io_config
+
+    def __call__(self, **inputs: Any) -> dict[str, Any]:
```
DingmaomaoBJTU, line 62: The `__call__` method receives **torch tensors** from `_inference_model` (which upcasts int dtypes) and forwards them to `self._session.run(inputs)`. I verified that `WinMLSession.run` -> `_prepare_inputs` handles torch->numpy conversion internally, so this is fine. But a brief inline comment noting 'inputs may be torch tensors; the session handles conversion' would help future readers avoid the same concern.

src/winml/modelkit/eval/tensor_similarity_evaluator.py
```diff
@@ -32,16 +36,65 @@
...
+        self.config = config
+
+        # Two-ONNX compare: build both raw ORT sessions directly, bypassing the
+        # HF PyTorch reference. ``model`` is None here (see evaluate._load_model).
+        if config.reference_path is not None:
+            self.model = _ONNXSessionModel(
+                str(config.model_path), device=config.device, ep=config.ep
+            )
+            self.reference_model = _ONNXSessionModel(
+                str(config.reference_path), device=config.device, ep=config.ep
```
DingmaomaoBJTU, line 96: The two-ONNX path bypasses `_load_reference_model` which validates `model_id is not None`. Good. But consider: if someone later passes `--input-data` together with `--reference`, the `prepare_data()` call will follow the `InputDataDataset` branch using `self.model.io_config` — which works, but there's no test covering this combination. Worth adding a test or a note?

src/winml/modelkit/commands/eval.py
```diff
@@ -702,6 +769,8 @@ def display_eval_report(result: EvalResult, console: Console) -> None:
     console.print(f"[dim]Samples:[/dim]    {samples}")
```
DingmaomaoBJTU, line 778: `cfg.model_id or cfg.model_path` — when `model_id` is `None` and `model_path` is a dict (composite), this will print the dict repr. Minor, but might be worth a `str(...)` or handling the composite case explicitly for a cleaner display.

DingmaomaoBJTU (commented):
Overall the `--reference` feature is well-designed. A few inline comments below.

tests/e2e/test_inspect_e2e.py
xieofxie: No lines were actually removed here -- the branch was based on `main` at #1221 and had fallen 2 commits behind (#1239, #1240), which is where these e2e tests landed. A two-dot `git diff main HEAD` on the stale branch renders those additions-on-main as deletions-on-branch. I've merged current `main` in, so `TestProcessExitCleanup`, `TestInspectWarningNoise` and the subprocess helpers are all present again and the diff now shows only the `--reference` files. Thanks for flagging.

src/winml/modelkit/eval/tensor_similarity_evaluator.py
xieofxie: Done in 6119142a -- `__call__` now carries a note that `inputs` may be torch tensors and that `WinMLSession.run` -> `_prepare_inputs` performs the torch->numpy conversion and dtype enforcement, so no explicit `.numpy()` is needed here.

src/winml/modelkit/commands/eval.py
xieofxie: Fixed in 6119142a. The header now joins the sub-model paths (`enc.onnx, dec.onnx`) instead of printing a dict repr, and the detail lines print one `ONNX (role): path` per sub-model. Covered by `TestDisplayEvalReportHeader`.

src/winml/modelkit/eval/tensor_similarity_evaluator.py
xieofxie: Added `TestONNXReferenceWithInputData` in 04cbf7c4 covering the combination: `test_prepare_data_uses_input_data_over_candidate_session` confirms `prepare_data` builds an `InputDataDataset` over the candidate `_ONNXSessionModel`'s `io_config` when both `--input-data` and `--reference` are set, and `test_compute_runs_real_inputs_through_both_sessions` runs the compare loop end-to-end on real `.npz` samples through both raw ORT sessions.

src/winml/modelkit/eval/tensor_similarity_evaluator.py
```diff
@@ -32,16 +36,65 @@
...
         self,
         config: WinMLEvaluationConfig,
         model: WinMLPreTrainedModel | WinMLCompositeModel,
     ) -> None:
         from ..models.winml.composite_model import WinMLCompositeModel
 
+        self.config = config
+
+        # Two-ONNX compare: build both raw ORT sessions directly, bypassing the
+        # HF PyTorch reference. ``model`` is None here (see evaluate._load_model).
```
DingmaomaoBJTU, line 90: When this two-ONNX path is taken, the `compute()` method (not in this diff) still references error/warning messages that say "ONNX and HF reference" — e.g. the `ValueError` at ~line 161 says `"ONNX and HF reference output names do not overlap"`, and the `logger.warning` says `"ONNX and HF reference output names differ"`. Both are misleading when both sides are raw ONNX sessions. Consider making those messages conditional on `self.config.reference_path`.

src/winml/modelkit/eval/tensor_similarity_evaluator.py
```diff
@@ -32,16 +36,65 @@
 logger = logging.getLogger(__name__)
```
DingmaomaoBJTU, line 36: Minor: `_ONNXSessionModel` stores `onnx_path` as a `Path` (line 49) but the config field `reference_path` and `model_path` are `str`. This mismatch is harmless (`prepare_data` calls `str(self.model.onnx_path)`), but it means `_ONNXSessionModel.onnx_path` has a different type than `WinMLPreTrainedModel.onnx_path` — worth a quick type check if you rely on duck typing elsewhere.

DingmaomaoBJTU (commented):
Second-pass review with a few additional findings.

tests/unit/commands/test_eval.py
```diff
@@ -335,6 +344,82 @@ def test_help_mentions_input_data(self, runner: CliRunner):
...
+            eval_cmd,
+            ["-m", str(onnx_file), "--reference", str(onnx_file)],
+            obj={"debug": False},
+        )
+        assert result.exit_code != 0
+        assert "--reference is only valid with --mode compare" in result.output
+
 
 class TestInputDataModeGuard:
     def test_input_data_requires_compare_mode(self, runner: CliRunner, onnx_file, tmp_path):
```
DingmaomaoBJTU, line 425: Good validation coverage for `_resolve_reference`. One gap: no test for the `display_eval_report` path where `cfg.model_id` is `None` (two-ONNX compare). The report header now uses `cfg.model_id or cfg.model_path` (eval.py line 769) — a test asserting the report renders correctly without `model_id` would catch regressions.

src/winml/modelkit/eval/tensor_similarity_evaluator.py
xieofxie: Good catch — fixed in 0b2c8bbb. `compute()` now derives a `reference_label` (`"reference ONNX"` when `self.config.reference_path` is set, else `"HF reference"`) and uses candidate/reference wording in both the no-overlap `ValueError` and the mismatch `logger.warning`. Added `TestReferenceLabelInDiagnostics` (with a `_FakePathAwareSession` that returns non-overlapping output names) to assert the two-ONNX path says "reference ONNX" and not "HF reference".

src/winml/modelkit/eval/tensor_similarity_evaluator.py
xieofxie: The types actually match, so there's no duck-typing mismatch here: `WinMLPreTrainedModel.onnx_path` is also typed as `Path` (base.py:102-104, `def onnx_path(self) -> Path: return self._onnx_path`, where `self._onnx_path = Path(onnx_path)`). `_ONNXSessionModel.onnx_path` is likewise a `Path`, so both duck-type consistently. Leaving as-is.

tests/unit/commands/test_eval.py
xieofxie: This is already covered — `test_two_onnx_compare_shows_candidate_path_without_model_id` (in `TestDisplayEvalReportHeader`, added in 6119142a) renders `display_eval_report` with `model_id=None`, `model_path="cand.onnx"`, `reference_path="ref.onnx"`, `mode="compare"` and asserts the header shows `Evaluation: cand.onnx`. Note the `cfg.model_id or cfg.model_path` at line 769 has since been replaced by an `eval_name` helper that renders composite dict paths readably instead of printing the dict repr.

DingmaomaoBJTU (commented):
Third-pass review — backward compatibility, error handling, and flag interactions deep-dive.

**Verified OK (no issues):**
- Backward compat: `from_dict()` uses `.get("reference_path")` → old config files without the field load fine (defaults to `None`)
- `to_dict()` only emits `reference_path` when non-None → minimal serialization, no noise for existing configs
- Error recovery: `WinMLSession` failures in `_ONNXSessionModel.__init__` bubble up cleanly through `evaluate()`'s exception handler
- `--skip-build` interaction: two-ONNX path bypasses entire build pipeline, no conflict
- `--ep`/`--device`: both models use same EP/device by design (fair comparison isolating quantization/optimization diffs)
- `_resolve_reference` mutation: safe — config is `replace()`-copied before `evaluate()`
- `allow_missing_model_id` with non-ONNX: gated by `_resolve_reference`'s `.onnx` suffix check, so HF model IDs are blocked upstream
- Click option ordering: decorators are independent, no ordering dependency

DingmaomaoBJTU (approved):
LGTM — approving after three review passes.

**Non-blocking follow-ups** (can be addressed in a later PR):
- `compute()` error/warning messages still say "ONNX and HF reference" in two-ONNX mode — cosmetic, doesn't affect functionality
- Consider adding a test for `display_eval_report` when `model_id` is None (two-ONNX path)

All other areas verified: backward compat, config serialization, error recovery, flag interactions, torch/numpy handling — all clean.

DingmaomaoBJTU (approved):
LGTM

Issue #369, opened 2026-04-20
Refactor: extract common CLI options into shared decorators


DingmaomaoBJTU:
## Summary

Extract commonly duplicated CLI options (`--model`, `--device`, `--ep`, `--task`, `--output`, `--verbose`) into shared reusable decorators in `src/winml/modelkit/utils/cli.py`, and adopt them across all commands to eliminate inconsistencies.

## Context

Raised in [PR #354 review](https://github.com/microsoft/WinML-ModelKit/pull/354) by @xieofxie: *"should we add common options for others that are reused across different commands?"* — @DingmaomaooBJTU confirmed this should be done.

A shared utility module already exists at `src/winml/modelkit/utils/cli.py` with four decorators (`model_option`, `ep_option`, `device_option`, `verbosity_options`), but **only `analyze.py` uses them**. The remaining 11 commands each define their own copies of these options with divergent types, defaults, help text, and casing.

### Existing shared decorators in `src/winml/modelkit/utils/cli.py`
- `model_option(required=True)` — `--model/-m` as `click.Path(exists=True, path_type=Path)`
- `ep_option(required=True, optional_message=None)` — `--ep` as `click.Choice(ALL_EP_NAMES)`
- `device_option(required=True, optional_message=None, default="NPU")` — `--device` as `click.Choice(SUPPORTED_DEVICES)`
- `verbosity_options(f)` — stackable `--verbose/-v` (count) + `--quiet/-q` (flag)

### Only consumer: `analyze.py`
All other commands duplicate these options inline.

### Key inconsistencies across commands

| Option | Commands using it | Inconsistencies |
|--------|------------------|-----------------|
| `--model/-m` | analyze, build, compile, eval, export, hub, inspect, perf, quantize (9) | Type varies: `click.Path` vs `str` (for HF model IDs); required flag differs |
| `--device/-d` | analyze, build, compile, config, eval, perf (6) | Default varies: `"NPU"` (analyze), `"npu"` (compile), `"auto"` (config/eval/perf); case sensitivity differs |
| `--ep` | analyze, build, compile, config, perf (5) | Choice source differs: `ALL_EP_NAMES` vs `VALID_EPS` from config module vs manual strings |
| `--task/-t` | config, eval, export, inspect, perf, quantize (6) | All optional `str` with auto-detection — duplicated definition |
[...]

PR #1141, opened 2026-07-20
feat: Add export override options to winml eval command

xieofxie:
## Summary

Brings `winml eval` to parity with `winml build` and `winml perf` by adding four export override options:

- `--shape-config` — JSON shape overrides for the auto-generated HuggingFace export config
- `--input-specs` — JSON input tensor specs merged into the HuggingFace export config (symbolic dims infer dynamic axes)
- `--export-config` — JSON ONNX export config overrides (opset version, constant folding, etc.)
- `--dynamic-axes` — JSON dynamic axes mapping for HuggingFace ONNX export

These options shape the ONNX export that `eval` generates when `-m` is a HuggingFace model ID. When `-m` is a pre-built `.onnx` file there is no export step, so the flags are ignored and the command prints a warning.

## Changes

- **`eval/config.py`** — added `shape_config` and `export_overrides` fields to `WinMLEvaluationConfig`, JSON-safe serialization of `InputTensorSpec` values, and `to_dict`/`from_dict` support.
- **`eval/evaluate.py`** — `_load_model` threads `{"export": ...}` overrides plus `shape_config` into `WinMLAutoModel.from_pretrained` on the HuggingFace build path; the `from_onnx` path is unchanged.
- **`commands/eval.py`** — added the four option decorators/params and an `_apply_export_overrides` helper that parses the JSON and warns + skips for pre-built ONNX inputs.
- **`docs/commands/eval.md`** — documented the four flags and added a pitfall note.
- **Tests** — added `TestEvalExportOverrides` (10 cases) and updated the existing `from_pretrained` assertion for the new `shape_config` kwarg.

## Verification

- `uv run pytest tests/unit/commands/test_eval.py tests/unit/eval/test_eval.py` → 103 passed
- `uv run ruff check` → clean
- Doc flag/code alignment check → passing


src/winml/modelkit/eval/evaluate.py
```diff
@@ -270,14 +272,28 @@ def _load_model(config: WinMLEvaluationConfig) -> WinMLPreTrainedModel | WinMLCo
...
         return model
 
+    # HuggingFace build path — export overrides (--input-specs/--export-config/
+    # --dynamic-axes) are merged under the build config's ``export`` section as a
+    # sparse dict so from_pretrained routes them through merge_export_overrides
+    # (patching auto-resolved input_tensors by name / re-deriving dynamic_axes).
+    # Passing a dict rather than a WinMLBuildConfig avoids clobbering the
+    # auto-resolved export config with default fields. Mirrors winml build/perf.
+    build_override: Any = quant_override
+    if config.export_overrides:
```
zhenchaoni: Updated todo list

Here's a PR-ready comment:

---

## Review: input option precedence

I checked the implementation against the intended precedence order:

**CLI explicit > config-file explicit > CLI default > config-class default**

The four new flags (`--shape-config`, `--input-specs`, `--export-config`, `--dynamic-axes`) bypass the standard `collect_cli_overrides`/`merge_config` machinery (they bind to `shape_config_path`/`input_specs`/`export_config`/`dynamic_axes`, which aren't `WinMLEvaluationConfig` field names) and are handled entirely by the new `_apply_export_overrides`.

### ✅ What's correct

- **CLI default > config-class default** — the options default to `None`; `_apply_export_overrides` returns early when no flag is provided, so CLI defaults never clobber anything.
- **config-file explicit > CLI default** — the config-file `eval` section can carry `shape_config`/`export_overrides`, and they survive when the CLI omits the flags. They then flow into `from_pretrained(config={"export": ...})` → `generate_hf_build_config`, which routes them through `merge_export_overrides` (patch `input_tensors` by name, re-derive dynamic axes), correctly preserving the auto-generated/Optimum defaults tier.
- **CLI explicit > config-file explicit for `shape_config`** — single-value override, fine.

### ❌ Precedence bug: `export_overrides` is replaced wholesale, not merged

In `_apply_export_overrides`:

```python
export_overrides = cli_utils.load_export_overrides(
    export_config=export_config, input_specs=input_specs, dynamic_axes=dynamic_axes,
)
if export_overrides:
    cfg.export_overrides = export_overrides   # wholesale replace
```

`load_export_overrides` returns a **sparse dict containing only the CLI-provided sub-keys**, so assigning it **discards any config-file `export_overrides` sub-keys the CLI didn't set**.

Repro: config file sets `export_overrides = {"opset_version": 17, "dynamic_axes": {...}}`, then run with only `--input-specs foo.json`. Result: `cfg.export_overrides = {"input_tensors": [...]}` — the config-file `opset_version` and `dynamic_axes` are **lost**, even though the CLI never set them. Per the intended ordering, those config-file values should win (config-file explicit > CLI default).

This is also a parity gap with `build`, which deliberately avoids exactly this by calling `merge_export_overrides(cfg, export_overrides)` on the loaded config *"instead of merge_config replacing the list wholesale"*.

**Suggested fix** — merge instead of replace:

```python
if export_overrides:
    merged = dict(cfg.export_overrides or {})
    merged.update(export_overrides)
    cfg.export_overrides = merged
```

### Related concerns (not strictly precedence)

1. **Config-file `input_tensors` will crash.** Config-file `export_overrides` is merged as a raw dict via `merge_config(cfg, eval_data)` (not through `from_dict`/`load_input_tensor_specs`), so its `input_tensors` stay plain dicts. Downstream `_patch_input_tensors` accesses `patch.name`/`patch.dtype`, which raises `AttributeError` on a dict — so config-file-supplied `input_tensors` is effectively unusable.
2. **Pre-built ONNX path:** the warn-and-skip only covers CLI flags. If the config file carries `shape_config`/`export_overrides` alongside a pre-built `.onnx`, they stay on `cfg` (ignored downstream, but with no warning) — minor inconsistency.

Overall the ordering is correct for `shape_config` and for the config-file/default tiers, but `export_overrides` needs field-level merging between the CLI and config-file layers to satisfy the intended precedence.

src/winml/modelkit/commands/eval.py
```diff
@@ -88,6 +88,32 @@
...
 )
 @cli_utils.max_optim_iterations_option(optional_message="Ignored for pre-built ONNX inputs.")
+@cli_utils.shape_config_option(
+    param_name="shape_config_path",
+    help_text=(
+        "JSON with shape overrides for auto-generated HuggingFace export configs. "
+        "Ignored for pre-built ONNX inputs."
+    ),
+)
+@cli_utils.input_specs_option(
```
zhenchaoni, line 98: An optional suggestion: as we are having more and more options, do you think they can be wrapped into a single "--additional-options" for all those optional command line inputs?

xieofxie: code wise we have an item like this https://github.com/microsoft/winml-cli/issues/369

but I don't think we should nested them one level lower to make it complex to use

src/winml/modelkit/eval/evaluate.py
xieofxie: Good catch — this was still valid on the current code (the anchor was on an older commit, but `_apply_export_overrides` was unchanged). Fixed in ee8e66a1: the CLI overrides now shallow-merge over any config-file `export_overrides` instead of replacing wholesale, so CLI-set sub-keys win while untouched config-file sub-keys (e.g. `opset_version`/`dynamic_axes`) survive — matching build's `merge_export_overrides` intent. Added two tests covering the merge and the same-key-CLI-wins cases.

zhenchaoni (approved):

PR #1142, opened 2026-07-20
feat(optim): add static Split-to-Slice rewrite

xieofxie:
## Summary

Add an opt-in `static-split-to-slice` algebraic rewrite that replaces eligible ONNX `Split` nodes with equivalent input-form `Slice` nodes while preserving output tensor names.

The rewrite only applies when the input shape, axis, and split boundaries are statically known. It supports explicit split sizes and equal splits, avoids malformed or dynamic cases, handles name collisions, preserves nested-subgraph captures, removes dead generated nodes and constants, and remains idempotent through the public optimization path.

## Why this is not a pattern rewrite

The existing pattern framework is designed for fixed-topology replacements with compatible source and target schemas. Its match result exposes a single pattern output, and `PatternRewriter` constructs the replacement using that fixed output mapping.

A `Split` node is variadic: one matched node may have any number of outputs, and the replacement must generate one `Slice` node plus four boundary initializers for each output. The replacement topology and constants therefore depend on the specific matched node rather than on a fixed target pattern. Representing this correctly as a pattern would first require broader framework support for variadic output mappings and match-parameterized replacement topology.

For that reason, this transformation is implemented as a focused algebraic rewrite rather than adding model-specific behavior or forcing it into the current fixed-shape pattern abstraction.


src/winml/modelkit/optim/pipes/algebraic.py
```diff
@@ -0,0 +1,423 @@
...
+# --------------------------------------------------------------------------
+"""Conservative, opt-in algebraic ONNX graph rewrites."""
+
+from __future__ import annotations
+
+from dataclasses import dataclass
+from typing import Any, ClassVar
+
+import numpy as np
+import onnx
```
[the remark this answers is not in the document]
xieofxie: Updated the implementation to use the existing `import onnx` consistently and reference `onnx.helper`/`onnx.numpy_helper` through that module.

tests/unit/optim/pipes/test_pipe_algebraic.py
```diff
@@ -0,0 +1,267 @@
...
+# Licensed under the MIT License.
+# --------------------------------------------------------------------------
+"""Generated-graph tests for static Split-to-Slice rewriting."""
+
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+import numpy as np
+import onnx
```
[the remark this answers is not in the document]
xieofxie: Updated the generated-graph tests to use only `import onnx`, including qualified helper, numpy helper, and TensorProto references.

src/winml/modelkit/optim/pipes/algebraic.py
```diff
@@ -0,0 +1,413 @@
...
+        shapes: dict[str, tuple[int | None, ...]] = {}
+        for value_info in (*graph.input, *graph.value_info, *graph.output):
+            shape = _value_info_shape(value_info)
+            if shape is not None:
+                shapes[value_info.name] = shape
+        for name, initializer in initializers.items():
+            shapes.setdefault(name, tuple(int(dim) for dim in initializer.dims))
+
+        for initializer in initializers.values():
+            onnx.numpy_helper.to_array(initializer)
```
DingmaomaoBJTU: This onnx.numpy_helper.to_array(initializer) call discards its result -- it doesn't feed into shapes or anything else. If it's meant as a validation guard, that's a bit surprising to bury silently inside _GraphIndex.build, and it's wasted work since _GraphIndex.build is rebuilt repeatedly inside the _prune_generated_slices/_prune_dead_constant_nodes convergence loops -- every initializer gets re-decoded on every iteration even though nothing here consumes the array. Worth removing (or hoisting into an explicit, intentional validation step) to avoid the redundant conversion cost on large models.

src/winml/modelkit/optim/pipes/algebraic.py
```diff
@@ -0,0 +1,413 @@
...
+    index: _GraphIndex,
+    node: onnx.NodeProto,
+    input_name: str,
+) -> tuple[int, list[tuple[int, int]]] | None:
+    """Return a static Split axis and output boundaries."""
+    input_shape = index.shapes.get(input_name)
+    if input_shape is None or len(node.output) == 0:
+        return None
+
+    axis_input_index = 2 if len(node.input) > 2 else None
```
DingmaomaoBJTU: Split's ONNX schema only ever has up to 2 inputs (input, split) -- xis is attribute-only in every opset (verified via onnx.defs.get_schema('Split').max_input == 2). So len(node.input) > 2 can never be true for a valid, checker-passing model, and xis_input_index is effectively dead code that never resolves to a real 'axis input' (there isn't one in the spec). Recommend dropping this input-based axis lookup and just reading the xis attribute directly -- it clarifies intent and avoids implying Split has an axis input, which could confuse future maintainers.

DingmaomaoBJTU (commented):
Reviewed the static Split-to-Slice rewrite. Overall design looks solid: opt-in flag, conservative bail-out for dynamic/malformed cases, dead-code pruning, and nested-subgraph capture handling are all done carefully. Left two non-blocking inline notes.

src/winml/modelkit/optim/pipes/algebraic.py
xieofxie: Removed the redundant initializer decoding from `_GraphIndex.build`, so repeated graph-index construction no longer converts every initializer without consuming the result.

src/winml/modelkit/optim/pipes/algebraic.py
xieofxie: Updated Split axis handling to read the attribute directly with its default of zero; the nonexistent third input is no longer considered.

DingmaomaoBJTU (approved):

PR #1154, opened 2026-07-21
Fix CTC model class resolution by passing model_type to TasksManager

codykk:
Pass `model_type` to Optimum's `TasksManager.get_model_class_for_task` in `resolve_task()` Stage 2, fixing model class resolution for all CTC-based ASR architectures (Wav2Vec2ForCTC, HubertForCTC, WavLMForCTC, etc.). One-line generalized fix — no per-model hardcoding. Add regression recipe for `jonatasgrosman/wav2vec2-large-xlsr-53-russian` (cpu/fp32, cpu/fp16).

Effort: L1-light | Goal: L1 (perf) | Outcome: L1

## Model metadata

- **What it does**: Wav2Vec2-Large-XLSR-53 fine-tuned on Russian Common Voice for CTC-based automatic speech recognition. Converts 16kHz raw audio waveforms to Russian text transcriptions.
- **Primary user stories**: Russian speech-to-text transcription for voice applications
- **Supported tasks**: `automatic-speech-recognition` (CTC decoding)
- **Architecture**:
  Wav2Vec2ForCTC
  ├── Wav2Vec2FeatureEncoder (7 CNN layers: raw audio → features)
  ├── Wav2Vec2FeatureProjection (LayerNorm + Linear → hidden_size=1024)
  ├── Wav2Vec2Encoder (24× transformer layers, 16 heads)
  └── Linear CTC head (1024 → 39 Cyrillic vocab)

## Validation and support evidence

**Baseline (main):** Build FAIL — `Unrecognized configuration class Wav2Vec2Config for this kind of AutoModel: AutoModelForSpeechSeq2Seq`

**Root cause:** `TasksManager.get_model_class_for_task("automatic-speech-recognition")` without `model_type` defaults to `AutoModelForSpeechSeq2Seq`; CTC models need `AutoModelForCTC`.

**Goal ceiling:** L1 (perf) — L2/L3 CLI-BLOCKED (`automatic-speech-recognition` not in `winml eval` supported tasks)

**Outcome:** L1 — generalized code fix + regression recipes

| EP | Device | Precision | Build | Perf |
|---|---|---|---|---|
| cpu | cpu | fp32 | ✅ PASS | 125.6ms avg, 7.96 samples/sec, +1268MB RAM |
| cpu | cpu | fp16 | ✅ PASS | 140.5ms avg, 7.12 samples/sec, +1276MB RAM |

**Delta:** `resolution.py` line 647 — added `model_type=model_type_norm or None`. When `model_type_norm` is available (from `config.model_type`), Optimum looks up the correct class per architecture family instead of defaulting to the first registered class. Empty string falls back to `None` (original behavior preserved).

**Reproduce:**
```bash
winml build -m jonatasgrosman/wav2vec2-large-xlsr-53-russian -o output/ --ep cpu --device cpu
winml perf -m output/model.onnx --ep cpu --device cpu

### DIFF
--- a/src/winml/modelkit/loader/resolution.py
+++ b/src/winml/modelkit/loader/resolution.py
@@ -569,7 +569,7 @@ def resolve_task(
         if resolved is None:
             try:
                 resolved = TasksManager.get_model_class_for_task(
-                    normalized, framework="pt", model_type=model_type_norm or None
+                    normalized, framework="pt", model_type=model_type or None
                 )
             except KeyError as e:
                 if composite is not None:
@@ -658,7 +658,7 @@ def resolve_task(
         if resolved is None:
             try:
                 resolved = TasksManager.get_model_class_for_task(
-                    opt_task, framework="pt", model_type=model_type_norm or None
+                    opt_task, framework="pt", model_type=model_type or None
                 )
             except Exception:
                 resolved = _resolve_model_class_from_config(config)  # arch fallback
--- a/src/winml/modelkit/loader/resolution.py
+++ b/src/winml/modelkit/loader/resolution.py
@@ -568,7 +568,9 @@ def resolve_task(
             ) or _get_custom_model_class(model_type_norm, normalized)
         if resolved is None:
             try:
-                resolved = TasksManager.get_model_class_for_task(normalized, framework="pt")
+                resolved = TasksManager.get_model_class_for_task(
+                    normalized, framework="pt", model_type=model_type_norm or None
+                )
             except KeyError as e:
                 if composite is not None:
                     # Pure composite (e.g. table-question-answering): no single model class
--- a/src/winml/modelkit/loader/resolution.py
+++ b/src/winml/modelkit/loader/resolution.py
@@ -655,7 +655,9 @@ def resolve_task(
         resolved = _get_custom_model_class(model_type_norm, opt_task)
         if resolved is None:
             try:
-                resolved = TasksManager.get_model_class_for_task(opt_task, framework="pt")
+                resolved = TasksManager.get_model_class_for_task(
+                    opt_task, framework="pt", model_type=model_type_norm or None
+                )
             except Exception:
                 resolved = _resolve_model_class_from_config(config)  # arch fallback
 
--- a/examples/recipes/jonatasgrosman_wav2vec2-large-xlsr-53-russian/cpu/cpu/automatic-speech-recognition_fp16_config.json
+++ b/examples/recipes/jonatasgrosman_wav2vec2-large-xlsr-53-russian/cpu/cpu/automatic-speech-recognition_fp16_config.json
@@ -1,0 +1,60 @@
+{
+  "export": {
+    "opset_version": 17,
+    "batch_size": 1,
+    "export_params": true,
+    "do_constant_folding": true,
+    "verbose": false,
+    "dynamo": false,
+    "enable_hierarchy_tags": true,
+    "clean_onnx": false,
+    "hierarchy_tag_format": "full",
+    "input_tensors": [
+      {
+        "name": "input_values",
+        "dtype": "float32",
+        "shape": [
+          1,
+          16000
+        ],
+        "value_range": [
+          -1,
+          1
+        ]
+      }
+    ],
+    "output_tensors": [
+      {
+        "name": "logits"
+      }
+    ]
+  },
+  "optim": {},
+  "quant": {
+    "mode": "fp16",
+    "samples": 10,
+    "calibration_method": "minmax",
+    "weight_type": "uint8",
+    "activation_type": "uint8",
+    "per_channel": false,
+    "symmetric": false,
+    "weight_symmetric": null,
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

**Review by DingmaomaoBJTU — APPROVED** [review]

File: src/winml/modelkit/loader/resolution.py [hunk 1]
```diff
@@ -644,7 +644,9 @@ def resolve_task(
         resolved = _get_custom_model_class(model_type_norm, opt_task)
         if resolved is None:
             try:
-                resolved = TasksManager.get_model_class_for_task(opt_task, framework="pt")
+                resolved = TasksManager.get_model_class_for_task(
+                    opt_task, framework="pt", model_type=model_type_norm or None
```
xieofxie: This only fixes the Stage 2 auto-detection path. When `task="automatic-speech-recognition"` is supplied without `model_class`, the earlier `USER_TASK` branch still calls `get_model_class_for_task(normalized, framework="pt")` without `model_type`, so a Wav2Vec2 CTC config can still resolve to `AutoModelForSpeechSeq2Seq`. Please pass `model_type=model_type_norm or None` in that branch as well and add a regression test for explicit-task resolution without a model-class override.

codykk: Thanks for catching this.

src/winml/modelkit/loader/resolution.py
```diff
@@ -557,7 +557,9 @@ def resolve_task(
             ) or _get_custom_model_class(model_type_norm, normalized)
         if resolved is None:
             try:
-                resolved = TasksManager.get_model_class_for_task(normalized, framework="pt")
+                resolved = TasksManager.get_model_class_for_task(
+                    normalized, framework="pt", model_type=model_type_norm or None
```
xieofxie: Please pass the raw `model_type` to `TasksManager` here rather than `model_type_norm`. The normalization is needed for WinML's internal mapping keys, but Optimum registers some model types with underscores. For example, `speech_to_text` is a supported Transformers model type; converting it to `speech-to-text` prevents Optimum from matching `AutoModelForSpeechSeq2Seq`, so explicit ASR resolution raises instead of resolving the class. The Stage 2 call added by this PR has the same issue. Please keep `model_type_norm` for `_get_custom_model_class`, pass `model_type` to both TasksManager calls, and add an underscore-containing regression case.

codykk: Fixed.

xieofxie (approved):
The earlier model-type resolution issues are addressed. No additional blocking findings in the latest revision.

ssss141414 (approved):

xieofxie:
I suggest not force push.. since we will squash merge to main, keep commit history will make human review pr easier

codykk:
> I suggest not force push.. since we will squash merge to main, keep commit history will make human review pr easier



> The earlier model-type resolution issues are addressed. No additional blocking findings in the latest revision.

Got it, will keep the commit history going forward.

DingmaomaoBJTU (commented):
## Review Summary

✅ **LGTM** — This is a correct, focused fix.

### What it does
Passes raw `model_type` (with underscores, e.g. `speech_to_text`) to `TasksManager.get_model_class_for_task()` at two call sites in `resolution.py` (lines 560 and 649). This allows TasksManager to disambiguate architectures sharing the same task (e.g. Wav2Vec2 CTC vs Whisper for `automatic-speech-recognition`).

### Verification

| Check | Result |
|-------|--------|
| `model_type or None` handles empty string | ✅ Correctly converts to `None` |
| Uses raw underscore format (not hyphens) | ✅ `speech_to_text` not `speech-to-text` |
| Deleted `_autoconfig` module has no stale references | ✅ Cleaned up |
| Deleted `_winml_generic_fallback` has no stale references | ✅ Cleaned up |
| API compatibility (`TasksManager` accepts `model_type` kwarg) | ✅ Confirmed |

No bugs, security issues, or incomplete refactoring found.

DingmaomaoBJTU (changes requested):
## ⚠️ Merge conflict resolution accidentally reverted recent main commits

The merge commit `e5ccb854` (Merge branch 'main' into yongyue/fix-ctc-model-class-resolution) appears to have dropped changes from two recent PRs on `main`:

### 1. **#1223** — Concurrent stderr drain reverted (🔴 Bug reintroduced)

`src/winml/modelkit/utils/native_stderr.py` reverts the background-thread pipe drain back to synchronous drain. The original fix (#1223, commit `57c2cdf2`) was necessary because a chatty native EP can emit more than the OS pipe buffer (~64 KB), causing the synchronous `os.read()` to deadlock since the pipe write-end is still open during `yield`.

### 2. **#1187** — OWLv2 recipes and Einsum op support deleted

- `examples/recipes/google_owlv2-large-patch14-ensemble/` configs deleted
- `src/winml/modelkit/pattern/op_input_gen/einsum_input_generator.py` deleted
- Related tests removed

### Recommended fix
Re-do the merge from main, being careful to keep changes from both sides during conflict resolution.

PR #1156, opened 2026-07-21
Add dynamo export support and make it the default exporter

DingmaomaoBJTU:
## Why

`winml export` previously rejected `--dynamo` even though PyTorch's dynamo ONNX exporter is the actively developed path. This PR makes dynamo a supported default while preserving the module hierarchy metadata used by inspection, reporting, benchmarking, and optimization tooling.

## What changed

- **Dynamo is the default:** `WinMLExportConfig` and `winml export` now default to `dynamo=True`. `--no-dynamo` remains available for the legacy TorchScript exporter, and generated submodule configs preserve an explicit parent exporter choice.
- **Registered exporter settings are preserved:** when Optimum supplies I/O specs, registered model configs still retain explicit exporter, opset, hierarchy, and related behavior settings. Batch size is applied before I/O generation using `user override > registry > default` precedence so the final config and tensor shapes stay consistent.
- **Hierarchy recovery from ONNX metadata:** `DynamoMetadataTagger` reads generic `pkg.torch.onnx.name_scopes` and `pkg.torch.onnx.class_hierarchy` metadata. It handles sparse scopes and same-class siblings without architecture-, operator-, or tensor-specific logic.
- **Source-aware export pipeline:** dynamo skips the hook-based TorchScript trace and reconstructs the hierarchy after export. Reports and metadata distinguish `onnx_metadata` recovery from `trace`, use discovery-order wording for dynamo, and source output names from the exported ONNX graph.
- **Visible hierarchy degradation:** a non-empty dynamo graph with no recoverable hierarchy metadata emits one concise warning and continues with root fallback tags; partial recovery remains valid without a warning.
- **Requested and actual opsets:** the requested opset is passed directly to PyTorch with no additional WinML conversion. The actual produced opset is read from the ONNX file and reported. Mismatches use exporter- and direction-aware warnings; detailed torch/onnxscript downgrade traces remain hidden unless verbose logging is enabled.
- **Compatibility:** existing JSON field names are retained, all checked-in recipes continue to pin `dynamo=false`, and `--no-dynamo` preserves the existing TorchScript path.

## Validation

- Export unit suite: 476 passed, 6 skipped, 1 xfailed.
- Config unit and integration suites: 357 passed.
- Broader affected unit scopes: 681 passed, 6 skipped, 1 xfailed.
- Mypy: 406 source files clean; Ruff clean.
- ResNet-50 end to end: ONNX and HTP metadata generated, 200/286 modules recovered, requested opset 17 and actual opset 18 reported with the concise WinML warning.


docs/commands/export.md
```diff
@@ -22,7 +22,7 @@ $ winml export [options]
 | `--output` | `-o` | path | *(required)* | Output ONNX file path (e.g., `model.onnx`). |
 | `--with-report/--no-with-report` | | flag | `false` | Generate full export reports: Markdown, JSON, and a console tree. |
 | `--hierarchy/--no-hierarchy` | | flag | `true` | Preserve `hierarchy_tag` metadata in ONNX nodes (use `--no-hierarchy` for a clean ONNX file). |
-| `--dynamo/--no-dynamo` | | flag | `false` | Enable PyTorch 2.9+ dynamo export for richer node metadata. (Experimental — currently logs a warning.) |
+| `--dynamo/--no-dynamo` | | flag | `true` | Use PyTorch's TorchDynamo ONNX exporter (default) for richer per-node module metadata. Pass `--no-dynamo` for the legacy TorchScript exporter (QNN-safe). |
```
xieofxie: what QNN-safe mean?

xieofxie: it is a little wierd. from my understanding, all NPU requires static shaped model? Maybe rephrase like `for the legacy TorchScript exporter (which generates static shaped model suitable for npu by default)`

And also add one part about how to generate static shaped model

src/winml/modelkit/commands/export.py
```diff
@@ -116,9 +116,10 @@ def _warn_partial_composite(completed: list[Path]) -> None:
 @click.option(
```
xieofxie, line 116: It should be added for all necessary commands? created an issue to track https://github.com/microsoft/winml-cli/issues/1161

src/winml/modelkit/commands/export.py
```diff
@@ -116,9 +116,10 @@ def _warn_partial_composite(completed: list[Path]) -> None:
...
     "--dynamo/--no-dynamo",
```
xieofxie, line 117: the parameter is automatically wrapped into config object?

docs/commands/export.md
DingmaomaoBJTU: Good callout -- "QNN-safe" was unclear jargon, so I have reworded it across the flag table, the notes section, the concept doc, and the `--dynamo` help text.

It does **not** mean static shapes (both exporters default to static shapes). It refers to the ONNX graph the QNN/NPU toolchain has been validated against. I confirmed the real difference by exporting `microsoft/resnet-50` both ways:

- `--no-dynamo` (TorchScript): **opset 17**, and the head lowers to `GlobalAveragePool` + `Flatten`
- default (dynamo): **opset 18** (torch's dynamo op library has a minimum opset of 18), and the same head lowers to `ReduceMean` + `Reshape`

So the differentiator is opset + op decomposition, not shapes. The opset-17 TorchScript graph is what QNN has been validated against, so `--no-dynamo` stays the validated path for QNN hand exports today.

DingmaomaoBJTU: I adapted the wording rather than applying it verbatim, because "generates static shaped model ... by default" isn't accurate as a *differentiator*: static shapes are the default for **both** exporters. Staticness is controlled by the input specs (`--shape-config` / `--input-specs`, with `--dynamic-axes` to opt into dynamic), not by the exporter choice. The real QNN-relevant difference is opset / op decomposition.

I reworded the docs to say that precisely, and added a static-shape export example showing how to produce one:

```bash
winml export -m bert-base-uncased -o bert.onnx --shape-config shape_config.json
# shape_config.json: {"sequence_length": 128}
```

Happy to adjust the framing if you'd prefer.

src/winml/modelkit/commands/export.py
DingmaomaoBJTU: Agreed. This PR intentionally scopes the CLI flag to `winml export` (only `export` exposes `--dynamo/--no-dynamo`; the shared `WinMLExportConfig.dynamo` field is already threaded so recipes/build configs can set it). Rolling the flag out to the other ONNX-producing commands (e.g. `build`) is follow-up work -- thanks for filing #1161 to track it.

src/winml/modelkit/commands/export.py
DingmaomaoBJTU: Yes. The click option is collected as the command's `dynamo` parameter, then wrapped into the config object: when the user explicitly passes the flag (guarded by `is_cli_provided(ctx, "dynamo")`) it's written to `config_kwargs["dynamo"]` and built via `WinMLExportConfig.from_dict(config_kwargs)` (export.py ~417-429). If neither `--dynamo/--no-dynamo` is given, the config's own default (now `True`) governs; a recipe's `export.dynamo` is applied earlier (~267). So the flag only overrides the config when actually specified.

src/winml/modelkit/core/onnx_node_tagger.py
```diff
@@ -311,3 +312,163 @@ def create_node_tagger_from_hierarchy(
...
+        segments: list[str] = []
+        for cls, scope in zip(classes, scopes, strict=False):
+            short = self._short_class(cls)
+            if not short:
+                continue
+            local = scope.rsplit(".", 1)[-1] if scope else ""
+            if local.isdigit():
+                segments.append(f"{short}.{local}")
+            else:
+                segments.append(short)
```
xieofxie: Named sibling modules that share a class collapse to the same hierarchy tag here. For example, attention `query`, `key`, and `value` scopes are each `torch.nn.Linear`, so all three become `/.../Linear`; the legacy path intentionally emitted distinct `/query`, `/key`, and `/value` tags. Because these tags drive per-module benchmarking and optimizer scoping, the default Dynamo path can no longer target those modules independently. Please preserve scope identity generically (for example, by incorporating/disambiguating the local scope component) and cover multiple same-class named siblings in the tests.

src/winml/modelkit/export/htp/exporter.py
```diff
@@ -229,13 +240,21 @@ def export(
...
-            # nested.
-            with self._get_optimum_patcher(model, task):
-                self._trace_model_hierarchy(model, inputs)
+            #
+            # Under dynamo the exporter records the module hierarchy natively on
+            # each ONNX node's metadata_props, so the TorchScript trace is
+            # redundant (and can fail for models that only export via dynamo).
+            # The hierarchy is recovered post-export from that metadata in Step 5.
+            if self._use_dynamo_hierarchy:
+                self._hierarchy_data = {}
```
xieofxie, line 250: The comment says the hierarchy is recovered post-export, but only per-node tags are recovered: `_hierarchy_data` stays empty and the `HIERARCHY` monitor update runs before the ONNX model exists. Consequently, default Dynamo exports report `hierarchy_modules == 0`, and `MetadataWriter` omits the module tree/tracing data even though the nodes contain enough metadata to reconstruct it. The skipped builder also makes the Step 4 report record `output_names=[]` instead of the authoritative config names. Please reconstruct/update the report hierarchy after loading the ONNX model and source reported output names from the config or graph.

src/winml/modelkit/core/onnx_node_tagger.py
DingmaomaoBJTU: Fixed in dc3c49c2. `_tag_from_metadata` now folds the local scope component into every level's segment (`Class.local`), unifying with the existing indexed rule, so same-class named siblings stay distinct: an attention block's query/key/value now tag as `.../Linear.query|key|value` instead of collapsing to one `/.../Linear`. Still fully generic (no hardcoded name/class lists). Added a same-class-sibling test.

src/winml/modelkit/export/htp/exporter.py
DingmaomaoBJTU: Fixed in dc3c49c2. Added `DynamoMetadataTagger.build_module_hierarchy()` that reconstructs the flat module hierarchy from node metadata. The exporter now recovers it after loading the exported ONNX (Step 5) and issues the previously-premature HIERARCHY monitor update there, so the report/console/metadata module tree and the `hierarchy_modules` stat are populated for dynamo (verified 200 modules on resnet-50, was 0). Dynamo `output_names` are now sourced from the ONNX graph outputs instead of the empty traced outputs (['logits'] on resnet-50). Added build_module_hierarchy unit tests.

src/winml/modelkit/export/htp/exporter.py
```diff
@@ -266,35 +286,71 @@ def export(
...
+            # Output names: dynamo exposes them authoritatively on the ONNX graph;
+            # the TorchScript path derives them from the traced outputs.
+            if self._use_dynamo_hierarchy:
+                output_names = [output.name for output in onnx_model.graph.output]
+            else:
+                traced_outputs = (
+                    self._hierarchy_builder.get_outputs() if self._hierarchy_builder else None
+                )
+                output_names = (
+                    infer_output_names(traced_outputs) if traced_outputs is not None else []
```
xieofxie: This assignment currently breaks the required mypy check: the Dynamo branch infers `output_names` as `list[str]`, while `infer_output_names()` returns `list[str] | None`, producing the CI error at this line. Normalize the legacy result (for example, `infer_output_names(...) or []`) or declare a compatible type before merging.

src/winml/modelkit/core/onnx_node_tagger.py
```diff
@@ -311,3 +312,223 @@ def create_node_tagger_from_hierarchy(
...
+
+    def build_module_hierarchy(self, onnx_model: onnx.ModelProto) -> dict[str, dict[str, Any]]:
+        """Reconstruct the module hierarchy from dynamo node metadata.
+
+        The dynamo exporter does not run the TorchScript trace that normally
+        feeds the export monitor's module tree, so this rebuilds the same flat
+        ``{scope_path -> module info}`` mapping directly from the per-node
+        ``name_scopes``/``class_hierarchy`` metadata. The result matches the
+        shape :class:`TracingHierarchyBuilder` produces, so the shared
+        report/console/metadata writers render the dynamo module tree
```
xieofxie, line 503: `build_module_hierarchy()` does not yet satisfy the traversal contract of the shared writers. The new test fixture produces `"" -> "blocks.0"` with no intermediate `"blocks"` entry, but both `find_immediate_children("", ...)` and `MetadataWriter._build_children_for_parent("", ...)` only accept root children whose path contains no dot. As a result, `blocks.0` and its entire subtree are absent from the console/report/metadata tree even though `hierarchy_modules` counts them. Please either make the writers handle compound root children/sparse parent scopes or normalize the reconstructed hierarchy, and exercise the actual writer output in a test rather than only asserting the flat map.

docs/commands/export.md
```diff
@@ -114,9 +123,21 @@ winml export -m microsoft/resnet-50 -o resnet50_clean.onnx --no-hierarchy
...
-- **`--dynamo` and `--torch-module` are experimental.** Both flags emit a
-  warning and have no effect in the current release. Do not rely on them in
-  automated pipelines yet.
+- **Dynamo is the default exporter.** `winml export` uses PyTorch's TorchDynamo
+  ONNX exporter, which records rich per-node module metadata that drives the
+  hierarchy tags. Pass `--no-dynamo` to select the legacy TorchScript exporter.
+  Shape staticness is independent of the exporter: both default to static shapes
+  and only emit dynamic axes when you ask for them. The QNN-relevant
+  difference is opset and op decomposition: torch's dynamo op library targets a
+  minimum opset of 18, so the dynamo path exports at opset 18, whereas the
```
xieofxie: This states the final Dynamo graph is always opset 18, but the implementation explicitly reads the produced opset because conversion to the requested version can succeed or fail. Dynamo builds against its minimum opset 18, then may successfully emit the requested opset; only models that cannot be lowered remain at 18. Please describe that conditional behavior here (and in the concept page) so users do not assume every default export is opset 18.

src/winml/modelkit/core/onnx_node_tagger.py
```diff
@@ -311,3 +312,223 @@ def create_node_tagger_from_hierarchy(
...
+            if not classes or not scopes:
+                continue
+
+            segments: list[str] = []
+            for scope, short, segment in self._module_segments(classes, scopes):
+                segments.append(segment)
+                if scope in hierarchy:
+                    continue
+                hierarchy[scope] = {
+                    "class_name": short,
```
xieofxie, line 528: The flat map keeps `query`/`key`/`value` distinct, but the persisted metadata tree still collapses them. `MetadataWriter._build_children_for_parent()` keys every non-indexed child as `children[module_info.class_name]`; these three entries all have `class_name == "Linear"`, so each assignment overwrites the previous sibling and only one survives. Please use a key that includes the local scope identity (or otherwise supports duplicate classes) and add a writer-level assertion that all same-class siblings are serialized.

docs/commands/export.md
DingmaomaoBJTU: Fixed in 9b360f88. Reworded both `docs/commands/export.md` and `docs/concepts/load-and-export.md`: dynamo targets a *minimum* opset of 18 and attempts to down-convert to a lower requested opset, which may succeed (saved at the requested opset) or fail (stays at 18); `winml export` reports the opset actually produced and warns on mismatch. The docs no longer claim the dynamo graph is always opset 18.

src/winml/modelkit/export/htp/exporter.py
DingmaomaoBJTU: Fixed in 9b360f88. Normalized the TorchScript branch to `infer_output_names(...) or []` so `output_names` is always `list[str]` at the merge point (the dynamo branch already yields `list[str]`). `uv run mypy -p winml.modelkit` is now clean (406 files, no issues).

src/winml/modelkit/core/onnx_node_tagger.py
DingmaomaoBJTU: Fixed in 9b360f88, at the writer level so it holds for any hierarchy producer (not only the reconstructed dynamo one). `find_immediate_children` now nests via a `_nearest_present_ancestor` helper: a path is an immediate child of the longest ancestor scope actually present in the hierarchy, falling back to root. A sparse `blocks.0` (no `blocks` container entry) therefore attaches to root instead of being dropped, while dense hierarchies keep their existing nesting. `MetadataWriter._build_children_for_parent` now reuses this shared traversal, so the JSON report gets the same fix. Added `tests/unit/export/test_dynamo_hierarchy_report.py`, which exercises the real `MetadataWriter` tree plus the sparse/dense `find_immediate_children` cases. Verified end to end on a real resnet-50 dynamo export (286 hierarchy modules, tree renders fully).

src/winml/modelkit/core/onnx_node_tagger.py
DingmaomaoBJTU: Fixed in 9b360f88. `_build_children_for_parent` now keys indexed children `Class.N` (unchanged) and disambiguates same-class *named* siblings as `Class.local` via a two-pass collision count, so an attention block's query/key/value Linears serialize as `Linear.query`/`Linear.key`/`Linear.value` instead of collapsing. The bare `Class` key is kept for the unique case, with a full-remainder uniqueness fallback for safety. The new writer-level test asserts all three attention Linears appear distinctly in the persisted tree.

docs/commands/export.md
```diff
@@ -22,7 +22,7 @@ $ winml export [options]
 | `--output` | `-o` | path | *(required)* | Output ONNX file path (e.g., `model.onnx`). |
 | `--with-report/--no-with-report` | | flag | `false` | Generate full export reports: Markdown, JSON, and a console tree. |
 | `--hierarchy/--no-hierarchy` | | flag | `true` | Preserve `hierarchy_tag` metadata in ONNX nodes (use `--no-hierarchy` for a clean ONNX file). |
-| `--dynamo/--no-dynamo` | | flag | `false` | Enable PyTorch 2.9+ dynamo export for richer node metadata. (Experimental — currently logs a warning.) |
+| `--dynamo/--no-dynamo` | | flag | `true` | Use PyTorch's TorchDynamo ONNX exporter (default) for richer per-node module metadata. Pass `--no-dynamo` for the legacy TorchScript exporter, whose opset-17 op decomposition is the validated path for QNN/NPU compilation today. |
```
xieofxie, line 25: After flipping this default, several user-facing workflow descriptions still say the default performs module-hook tracing and TorchScript export: the “How it works” paragraph below still lists `module-hierarchy tracing, TorchScript ONNX export`, `load-and-export.md` says hooks are always registered, and the CLI docstring says “TorchScript by default.” The Dynamo branch explicitly skips that trace and reconstructs hierarchy from ONNX metadata, so please update these descriptions to distinguish the two paths.

src/winml/modelkit/export/htp/exporter.py
```diff
@@ -266,35 +286,74 @@ def export(
...
+            # instead of a TorchScript trace. Recover it now that the graph exists
+            # and issue the deferred Step 3 update so the report/console/metadata
+            # module tree and the hierarchy_modules stat reflect the real modules.
+            if self._use_dynamo_hierarchy:
+                assert isinstance(self._node_tagger, DynamoMetadataTagger)
+                self._hierarchy_data = self._node_tagger.build_module_hierarchy(onnx_model)
+                monitor.update(
+                    ExportStep.HIERARCHY,
+                    hierarchy=self._hierarchy_data,
+                    execution_steps=len(self._hierarchy_data),
```
xieofxie: No execution trace exists on this branch, so `len(self._hierarchy_data)` is the number of reconstructed modules, not `execution_steps`. Passing it under that field makes the console/metadata claim “Tracing module execution…” and report a made-up execution-step count for every default Dynamo export. Please represent the hierarchy source explicitly (or leave execution steps unavailable/zero) and make the shared writers use source-appropriate wording instead of reporting a trace that did not happen.

docs/commands/export.md
DingmaomaoBJTU: Fixed in 3c07a438. The default (dynamo) and `--no-dynamo` paths are now distinguished in all three spots: the CLI docstring (steps 3–4), `docs/commands/export.md` ("How it works"), and `docs/concepts/load-and-export.md` ("How tags are built"). They now state that under the default TorchDynamo exporter the hierarchy is reconstructed from ONNX node metadata (no forward hooks), while `--no-dynamo` registers PyTorch forward hooks and traces the forward pass. The bert-tiny example is labelled as the `--no-dynamo` TorchScript-trace path.

src/winml/modelkit/export/htp/exporter.py
DingmaomaoBJTU: Fixed in 3c07a438. The dynamo path no longer fabricates an execution-step count. `HierarchyData` now carries an explicit `source` (`trace` vs `onnx_metadata`) and `execution_steps` is `int | None`; the dynamo HIERARCHY update sets `source=onnx_metadata` and leaves `execution_steps` unset. Writers now use source-appropriate wording: the console shows "Reconstructing module hierarchy from ONNX node metadata…" / "Recovered N modules" (no execution-step line), the export summary says "Recovered modules" instead of "Traced modules", and the metadata records `source=onnx_metadata`, `builder=DynamoMetadataTagger`, `execution_steps=0` (the schema documents the new `source` field). Verified end-to-end on resnet-50.

src/winml/modelkit/export/config.py
```diff
@@ -177,7 +177,7 @@ class WinMLExportConfig:
     export_params: bool = True
     do_constant_folding: bool = True
     verbose: bool = False
-    dynamo: bool = False  # Use TorchScript exporter by default (dynamo=True for PyTorch 2.x)
+    dynamo: bool = True  # TorchDynamo exporter by default; False = legacy TorchScript path
```
xieofxie, line 180: Flipping this default makes registered `dynamo=False` configs ineffective on the normal config-generation path. In `generate_hf_build_config`, a registered export config is only copied when it already contains `input_tensors`; otherwise `_resolve_export_config_from_specs()` returns a fresh config with this new `dynamo=True` default, and `_assemble_config()` uses it unchanged (`_merge_export_config()` is never called). The registered qwen3 transformer/embeddings/lm-head-only configs all specify `WinMLExportConfig(dynamo=False, opset_version=18)` without input tensors, so they now silently switch to Dynamo despite explicitly requiring TorchScript. The new submodule test only mutates an already-built parent and therefore misses this path. Please merge the registered export overrides (including an explicit false) into the resolved config and add a generation-level regression test.

DingmaomaoBJTU: Fixed in 66abb115. The normal generation path now lets Optimum resolve I/O and then merges every non-default registered export setting, so the qwen3 variant configs retain `dynamo=False` and opset 18. The merge handles explicit false values instead of using truthiness. Effective batch size is selected before I/O resolution with user override > registry > default precedence, keeping generated tensor shapes consistent. Added generation-level regressions for registered exporter settings, all behavior fields, both user override forms, and batch-shape consistency.

xieofxie (approved):

PR #1162, opened 2026-07-21
fix unknow pattern case

fangyangci:
related check results and rules parquet:
https://github.com/gim-home/ModelKitArtifacts/pull/156


src/winml/modelkit/pattern/gemm_patterns.py
```diff
@@ -19,6 +19,15 @@
 from .op_input_gen import InputShapeConstraint, InputValueConstraint
 
 
+def _is_scalar_constraint_value(value: Any) -> bool:
```
DingmaomaoBJTU, line 22: This scalar-vs-vector bias logic is the core behavioral change on the Gemm side, but there's no unit test for it in this PR (only the transpose generator got new tests). Please add coverage asserting `C_is_value_constraint` is `True` for a scalar bias and `False` for a 1-D/N-D vector constant and for `None` — that's exactly the distinction this fix hinges on, and it'd lock the regression in.

src/winml/modelkit/pattern/gemm_patterns.py
```diff
@@ -388,7 +397,7 @@ class GemmPatternInputGenerator(PatternInputGenerator):
     """PatternInputGenerator for Gemm patterns."""
 
     # Typed as the base Pattern so subclasses can set a different concrete pattern.
-    pattern: Pattern = ReshapeGemmReshapePattern()
+    pattern = ReshapeGemmReshapePattern()
```
DingmaomaoBJTU: This drops the explicit `: Pattern` annotation, but the comment right above still says "Typed as the base Pattern so subclasses can set a different concrete pattern." Without the annotation the attribute falls back to the base `Pattern | None` type, and this change looks unrelated to the unknown-pattern fix. Suggest restoring `pattern: Pattern = ReshapeGemmReshapePattern()` (or dropping the now-inaccurate comment).

src/winml/modelkit/pattern/gemm_patterns.py
```diff
@@ -447,9 +456,12 @@ def derive_properties(self, properties: dict) -> dict:
...
 
-        # Distinguish fixed-value bias constraints from shape-only constraints.
-        # This is finite (bool) and derives directly from existing row/query fields.
-        item["C_is_value_constraint"] = "C_value" in item and item["C_value"] is not None
+        # Distinguish scalar literal constraints from shape-based constraints.
+        # Vector/tensor constants should remain shape-constrained (not value-constrained).
+        c_value = item.get("C_value")
+        item["C_is_value_constraint"] = _is_scalar_constraint_value(c_value) and item[
+            "C_dim"
+        ] == 0
```
DingmaomaoBJTU, line 466: Is `and item["C_dim"] == 0` adding anything beyond `_is_scalar_constraint_value(c_value)`? A scalar / 0-d value already implies `C_shape == ()` → `C_dim == 0`, so this second clause looks redundant. If it's intentional defensive coding for rows where value and shape disagree, a one-line comment would help; otherwise it can be dropped to keep the intent clear.

src/winml/modelkit/pattern/transpose_patterns.py
```diff
@@ -338,20 +338,62 @@ def get_finite_attribute_sets(self) -> dict[str, list[Any]]:
...
                 "data": InputShapeConstraint((1, 256, 256, 96)),
                 "transpose_shape": (1, 32, 8, 32, 8, 96),
                 "perm": (0, 1, 3, 2, 4, 5),
                 "output_shape": (1024, 8, 8, 96),
             },
+            {
+                "data": InputShapeConstraint((1, 256, 256, 96)),
+                "transpose_shape": (1, 32, 8, 32, 8, 96),
+                "perm": (0, 1, 3, 2, 4, 5),
+                "output_shape": (1024, 64, 96),
```
DingmaomaoBJTU, line 357: Good — this is the missing 3-D output coverage (and the later case adds `merged_transpose_dim=5`). One thing to confirm before this actually fixes anything for users: the runtime matcher loads precomputed rule tables (parquet/JSON from a release, per `rules/runtime_check_rules/README.md`), not the generators directly. So these new combinations only change classification once the runtime-check rules are regenerated and re-released. Is that regeneration planned as part of this change, or does it happen out-of-band?

tests/unit/analyze/pattern/test_transpose_patterns.py
```diff
@@ -244,3 +246,50 @@ def test_low_dim_pattern_id_differs_from_schema_name(self) -> None:
         pattern = ReshapeTransposeReshapeLowDimPattern()
         schema_based = f"SUBGRAPH/{pattern.get_schema().name}"
         assert pattern.pattern_id != schema_based
+
+
+class TestRTRPatternInputGeneratorCoverage:
+    """Regression tests for RTR pattern input-generation coverage."""
+
+    def test_output_dim_coverage_includes_3d_and_4d(self) -> None:
```
DingmaomaoBJTU, line 255: These tests assert that the generator returns the combinations that were just hard-coded above — useful as a guard against accidental removal, but they don't exercise the path that actually produced the "unknown pattern" (property derivation → rule matching in `runtime_checker_query`). Consider adding a regression test that runs a representative previously-unknown subgraph through the matcher and asserts it's now classified, so the tests protect the behavior rather than the fixture.

DingmaomaoBJTU (commented):
Reviewed the fix for the "unknown pattern" cases. The overall approach is sound: broaden the RTR transpose input-generator to cover 3-D outputs and `merged_transpose_dim=5`, and make the Gemm generator treat only scalar biases as value-constraints so vector/tensor biases stay shape-constrained (instead of emitting value-based rules that real models never match). I hand-checked the added transpose combinations — they're all element-count consistent, and the PixelShuffle case does resolve to `merged_transpose_dim=5`.

A few things to look at (details inline):
- The Gemm-side behavioral change (`_is_scalar_constraint_value` / `C_is_value_constraint`) has no unit test — only the transpose generator got new tests.
- `pattern: Pattern` lost its type annotation in `GemmPatternInputGenerator`; that looks unrelated to this fix and now contradicts the comment directly above it.
- The new transpose tests assert the hard-coded fixture rather than the derivation → rule-matching path that actually produced "unknown".
- End-to-end: the matcher loads precomputed rule tables (parquet/JSON pulled from a release, per `rules/runtime_check_rules/README.md`), so these generator changes won't change real analyses until the runtime-check rules are regenerated and re-released. Is that regeneration planned as part of landing this?

Note: I couldn't run the suite locally (ARM64 host vs. the AMD64/linux lockfile), so please rely on CI to confirm the new tests pass.

src/winml/modelkit/pattern/gemm_patterns.py
fangyangci: Thanks for the review — correcting my earlier reply here. I have restored the explicit base annotation and aligned it with the parent declaration to avoid unintended type narrowing in subclasses: `pattern: Pattern | None = ReshapeGemmReshapePattern()`. This keeps the class attribute type consistent with `PatternInputGenerator` while allowing subclass overrides like `MatMulAddPattern` without mypy assignment incompatibility.

src/winml/modelkit/pattern/transpose_patterns.py
fangyangci: Great point. The runtime-check rule artifacts are generated and released outside this repo. This PR updates the in-repo generators/tests, and the corresponding rules regeneration is synced through the external artifacts pipeline, so user-visible classification will be updated with that artifact sync.

src/winml/modelkit/pattern/gemm_patterns.py
fangyangci: Good question. `_is_scalar_constraint_value(c_value)` only checks the value payload shape (for ndarray: `value.ndim == 0`), while `C_dim` comes from the separately derived `C_shape` metadata. They are related but not identical signals. We intentionally keep `and C_dim == 0` as a defensive consistency gate so scalar value constraints do not get classified as value-constrained when shape metadata disagrees. I added an inline comment in code to make this intent explicit.

src/winml/modelkit/pattern/op_input_gen/attention_input_generator.py
```diff
@@ -0,0 +1,59 @@
+# -------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+# --------------------------------------------------------------------------
+"""Input generator for ai.onnx Attention operator (opset 23+).
```
DingmaomaoBJTU, line 5: The docstring says `ai.onnx Attention operator (opset 23+)` but `available_ops_all.json` lists Attention with `start_version: 99999` and `valid_for_gene: false` - this is a **com.microsoft** contrib op, not a standard ai.onnx op.

Should this class set `op_domain = "com.microsoft"` so it registers under the domain-qualified key `com.microsoft::Attention` (same pattern as the Gelu generator)? Without it, the op registers under the default ai.onnx domain which seems incorrect.

src/winml/modelkit/pattern/op_input_gen/attention_input_generator.py
```diff
@@ -0,0 +1,59 @@
...
+                "Q": InputShapeConstraint((1, 2, 4, 8)),
+                "K": InputShapeConstraint((1, 2, 6, 8)),
+                "V": InputShapeConstraint((1, 2, 6, 8)),
+                "attn_mask": InputShapeConstraint((1, 2, 4, 6)),
+            },
+            # Larger shape with explicit scale attribute.
+            {
+                "Q": InputShapeConstraint((2, 4, 8, 16)),
+                "K": InputShapeConstraint((2, 4, 8, 16)),
+                "V": InputShapeConstraint((2, 4, 8, 16)),
```
DingmaomaoBJTU, line 55: `scale: 1.0 / 16` is a raw float mixed with `InputShapeConstraint` entries in the dict. Worth adding a comment clarifying that non-`InputShapeConstraint` keys are forwarded as op attributes, since the other entries are all shapes and it's not immediately obvious.

src/winml/modelkit/pattern/base.py
```diff
@@ -925,7 +925,8 @@ def get_onnx_model(
                 output_dtypes[output_idx]
             ).tensor_proto_type
 
-            output_tensor = helper.make_tensor_value_info(output_name, elem_type, None)
+            # Keep shape present for ONNX checker while leaving dimensions unknown.
+            output_tensor = helper.make_tensor_value_info(output_name, elem_type, [None])
```
DingmaomaoBJTU: This changes the output shape from `None` (unknown rank) to `[None]` (rank-1 with one unknown dim). For patterns whose output is multi-dimensional (e.g., matmul -> 2-D, conv -> 4-D), a hardcoded rank-1 shape declaration is semantically incorrect and may cause ONNX checker warnings.

Consider using `[None] * expected_output_rank` derived from pattern metadata, or documenting why rank-1 is always safe here.

src/winml/modelkit/pattern/transpose_patterns.py
```diff
@@ -338,20 +338,146 @@ def get_finite_attribute_sets(self) -> dict[str, list[Any]]:
...
         return [
             {
                 "data": InputShapeConstraint((1, 256, 256, 96)),
                 "transpose_shape": (1, 32, 8, 32, 8, 96),
                 "perm": (0, 1, 3, 2, 4, 5),
                 "output_shape": (1024, 8, 8, 96),
             },
+            {
+                "data": InputShapeConstraint((1, 256, 256, 96)),
+                "transpose_shape": (1, 32, 8, 32, 8, 96),
```
DingmaomaoBJTU, line 355: Nit: This method grew from 2 entries to 20+. Many entries share the same `perm=(0,1,3,2,4,5)` and only vary spatial/hidden dims and output shape. Consider generating these programmatically from a compact `(spatial, hidden_size, output_dims)` spec to reduce maintenance burden.

src/winml/modelkit/pattern/gemm_patterns.py
```diff
@@ -447,9 +456,14 @@ def derive_properties(self, properties: dict) -> dict:
         else:
             item["C_dim"] = 0
 
-        # Distinguish fixed-value bias constraints from shape-only constraints.
-        # This is finite (bool) and derives directly from existing row/query fields.
-        item["C_is_value_constraint"] = "C_value" in item and item["C_value"] is not None
+        # Distinguish scalar literal constraints from shape-based constraints.
+        # Vector/tensor constants should remain shape-constrained (not value-constrained).
+        c_value = item.get("C_value")
+        # Keep the shape gate as a defensive consistency check in case value/shape
```
DingmaomaoBJTU, line 462: :+1: Nice fix - the old check incorrectly treated vector constants as value-constrained. The new scalar-only gate with the `C_dim == 0` defensive check is correct and well-tested.

tests/unit/analyze/pattern/conftest.py
```diff
@@ -148,17 +152,24 @@ def add_first_op_output_to_graph_output(
...
+
     assert output_type is not None, "Expected to find type info for first op output"
 
     new_output = helper.make_tensor_value_info(
         first_output_name,
         output_type.tensor_type.elem_type,
         None,
     )
-
+    new_nodes = list(model.graph.node)
```
DingmaomaoBJTU: Nit: `new_nodes = list(model.graph.node)` is the same expression that was already inlined. The extra local variable doesn't seem to serve any purpose - was there an intent to mutate it before passing?

tests/unit/analyze/test_input_generators.py
```diff
@@ -40,8 +40,24 @@ class TestInputGeneratorRegistry:
 
     def test_all_operators_registered(self) -> None:
         """Test that all operators are registered."""
-        # Verify count
-        assert len(get_registered_operators()) == 120
+        registered_ops = get_registered_operators()
+
+        # Keep a lower bound so additive registrations do not break this test.
```
DingmaomaoBJTU, line 45: :+1: Good change - `>= 120` lower bound + uniqueness check makes this test resilient to additive registrations without masking accidental removals.

DingmaomaoBJTU (commented):
Overall solid fix for unknown patterns with good test coverage. A few concerns below, primarily about the Attention domain and the base.py shape change.

fangyangci:
Tracking issue: #1236

src/winml/modelkit/pattern/op_input_gen/attention_input_generator.py
fangyangci: Thanks for flagging this. I rechecked the current source and ONNX schema: `available_ops_all.json` records Attention with `start_version: 23` and `valid_for_gene: true`, and ONNX reports `Attention` opset 23 in the default domain. The Q/K/V signature here is the standard `ai.onnx::Attention`; `com.microsoft::Attention` has a different signature, so no `op_domain` change is needed.

src/winml/modelkit/pattern/op_input_gen/attention_input_generator.py
fangyangci: Added the requested clarification in `2536ec45`: non-input entries in these combinations are forwarded as operator attributes.

src/winml/modelkit/pattern/base.py
fangyangci: Fixed in `2536ec45`. The graph output now starts with unknown rank (`None`) and model generation uses the existing symbolic-first `infer_shapes` helper instead of standard-only `infer_onnx_shapes`, so the true output rank is inferred without hardcoding. I added a rank-2 `com.microsoft::Gelu` regression that validates the inferred `[2, 4]` shape with `onnx.checker.check_model(..., full_check=True)`. The affected test scope passes: 685 passed, 7 existing runtime-limitation skips.

src/winml/modelkit/pattern/transpose_patterns.py
fangyangci: Thanks for the suggestion. I am keeping these as explicit regression fixtures: several cases intentionally vary `perm`, use `-1`, or cover the PixelShuffle-style merged-rank path, and the comments tie each concrete layout to its coverage intent. A compact generated product would reduce lines but make the previously unknown layouts and future fixture-level regressions less obvious.

src/winml/modelkit/pattern/gemm_patterns.py
fangyangci: Thanks for confirming the scalar-only behavior and its coverage.

tests/unit/analyze/pattern/conftest.py
fangyangci: There was no intended mutation. I removed the redundant `new_nodes` local and now pass `list(model.graph.node)` directly in `2536ec45`.

tests/unit/analyze/test_input_generators.py
fangyangci: Thanks for confirming the registry assertion behavior.

DingmaomaoBJTU (approved):
Reviewed. No high-confidence blocking issues found.

PR #1163, opened 2026-07-21
Make e2e eval recipe/precision expansion NPU-only

KayMKM:
## Why

In the e2e eval harness, recipe-driven builds and multi-precision job expansion were applied on every device. That only makes sense on the NPU: recipes target NPU quantization, and expanding a model into several quantized precisions has no meaning on CPU/GPU. This change scopes that behavior to `--device npu` and gives NPU models without a recipe a sensible default matrix.

## What changed

- **Job expansion is now NPU-only.** `_build_jobs` takes the target `device`. Recipes are only consulted when `device == "npu"`. On CPU/GPU/auto, recipes are ignored and each model builds as a single `winml config` fallback job (prior behavior).
- **NPU fallback expands to w8a8 + w8a16.** An NPU model with no recipe and no explicit per-model precision now expands into two jobs (w8a8 and w8a16). An explicit per-model `precision` still wins and produces a single job.
- **Forward `--precision` to `winml build`.** The fallback path invoked `winml build -c <config> --device npu` without `--precision`. Because `--device` is always passed, `winml build` re-resolves quant from device+precision and overwrites the quant baked into the config by `winml config`. With precision defaulting to auto, the NPU build reverted to w8a16, so a w8a8 job and a w8a16 job produced the same cached artifact (identical config hash) and perf reported w8a16 for both. The build call now forwards the resolved precision, mirroring the `winml config` invocation.

## Notes for reviewers

- NPU detection is an exact `device == "npu"` match (auto does not count), consistent with `_resolve_precision`.
- The recipe build path (`_run_recipe_build`) passes neither `--device` nor `--precision`, so the config's quant stays authoritative there. It was never affected by the overwrite issue.
- The `--precision` overwrite mechanism predates this PR and was latent on main (NPU fallback only ever resolved to w8a16, which matched the build's auto default). Introducing the w8a8 variant is what surfaced it; the fix also closes the latent case where a per-model `fp16` entry built on NPU would be silently coerced to w8a16.
- `--list` still prints one line per model and does not preview expanded jobs. That is pre-existing behavior and unchanged here.

Docs (`scripts/e2e_eval/README.md`) and unit tests are updated to cover the NPU-only expansion, the w8a8 + w8a16 fallback, and the `--precision` forwarding.

### DIFF
--- a/scripts/e2e_eval/run_eval.py
+++ b/scripts/e2e_eval/run_eval.py
@@ -6,12 +6,14 @@
 """E2E evaluation runner — unified, recipe-driven perf + accuracy.
 
 Batch-builds models and runs winml perf, then (when perf passes) winml eval,
-writing one unified eval_result.json per (model, task, precision). Recipes and
-precision expansion are NPU-only: on ``--device npu`` builds prefer an authored
-recipe under ``examples/recipes/`` (``winml build -c`` for each precision variant),
-and an NPU model without one falls back to ``winml config`` expanded into ``w8a8``
-+ ``w8a16`` jobs (an explicit per-model precision overrides this). CPU/GPU ignore
-recipes and build a single ``winml config`` variant.
+writing one unified eval_result.json per (model, task, precision). Recipes drive
+the build on every device (they carry the accuracy eval/dataset config), but
+quantized precision variants are NPU-only. On ``--device npu`` the runner builds
+every authored recipe variant under ``examples/recipes/`` (``winml build -c``),
+and a recipe-less NPU model falls back to ``winml config`` expanded into ``w8a8``
++ ``w8a16`` jobs (an explicit per-model precision overrides this). CPU/GPU build
+only the non-quantized recipe variants (e.g. ``fp16``), or a single ``winml
+config`` fallback when a model has no applicable recipe.
 
 The runner records facts only (perf output + the winml-eval metrics/dataset).
 The per-model report HTML — perf latency and the "Model Accuracy Report" delta
@@ -95,8 +97,8 @@ _DEFAULT_PRECISION_NPU = "w8a16"
 
 # NPU fallback precisions used when a model has no authored recipe: the harness
 # expands into one winml-config job per NPU quantization scheme (w8a8 + w8a16).
-# An explicit per-model precision (``ModelEntry.precision``) overrides this, and
-# non-NPU devices never expand (they keep the single winml-config fallback).
+# An explicit per-model precision (``ModelEntry.precision``) overrides this. This
+# expansion is NPU-only; off-NPU devices never build quantized variants.
 _NPU_FALLBACK_PRECISIONS: tuple[str, ...] = ("w8a8", "w8a16")
 
 # EPs whose eval track keeps the model unquantized (the "fp" variant)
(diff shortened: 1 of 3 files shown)


DingmaomaoBJTU (changes requested):
CPU/GPU also need recipes for winml eval. Please just ignore quantized version for them and add fallbacks

DingmaomaoBJTU (commented):
Reviewed the NPU-only recipe/precision expansion. The core change is sound and the `--precision` forwarding fix is well-reasoned and well-tested. A few observations below — one edge case worth a look (NPU + VitisAI), plus two low-severity notes.

scripts/e2e_eval/run_eval.py
```diff
@@ -2301,34 +2320,80 @@ class EvalJob:
...
             if recipes_dir is not None
             else []
         )
+        if not npu:
+            # Off-NPU still uses recipes for their eval config, but drops
+            # quantized variants -- quantization is an NPU-only concern here.
+            variants = [v for v in variants if not _is_quantized_precision(v.precision)]
         if variants:
             jobs.extend(EvalJob(entry, variant) for variant in variants)
+        elif npu and entry.precision is None:
```
DingmaomaoBJTU: On NPU paired with a skip-quant EP (VitisAI — the documented `vitisai_npu` combo), a recipe-less model with no per-model precision still expands here into two jobs (`w8a8` + `w8a16`). But `_resolve_precision` forces precision to `None` for `_EPS_SKIP_WINML_QUANT` EPs, so both jobs build the *same* unquantized artifact and write two `eval_result.json` (`…__w8a8` and `…__w8a16`) with identical perf — the same "two precisions collapse onto one artifact" class of bug this PR fixes elsewhere. Pre-PR this was a single fallback job, so `run_eval.py --device npu --ep vitisai` on a recipe-less model regresses to duplicate/mislabeled results.

Since `_build_jobs` only receives `device` (not `ep`), consider also threading `ep` through and skipping the expansion when `_should_skip_winml_quant(ep)` (emit a single fallback instead). Not the default path (`--ep` defaults to None), so low severity, but it's a supported invocation.

scripts/e2e_eval/run_eval.py
```diff
@@ -2301,34 +2320,80 @@ class EvalJob:
...
     for entry in entries:
         variants = (
             discover_recipe_variants(recipes_dir, entry.hf_id, entry.task)
             if recipes_dir is not None
             else []
         )
+        if not npu:
+            # Off-NPU still uses recipes for their eval config, but drops
+            # quantized variants -- quantization is an NPU-only concern here.
+            variants = [v for v in variants if not _is_quantized_precision(v.precision)]
```
DingmaomaoBJTU, line 2398: Off-NPU, a model whose recipe has only quantized variants (e.g. a `w8a16` recipe with no `fp16`) is filtered to an empty list and falls through to the `winml config` fallback (`variant=None` → `recipe_meta=None`). That silently drops the recipe's eval/dataset config for accuracy runs — which is the very reason the PR now consults recipes off-NPU. Likely rare (most recipes ship an `fp16` variant), but worth confirming it's acceptable that such models lose their authored accuracy config on CPU/GPU rather than, say, warning.

scripts/e2e_eval/run_eval.py
```diff
@@ -2301,34 +2320,80 @@ class EvalJob:
...
+    fallback_precision: str | None = None
 
     @property
     def precision(self) -> str | None:
-        """Recipe precision (e.g. ``fp16``/``w8a16``), or None for the fallback."""
-        return self.variant.precision if self.variant is not None else None
+        """Recipe precision, else the pinned fallback precision, else None."""
+        if self.variant is not None:
+            return self.variant.precision
+        return self.fallback_precision
```
DingmaomaoBJTU: Minor / pre-existing, but the PR reworks this property so worth aligning: for a recipe-less NPU model with an explicit per-model precision (e.g. `fp16`), `fallback_precision` is `None`, so this returns `None`. Yet `_build_for_job` builds with `job.fallback_precision or job.entry.precision` = `entry.precision`. The result-dir slug, display label, and `build_eval_result(precision=…)` then record no precision even though the artifact is actually `fp16`. Returning `self.fallback_precision or self.entry.precision` would keep the label/slug/recorded precision in sync with what's built (note: that would add a precision suffix to those models' output dirs, so it's a deliberate behavior change).

scripts/e2e_eval/run_eval.py
KayMKM: Good catch — fixed in 1f3fc30c. `_build_jobs` now takes `ep` and skips the recipe-less NPU `w8a8`+`w8a16` expansion when `_should_skip_winml_quant(ep)` is true, emitting a single `winml config` fallback instead. Added `test_npu_skip_quant_ep_no_recipe_single_fallback` covering `--device npu --ep vitisai`, and noted the behavior in the README.

scripts/e2e_eval/run_eval.py
KayMKM: Fixed in 1f3fc30c. `EvalJob.precision` now returns `self.fallback_precision or self.entry.precision`, so a recipe-less job with an explicit per-model precision keeps its slug/label/recorded precision in sync with the built artifact. This does add a precision suffix to those models' output dirs, as you noted — intentional. Updated `test_npu_no_recipe_honors_explicit_precision` accordingly.

scripts/e2e_eval/run_eval.py
KayMKM: Confirmed it's acceptable for now. I scanned all recipes under `examples/recipes`: across 82 `(model, task)` groups, none is quantized-only — every group ships a non-quantized (`fp16`/`fp32`) variant, so no model loses its authored eval/dataset config off-NPU today. Leaving the current fall-through behavior as-is rather than adding a warning for a case that can't currently trigger; happy to revisit if a quantized-only recipe lands later.

DingmaomaoBJTU (approved):

PR #1167, opened 2026-07-22
feat(optim): add Conv channel-affine folding

xieofxie:
## Summary

Add an opt-in `conv-channel-affine-folding` algebraic rewrite that folds safe scalar or channel-wise constant `Mul`/`Add` operations into Conv weights and bias.

The rewrite supports direct affine chains and channel routing through static `Split`/`Slice` and channel-preserving `Reshape`/`Squeeze`/`Unsqueeze` operations. It conservatively rejects dynamic shapes and constants, shared or public intermediate tensors, overlapping channel ranges, unsupported broadcasts, and non-floating Conv parameters. Multiple independent matches are handled, generated parameters use collision-safe names, dead constants and replaced initializers are pruned, and repeated optimization is idempotent.

BatchNormalization folding from the source change is intentionally excluded.

## Why this is not a pattern rewrite

The current pattern rewriter replaces a fixed, removable source subgraph with a fixed target topology whose source and target schemas must match. Conv channel-affine folding requires different behavior:

- It computes replacement Conv weights and bias from matched initializers and affine values.
- Routed matches can span a variable number of `Split`, `Slice`, and shape-only view branches.
- Affine updates from disjoint branches must be aggregated by Conv output-channel interval before changing the Conv.
- Routing nodes may need to remain in the graph while only terminal `Mul`/`Add` nodes are removed and router outputs are rewired.
- Shared or externally visible intermediates must make only the affected candidate ineligible rather than treating the entire routed topology as a removable pattern.

These requirements are match-parameterized parameter surgery over partially retained topology, not fixed subgraph substitution. Supporting them as a pattern would require broader framework capabilities for mutable initializer transforms, variable topology, partial-match retention, and cross-branch aggregation. Therefore this remains a focused algebraic rewrite.

### DIFF
--- a/src/winml/modelkit/optim/pipes/algebraic.py
+++ b/src/winml/modelkit/optim/pipes/algebraic.py
@@ -266,6 +266,17 @@ def _prune_unused_initializers(model: onnx.ModelProto) -> None:
     model.graph.initializer.extend(remaining)
 
 
+def _prune_stale_value_info(model: onnx.ModelProto) -> None:
+    """Remove value metadata for tensors no longer defined by the graph."""
+    graph = model.graph
+    defined = {value.name for value in graph.input if value.name}
+    defined.update(initializer.name for initializer in graph.initializer)
+    defined.update(output for node in graph.node for output in node.output if output)
+    remaining = [value for value in graph.value_info if value.name in defined]
+    del graph.value_info[:]
+    graph.value_info.extend(remaining)
+
+
 def _prune_generated_slices(model: onnx.ModelProto, introduced: set[str]) -> None:
     """Remove generated Slice nodes whose outputs are entirely dead."""
     while True:
@@ -486,12 +497,13 @@ def _collect_affine_chain(
     output_shape: tuple[int, ...],
     start: int,
     end: int,
+    calculation_dtype: np.dtype[Any],
 ) -> _AffineCandidate | None:
     """Collect a safe consecutive Mul/Add chain from one routed branch."""
     current = first
     current_input = source_name
-    scale = np.ones(end - start, dtype=np.float32)
-    offset = np.zeros(end - start, dtype=np.float32)
+    scale = np.ones(end - start, dtype=calculation_dtype)
+    offset = np.zeros(end - start, dtype=calculation_dtype)
     matched: list[onnx.NodeProto] = []
 
     while current.op_type in {"Mul", "Add"}:
@@ -503,7 +515,7 @@ def _collect_affine_chain(
         values = _affine_operand(index, current, current_input, output_shape, end - start)
         if values is None:
             return None
-        values = values.astype(np.result_type(values.dtype, np.float32), copy=False)
+        values = values.astype(calculation_dtype, copy=False)
         if current.op_type == "Mul":
             scale *= values
             offset *= values
@@ -543,6 +555,7 @@ def _collect_routed_affine_candidates(
     source_output_index: int,
     start: int,
     end: int,
+    calculation_dtype: np.dtype[Any],
 ) -> list[_AffineCandidate]:
     """Collect affine leaves below safe views and disjoint channel slices."""
     if source_output_index >= len(source_node.output):
@@ -587,6 +600,7 @@ def _collect_routed_affine_candidates(
             current_shape,
             start,
             end,
+            calculation_dtype,
         )
         if candidate is None:
             return []
@@ -618,6 +632,7 @@ def _collect_routed_affine_candidates(
                 0,
                 start + local_start,
                 start + local_end,
+                calculation_dtype,
             )
         )
     return candidates
@@ -716,6 +731,7 @@ def _fold_channel_affine(
         weight_dtype = onnx.numpy_helper.to_array(weight_initializer).dtype
         if channels <= 0:
             continue
+        calculation_dtype = np.result_type(weight_dtype, np.float32)
 
         route_name = conv_output
         route_shape = conv_shape
@@ -750,6 +766,7 @@ def _fold_channel_affine(
                 route_shape,
                 0,
                 channels,
+                calculation_dtype,
             )
             if direct is not None:
                 direct.source_node = route_source_node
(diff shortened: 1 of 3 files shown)


xieofxie:
### Simple example

Assume a Conv produces 10. The graph then multiplies it by 2 and adds 3:

```text
Conv result: 10
Multiply:    10 * 2 = 20
Add:         20 + 3 = 23
```

Instead, ModelKit changes the Conv's stored numbers:

```text
new Conv weight = old Conv weight * 2
new Conv bias   = old Conv bias * 2 + 3
```

The modified Conv directly produces 23, so the separate `Mul` and `Add` are no
longer needed.

src/winml/modelkit/optim/pipes/algebraic.py
```diff
@@ -313,6 +342,477 @@ def _split_boundaries(
...
+    source_name: str,
+    output_shape: tuple[int, ...],
+    start: int,
+    end: int,
+) -> _AffineCandidate | None:
+    """Collect a safe consecutive Mul/Add chain from one routed branch."""
+    current = first
+    current_input = source_name
+    scale = np.ones(end - start, dtype=np.float32)
+    offset = np.zeros(end - start, dtype=np.float32)
```
DingmaomaoBJTU: These accumulators are hard-coded to `float32`, but the rest of the fold respects the weight dtype (`calculation_dtype = np.result_type(weight_dtype, np.float32)` in `_fold_channel_affine`). For a float64 Conv the affine values get truncated to float32 here via the in-place `*=`/`+=`, so the folded weight loses precision.

Repro: a float64 `scale = 1 + 2**-30` folds into a weight of exactly `1.0` (~9.3e-10 error). Consider sizing `scale`/`offset` from the operand/weight dtype instead of hard-coding float32, and adding a float64/float16 dtype test.

src/winml/modelkit/optim/pipes/algebraic.py
```diff
@@ -313,6 +342,477 @@ def _split_boundaries(
...
+        offset = np.zeros(channels, dtype=calculation_dtype)
+        for candidate in candidates:
+            scale[candidate.start : candidate.end] = candidate.scale
+            offset[candidate.start : candidate.end] = candidate.offset
+        if not _copy_conv_parameters(model, allocator, conv, scale, offset):
+            continue
+
+        removed = {id(node) for candidate in candidates for node in candidate.nodes}
+        for candidate in candidates:
+            candidate.source_node.output[candidate.source_output_index] = candidate.final_output
```
DingmaomaoBJTU, line 828: Renaming the source output here (and removing the chain nodes) leaves the old tensor names' `value_info` entries dangling — e.g. folding `Conv→Mul` leaves a stale `value_info` for the old `conv_out`. The checker tolerates it and it matches the existing split-to-slice behavior, but since this pass otherwise prunes initializers/constants/slices carefully, it'd be consistent to also drop `value_info` for names that are no longer produced.

src/winml/modelkit/optim/pipes/algebraic.py
```diff
@@ -313,6 +342,477 @@ def _split_boundaries(
...
+        _remove_nodes(model, removed)
+        index = _GraphIndex.build(model)
```
DingmaomaoBJTU, line 830: This rebuilds the entire graph index after every folded Conv — O(folds × graph size). It's required for correctness (a later Conv can consume a renamed output in a `Conv→Mul→Conv` chain), so it can't simply be hoisted out of the loop; but on large graphs with many foldable Convs it could get costly. Incremental index maintenance would avoid the repeated full rebuilds.

DingmaomaoBJTU (commented):
Reviewed the Conv channel-affine folding. The core algebra is correct and nicely conservative — I checked direct chains, Split/Slice routing, channel-preserving views, nested slices, partial (single-branch) folds, shared-weight Convs, multi-consumer terminals, and chained Conv→Mul→Conv, and they all stay numerically exact; the 17 tests pass. A few non-blocking notes below.

src/winml/modelkit/optim/pipes/algebraic.py
xieofxie: The Conv weight-derived calculation dtype is now propagated through direct and routed affine collection, preventing float64 operands from being truncated. Added a float64 precision regression test.

src/winml/modelkit/optim/pipes/algebraic.py
xieofxie: Added post-rewrite pruning for `value_info` entries whose tensors are no longer defined, with the direct affine test verifying stale intermediate metadata is removed.

src/winml/modelkit/optim/pipes/algebraic.py
xieofxie: Agreed that rebuilding has an O(folds Ã— graph size) cost. I am retaining it here because each rewrite changes producers, consumers, initializers, and names used by later candidates; incremental maintenance would add substantial synchronization risk and is better handled as a separately measured optimization.

DingmaomaoBJTU (commented):
Re-reviewed `4209ab2` (fix: preserve affine precision and metadata). Both earlier points are resolved and verified:

- **Affine precision** — `calculation_dtype = np.result_type(weight_dtype, np.float32)` is now threaded through `_collect_affine_chain` / `_collect_routed_affine_candidates`, so the accumulators no longer truncate to float32. My repro (float64 `scale = 1 + 2**-30`) now folds to `1.0000000009313226` exactly instead of `1.0`, and `test_float64_affine_values_preserve_weight_precision` locks it in. float16 still accumulates in float32 as expected.
- **Stale value_info** — `_prune_stale_value_info` cleanly drops metadata for tensors no longer produced (only removes names absent from inputs/initializers/node outputs), and the new `assert not transformed.graph.value_info` guards it.

All 18 algebraic tests pass and my direct/routed/view/nested-slice/shared-weight/chained probes stay numerically exact. The one remaining note (the per-fold `_GraphIndex.build`) is a low-priority perf item and safe to leave. LGTM from my side.

DingmaomaoBJTU (approved):

PR #1168, opened 2026-07-22
Add --duration flag to perf for time-budgeted benchmarking

xieofxie:
## Summary

Adds a `--duration <seconds>` flag to `winml perf`. When set, the benchmark runs for a wall-clock time budget instead of a fixed `--iterations` count. After warmup, inference loops until the duration elapses (always running at least one benchmark iteration so stats are never empty).

This is ideal with `--monitor`, whose PDH hardware counters need time to emit real utilization data — a small fixed iteration count often finishes before the counters produce meaningful samples.

```bash
winml perf -m model.onnx --device npu --monitor --duration 15
```

## Details

- **New `--duration FLOAT` option** (must be `> 0`), plumbed through `BenchmarkConfig.duration`.
- **`_benchmark_indices()` helper** centralizes the loop: warmup first, then either a fixed iteration count (default) or a timed loop. Used by the single-model, monitored, and per-module paths.
- **Mutually exclusive with `--op-tracing`** (op-tracing runs its own fixed, small iteration count) — rejected with a clear `UsageError`.
- **Live `--monitor` chart** shows elapsed/total time progress (`Time: 12.3/30s`) in duration mode instead of an iteration counter.
- **Accurate iteration reporting**: in duration mode the reported iteration count reflects the actual number of timed samples rather than the unused `--iterations` default. For per-module runs (where each module runs a different count under a shared budget), the count is recorded per instance and the top-level value is `null`; both single-model and module reports also record `duration_sec`.
- `--duration` is added to the genai "ignored flags" warning (genai benchmarks by token generation).

## Tests

- `--duration` help/forwarding/default, `--op-tracing` conflict, non-positive rejection.
- `_benchmark_indices` iteration mode, timed budget, and at-least-one-iteration guarantee (deterministic via a faked clock).
- `LiveMonitorDisplay` time-based vs iteration-based progress.
- `benchmark_info.iterations` reports configured count without duration and actual sample count with duration.

All perf suites pass (`test_perf_cli`, `test_perf_module`, `test_perf_composite`, `test_ep_monitor`): 200 passed, 2 NPU-only skips; `ruff` clean.


xieofxie:
Needed for https://github.com/microsoft/winml-cli/pull/1149

src/winml/modelkit/commands/perf.py
```diff
@@ -1747,9 +1818,16 @@ def _run_simple_loop(
...
-    for i in range(total_iterations):
+    """Run the benchmark iteration loop with periodic debug logging.
+
+    When ``duration_sec`` is set, the benchmark phase (after ``warmup``) runs
+    until the wall-clock duration elapses instead of a fixed iteration count.
+    """
+    for i in _benchmark_indices(total_iterations, warmup, duration_sec):
         session.run(inputs)
 
         if (i + 1) % max(1, total_iterations // 10) == 0:
```
DingmaomaoBJTU: In `--duration` mode `i` keeps counting past `total_iterations` (the loop is time-bounded, not count-bounded), so this progress log can read e.g. `Progress: 350/110`. The `total_iterations` denominator here — and on the `logger.debug` line just below — is meaningless once the run is timed. Minor since it's debug-only, but might be worth switching to an elapsed/duration readout in duration mode.

src/winml/modelkit/commands/_live_chart.py
```diff
@@ -95,6 +102,15 @@ def update(
...
             return
 
+        # Stamp the start of the timed benchmark phase so duration-based progress
+        # is measured from the first post-warmup iteration.
+        if (
+            self._duration_sec is not None
+            and self._bench_start is None
+            and iteration > self._warmup
+        ):
+            self._bench_start = time.perf_counter()
```
DingmaomaoBJTU: This stamps `_bench_start` on the first post-warmup `update()`, which runs *after* that iteration's `session.run()`. But `_benchmark_indices` starts its own budget clock *before* that run (`perf.py` L1771) and the loop terminates on the generator's clock. Net effect: this display clock lags by ~one inference, so `Time: x/Ns` tends to finish a bit under 100%. Purely cosmetic — if you want the bar to land on 100%, share the generator's start timestamp with the display instead of re-stamping here.

DingmaomaoBJTU (commented):
Took a pass through this — the overall approach looks good. The warmup→timed-loop split, the at-least-one-iteration guarantee, the `--op-tracing` conflict, and the genai ignored-flag warning are all correct, and the tests cover the important paths.

A few minor nits (none blocking):

**`perf.py` L732-734** — `logger.info("Running benchmark: %d iterations + %d warmup", ...)` still logs the fixed `iterations` count in `--duration` mode, so it reads e.g. *"Running benchmark: 100 iterations + 10 warmup"* for a run that's actually time-bounded. Might be worth branching to log the time budget there.

Two more inline below.

src/winml/modelkit/commands/perf.py
xieofxie: Good catch. In duration mode _run_simple_loop now logs an elapsed/total-seconds readout (Progress: %.1f/%.0fs) driven by the shared benchmark clock, instead of the meaningless i/total_iterations count. Iteration mode keeps the original count log. Fixed in 6b3f7448.

src/winml/modelkit/commands/_live_chart.py
xieofxie: Fixed in 6b3f7448. Introduced a shared _BenchmarkClock that _benchmark_indices stamps the instant warmup ends (the same instant its budget clock starts). The loop passes that clock to LiveMonitorDisplay, which now reads clock.start instead of self-stamping on the first post-warmup update(), so the "Time: x/Ns" bar tracks the exact budget the loop terminates on.

xieofxie:
Thanks for the review @DingmaomaoBJTU. Addressed all three nits in 6b3f7448:

1. **Info log** — the "[3] Run benchmark" log now reports `%gs duration + %d warmup` in duration mode instead of the iteration count.
2. **`_run_simple_loop` debug progress** — logs an elapsed/total-seconds readout in duration mode instead of the meaningless `i/total_iterations` count (resolved inline).
3. **Live chart clock** — added a shared `_BenchmarkClock` stamped by `_benchmark_indices` at the warmup boundary; the display reads it instead of self-stamping, so the time bar tracks the exact budget the loop stops on (resolved inline).

src/winml/modelkit/commands/perf.py
```diff
@@ -1747,13 +1849,35 @@
...
+            if (i + 1) % max(1, total_iterations // 10) == 0:
+                logger.debug("Progress: %d/%d", i + 1, total_iterations)
+        return
 
-        if (i + 1) % max(1, total_iterations // 10) == 0:
-            logger.debug("Progress: %d/%d", i + 1, total_iterations)
+    clock = _BenchmarkClock()
+    next_log = 0.0
+    log_step = max(duration_sec / 10.0, 0.1)
+    for _i in _benchmark_indices(total_iterations, warmup, duration_sec, clock):
```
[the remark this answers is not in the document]
xieofxie: Fixed in 6519b1ea. The duration-mode loop does not use the yielded index, so I renamed the loop variable to the throwaway _.

src/winml/modelkit/commands/perf.py
```diff
@@ -2129,6 +2254,14 @@ def _resolve_composite_components_for_perf(model: str, task: str | None) -> dict
     show_default=True,
     help="Number of warmup iterations (excluded from statistics; must be >= 0)",
 )
+@click.option(
+    "--duration",
+    type=click.FloatRange(min=0, min_open=True),
```
DingmaomaoBJTU, line 2494: Click `FloatRange` accepts non-finite values because comparisons such as `nan <= 0` are false. That value reaches `_benchmark_indices`, where `elapsed >= duration_sec` is permanently false, so `winml perf --duration nan` never terminates; `inf` likewise creates an unbounded run. Please reject non-finite durations (for example with `math.isfinite`) and add a CLI regression test.

src/winml/modelkit/commands/perf.py
```diff
@@ -2346,6 +2480,15 @@ def perf(
     # the inference session for both HF model IDs and ONNX file inputs.
     ep_provider_options = cli_utils.parse_ep_options(ep_options)
 
+    # --duration replaces the fixed iteration count with a wall-clock budget.
+    # Op-tracing runs its own fixed, small iteration count, so the two are
+    # mutually exclusive.
+    if duration is not None and op_tracing:
```
DingmaomaoBJTU: This validation runs before the `winml-genai` dispatch, but both `--duration` and `--op-tracing` are registered in `_GENAI_IGNORED_FLAGS`. Passing either alone warns and continues, while passing both now fails before the ignored-flag warning. Please move the conflict check to the WinML path (after the GenAI early return), or explicitly exclude the GenAI runtime, so ignored options remain consistently non-fatal.

DingmaomaoBJTU (changes requested):
Found one termination bug and one runtime-dispatch inconsistency in the new flag handling; details inline.

src/winml/modelkit/commands/perf.py
xieofxie: Good catch. FloatRange lets nan/inf through (nan <= 0 and inf <= 0 are both false), so I added a Click callback that rejects any non-finite --duration up front, plus a parametrized CLI regression test (nan/inf). Fixed in f62cb259.

src/winml/modelkit/commands/perf.py
xieofxie: Agreed. Moved the --duration/--op-tracing conflict check into the WinML path (after the genai early return) so both flags stay non-fatal (ignored) for winml-genai, and added a genai dispatch test asserting it warns rather than aborting. Fixed in f62cb259.

src/winml/modelkit/commands/_live_chart.py
```diff
@@ -315,15 +326,26 @@ def _render_status(
...
+            # benchmark iteration count is not known ahead of time. The start
+            # reference is the loop's shared clock, so this tracks the same
+            # budget the loop terminates on.
+            start = getattr(self._clock, "start", None)
+            elapsed = time.perf_counter() - start if start else 0.0
+            pct = min(elapsed / self._duration_sec, 1.0) if self._duration_sec > 0 else 0.0
+            shown = min(elapsed, self._duration_sec)
+            progress = f"[green]Time: {shown:.1f}/{self._duration_sec:.0f}s[/green]"
+        else:
+            pct = iteration / self._total if self._total > 0 else 0
```
DingmaomaoBJTU: Bug: When `duration_sec` is set and we're in the warmup phase, this `else` branch computes `pct = iteration / self._total`. But `self._total = warmup + iterations` where `iterations` is the meaningless default 100 in duration mode. So warmup iter 5/10 shows as ~5% instead of 50%.

Suggest splitting warmup out first:
```python
if phase == "warmup":
    pct = iteration / self._warmup if self._warmup > 0 else 0
    progress = f"[yellow]Warmup: {iteration}/{self._warmup}[/yellow]"
elif self._duration_sec is not None:
    # duration-mode benchmark progress ...
else:
    # iteration-mode benchmark progress ...
```

src/winml/modelkit/commands/perf.py
```diff
@@ -1914,17 +2017,23 @@ def _run_monitored_loop(
...
     display = LiveMonitorDisplay(
         total_iterations=total_iterations,
         warmup=warmup,
         model_id=model_id,
         device=device,
         device_kind=getattr(hw, "device_kind", None),
+        duration_sec=duration_sec,
+        clock=clock,
     )
     with display:
```
DingmaomaoBJTU, line 2035: In duration mode, `total_iterations = warmup + self.config.iterations` is passed to `LiveMonitorDisplay` as `self._total`. Since warmup progress in `_render_status` uses `iteration / self._total`, the warmup bar will be wrong (denominator includes the unused 100 default iterations). This is the other side of the bug noted in `_live_chart.py`.

src/winml/modelkit/commands/perf.py
```diff
@@ -1904,6 +1955,58 @@ def _print_save_to_footer(
...
+    if duration_sec is None:
+        while i < total_iterations:
+            yield i
+            i += 1
+        return
+    while True:
+        yield i
+        i += 1
+        if time.perf_counter() - start >= duration_sec:
+            return
```
DingmaomaoBJTU, line 2007: Nit: `_benchmark_indices` checks the budget **after** yielding, so actual wall-clock time can exceed `--duration` by one inference latency. The help text says "Run the benchmark for this many seconds" which could be read as strict. Consider clarifying it's a minimum budget.

src/winml/modelkit/commands/perf.py
```diff
@@ -1946,13 +2055,35 @@ def _run_simple_loop(
...
+    When ``duration_sec`` is set, the benchmark phase (after ``warmup``) runs
+    until the wall-clock duration elapses instead of a fixed iteration count,
+    and progress is logged as elapsed/total time (the iteration count is
+    unbounded and its ``total_iterations`` denominator is meaningless).
+    """
+    if duration_sec is None:
+        for i in _benchmark_indices(total_iterations, warmup, duration_sec):
+            session.run(inputs)
+            if (i + 1) % max(1, total_iterations // 10) == 0:
+                logger.debug("Progress: %d/%d", i + 1, total_iterations)
```
DingmaomaoBJTU: Nit: Now that the refactor makes the warmup boundary explicit via `_benchmark_indices`, this progress log still uses `total_iterations` (warmup + bench) as denominator. Not a regression, but a follow-up could log benchmark-only progress as `(i - warmup + 1) / (total_iterations - warmup)`.

src/winml/modelkit/commands/perf.py
```diff
@@ -2263,6 +2395,21 @@ def _resolve_composite_components_for_perf(model: str, task: str | None) -> dict
         raise click.ClickException(f"Composite model detection failed unexpectedly: {e}") from e
 
 
+def _validate_duration(
+    ctx: click.Context, param: click.Parameter, value: float | None
+) -> float | None:
+    """Reject non-finite ``--duration`` values.
+
+    ``click.FloatRange`` lets ``nan``/``inf`` through because ``nan <= 0`` and
+    ``inf <= 0`` are both false, but such a budget never terminates the timed
```
DingmaomaoBJTU: Nit: The docstring says `nan <= 0` and `inf <= 0` are both false. Correct for `nan` and `+inf`, but `float('-inf') <= 0` is True so FloatRange rejects it on its own. Consider saying "+inf" explicitly.

src/winml/modelkit/commands/perf.py
```diff
@@ -1984,6 +2115,7 @@ def _run_simple_loop(
     "memory": "--memory",
     "op_tracing": "--op-tracing",
     "batch_size": "--batch-size",
+    "duration": "--duration",
```
DingmaomaoBJTU: Nit: Consider placing `duration` near the other benchmark-control flags (`batch_size`, etc.) rather than at the end, for discoverability as this dict grows.

DingmaomaoBJTU (commented):
Overall well-structured feature with good test coverage. 1 bug (warmup progress bar in duration mode) and several minor suggestions - see inline comments.

src/winml/modelkit/commands/_live_chart.py
xieofxie: Good catch, that's a real bug. Split warmup into its own branch in _render_status so it scales by the warmup count (5/10 -> 50%) instead of total_iterations, and made the iteration-mode benchmark bar match its label. Added a regression test (test_duration_mode_warmup_progress_scales_by_warmup). Fixed in 0fe917ec.

src/winml/modelkit/commands/perf.py
xieofxie: Right, same bug from the caller side. Fixed in _render_status (0fe917ec) so the warmup bar no longer uses total_iterations as its denominator; total_iterations still feeds the benchmark-phase math where it's meaningful. Regression test added.

src/winml/modelkit/commands/perf.py
xieofxie: Agreed. Reworded the help text to 'Run the benchmark for at least this many seconds ... it is a minimum budget, so the final inference may overrun it slightly.' Fixed in 0fe917ec.

src/winml/modelkit/commands/perf.py
xieofxie: Applied it. The simple-loop iteration branch now logs benchmark-only progress (bench_done / bench_total, warmup excluded). Fixed in 0fe917ec.

src/winml/modelkit/commands/perf.py
xieofxie: Fixed the docstring to say +inf explicitly and note that -inf is already rejected by FloatRange (since -inf <= 0). 0fe917ec.

src/winml/modelkit/commands/perf.py
xieofxie: Regrouped: batch_size and duration (run-size controls) now sit together ahead of the monitor/memory/op-tracing flags, so duration is no longer tacked on at the very end. 0fe917ec.

DingmaomaoBJTU (approved):
LGTM! All review comments addressed cleanly:
- Warmup progress bar bug fixed with correct branch restructuring
- Progress log now excludes warmup from denominator
- Help text clarifies minimum-budget semantics
- Docstring accurately describes +inf/-inf behavior
- Dict entries regrouped logically
- Regression test added for the warmup bar fix

Nice work!

PR #1171, opened 2026-07-23
feat(optim): fold Conv Add BatchNormalization pattern

xieofxie:
## Summary

- add source patterns for both operand orders of `Add(Conv(...), static)` followed by inference `BatchNormalization`
- fold BatchNormalization parameters into copied Conv weights/bias and the static Add tensor
- support Conv with or without an existing bias while preserving Conv attributes
- reject training mode, dynamic parameters, invalid variance, and non-removable/shared intermediates
- register the opt-in `conv-add-batch-normalization-folding` rewrite capability

## Why this uses the pattern framework

Unlike the static Split-to-Slice rewrite and channel-affine folding, this transformation has fixed source and target topology. It always replaces `Conv -> Add -> BatchNormalization` with `Conv -> Add`, has one output, and removes the complete matched subgraph without topology-dependent node generation or graph-wide route aggregation. The pattern framework can therefore express removability, support both commutative Add input orders as source variants, and share one target implementation. Match-specific constants and optional Conv bias are validated and captured by the source pattern before the target generates the folded parameters.


src/winml/modelkit/pattern/conv_batchnorm_patterns.py
```diff
@@ -0,0 +1,349 @@
...
+# Licensed under the MIT License.
+# --------------------------------------------------------------------------
+"""Patterns for folding inference BatchNormalization after Conv and Add."""
+
+from __future__ import annotations
+
+from typing import TYPE_CHECKING, Any
+
+import numpy as np
+import onnx
```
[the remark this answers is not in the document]
xieofxie: Updated the source module to import `ModelProto` explicitly alongside the existing ONNX symbols, removing the mixed `import onnx` / `from onnx import ...` usage.

tests/unit/pattern/test_conv_batchnorm_patterns.py
```diff
@@ -0,0 +1,161 @@
+# -------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+# --------------------------------------------------------------------------
+"""Tests for Conv/Add/BatchNormalization folding patterns."""
+
+from __future__ import annotations
+
+import numpy as np
+import onnx
```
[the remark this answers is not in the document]
xieofxie: Updated the test module to import `ModelProto`, `ValueInfoProto`, and `checker` explicitly, removing the mixed ONNX import style.

xieofxie:
### Simple example

Assume:

```text
Conv produces:       10
Static tensor adds:   2
BatchNormalization:  halve the value, then subtract 1
```

The original graph calculates:

```text
(10 + 2) * 0.5 - 1 = 5
```

ModelKit moves the fixed BatchNormalization work into the earlier operations:

```text
Modified Conv result: 10 * 0.5 - 1 = 4
Modified static value: 2 * 0.5     = 1
Final Add:             4 + 1       = 5
```

Both paths produce 5.

src/winml/modelkit/pattern/conv_batchnorm_patterns.py
```diff
@@ -0,0 +1,348 @@
...
+        if epsilon < 0 or np.any(denominator <= 0) or not np.all(np.isfinite(denominator)):
+            return None
+        gamma = scale / np.sqrt(denominator)
+        if not np.all(np.isfinite(gamma)):
+            return None
+
+        scaled_add = _scale_broadcast_tensor(add_tensor, output_shape, gamma)
+        if scaled_add is None:
+            return None
+        scaled_weight = weight * gamma.reshape((channels,) + (1,) * (weight.ndim - 1))
```
DingmaomaoBJTU: Only `gamma` is checked for finiteness, but the dtype-preserving fold can still overflow. With float16 `W=40000`, `X=0.25`, `scale=2`, `var=1`, and `epsilon=0`, the original graph produces finite `20000`, while this multiplication creates an `inf` weight and the rewritten graph returns `inf`. Please reject the match when any generated weight, bias, or scaled Add constant is non-finite.

src/winml/modelkit/pattern/conv_batchnorm_patterns.py
```diff
@@ -0,0 +1,348 @@
...
+            "folded_weight",
+            "folded_bias",
+            "scaled_add",
+        )
+        if any(name not in attributes for name in required):
+            raise PatternMismatchedError("Missing folded Conv/Add attributes")
+        if input_names is None or output_names is None:
+            raise PatternMismatchedError("Input and output names are required")
+
+        weight_name = f"{prefix}weight"
```
DingmaomaoBJTU, line 319: `prefix` is unique only among rewrites, not against tensor names already in the model. If the graph already has `Rewrite_FoldedConvAddPattern_0_weight`, `PatternRewriter` skips appending this new initializer and the Conv silently consumes the existing value; a checker-valid probe produced different outputs. Please allocate generated tensor names against all names in the parent graph, or remap them before insertion.

src/winml/modelkit/pattern/conv_batchnorm_patterns.py
```diff
@@ -0,0 +1,348 @@
...
+        return None
+
+    expanded_shape = list(padded_shape)
+    expanded_shape[1] = output_shape[1]
+    try:
+        broadcast_values = np.broadcast_to(values.reshape(padded_shape), tuple(expanded_shape))
+    except ValueError:
+        return None
+    factors = gamma.reshape((1, len(gamma)) + (1,) * (len(output_shape) - 2))
+    return np.asarray(broadcast_values * factors, dtype=values.dtype)
```
DingmaomaoBJTU: `broadcast_to` is a view, but this multiplication materializes the full expanded array. For output `[1,512,1024,1024]` and static Add input `[1,1,1024,1024]`, matching allocates a 2-GiB array and rewriting embeds it as an initializer; this pattern is enabled for default analysis. Please reject or cap expansions that would create excessively large constants.

src/winml/modelkit/pattern/conv_batchnorm_patterns.py
xieofxie: Added a final finiteness check after casting the folded weight, bias, and Add constants to their stored dtype. The rewrite now leaves the original graph unchanged when any generated constant overflows.

src/winml/modelkit/pattern/conv_batchnorm_patterns.py
xieofxie: Updated `PatternRewriter` to allocate unique names for all generated internal tensors and remap node references consistently. Schema boundary names remain unchanged, and a collision regression test verifies numerical equivalence.

src/winml/modelkit/pattern/conv_batchnorm_patterns.py
xieofxie: Added a preflight size check before broadcast materialization, using the repository's 100 MiB external-data threshold. Matches that would generate an oversized Add initializer are now rejected without allocating the expanded array.

src/winml/modelkit/pattern/conv_batchnorm_patterns.py
```diff
@@ -0,0 +1,363 @@
...
+    gamma: np.ndarray,
+) -> np.ndarray | None:
+    """Scale a broadcastable Add operand along the Conv channel axis."""
+    if values.ndim > len(output_shape):
+        return None
+    padded_shape = (1,) * (len(output_shape) - values.ndim) + tuple(values.shape)
+    if any(dimension not in (1, output_shape[axis]) for axis, dimension in enumerate(padded_shape)):
+        return None
+    if padded_shape[1] not in (1, output_shape[1]):
+        return None
```
DingmaomaoBJTU: This check appears redundant — the `any(...)` guard two lines above already validates every dimension (including index 1) against `(1, output_shape[axis])`. If `padded_shape[1]` failed this condition, the earlier check would have already returned `None`.

If this is intended as an explicit assertion that the channel axis is specifically validated, consider adding a brief comment to clarify the intent, or removing the duplicate check to simplify.

src/winml/modelkit/pattern/base.py
```diff
@@ -2130,9 +2183,8 @@ def _allocate_graph_node_key(node: Any) -> str:
 
                 # Append new initializers (constants) from the new subgraph
                 for initializer in new_subgraph_model.graph.initializer:
-                    # Check if initializer already exists (by name)
-                    existing_names = {init.name for init in graph.initializer}
-                    if initializer.name not in existing_names:
+                    # Schema-input constants already exist in the parent graph.
+                    if initializer.name not in boundary_names:
```
DingmaomaoBJTU, line 2187: Good improvement — the old code rebuilt `existing_names = {init.name for init in graph.initializer}` inside the loop on every iteration (O(n²)), and guarded against name collisions with existing graph initializers. The new approach relies on the invariant that `_allocate_graph_tensor_name` has already guaranteed uniqueness for all non-boundary tensor names, so only boundary names (which refer to tensors already in the parent graph) need to be excluded. This is both more correct and more efficient.

Worth noting: this invariant is subtle — if a future target pattern's `get_onnx_model` bypasses the prefix convention and emits initializer names that collide with existing graph initializers, the remapping step would catch and rename them before this point, so the guard remains safe.

src/winml/modelkit/pattern/conv_batchnorm_patterns.py
```diff
@@ -0,0 +1,363 @@
...
+        except (TypeError, ValueError):
+            return None
+
+        denominator = variance + epsilon
+        if epsilon < 0 or np.any(denominator <= 0) or not np.all(np.isfinite(denominator)):
+            return None
+        gamma = scale / np.sqrt(denominator)
+        if not np.all(np.isfinite(gamma)):
+            return None
+
```
DingmaomaoBJTU, line 216: The folded bias formula `bias * gamma + beta - gamma * mean` is mathematically correct:

`
BN(Conv(X,W,b) + A) = gamma * (Conv(X,W,b) + A - mean) + beta
                     = Conv(X, W*gamma, b*gamma + beta - gamma*mean) + gamma*A
`

where `gamma = scale / sqrt(var + eps)`. The separate `scaled_add = gamma * A` ensures the Add operand is correctly scaled. Verified that all intermediate and final results are checked for finiteness before committing the rewrite. 👍

DingmaomaoBJTU (approved):
LGTM - code quality is high, math is correct, tests are thorough. No significant issues found. Left a few inline observations.

src/winml/modelkit/pattern/conv_batchnorm_patterns.py
xieofxie: Agreed. I removed the redundant channel-axis check because the preceding dimension-wise validation already covers it.

src/winml/modelkit/pattern/base.py
xieofxie: Thanks for confirming. The remapping step guarantees uniqueness for every non-boundary target tensor before initializers are inserted, while matched schema boundaries intentionally retain their parent-graph names.

src/winml/modelkit/pattern/conv_batchnorm_patterns.py
xieofxie: Thanks for verifying the derivation and finiteness handling.

DingmaomaoBJTU (approved):
Latest commit addresses the redundant shape check feedback. LGTM.

PR #1209, opened 2026-07-24
Implement PPL evaluator for qwen 3 0.6B

zhenchaoni:
### Summary

Adds a `text-generation` eval task to `winml eval` that computes teacher-forced, disjoint fixed-length **perplexity (PPL)** for autoregressive causal LMs. The same evaluator scores both **onnxruntime-genai bundles** (ONNX, on-device) and **PyTorch fp32 baselines** through a single model-agnostic contract, so ONNX-vs-baseline accuracy regression can be tracked in the existing e2e harness.

Validated on **Qwen/Qwen3-0.6B** against wikitext-2-raw-v1 (test split):
- ONNX genai bundle PPL = **25.20** 
- PyTorch fp32 baseline PPL = **23.65**

### What's included

**Evaluator (`eval/text_generation_evaluator.py`)**
- `WinMLTextGenerationEvaluator`: model-agnostic, zero-branch. Builds a token corpus, splits it into non-overlapping `seqlen` blocks, and computes PPL via a numerically stable log-sum-exp NLL accumulator.
- Relies on a uniform scoring contract — `encode(text) -> list[int]` and `forward(ids) -> out` where `out.logits` is `(1, len(ids)-1, vocab)` with row *i* predicting token *i+1*. No per-model conditionals.

**Model layer (`models/winml/genai_causal_lm.py`)**
- `WinMLGenaiCausalLM`: causal-LM inference over an onnxruntime-genai bundle directory (loaded as-is, `compile=False` — building/compiling is `winml build`'s job).
- `HFCausalLM`: PyTorch fp32 adapter implementing the same contract (lazy torch/transformers import, `add_special_tokens=False` to reproduce POC tokenization). Lives in `src` so the baseline script and the evaluator share one code path.
- `CausalLMOutput` shared output dataclass.

**Wiring**
- `eval/evaluate.py`: registers the `text-generation` evaluator, a default wikitext-2-raw-v1 dataset, and `_load_genai_causal_lm` (validates the bundle dir has `genai_config.json` + `.onnx` files before loading).
- `commands/eval.py`: `_resolve_model_path` now routes an existing local **directory** in `-m` to `model_path` (so a genai bundle dir is read from disk instead of being treated as a Hub id).
- `utils/eval_utils.py`: `_TEXT_GENERATION_SCHEMA` with `input_column` (default `text`), `num_tokens` (default 8192), `seqlen` (default 2048).

**Baseline / accuracy harness**
- `run_pytorch_baseline.py`: text-generation branch uses the shared `HFCausalLM` adapter.
- `accuracy.py`: adds a `perplexity` compare strategy (`delta_relative`, warn 0.05 / fail 0.10, lower-is-better).
- `testsets/models_with_acc.json` + `cache/baseline_cache.json`: Qwen/Qwen3-0.6B entry with cached fp32 baseline (PPL 23.645425).

**Tests**
- `test_text_generation_evaluator.py` (14): block-chunking protocol, corpus loading, uniform-logits → PPL = vocab-size sanity, NLL cross-entropy reference.
- `test_genai_causal_lm.py` (9): `CausalLMOutput`, encode/forward contract, float32 conversion, last-row trimming.


src/winml/modelkit/commands/eval.py
```diff
@@ -533,6 +533,13 @@ def _resolve_model_path(
                 "for preprocessor and config resolution."
             )
         return value, model_id
+
+    # A local directory (e.g. an onnxruntime-genai bundle) is a model *path*,
+    # not a Hub model id. Route it to model_path so downstream loaders read the
+    # bundle from disk instead of treating the directory string as a Hub ref.
+    if Path(value).expanduser().is_dir():
+        return str(Path(value).expanduser()), model_id
```
DingmaomaoBJTU: This condition also matches local Hugging Face checkpoints, which were previously handled as `model_id` and passed to `from_pretrained()`. With `winml eval -m ./saved-hf-model --task image-classification`, `_resolve_model_path()` now returns a `model_path` with `model_id=None`, so `_load_model()` fails with `model_id is required`; supplying `--model-id` instead routes the directory to `from_onnx()`. Could we only classify a directory as a GenAI bundle after checking for `genai_config.json` (or gate this on `text-generation`)?

src/winml/modelkit/eval/evaluate.py
```diff
@@ -282,6 +295,51 @@ def _load_model(config: WinMLEvaluationConfig) -> WinMLPreTrainedModel | WinMLCo
...
+
+    bundle_dir = Path(bundle_path).expanduser()
+    if not bundle_dir.is_dir():
+        raise ValueError(f"Genai bundle directory not found: {bundle_dir}")
+    if not (bundle_dir / "genai_config.json").is_file():
+        raise ValueError(
+            f"'{bundle_dir}' is not a genai bundle: no genai_config.json found. "
+            "Point -m at a bundle built with 'winml build ... --device npu --ep qnn'."
+        )
+    if not any(bundle_dir.glob("*.onnx")):
```
DingmaomaoBJTU: This root-only glob rejects valid GenAI bundles whose config uses bundle-relative subpaths, for example `models/decoder.onnx`. `GenaiSession`/ORT GenAI resolve each stage's `filename` relative to the bundle directory, so the ONNX file does not have to live at the root. Please validate the filenames referenced by `genai_config.json` instead of requiring `bundle_dir/*.onnx`.

src/winml/modelkit/eval/text_generation_evaluator.py
```diff
@@ -0,0 +1,152 @@
...
+    def _load_corpus_tokens(self, num_tokens: int) -> list[int]:
+        """Concatenate the dataset text column and tokenize with the model.
+
+        Uses the model's own tokenizer (``model.encode``) so the token stream
+        matches the model under test exactly.
+        """
+        from datasets import load_dataset
+
+        from ..utils.eval_utils import get_default
+
```
DingmaomaoBJTU, line 157: By overriding `prepare_data()`, this loader bypasses the base evaluator's `streaming`, `shuffle`, and `samples` handling. `load_dataset()` is always eager here and the full `dataset[column]` is joined before token truncation, so `--streaming` and `--samples` have no effect and a large split can be fully downloaded/materialized just to score 8,192 tokens. Please honor those `DatasetConfig` controls and stop iterating once enough corpus tokens have been collected.

scripts/e2e_eval/run_pytorch_baseline.py
```diff
@@ -152,16 +152,28 @@ def _measure_pytorch_latency(task_evaluator: Any, warmup: int, iterations: int)
...
+
+    # text-generation resolves to a composite (embeddings/ctx/iter/lm_head) that
+    # is not a runnable HF model. Wrap a plain causal LM in the perplexity
+    # scoring contract so the evaluator scores it through the same path as a
+    # genai bundle.
+    if task == "text-generation":
+        from winml.modelkit.models.winml import HFCausalLM
+
+        _out(f"Loading causal LM baseline adapter for {model_id} on {device}")
+        return HFCausalLM(model_id, device)
```
DingmaomaoBJTU, line 168: `HFCausalLM` is paired with `WinMLTextGenerationEvaluator`, whose `data` items are token-ID lists and whose `pipe` is `None`. The existing `--perf-iterations` path calls `_extract_pipeline_input(task_evaluator.data[0], ...)` as if the item were a mapping, then invokes `task_evaluator.pipe(...)`, so text-generation baselines always fail when latency is requested. Please add a latency path that calls this adapter directly, or explicitly reject the option for this task.

src/winml/modelkit/models/winml/genai_causal_lm.py
```diff
@@ -0,0 +1,231 @@
...
+                target = input_ids[i]
+                # Pin the next input to the given token so the pass follows
+                # input_ids regardless of the model's own argmax.
+                if int(step.argmax()) != target:
+                    forced = np.full((1, 1, step.shape[0]), _FORCE_LOW, dtype=np.float32)
+                    forced[0, 0, target] = _FORCE_HIGH
+                    gen.set_logits(forced)
+                gen.generate_next_token()
+
+        return CausalLMOutput(logits=np.stack(rows)[None, ...])
```
DingmaomaoBJTU: Returning all per-token vocabulary logits makes the default evaluation extremely memory-heavy. With `seqlen=2048`, this stacks `(2047, vocab_size)` float32 values (about 1.16 GiB for Qwen3's 151,936-token vocabulary); `np.stack()` duplicates the row storage, and `_block_nll()` then casts the whole tensor to float64 and allocates another exponential buffer. This is likely to OOM on the target model. Since PPL only needs each target logit and log-sum-exp, can we accumulate NLL per decode step instead of retaining the full logits tensor?

DingmaomaoBJTU (commented):
Please see the inline comments.

src/winml/modelkit/commands/eval.py
zhenchaoni: Fixed. A directory is now only treated as a GenAI bundle when it contains genai_config.json; otherwise it stays a local HF checkpoint routed by model_id as before. Added regression tests for both cases.

src/winml/modelkit/eval/evaluate.py
zhenchaoni: Fixed. The bundle ONNX check now uses 
glob(\*.onnx\), so bundles whose genai_config.json references files in subdirectories (e.g. \models/decoder.onnx\) are accepted.

scripts/e2e_eval/run_pytorch_baseline.py
zhenchaoni: Fixed. \--perf-iterations\ is now explicitly rejected for text-generation, whose data items are token-ID lists (not pipeline mappings), so the latency path no longer fails on this task.

src/winml/modelkit/eval/text_generation_evaluator.py
zhenchaoni: Addressed. The loader no longer overrides `prepare_data` in a way that bypasses the DatasetConfig controls:

- **streaming**: now respects `DatasetConfig.streaming` (default no-streaming = download once + reuse the local cache; `--streaming` streams). I verified the in-order non-blank rows are byte-identical between the two modes, so the token stream and score are unchanged either way.
- **early-stop**: iteration is in dataset order and stops as soon as enough rows cover `num_tokens` (~83 rows for 8192 tokens), so a large split is not materialized just to score a few thousand tokens.
- **samples**: not applied here on purpose. Perplexity needs one contiguous corpus, so the budget is `num_tokens` (a token count), not N independent row samples; and we keep dataset order (no shuffle) because splicing shuffled passages would distort a teacher-forced score.

src/winml/modelkit/models/winml/genai_causal_lm.py
zhenchaoni: Fixed. `forward` no longer materializes the full `(seqlen-1, vocab)` logits tensor. It now yields one `(vocab,)` logit vector per position, and the evaluator accumulates NLL per step (log-sum-exp minus the target logit), casting only a single vector to float64 at a time. Peak memory drops from ~GiB per block to one vocab row. This is a term-by-term identical sum, so the perplexity score is unchanged. Dropped the now-unused `CausalLMOutput` wrapper.

src/winml/modelkit/eval/evaluate.py
```diff
@@ -282,6 +298,51 @@ def _load_model(config: WinMLEvaluationConfig) -> WinMLPreTrainedModel | WinMLCo
...
+            "Point -m at a bundle built with 'winml build ... --device npu --ep qnn'."
+        )
+    if not any(bundle_dir.rglob("*.onnx")):
+        raise ValueError(f"'{bundle_dir}' contains no .onnx files; not a valid genai bundle.")
+
+    return WinMLGenaiCausalLM(
+        bundle_dir,
+        config.ep,
+        device=config.device,
+        compile=False,
```
DingmaomaoBJTU: This disables the compilation path required by the only currently supported GenAI bundle target. The Qwen3 bundle contains raw QNN graphs, and `GenaiSession` only creates/reuses the isolated EPContext bundle when `compile=True`; the existing Qwen3 documentation warns that loading without compilation can native-fault during model creation. As written, `winml eval` bypasses the safety mechanism used by `winml perf`. Please enable compilation here (with a configurable timeout) rather than forcing it off.

scripts/e2e_eval/testsets/models_with_acc.json
```diff
@@ -1246,6 +1246,26 @@
       }
     }
   },
+  {
+    "hf_id": "Qwen/Qwen3-0.6B",
```
DingmaomaoBJTU: The standard E2E runner cannot successfully execute this new entry. On the QNN/NPU path, `_run_build()` invokes `winml build` with both `-c` and `--use-cache`, but optimized GenAI bundle builds reject those options. On every generic build path it returns one or more ONNX file paths, and `_run_winml_eval()` passes those to `eval`, while the new text-generation loader accepts only a bundle directory. Please add bundle-aware build/output handling in `run_eval.py` and pass the resulting directory to `winml eval`; otherwise this registry case always fails before PPL is computed.

scripts/e2e_eval/cache/baseline_cache.json
```diff
@@ -659,6 +659,16 @@
     "elapsed": 1319.8,
     "command": "python.exe run_pytorch_baseline.py --model bert-base-uncased --task fill-mask --device cpu --num-samples 100 --dataset wikitext --split test --dataset-config wikitext-2-raw-v1 --columns-mapping {\"input_column\": \"text\"}"
   },
+  "Qwen/Qwen3-0.6B|text-generation|Salesforce/wikitext|wikitext-2-raw-v1|test|100": {
```
DingmaomaoBJTU: This PPL baseline is not safely keyed. `_baseline_cache_key()` includes `samples` but omits `columns_mapping`; for this evaluator `samples` is ignored, while `num_tokens` and `seqlen` define the scoring protocol and can change the resulting PPL. Changing either parameter will therefore silently reuse this `23.645425` value and compare mismatched evaluations. Please include a deterministic representation of the protocol-relevant mapping in the cache key (and migrate this entry).

src/winml/modelkit/eval/text_generation_evaluator.py
```diff
@@ -0,0 +1,209 @@
...
+        model: Any = self.model
+        ds_config = self.config.dataset
+        column = ds_config.columns_mapping.get(
+            "input_column", get_default(self._TASK, "input_column")
+        )
+        dataset = load_dataset(
```
DingmaomaoBJTU: Unlike `WinMLEvaluator.prepare_data()`, this always calls `load_dataset()`. A local dataset directory produced by Hugging Face `Dataset.save_to_disk()`—including the output of `--dataset-script`—must be opened with `load_from_disk()`, so text-generation fails for a dataset source supported by every other evaluator. Please mirror the base loader's local-directory branch before applying the ordered token-budget iteration.

DingmaomaoBJTU (commented):
A few additional issues found while reviewing the updated changes.

src/winml/modelkit/eval/evaluate.py
zhenchaoni: Fixed in deceb7ff. Dropped the `compile=False` override so the loader uses the default `compile=True`, matching the safety path `winml perf` relies on: `GenaiSession` builds/reuses the isolated EPContext bundle instead of native-faulting on raw QNN graphs. Verified end-to-end against the uncompiled `out/qwen3-bundle` (QNN EP, no fault).

scripts/e2e_eval/testsets/models_with_acc.json
zhenchaoni: Agreed. E2E runner build support for GenAI bundles is out of scope for this PR, so I removed this registry entry (and its baseline-cache line) in deceb7ff. We will re-add it with proper bundle-aware build/output handling when E2E build support lands.

scripts/e2e_eval/cache/baseline_cache.json
zhenchaoni: Since the E2E entry is being removed (see the registry comment), I dropped this baseline line in deceb7ff rather than re-key it. The cache-key protocol coverage (num_tokens/seqlen) will be addressed together with E2E build support when the entry returns.

src/winml/modelkit/eval/text_generation_evaluator.py
zhenchaoni: Fixed in deceb7ff. `_load_corpus_tokens` now mirrors the base loader: a local directory is opened with `load_from_disk`, otherwise `load_dataset` is used. Verified end-to-end against a `save_to_disk` output directory.

src/winml/modelkit/eval/evaluate.py
```diff
@@ -282,6 +298,50 @@ def _load_model(config: WinMLEvaluationConfig) -> WinMLPreTrainedModel | WinMLCo
...
+        raise ValueError(
+            f"'{bundle_dir}' is not a genai bundle: no genai_config.json found. "
+            "Point -m at a bundle built with 'winml build ... --device npu --ep qnn'."
+        )
+    if not any(bundle_dir.rglob("*.onnx")):
+        raise ValueError(f"'{bundle_dir}' contains no .onnx files; not a valid genai bundle.")
+
+    return WinMLGenaiCausalLM(
+        bundle_dir,
+        config.ep,
```
DingmaomaoBJTU, line 404: Passing `config.ep` through unchanged makes `--device` ineffective for GenAI bundles. `_resolve_device()` has already resolved an explicit `--device cpu` to `"cpu"`, but `GenaiSession` leaves the bundle routing unchanged whenever `ep is None`; its `device` value is only consulted after an EP override exists. Consequently a QNN bundle still registers/runs QNN even when eval is invoked with `--device cpu`. Please mirror the `winml-genai` perf precedence (or otherwise preserve option explicitness) so a concrete user-selected device produces an EP override instead of being ignored.

src/winml/modelkit/eval/evaluate.py
```diff
@@ -199,6 +202,14 @@ def get_evaluator_class(config: WinMLEvaluationConfig) -> type[WinMLEvaluator]:
         "path": "mattmdjaga/human_parsing_dataset",
         "split": "train",
     },
+    "text-generation": {
+        # Raw wikitext-2 test split scored token-by-token for perplexity;
+        # ``input_column`` names the text field the corpus is built from.
+        "path": "Salesforce/wikitext",
+        "name": "wikitext-2-raw-v1",
+        "split": "test",
+        "columns_mapping": {"input_column": "text"},
```
DingmaomaoBJTU, line 213: The evaluator advertises `num_tokens` and `seqlen` through `--column`, but this full default mapping makes those parameters unusable with the default dataset. `evaluate()` copies the default fields wholesale whenever `dataset.path is None`, replacing the user's mapping, so `--column num_tokens=1024 --column seqlen=512` still scores the schema defaults unless the user redundantly repeats `--dataset Salesforce/wikitext`. Please merge only missing defaults, or otherwise preserve user-supplied scoring parameters while filling `input_column`.

src/winml/modelkit/models/winml/genai_causal_lm.py
```diff
@@ -0,0 +1,216 @@
...
+        import onnxruntime_genai as og
+
+        # set_logits is only on the low-level generator, not the session's
+        # generate path, so reach the session's loaded model directly.
+        model = self._session._model
+
+        params = og.GeneratorParams(model)
+        # Fixed max_length is required by the compiled pipeline; do_sample=False
+        # keeps the decode deterministic.
+        params.set_search_options(max_length=max_length, do_sample=False)
```
DingmaomaoBJTU: Setting only `max_length` and `do_sample` does not isolate teacher forcing from the search policy in `genai_config.json`. `GeneratorParams` still inherits options such as `min_length` and `no_repeat_ngram_size`, and ORT GenAI applies those masks inside `generate_next_token()` *after* `set_logits()`. If the corpus target is EOS before `min_length`, or closes a forbidden n-gram, the forced target is masked, another token is appended, and every later logit is conditioned on the wrong prefix. Please explicitly neutralize the bundle search constraints for scoring (for example `min_length=0`, `no_repeat_ngram_size=0`, `repetition_penalty=1`, `num_beams=1`) and cover a non-default search config.

src/winml/modelkit/eval/evaluate.py
zhenchaoni: @temp/reply_e1.md

src/winml/modelkit/eval/evaluate.py
zhenchaoni: @temp/reply_e2.md

src/winml/modelkit/models/winml/genai_causal_lm.py
zhenchaoni: @temp/reply_e3.md

DingmaomaoBJTU (approved):
LGTM. Well-structured text-generation perplexity evaluator. Key observations (non-blocking):

1. **Teacher forcing via set_logits is clever** — the one-token-at-a-time replay with forced logits correctly scores all positions while reusing the genai generator's KV cache. The neutralization of bundle search constraints (min_length=0, 
o_repeat_ngram_size=0, etc.) is essential and properly tested.

2. **_step_nll is numerically sound** — the log-sum-exp with max subtraction avoids overflow, and the f64 upcast avoids precision issues on large vocabs.

3. **Default dataset column merge fix** is a nice UX improvement — preserving user --column values while filling in defaults means --column num_tokens=1024 --column seqlen=512 works without needing to repeat --dataset.

4. **Minor**: _load_corpus_tokens uses model.encode(text) + 2 per row as an approximation, which could over-shoot significantly for short rows (many BPE boundary tokens). Fine for the first iteration, but a comment noting this is conservative (overcounts) would help.

5. **Minor**: HFCausalLM.forward computes the full sequence in one shot (model(input_ids=...)) which works for short seqlens but could OOM on very long sequences on GPU. Since the default seqlen is 2048 and this is a baseline path, acceptable — just noting for awareness.

DingmaomaoBJTU (approved):

src/winml/modelkit/eval/evaluate.py
```diff
@@ -243,11 +256,15 @@ def _load_model(config: WinMLEvaluationConfig) -> WinMLPreTrainedModel | WinMLCo
     from ..session import EPDeviceTarget, WinMLEPRegistry, resolve_device
     from ..utils import cli as cli_utils
 
+    if config.task == "text-generation":
+        return _load_genai_causal_lm(config)
```
xieofxie, line 260: Alternative: use runtime as perf to distinguish

PR #1131, opened 2026-07-18
recipe(open-vakgyata): add CPU audio-classification recipes

ssss141414:
## Summary

This adds CPU fp32 and fp16 coverage recipes for `onecxi/open-vakgyata`, a Wav2Vec2 audio-classification checkpoint for identifying 10 Indian speech languages. Effort L0 / Outcome L0 ships recipe-only coverage for both required tuples. The committed Goal L2 was reached with `L2 PASS` for both precisions; supplementary task evaluation remains `CLI-BLOCKED` because `winml eval` has no registered `audio-classification` evaluator.

### What the model does

`open-vakgyata` is a speech language-identification model: it accepts a normalized 16 kHz mono waveform and returns logits over 10 Indian speech-language labels (`en-IN`, `hi-IN`, `or-IN`, `bn-IN`, `ta-IN`, `te-IN`, `kn-IN`, `ml-IN`, `mr-IN`, and `gu-IN`). **Confidence: verified.**

- Evidence: the pinned [model card](https://huggingface.co/onecxi/open-vakgyata/blob/f2754058e485dfc65cc62589b8d0e21c1d328399/README.md) documents the model overview, supported languages, 16 kHz input, and inference; the pinned [checkpoint config](https://huggingface.co/onecxi/open-vakgyata/blob/f2754058e485dfc65cc62589b8d0e21c1d328399/config.json) declares `Wav2Vec2ForSequenceClassification`, 10 labels, and `id2label`; the pinned [preprocessor config](https://huggingface.co/onecxi/open-vakgyata/blob/f2754058e485dfc65cc62589b8d0e21c1d328399/preprocessor_config.json) declares 16 kHz normalization. The measured baseline interface was `input_values float32[1,16000] -> logits float32[1,10]`.

### Primary user stories

- A user supplies a 16 kHz speech waveform to obtain the predicted language label for selecting language-aware downstream speech processing. **Confidence: verified.** Evidence: the pinned model card describes language identification, recommended audio format, and logits-to-`id2label` inference; the pinned checkpoint config defines the 10 locale-code labels.

### Supported tasks

[...]

PR #1212, opened 2026-07-26
feat(eval): support audio classification

### LINKED ISSUE #1175: Add winml eval support for audio-classification
`winml eval` currently rejects the `audio-classification` task before dataset selection, even when `winml inspect` and `winml build` resolve/build a standard audio-classification model.

Repro from `facebook/mms-lid-256` on current main:

```powershell
uv run winml eval --schema --task audio-classification
uv run winml eval -m temp\facebook_mms-lid-256_tester\cpu\fp32\model.onnx --model-id facebook/mms-lid-256 --task audio-classification --ep cpu --device cpu
```

Observed:

```text
Error: Task 'audio-classification' is not supported by `winml eval`.
Error: Evaluation failed: Task 'audio-classification' is not supported.
```

Expected: an audio-classification eval schema/adapter with dataset preprocessing for raw audio or feature-extractor-normalized waveform input and a task metric such as accuracy.

This blocks Goal-L3 validation for standard `Wav2Vec2ForSequenceClassification` audio classifiers such as `facebook/mms-lid-256`.

ssss141414:
## Summary

Adds generalized `audio-classification` evaluation support and closes #1175 and unblocks supplementary evaluation for #1131.

Before this change, `winml eval --task audio-classification` failed before dataset loading with `Task 'audio-classification' is not supported`. This change registers a class-wide evaluator and schema without model-ID or dataset-specific branching.

The evaluator:
- requires an explicit semantically compatible labeled dataset;
- validates authoritative checkpoint/dataset label alignment;
- decodes remote/archive audio through the dataset opener plus SoundFile without requiring TorchCodec;
- supports static rank-2 waveform inputs through deterministic windowing and utterance mean-logit aggregation;
- preserves and validates static rank-3+ extracted feature tensors;
- reports accuracy, macro-F1, confusion matrix, and complete requested/eligible/selected/processed/rejected accounting.

## Unsupported to supported

- Baseline: audio-classification Eval exited before dataset loading as unsupported.
- Candidate: direct pinned `google/fleurs` test rows execute successfully for both existing fp32 and fp16 CPU artifacts.
- No model ID or FLEURS literal is present in shared evaluator source.

## Dataset and metrics

Checkpoint: `onecxi/open-vakgyata@f2754058e485dfc65cc62589b8d0e21c1d328399`

Dataset: `google/fleurs@70bb2e84b976b7e960aa89f1c648e09c59f894dd`, test split, streaming, no shuffle, first 10 deterministic rows from each exact locale config:

`bn_in`, `hi_in`, `or_in`, `ta_in`, `te_in`, `kn_in`, `ml_in`, `mr_in`, `gu_in`

`en_us` is deliberately excluded because US English is not semantically equivalent to checkpoint class `en-IN`.

| Precision | Requested / processed / rejected | Correct | Accuracy | 9-class macro-F1 |
|---|---:|---:|---:|---:|
| fp32 | 90 / 90 / 0 | 52 | 0.5777777778 | 0.4714699307 |
| fp16 | 90 / 90 / 0 | 52 | 0.5777777778 | 0.4714699307 |

The fp32/fp16 confusion matrices are identical; both metric deltas are zero. Macro-F1 is recomputed from the merged nine-class confusion matrix, not averaged across per-locale macro-F1 values.

Accuracy measures total utterance correctness. Macro-F1 gives each eligible locale equal weight and exposes class collapse (`or-IN`, `te-IN`, and `ml-IN` have zero recall in this bounded smoke sample). These are meaningful balanced smoke metrics, not a population FLEURS benchmark.

## Performance

CPUExecutionProvider, 100 measured iterations after 10 warmups:

| Precision | Mean | P50 | P90 | Throughput |
|---|---:|---:|---:|---:|
| fp32 | 44.948 ms | 47.373 ms | 49.018 ms | 22.25 samples/s |
| fp16 | 47.749 ms | 45.324 ms | 54.581 ms | 20.94 samples/s |

## Numerical parity

Same pinned checkpoint, feature extractor, named 16 kHz one-second waveform, and logits interpretation for PyTorch and ONNX:

| Precision | Cosine | Max abs | Top-1 |
|---|---:|---:|---|
| fp32 | 0.99999999967 | 0.0003737211 | `ta-IN` preserved |
| fp16 | 0.99999986842 | 0.0055872202 | `ta-IN` preserved |

## Component analysis

Semantic regions: convolutional feature encoder; feature projection; Wav2Vec2 Transformer encoder; two stable-LayerNorm Transformer layers; temporal mean pooling; classification projector; ten-language classifier.

- fp32: 129 mapped, 0 partial, 0 unmapped nodes.
- fp16: 129 mapped semantic nodes plus two topology-level Cast nodes without a unique component tag; no actionable EP finding.

## Operator-level analysis

- fp32: 129 operators, 14 unique types.
- fp16: 131 operators, 15 unique types; the delta is two Cast nodes.
- Dominant counts for both: Reshape=32, Transpose=23, Gemm=19, LayerNormalization=15, Gelu=10, Conv=8, Add=7, MatMul=4.

Static rules classify all operator types as supported for NvTensorRTRTX GPU, QNN NPU/GPU, and OpenVINO NPU/GPU/CPU. CUDA, MIGraphX, DML, CPU, and VitisAI rows are unknown where the pinned rule set has no classification; this is static rule evidence, not runtime coverage.

## Validation

- Direct pinned FLEURS: 18/18 locale/precision result files valid; 180/180 processed; zero rejected.
- Focused evaluator tests: 22 passed.
- `mypy -p winml.modelkit`: success, 411 source files.
- `ruff check src tests`: passed.
- `git diff --check`: passed.

This PR is intentionally kept in Draft state pending final independent review and CI verification.

## August 11 update

- Synced the branch with current `main`.
- Added direct support for feature-extractor-normalized 1-D waveform dataset columns.
- Updated the saved-ONNX CLI test for the current config-loading path.
- Preserved the established channels-first convention and added one-frame regression coverage; the ambiguous smallest-dimension heuristic from #1302 was intentionally not carried forward.


timenick:
Compared with closed PR #1302 and incorporated the safe improvements: pre-normalized waveform input, current-main compatibility, schema updates, and regression coverage. The smallest-dimension stereo heuristic was excluded because it misclassifies known channels-first one-frame audio. The branch is also synced with current main.

src/winml/modelkit/eval/audio_classification_evaluator.py
```diff
@@ -0,0 +1,622 @@
...
+        count = math.ceil(waveform.size / window_length)
+        windows = np.full((count, window_length), padding_value, dtype=np.float32)
+        for index in range(count):
+            chunk = waveform[index * window_length : (index + 1) * window_length]
+            windows[index, : chunk.size] = chunk
+        return windows
+
+    def _model_input_contract(self) -> tuple[str, list[int]]:
+        """Return the single static ONNX input name and shape."""
+        io_config = getattr(self.model, "io_config", None) or {}
```
zhenchaoni: `run_pytorch_baseline.py` loads a native Hugging Face `AutoModelForAudioClassification` and passes it to this evaluator. Native HF models do not expose `io_config`, so this produces an empty contract and every sample is rejected before `forward`; the 10-sample FLEURS smoke at this commit ended with `No audio samples were successfully processed`. This also rejects otherwise valid dynamic or multi-input WinML exports, such as models that consume an `attention_mask`. Please define a shared forward-facing model/adapter contract for HF and WinML model instances, and keep ONNX input-name/shape adaptation inside the WinML model implementation rather than the evaluator.

src/winml/modelkit/eval/audio_classification_evaluator.py
```diff
@@ -0,0 +1,622 @@
...
+        """Load the checkpoint's audio feature extractor.
+
+        A Transformers audio pipeline processes one fixed input at a time and
+        would silently truncate long utterances for a static ONNX graph. This
+        evaluator instead uses only its feature extractor and owns deterministic
+        windowing and utterance-level logit aggregation.
+        """
+        from transformers import AutoFeatureExtractor
+
+        return AutoFeatureExtractor.from_pretrained(self.config.model_id)
```
zhenchaoni: This returns only an `AutoFeatureExtractor`, but `_measure_pytorch_latency()` in `run_pytorch_baseline.py` treats `task_evaluator.pipe` as the complete inference pipeline and calls it directly. For audio classification, `--perf-iterations` therefore either fails on the raw audio value or measures preprocessing without the model forward/postprocessing that the baseline claims to measure. Please expose a full inference callable, or provide a task-specific latency hook that uses the same inference path as `compute()`.

src/winml/modelkit/eval/audio_classification_evaluator.py
```diff
@@ -0,0 +1,622 @@
...
+        streams stop once every authoritative overlapping class has supplied
+        its balanced quota.
+        """
+        from datasets import load_dataset, load_from_disk
+
+        ds = self.config.dataset
+        try:
+            ds_path = Path(ds.path).expanduser() if ds.path else None
+            if ds_path and ds_path.is_dir():
+                dataset = load_from_disk(str(ds_path))
```
zhenchaoni, line 289: `load_from_disk()` can return a `DatasetDict`, but this branch never selects `ds.split`. In that case `_validate_and_resolve_labels()` sees split names such as `train` and `test` as the top-level columns and reports that the audio/label columns are missing. Please detect dataset dictionaries, select the configured split, and report a clear error when that split is absent.

src/winml/modelkit/eval/audio_classification_evaluator.py
```diff
@@ -0,0 +1,622 @@
...
+                else:
+                    raise ValueError("selected audio sample has no row or dataset index")
+                windows = self._preprocess_audio(sample[self._audio_col])
+            except (TypeError, ValueError, RuntimeError) as error:
+                rejected_by_reason[type(error).__name__] += 1
+                logger.warning("Skipping audio sample: %s", error)
+                continue
+
+            window_logits = [self._infer_window(window) for window in windows]
+            utterance_logits = np.mean(np.stack(window_logits), axis=0)
```
zhenchaoni: Equal-weight mean-logit aggregation gives a short, mostly padded final window the same influence as a full window. For an input of `window_length + 1` samples, the one-real-sample tail can therefore flip the utterance prediction, and one dataset example now causes multiple forwards instead of the direct one-example/one-prediction flow. Chunking and aggregation should live behind the model/preprocessing adapter; if chunking remains supported, it needs an explicit policy that accounts for valid lengths or attention masks.

src/winml/modelkit/eval/audio_classification_evaluator.py
```diff
@@ -0,0 +1,622 @@
...
+                for label, (actual, required) in sorted(insufficient.items())
+            )
+            raise DatasetValidationError(
+                f"Too few usable samples after audio decoding/preprocessing: {details}.",
+            )
+
+        metric = ClassificationMetric().compute(
+            predictions,
+            references,
+            self._eligible_model_labels,
```
zhenchaoni: `_eligible_model_labels` contains every mapped class, including classes for which the selected data contains no reference sample. Because `ClassificationMetric` passes this full list to `f1_score(..., zero_division=0)`, unevaluated classes are scored as zero: for example, two perfect predictions with ten mapped classes produce macro-F1 `0.2`. Please either require coverage for every class when reporting full-class macro-F1, or compute macro-F1 over represented reference classes and report class coverage separately.

zhenchaoni (changes requested):
I am requesting changes based on the evaluator contract and an end-to-end smoke test at this exact commit. Using `onecxi/open-vakgyata` with 10 deterministic `google/fleurs` `bn_in` samples, the WinML model processed 10/10 samples with no rejections on both `CPUExecutionProvider` and `DmlExecutionProvider`. The same evaluator rejected all 10 samples when invoked through `scripts/e2e_eval/run_pytorch_baseline.py` with the native Hugging Face model. The inline comments cover that compatibility blocker and the other correctness issues found during review.

src/winml/modelkit/eval/audio_classification_evaluator.py
ssss141414: Fixed in bbd266e9. `prepare_pipeline()` now returns a shared callable `_AudioModelAdapter`: native HF models (no `io_config`) receive all feature-extractor outputs, while WinML models select and validate every declared input, including `attention_mask`. Added regressions for native HF and two-input WinML forwarding. Focused post-merge validation: 246 passed; Ruff passed.

src/winml/modelkit/eval/audio_classification_evaluator.py
ssss141414: Fixed in bbd266e9. The adapter now owns decode/preprocess/forward/logit extraction, exposes `.model`, and `compute()` calls the same `self.pipe(raw_audio)` path used by `_measure_pytorch_latency()`. I also taught the baseline input extractor to recognize the `audio` column and added a regression that invokes `_measure_pytorch_latency()` against the evaluator. Focused post-merge validation: 246 passed; Ruff passed.

src/winml/modelkit/eval/audio_classification_evaluator.py
ssss141414: Fixed in bbd266e9. Saved `DatasetDict` values now select the configured `ds.split`; a missing split raises a `DatasetValidationError` listing available splits. Added tests for both requested-split selection and the missing-split diagnostic.

src/winml/modelkit/eval/audio_classification_evaluator.py
ssss141414: Fixed in bbd266e9. The evaluator no longer chunks a row and mean-aggregates equal-weight logits. The shared adapter performs one forward per selected row and uses feature-extractor truncation/padding for a fixed waveform contract. Added an overlength regression asserting exactly one forward and the expected padded/truncated multi-input tensors.

src/winml/modelkit/eval/audio_classification_evaluator.py
ssss141414: Fixed in bbd266e9. Single-label macro-F1 is now computed over represented reference classes, and results include `represented_classes`, `total_classes`, and `class_coverage`. The end-to-end regression covers a two-of-three represented-class case and verifies coverage `2/3`.

src/winml/modelkit/eval/audio_classification_evaluator.py
```diff
@@ -0,0 +1,801 @@
...
+                waveform,
+                target_rate // divisor,
+                sampling_rate // divisor,
+            ).astype(np.float32)
+
+        extractor_kwargs: dict[str, Any] = {
+            "sampling_rate": target_rate,
+            "return_tensors": "pt",
+        }
+        waveform_contract = self._input_contracts.get("input_values")
```
zhenchaoni: This makes preprocessing depend on whether the model exposes WinML `io_config`: a native HF model has no contract and receives the full utterance, while the corresponding static WinML model is padded/truncated to its exported length. At this head, the same first 10 pinned FLEURS `bn_in` rows scored 0.9 through the PyTorch baseline but 0.3 through the same fp32 ONNX on both CPU and DML. CPU/DML agreement rules out an EP-specific delta; the two runtimes are simply not evaluating identical model inputs, so the baseline comparison is not meaningful. Please make the preprocessing/input-length contract explicit and shared by both model adapters (or use an equivalent dynamic export) before comparing their metrics.

src/winml/modelkit/eval/audio_classification_evaluator.py
```diff
@@ -0,0 +1,801 @@
...
+        if not names and not shapes:
+            return {}
+        if len(names) != len(shapes) or not names:
+            raise ValueError(
+                "audio-classification input names and shapes must have equal non-zero lengths",
+            )
+        contracts: dict[str, list[int]] = {}
+        for name, raw_shape in zip(names, shapes, strict=True):
+            shape = list(raw_shape)
+            if len(shape) < 2 or any(not isinstance(value, int) for value in shape[1:]):
```
zhenchaoni: This rejects every model with a symbolic or otherwise dynamic non-batch dimension before `forward`. Dynamic waveform length is a normal audio export contract and is supported by the CLI's dynamic-axis/export configuration. Please preserve symbolic dimensions, apply padding only when a concrete length exists, and validate only concrete dimensions at runtime.

src/winml/modelkit/eval/audio_classification_evaluator.py
```diff
@@ -0,0 +1,801 @@
...
+
+    def _select_model_inputs(self, encoded: Any) -> dict[str, Any]:
+        values = dict(encoded)
+        if not self._input_contracts:
+            if not values:
+                raise ValueError("audio feature extractor produced no model inputs")
+            return values
+        selected: dict[str, Any] = {}
+        for name, expected_shape in self._input_contracts.items():
+            if name not in values:
```
zhenchaoni: This still couples evaluator preprocessing to raw ONNX tensor names. A valid single-input classifier whose ONNX input is named `audio` or `waveform` is rejected because the feature extractor returns its semantic key (typically `input_values`). Input-name adaptation belongs in the WinML model/adapter contract; the evaluator should pass semantic forward inputs without requiring processor keys to equal serialized ONNX names.

src/winml/modelkit/eval/audio_classification_evaluator.py
```diff
@@ -0,0 +1,801 @@
...
+        )
+        model_inputs = {
+            name: value.to(device)
+            if hasattr(value, "to")
+            else torch.as_tensor(value, device=device)
+            for name, value in model_inputs.items()
+        }
+        output = self.model(**model_inputs)
+        logits = (
+            output.get("logits")
```
zhenchaoni: Audio classification currently falls back to `WinMLModelForGenericTask`, which returns a dict keyed by the actual ONNX output names. Requiring the literal key `logits` therefore rejects a valid one-output classifier named `scores` or `output_0` (including prebuilt ONNX models). Please normalize the classification output in a task-level WinML model wrapper, as image/text classification do, or at minimum accept the unique model output without hardcoding its tensor name.

src/winml/modelkit/eval/audio_classification_evaluator.py
```diff
@@ -0,0 +1,801 @@
...
+        waveform_contract = self._input_contracts.get("input_values")
+        if waveform_contract is not None and len(waveform_contract) == 2:
+            extractor_kwargs.update(
+                padding="max_length",
+                truncation=True,
+                max_length=waveform_contract[1],
+            )
+        encoded = self._feature_extractor(waveform, **extractor_kwargs)
+        model_inputs = self._select_model_inputs(encoded)
+        device = (
```
zhenchaoni, line 116: The public PyTorch baseline still constructs `WinMLEvaluationConfig(device="cuda")` with the default `runtime="winml"` and instantiates the evaluator directly. Consequently this branch selects CPU tensors even though `_load_pytorch_model()` moved the model to CUDA, so `--device cuda` fails with a device mismatch. Please route the script through `evaluate(..., pytorch_model=model)` and normalize its CLI `cuda` value to the evaluation API's `gpu` device so `_prepare_supplied_pytorch_model()` sets the exact pipeline device.

src/winml/modelkit/eval/audio_classification_evaluator.py
```diff
@@ -0,0 +1,801 @@
...
+            if getattr(self._config, "runtime", None) == "pytorch"
+            else "cpu"
+        )
+        model_inputs = {
+            name: value.to(device)
+            if hasattr(value, "to")
+            else torch.as_tensor(value, device=device)
+            for name, value in model_inputs.items()
+        }
+        output = self.model(**model_inputs)
```
zhenchaoni: Native HF forwards run here with autograd enabled. `.eval()` only changes module behavior; it does not disable graph construction, so accuracy evaluation uses unnecessary memory and `--perf-iterations` measures training-mode autograd overhead rather than inference latency. Please execute the forward under `torch.inference_mode()` (or `torch.no_grad()`).

zhenchaoni (changes requested):
Re-reviewed the complete PR at `bbb3f509`. The previous CPU Hugging Face failure, local `DatasetDict` handling, equal-weight window aggregation, and represented-class macro-F1 issues are fixed. The focused evaluator suite passes (25 tests), and a real 10-sample PyTorch CPU run now succeeds. However, the same smoke exposed a new cross-runtime preprocessing mismatch: PyTorch scored 0.9 while the same fp32 ONNX scored 0.3 on both CPU and DML because the native model receives the full utterance while the static WinML model receives only the first 16,000 samples. The new inline comments cover this and the remaining model-contract issues.

src/winml/modelkit/eval/audio_classification_evaluator.py
ssss141414: Fixed in 9d5775a3. The baseline now evaluates through the public evaluate(..., pytorch_model=model) path, maps an available CUDA request to the evaluation API gpu device, and sets runtime=pytorch. Audio baselines also require and carry the same explicit waveform length as the ONNX contract, so native and WinML accuracy use identical input duration. Exact-head validation: 347 focused tests, Ruff, package mypy, detached tester, and all 9 fresh GitHub checks passed.

ssss141414:
APPROVE: reviewed exact head 9d5775a361f4664e889122101ff29670932bc221. Audio preprocessing now preserves the PyTorch/WinML tensor boundary, fixed and dynamic input contracts, custom ONNX names, inference mode, and baseline device/input-length parity. Independent detached validation passed 347 focused tests; Ruff and package-wide mypy passed; all 9 fresh GitHub checks are green; no active review threads remain.

src/winml/modelkit/eval/audio_classification_evaluator.py
ssss141414: Fixed in 9d5775a3 with a typed shared audio_input_length contract. Native PyTorch and static ONNX paths now apply identical truncation/padding; the value is serialized, exposed by CLI, forwarded by the E2E baseline/WinML runners, and included in baseline cache identity.

src/winml/modelkit/eval/audio_classification_evaluator.py
ssss141414: Fixed in 9d5775a3. Symbolic input dimensions are preserved and only concrete dimensions are validated; dynamic waveform inputs keep their actual runtime length. Added a dynamic-shape regression.

src/winml/modelkit/eval/audio_classification_evaluator.py
ssss141414: Fixed in 9d5775a3. A sole serialized ONNX input can consume the semantic input_values feature even when its tensor name is audio/waveform. Added a custom-input-name regression.

src/winml/modelkit/eval/audio_classification_evaluator.py
ssss141414: Fixed in 9d5775a3. Dict outputs still prefer logits, but a one-output classifier now accepts its unique output regardless of serialized tensor name.

src/winml/modelkit/eval/audio_classification_evaluator.py
ssss141414: Fixed in 9d5775a3. Native model forward now runs under torch.inference_mode(); the multi-input native regression asserts gradients are disabled.

src/winml/modelkit/eval/config.py
```diff
@@ -48,6 +49,7 @@ class DatasetConfig:
     samples: int = 100
     shuffle: bool = True
     seed: int = 42
+    audio_input_length: int | None = None
```
DingmaomaoBJTU: `DatasetConfig` is publicly exported, so inserting this field before `columns_mapping` changes its positional constructor contract. Existing code such as `DatasetConfig(path, name, split, samples, shuffle, seed, columns_mapping)` now silently binds that dictionary to `audio_input_length` and leaves `columns_mapping` empty. Please append the new field after the existing fields (or make the dataclass keyword-only in a separately managed breaking change).

DingmaomaoBJTU (changes requested):
The prior audio-evaluation correctness issues appear addressed and CI is green, but one public API compatibility regression remains.

src/winml/modelkit/eval/config.py
ssss141414: Fixed in 855725e2 by moving `audio_input_length` after all existing `DatasetConfig` fields, preserving the public positional constructor contract. Added `test_dataset_config_preserves_existing_positional_arguments`; the focused suite and exact-head CI pass.

DingmaomaoBJTU (approved):
Re-reviewed the latest head. The DatasetConfig positional compatibility issue is fixed with regression coverage, the prior audio evaluation concerns are addressed, and all CI checks pass.

PR #1218, opened 2026-07-27
feat(analyze): use pattern rules for runtime support check

fangyangci:
## Summary

Update `winml analyze` to evaluate matched subgraphs with pattern-level runtime rules before falling back to operator-level checks. This allows runtime support to be reported at the pattern level while avoiding duplicate checks for operators already covered by a matched pattern.

## Changes

- Resolve pattern rule Parquet tables by execution provider, device, domain, and model opset.
- Evaluate pattern rule conditions using matched input types, shapes, constant values, attributes, and dynamic-axis information.
- Prefer EP-specific pattern sources over default sources and deduplicate overlapping matches by stable node key.
- Cache deduplicated matches per model and EP so targets on different devices can reuse pattern extraction results.
- Determine pattern runtime support from compile and run results, then propagate the selected result into analysis recommendations and optimization actions.
- Select the best configured pattern alternative by support status and priority.
- Mark pattern-covered nodes as `pattern_match` and skip their operator-level Parquet lookup and OP CHECK accounting.
- Add PATTERN CHECK progress, support summaries, and internal operator coverage to the CLI output.
- Extend runtime debug output with `match_status` to distinguish `pattern_match` from `op_match`.
- Remove obsolete pattern and rule-query helper logic.

## Why remove `--htp-metadata`

- There is no explicit, stable mapping between HTP metadata entries and pattern skeleton definitions. As a result, metadata-only matches often cannot resolve to a pattern rule table, so compile and run results cannot be determined consistently.
- Most useful signal from HTP metadata is already covered by the current pattern matching flow. Remaining gaps should be addressed by improving pattern matching coverage, instead of maintaining a parallel metadata path.
- Multiple layered matching sources can overlap and override each other, which introduces ambiguous precedence and makes the logic harder to reason about and maintain. A single pattern-rule-driven path is more deterministic and maintainable.

## Compatibility

- Remove the obsolete `--htp-metadata` option from `winml analyze`.

image

## Pattern count output

- Pattern counts are execution-provider-specific because pattern sources and node deduplication can differ by EP.
- Keep `metadata.detected_pattern_count` in its existing location and represent it as `EP -> pattern ID -> count`.
- Read one EP total with `sum(metadata.detected_pattern_count[ep].values())`.
- Do not expose a single all-EP count because summing across EPs may count the same model pattern more than once.

## Compatibility update

- The `metadata.detected_pattern_count` field remains in place; its value changes from `pattern ID -> count` to `EP -> pattern ID -> count`.

### DIFF
--- a/src/winml/modelkit/analyze/analyzer.py
+++ b/src/winml/modelkit/analyze/analyzer.py
@@ -213,9 +213,7 @@ def _build_match_status_by_match_id(
             continue
 
         raw_status = str(entry.get("support_status", "")).strip().lower()
-        if raw_status == "unknown":
-            raw_status = "unknown"
-        elif raw_status == "unknow":
+        if raw_status == "unknown" or raw_status == "unknow":
             raw_status = "unknown"
 
         if raw_status in _PATTERN_STATUS_QUALITY:
@@ -295,7 +293,11 @@ def _build_pattern_matching_summary(
 
         node_breakdown: list[dict[str, Any]] = []
         for op_type, total_count in sorted(op_counts.items(), key=lambda item: (-item[1], item[0])):
-            per_instance_count = total_count // instances if instances > 0 and total_count % instances == 0 else None
+            per_instance_count = (
+                total_count // instances
+                if instances > 0 and total_count % instances == 0
+                else None
+            )
             node_breakdown.append(
                 {
                     "op_type": op_type,
@@ -330,7 +332,11 @@ def _build_operator_counts_excluding_pattern_nodes(
 ) -> dict[str, int]:
     """Subtract pattern-matched nodes from operator totals for OP CHECK display."""
     if not matched_node_keys:
-        return {str(op_type): int(count) for op_type, count in operator_counts.items() if int(count) > 0}
+        return {
+            str(op_type): int(count)
+            for op_type, count in operator_counts.items()
+            if int(count) > 0
+        }
 
     matched_counts_by_op: dict[str, int] = {}
     for node_key in matched_node_keys:
@@ -1130,19 +1136,26 @@ class ONNXStaticAnalyzer:
         for current_ep in eps_to_analyze:
             logger.info("Checking runtime support for %s...", current_ep)
 
-            def _on_pattern_query_start_for_ep(pattern_counts: Mapping[str, int]) -> None:
+            def _on_pattern_query_start_for_ep(
+                pattern_counts: Mapping[str, int],
+                _ep: EPName = current_ep,
+            ) -> None:
                 if on_pattern_query_start is None:
                     return
                 try:
-                    on_pattern_query_start(current_ep, dict(pattern_counts))
+                    on_pattern_query_start(_ep, dict(pattern_counts))
                 except Exception:
                     logger.debug("on_pattern_query_start callback failed", exc_info=True)
 
-            def _on_pattern_query_result_for_ep(pattern_id: str, support_status: str) -> None:
+            def _on_pattern_query_result_for_ep(
+                pattern_id: str,
+                support_status: str,
+                _ep: EPName = current_ep,
+            ) -> None:
                 if on_pattern_query_result is None:
                     return
                 try:
-                    on_pattern_query_result(current_ep, pattern_id, support_status)
+                    on_pattern_query_result(_ep, pattern_id, support_status)
                 except Exception:
                     logger.debug("on_pattern_query_result callback failed", exc_info=True)
 
--- a/src/winml/modelkit/analyze/analyzer.py
+++ b/src/winml/modelkit/analyze/analyzer.py
@@ -1166,6 +1166,22 @@ class ONNXStaticAnalyzer:
         from .core.pattern_extractor import PatternExtractor
         from .core.runtime_checker import RuntimeChecker
 
+        def _make_local_pattern_checker(
+            runtime_checker: RuntimeChecker,
+        ) -> Callable[[PatternMatchResult, str, bool], RuntimeTestResult | None]:
+            def _check_pattern_locally(
+                pattern_match: PatternMatchResult,
+                fallback_reason: str,
+                debug: bool,
+            ) -> RuntimeTestResult | None:
+                return runtime_checker.check_pattern_locally(
+                    pattern_match,
+                    fallback_reason=fallback_reason,
+                    for_debug=debug,
+                )
+
+            return _check_pattern_locally
+
         # Normalize EP name (convert aliases to full names)
         total_start = time.perf_counter()
         ep_normalized = normalize_ep_name(cast("EPNameOrAlias | None", ep))
@@ -1239,6 +1255,22 @@ class ONNXStaticAnalyzer:
         for current_ep in eps_to_analyze:
             logger.info("Checking runtime support for %s...", current_ep)
 
+            # TODO: add VitisAIExecutionProvider back once non-QDQ
+            # data is ready, and run_unknown_op is supported for QDQ ops
+            run_unknown_op_for_ep = run_unknown_op
+            if current_ep == "VitisAIExecutionProvider":
+                run_unknown_op_for_ep = False
+
+            pattern_runtime_checker = (
+                RuntimeChecker(
+                    ep=current_ep,
+                    device=device_to_use,
+                    model=onnx_model,
+                )
+                if run_unknown_op_for_ep
+                else None
+            )
+
             def _on_pattern_query_start_for_ep(
                 pattern_counts: Mapping[str, int],
                 pattern_lookup_supported: bool = True,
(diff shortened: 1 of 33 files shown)


src/winml/modelkit/analyze/analyzer.py
```diff
@@ -89,6 +103,259 @@ def _iter_runtime_test_results(pattern_runtime: PatternRuntime) -> list[RuntimeT
...
+def _build_pattern_status_by_node_key(
+    subgraph_patterns: list[PatternMatchResult],
+    merge_prep_entries: Sequence[Mapping[str, Any]],
+) -> dict[str, str]:
+    """Build per-node pattern status map for matched-node runtime short-circuit."""
+    status_by_match_id = _build_match_status_by_match_id(merge_prep_entries)
+    status_by_node_key: dict[str, str] = {}
+
+    for pattern_match in subgraph_patterns:
+        status = status_by_match_id.get(pattern_match.match_id, "unknown")
```
DingmaomaoBJTU, line 249: This makes every matched node with no concrete pattern-rule result look pattern-covered with `unknown`. `RuntimeCheckerQuery.run_for_node()` then short-circuits those nodes to `no_data` with `reason="pattern_matched"`, so if the pattern table is missing, invalid for the EP/device, or has no matching row, we skip the normal operator parquet/local probe and regress known operator support to unknown just because a pattern matched.

Can we only put nodes in this map when the pattern rule produced a concrete `supported`/`partial`/`unsupported` result, and let missing/unknown pattern results fall through to the regular op check?

src/winml/modelkit/analyze/analyzer.py
```diff
@@ -854,17 +1135,85 @@ def analyze_from_proto(
...
+                ep=current_ep,
+                device=device_to_use,
+                for_debug=for_debug,
+                on_pattern_query_start=_on_pattern_query_start_for_ep,
+                on_pattern_query_result=_on_pattern_query_result_for_ep,
+            )
+            # In single-EP mode, keep metadata pattern counts aligned with the
+            # selected EP summary while still avoiding duplicate upfront extraction.
+            # Also tolerate minimal test doubles that don't stub model_summary().
+            if ep_normalized is not None or not isinstance(metadata, ModelStats):
```
DingmaomaoBJTU: In all-EP mode (`ep` is `None`), this condition leaves `metadata` as the initial `pattern_extractor.model_summary()` from line 1120, which has an empty `detected_pattern_count`. Each EP summary computes pattern counts, but none are copied into the output metadata, so `AnalysisResult.output.metadata.detected_pattern_count` regresses to `{}` for public calls that analyze all EPs.

Can we populate metadata from the first `ep_pattern_summary["summary"]` or aggregate the counts across EP summaries in the multi-EP path too?

src/winml/modelkit/analyze/analyzer.py
fangyangci: This is intentional. Pattern matches are deduplicated by covered nodes before their parquet rules are queried. If an `unknown` pattern were removed afterward, previously discarded overlapping patterns might need to be restored, which would require backtracking and rerunning deduplication. Simply falling back to per-op checks would therefore leave the pattern selection inconsistent.

Pattern coverage is expected to be complete. Here, `unknown` means that a valid matched pattern is missing runtime rule data for the EP/device, not that the match is invalid. We want that data gap to remain visible and be fixed at the pattern level rather than silently falling back to operator rules, which cannot establish support for the complete subgraph.

src/winml/modelkit/analyze/analyzer.py
fangyangci: Good catch. Pattern counts are EP-specific because pattern sources and node deduplication can differ by EP. Using the first EP would be arbitrary, while aggregating counts may count the same model pattern more than once. I kept `metadata.detected_pattern_count` in its existing location and changed its value to `EP -> pattern ID -> count`, so callers can read one EP total with `sum(metadata.detected_pattern_count[ep].values())`.

fangyangci: Implemented in `67e62f3f`. `metadata.detected_pattern_count` now preserves the existing field location while storing counts as `EP -> pattern ID -> count`. The analyzer merges every EP summary in all-EP mode, so callers can read an individual EP total with `sum(metadata.detected_pattern_count[ep].values())` without selecting an arbitrary first EP or exposing a misleading cross-EP aggregate. Updated the schema, analyzer/aggregator, docs, fixtures, and multi-EP coverage; the affected 145 tests pass.

DingmaomaoBJTU (changes requested):
I don't think this is ready to approve yet. This change removes the `winml analyze --htp-metadata` CLI option and the corresponding `ONNXStaticAnalyzer.analyze(..., htp_metadata_path=...)` plumbing, but the public docs still document the option and the behavior:

- `docs/commands/analyze.md` still lists `--htp-metadata` and includes examples using it.
- `docs/concepts/analyze-and-optimize.md` still says HTP metadata is consumed by `PatternExtractor`.
- In the new `src/winml/modelkit/commands/analyze.py`, the analyze command no longer declares `--htp-metadata`, so existing documented invocations now fail with an unknown option.

Please either restore the option/API path and wire it into the new pattern-rule flow, or intentionally remove/deprecate the feature and update the docs/tests accordingly. As-is this is a documented CLI/API regression.

src/winml/modelkit/commands/analyze.py
```diff
@@ -879,12 +1104,6 @@ def _build_runtime_debug_output_path(model_path: Path, ep_name: str, device_name
     default=True,
     help="Include detailed recommendations (default: enabled)",
 )
-@click.option(
-    "--htp-metadata",
```
DingmaomaoBJTU, line 883: This option is still documented and appears in the public command reference/examples, but this PR removes the CLI option and the analyzer API plumbing. Existing documented invocations like `winml analyze ... --htp-metadata ...` will now fail with "No such option". Please either restore this option and wire it into the new pattern-rule path, or intentionally remove/deprecate it and update the docs/tests in the same change.

src/winml/modelkit/analyze/analyzer.py
```diff
@@ -854,17 +1238,117 @@ def analyze_from_proto(
...
+                    )
+                    on_ep_start(
+                        current_ep,
+                        op_counts_for_display,
+                        not pattern_lookup_supported,
+                    )
                 except Exception:
                     logger.debug("on_ep_start callback failed", exc_info=True)
+
+            if not pattern_lookup_supported:
```
DingmaomaoBJTU: This skip also bypasses `--run-unknown-op`: once `pattern_lookup_supported` is false, we never construct `RuntimeChecker` or reach the `runtime_checker.summary(... run_unknown_op=...)` path below. That means a valid local EP/device target with pattern parquet lookup disabled cannot use local probing for unknown ops even when the user explicitly requested it. Pattern parquet lookup support should be decoupled from op runtime checking; if `run_unknown_op` is true, this should still run the runtime checker, likely without the pattern short-circuit map.

src/winml/modelkit/commands/analyze.py
fangyangci: Fixed in 2a57b29c. Removed the obsolete analyzer `--htp-metadata` option reference, example, and related analyzer concept text. Export-side HTP metadata documentation remains because that export behavior is still supported.

src/winml/modelkit/analyze/analyzer.py
fangyangci: Fixed in 2a57b29c. Parquet lookup is now decoupled from local runtime probing: when `--run-unknown-op` is enabled, an unknown matched base pattern is checked locally as its complete subgraph before remaining unmatched nodes proceed to op-level probing. Pattern ownership and node deduplication are preserved, alternatives remain parquet-only, and the existing VitisAI guard remains. Added coverage for no-parquet behavior, pattern-first ordering, progress reporting, and runner cleanup.

tests/unit/analyze/runtime_checker/test_ep_checker.py
```diff
@@ -3,7 +3,12 @@
 # Licensed under the MIT License.
 # --------------------------------------------------------------------------
 
+from pathlib import Path
+from unittest.mock import MagicMock, patch
+
+import onnx
```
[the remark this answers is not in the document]
fangyangci: Resolved by reverting the superseded EPChecker fallback test changes. The current test file no longer imports `onnx` using both `import onnx` and `from onnx import ...`, so the flagged mixed-import pattern is no longer present. The thread is now outdated.

src/winml/modelkit/commands/analyze.py
DingmaomaoBJTU: Removing the docs reference fixes the stale documentation part, but it does not fix the compatibility regression: this PR still removes the public `winml analyze --htp-metadata` CLI option and the `ONNXStaticAnalyzer.analyze(..., htp_metadata_path=...)` API plumbing that exist on `main`. Existing scripts/callers using the released option will now fail at option/API parsing. Please either restore the option/API path (possibly deprecated) or move this removal into an explicitly accepted breaking-change PR.

DingmaomaoBJTU: Given the current direction to fully remove analyze-side HTP metadata support, I am no longer treating the compatibility impact as blocking. With that product decision, removing the CLI/API/docs/tests together is consistent.

DingmaomaoBJTU (approved):
Approved latest head 1fad6b9f. The previous `--run-unknown-op` blocker is addressed, CI is green, and the HTP metadata removal is accepted as an intentional full removal rather than a compatibility blocker.

PR #1221, opened 2026-07-27
feat: Add --input-data to winml eval --mode compare

xieofxie:
## Summary

Adds `--input-data` to `winml eval --mode compare`, letting you compare an ONNX candidate against its HuggingFace reference on **real input tensors** from a `.npz` archive instead of randomly generated ones — mirroring `winml perf`'s `--input-data`.

The archive keys must match the candidate model's input names and are validated/dtype-cast against them. The **leading axis of each array is the sample axis**, so an archive shaped `(N, ...)` runs `N` samples and the similarity table reflects a real distribution (mean/std/min/max); all inputs must share the same `N`. The flag is only valid with `--mode compare`.

```bash
winml eval --mode compare -m cand.onnx --model-id microsoft/resnet-50 --input-data inputs.npz
```

## Changes

- **`datasets/input_data.py`** (new): shared `load_input_data` (extracted from `perf`) plus a multi-sample `InputDataDataset` wrapper (leading axis = sample axis, each sample sliced to a batch of 1).
- **`commands/perf.py`**: `load_input_data` now delegates to the shared module via a thin lazy import (same public symbol, unchanged behavior).
- **`eval/config.py`**: new `input_data` field with `to_dict`/`from_dict` support.
- **`commands/eval.py`**: `--input-data` option, compare-mode guard (`UsageError` when `mode != compare`, checked before model resolution), docstring example, and the report header now shows `Input data` / the real sample count.
- **`eval/tensor_similarity_evaluator.py`**: `prepare_data` builds `InputDataDataset` when `input_data` is set and propagates `len(dataset)` into the effective config so the report/JSON reflect the real `N`.
- **`eval/evaluate.py`**: `print_config` prints the input-data path in the config summary.
- **docs**: `docs/commands/eval.md` flag row + example.
- **tests**: new `tests/unit/datasets/test_input_data.py`; input-data cases added to the eval command/config/evaluator test suites; `test_perf_cli.py` logger target retargeted to the shared module.

## Notes

This cherry-picks **only** the `--input-data` parts of #1139, decoupled from that PR's two-ONNX `--reference` feature (not included here). The guard runs before model resolution so it fires cleanly even without `--model-id`.

### DIFF
--- a/src/winml/modelkit/datasets/input_data.py
+++ b/src/winml/modelkit/datasets/input_data.py
@@ -1,0 +1,187 @@
+# -------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+# --------------------------------------------------------------------------
+
+"""Real input tensors loaded from a ``.npz`` archive.
+
+Shared by ``winml perf`` (benchmark on real tensors instead of random ones)
+and ``winml eval --mode compare`` (compare a candidate and reference on the
+same real inputs). :func:`load_input_data` validates and dtype-casts the
+archive against a model's I/O config; :class:`InputDataDataset` wraps the
+loaded archive as a torch dataset (leading axis = sample axis) the compare
+loop can iterate.
+"""
+
+from __future__ import annotations
+
+import logging
+from pathlib import Path
+from typing import TYPE_CHECKING, Any
+
+import click
+import numpy as np
+
+
+if TYPE_CHECKING:
+    import torch
+
+logger = logging.getLogger(__name__)
+
+
+def load_input_data(
+    path: Path,
+    io_config: dict[str, Any],
+) -> dict[str, np.ndarray]:
+    """Load model inputs from a ``.npz`` file, validated against the model.
+
+    Lets ``winml perf`` / ``winml eval`` run with real input tensors instead
+    of randomly generated ones. Only ``.npz`` (a named-array archive) is
+    supported today; a single-array ``.npy`` carries no input names to bind
+    against and is rejected with guidance to repackage as ``.npz``.
+
+    Validation:
+
+    * the archive's keys must exactly match the model's input names -- any
+      missing or unexpected key is an error (an unexpected key is usually a
+      typo that would otherwise leave a required input silently unset);
+    * an array whose dtype differs from the model's expected input dtype is
+      cast to the expected dtype with a warning, matching the silent casting
+      ``WinMLSession._prepare_inputs`` does on a normal run (e.g. numpy's
+      default int64 literals binding to an int32 input).
+
+    Shapes are taken from the arrays as-is; correctness beyond dtype (e.g. a
+    static dimension the data violates) surfaces as a runtime error from the
+    inference session.
+
+    Args:
+        path: Path to the ``.npz`` file.
+        io_config: Model I/O configuration (``input_names``, ``input_types``).
+
+    Returns:
+        Dictionary of ``input_name -> numpy array``.
+
+    Raises:
+        click.UsageError: On a non-``.npz`` file or a key mismatch.
+    """
+    path = Path(path)
+    if path.suffix.lower() == ".npy":
+        raise click.UsageError(
+            f"--input-data does not support .npy files ({path.name}). A single "
+            f"array carries no input names; save your inputs as a named .npz "
+            f"archive instead (e.g. np.savez('inputs.npz', input_ids=..., "
+            f"attention_mask=...))."
+        )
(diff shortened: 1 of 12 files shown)


src/winml/modelkit/datasets/input_data.py
```diff
@@ -0,0 +1,187 @@
...
+        """Number of samples (the shared leading-axis length of the inputs)."""
+        return self._num_samples
+
+    def __getitem__(self, idx: int) -> dict[str, torch.Tensor]:
+        """Return sample ``idx`` with each input sliced to a batch of 1."""
+        if not 0 <= idx < self._num_samples:
+            raise IndexError(
+                f"InputDataDataset index {idx} out of range for {self._num_samples} samples."
+            )
+        return {name: tensor[idx : idx + 1] for name, tensor in self._arrays.items()}
```
zhenchaoni: The semantics here differ from `perf`'s `--input-data`, despite the "mirroring perf" wording: this treats the leading axis as the *sample* axis and forces every sample to batch=1 (`arr[i:i+1]`), while `perf` runs the whole `.npz` as a single batch. Two concerns:

1. Same `.npz`, different interpretation between `perf` and `eval` — worth calling out explicitly in the docs so users don't reuse an archive expecting identical behavior.
2. `RandomDataset` sizes the batch dim from the ONNX I/O spec (static dims preserved, dynamic -> 1), so a static-batch candidate (e.g. fixed batch=4) works on the random path but will shape-mismatch here. Consider honoring the model's batch dim instead of unconditionally slicing to 1 (or reject/handle static-batch explicitly).

src/winml/modelkit/eval/tensor_similarity_evaluator.py
```diff
@@ -80,7 +83,23 @@ def _load_reference_model(self) -> Any:
...
+        :class:`RandomDataset` of synthetic inputs sized by ``config.dataset``.
+        """
+        if self.config.input_data is not None:
+            from ..datasets.input_data import InputDataDataset
+
+            dataset = InputDataDataset(self.config.input_data, self.model.io_config)
+            # Reflect the real sample count (leading axis of the .npz) in the
+            # effective config so the report header / JSON show N, not the
+            # unused dataset default.
+            self.config.dataset.samples = len(dataset)
```
zhenchaoni: Two things on this line:

1. Mutating `self.config` breaks the immutability contract `evaluate()` documents (it deep-copies config specifically to avoid in-place mutation). It's safe today because `dataset` is deep-copied, but prefer propagating the real N to the report without writing it back into config.
2. This silently overrides a user-provided `--samples` (or config `samples`) with the archive's N. Since it discards an explicit user value, emit a warning — consistent with how eval warns for other ignored options.

zhenchaoni (commented):
A couple of implementation notes on using real inputs for tensor similarity. The main link (feed the same inputs to the ONNX candidate and the HF reference, run per-sample, accumulate the similarity metrics) looks correct; comments are about the input semantics and a config side-effect.

src/winml/modelkit/datasets/input_data.py
xieofxie: yes, for 1 worth calling out as perf only need one sample

yes. we should also check perf like input model has batch 4 and input data is 8, we could just use [0:4] for that

xieofxie: Done in 9594ecdc.

- Static batch is now honored: `InputDataDataset` reads the leading (batch) axis of the candidate's `input_shapes` via a new `_model_batch_size` helper and chunks the sample axis into batches of `B` (`arr[i*B:(i+1)*B]`), yielding `N // B` samples. A dynamic batch dim still runs one row per sample, `N < B` is an error, a non-multiple `N` drops the trailing rows with a warning, and inputs declaring conflicting static batch sizes are rejected.
- Documented the perf-vs-eval difference (perf runs the whole archive as a single batch; eval treats the leading axis as the sample axis and chunks it to the model's batch) in the module docstring and `docs/commands/eval.md`.

src/winml/modelkit/eval/tensor_similarity_evaluator.py
xieofxie: Done in 9594ecdc.

- `prepare_data` no longer mutates `config`. The effective sample count is now carried on a new `EvalResult.num_samples` field (computed in `evaluate()` from the dataset length), and both the console report and `to_dict()`/JSON read it instead of writing back into the deep-copied config.
- `eval()` now warns when `--samples` is passed together with `--input-data`, since the count comes from the archive's leading axis (mirrors perf's `--batch-size` warning).

zhenchaoni (approved):

PR #1228, opened 2026-07-27
Fix analyze default auto target selection

DingmaomaoBJTU:
Default `winml analyze` should choose the strongest exact local EP/device target, not inherit a lower-priority device axis from the session fallback before EP selection. On hosts where one provider exposes both NPU and GPU bindings, the previous auto/auto path could settle on the GPU binding even when the NPU binding was locally available.

This updates analyze's default target selection to rank exact local EP/device bindings by `DEVICE_PRIORITY` before resolving the EP. Explicit targets and existing fan-out behavior are preserved. The regression test derives a multi-device EP from `EP_SUPPORTED_DEVICES` so the coverage follows the provider matrix rather than hardcoding a provider-specific rule.

Validation:
- `ruff check src\winml\modelkit\commands\analyze.py tests\unit\analyze\test_static_analyzer_cli.py`
- `pytest tests\unit\analyze\test_static_analyzer_cli.py -q`

### DIFF
--- a/src/winml/modelkit/commands/analyze.py
+++ b/src/winml/modelkit/commands/analyze.py
@@ -38,6 +38,7 @@ from rich.text import Text
 from ..session import DEVICE_TYPE_TO_DEVICE
 from ..utils import cli as cli_utils
 from ..utils.constants import (
+    DEVICE_PRIORITY,
     EP_SUPPORTED_DEVICES,
     SUPPORTED_DEVICES,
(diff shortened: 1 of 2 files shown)


xieofxie (approved):
The default auto/auto path now ranks exact local bindings by the shared device priority without changing explicit targets or fan-out behavior. The provider-matrix-derived regression test is appropriately architecture-agnostic.

PR #1232, opened 2026-07-27
Hide inspect warning noise

DingmaomaoBJTU:
## Summary
- Hide Hugging Face, Transformers, and native ONNX Runtime warning noise during normal `winml inspect` runs.
- Preserve diagnostics through `-v`, `-vv`, and `WINMLCLI_SHOW_ALL_WARNINGS=1`.
- Scope Hugging Face warning suppression to inspect and restore logger/env/library verbosity state afterward.
- Add regression coverage for suppression, opt-outs, native stderr behavior, and state restoration.

## Validation
- `python -m ruff check src\winml\modelkit\commands\inspect.py src\winml\modelkit\utils\logging.py src\winml\modelkit\utils\native_stderr.py tests\cli\test_import_time.py tests\unit\commands\test_inspect_cli.py tests\unit\utils\test_logging.py tests\unit\utils\test_native_stderr.py`
- `python -m pytest tests\unit\utils\test_logging.py tests\unit\commands\test_inspect_cli.py tests\unit\utils\test_native_stderr.py tests\cli\test_import_time.py::TestCommandHelp::test_inspect_model_type_hides_onnxruntime_startup_warning -q`
- `python -m winml.modelkit inspect -m facebook/convnext-tiny-224` default / `-v` / `WINMLCLI_SHOW_ALL_WARNINGS=1` smoke checks

### DIFF
--- a/src/winml/modelkit/commands/inspect.py
+++ b/src/winml/modelkit/commands/inspect.py
@@ -16,6 +16,7 @@ Usage:
 
 from __future__ import annotations
 
+import importlib
 import logging
 from typing import TYPE_CHECKING
 
@@ -297,6 +298,8 @@ def inspect(
             from ..inspect import InspectError, ModelNotFoundError, NetworkError
             from ..inspect.formatter import output_json, output_table
 
+            _load_inspect_model_v2_dependencies()
+
         try:
             if quiet or json_mode:
                 result = _inspect_model_v2(
@@ -305,7 +308,7 @@ def inspect(
                     model_type_override=model_type,
                     model_class_override=model_class,
                     include_hierarchy=hierarchy,
-                    suppress_native_stderr_output=suppress_third_party_stderr,
+                    suppress_native_stderr_output=False,
                 )
             else:
                 with _stderr_console.status(
@@ -318,7 +321,7 @@ def inspect(
                         model_type_override=model_type,
                         model_class_override=model_class,
                         include_hierarchy=hierarchy,
-                        suppress_native_stderr_output=suppress_third_party_stderr,
+                        suppress_native_stderr_output=False,
                     )
 
             if output_format == "json":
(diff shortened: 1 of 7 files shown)


src/winml/modelkit/utils/logging.py
```diff
@@ -98,14 +110,118 @@ def configure_logging(
...
+        for env_name in _HUGGINGFACE_VERBOSITY_ENVS:
+            os.environ[env_name] = library_verbosity
+        _sync_imported_huggingface_verbosity(huggingface_level)
+        yield
+    finally:
+        for env_name, value in saved_env.items():
+            if value is _MISSING:
+                os.environ.pop(env_name, None)
+            else:
+                os.environ[env_name] = value
```
xieofxie, line 155: This assignment currently fails the required mypy job because saved_env is inferred as dict[str, object] from the sentinel. Please narrow value to str (or type the saved-state representation so the sentinel and string cases discriminate) before assigning it back to os.environ.

src/winml/modelkit/commands/inspect.py
```diff
@@ -288,50 +290,54 @@ def inspect(
     if not quiet and not json_mode:
         _stderr_console.print(f"[dim]Inspecting [bold]{target}[/bold] …[/dim]")
 
-    from ..inspect import InspectError, ModelNotFoundError, NetworkError
-    from ..inspect.formatter import output_json, output_table
-
+    suppress_third_party_stderr = not verbose and not env_flag_enabled("WINMLCLI_SHOW_ALL_WARNINGS")
     configure_logging(verbosity=verbose, quiet=quiet)
+    with suppress_huggingface_warning_logs(verbosity=verbose, quiet=quiet):
```
xieofxie, line 296: On Windows this redirects fd 2 for the entire inspection, including the _stderr_console.status(...) block and WinML logging below. That means the existing resolving spinner and any application warnings are discarded together with ORT noise in normal mode. Can we scope fd-level suppression to the third-party/native operations so command-owned stderr remains visible?

xieofxie (changes requested):
Two issues need addressing before merge.

src/winml/modelkit/utils/logging.py
DingmaomaoBJTU: Fixed by replacing the object sentinel with `None` for the saved env state, so mypy narrows restored values to `str` before assigning back to `os.environ`. Local `uv run mypy -p winml.modelkit` passes.

src/winml/modelkit/commands/inspect.py
DingmaomaoBJTU: Fixed by preloading native-warning-prone inspect dependencies under fd-level suppression before entering the Rich status context, then calling the core inspect path with native stderr suppression disabled while status is active. Added a regression that fails if native stderr suppression is entered while the status context is active.

xieofxie (approved):
Re-reviewed the two follow-up commits. The environment restore now type-narrows cleanly, and native stderr suppression is limited to dependency preloading so the Rich status and command-owned diagnostics remain visible. The added regressions cover both paths, and all required checks are green.

PR #1235, opened 2026-07-28
fix(export): accept traced 0-dim tensors as KV cache dimensions

KayMKM:
## Problem

`TestBuildT5Composite::test_composite_fanout` and
`TestExportT5Composite::test_composite_fanout` fail on **every** execution
provider (reproduced on both the QNN and OpenVINO E2E jobs of run
`20260727.1`):

```
File "src/winml/modelkit/models/hf/t5.py", line 164, in forward
    self_attn_cache.early_initialization(
File "transformers/cache_utils.py", line 1394, in early_initialization
    if len(num_heads) != len(self.layers):
File "optimum/exporters/onnx/model_patcher.py", line 515, in <lambda>
    PatchingSpec(torch.Tensor, "__len__", lambda x: x.shape[0], torch.Tensor.__len__),
IndexError: tuple index out of range
```

## Root cause

Decoder export wrappers derive the cache dimensions from the past-KV graph
inputs with `Tensor.size(dim)`. Under the TorchScript tracer used by
`torch.onnx.export`, that call returns a **0-dim tensor** rather than an
`int` (verified directly with `torch.jit.trace`).

`Cache.early_initialization` therefore skips its `isinstance(..., int)`
broadcast and falls through to `len(num_heads)`. Optimum's model patcher
replaces `torch.Tensor.__len__` with `lambda x: x.shape[0]`, and a 0-dim
tensor has an empty shape, so the lookup raises `IndexError`.

## Fix

Override `early_initialization` on `WinMLCache` to normalize the shape
arguments back to Python ints before delegating upstream. These dimensions
are static for an exported graph, so the conversion is lossless.

Placing the override on the base class covers every decoder wrapper (T5,
BART, Marian, Qwen, Mu2 and the shared decoder adapter) and any future one,
with no model-specific handling.

## Verification

- `tests/e2e/test_export_e2e.py::TestExportT5Composite::test_composite_fanout` — failed before, passes after
- `tests/e2e/test_build_e2e.py::TestBuildT5Composite` — 3/3 pass
- Full unit suite — 7820 passed, 77 skipped, 3 xfailed
- New regression test `TestWinMLCacheEarlyInitialization` covers both cache classes with 0-dim tensor dimensions

### DIFF
--- a/src/winml/modelkit/models/winml/kv_cache.py
+++ b/src/winml/modelkit/models/winml/kv_cache.py
@@ -55,6 +55,17 @@ if TYPE_CHECKING:
     from transformers.cache_utils import CacheLayerMixin
 
 
+def _as_static_dim(value: Any) -> Any:
+    """Normalize a (possibly traced) shape value to Python ``int``/``list[int]``.
+
+    ``Tensor.size(dim)`` returns a 0-dim tensor while TorchScript tracing, which
+    breaks callers that type-check for ``int``.
+    """
+    if isinstance(value, (list, tuple)):
+        return [int(item) for item in value]
+    return int(value)
+
+
 # =============================================================================
 # WinMLCache — common interface
 # =============================================================================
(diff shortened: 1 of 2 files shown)


DingmaomaoBJTU (approved):
LGTM — clean fix normalizing traced 0-dim tensors at the WinMLCache boundary. Minor suggestions (non-blocking): consider explicit torch.Tensor handling in _as_static_dim for robustness, and asserting the full shape (including seq_len dim) in the test.

DingmaomaoBJTU (approved):
LGTM — clean fix normalizing traced 0-dim tensors at the WinMLCache boundary. Minor suggestions (non-blocking): consider explicit torch.Tensor handling in _as_static_dim for robustness, and asserting the full shape (including seq_len dim) in the test.

PR #1238, opened 2026-07-28
feat(optimize/analyze): preview applicable optimizations and check their produced-operator support

xieofxie:
## Summary

Two related, opt-in features for reasoning about optimizations before applying them, sharing one consistent flag name `--check-optim`:

1. **`winml optimize --check-optim`** — answers "which optimizations actually apply to *this* model, and on which nodes?" without writing any output.
2. **`winml analyze --check-optim`** — for each optimization that would change the model, checks whether the operators it *introduces* are supported on the resolved EP/device.

### `optimize --check-optim`

Probes **every** registered optimization capability independently against the given model and reports only the ones that would genuinely change it, along with the specific nodes and constants affected.

```bash
$ winml optimize -m model.onnx --check-optim
2 applicable optimization(s):

--enable-matmul-add-fusion  (matmul)
  Fuse MatMul+Add operations into single kernel
  nodes: -1 ~1
    removed: MatMul (1)
    modified: Gemm (1)

--enable-clamp-constant-values  (surgery)
  constants: ~1
```

Add `-v` to list every affected node/constant instead of a sample. `--check-optim` requires `--model` and writes nothing to disk.

### `analyze --check-optim`

Reuses the optimize probe engine to obtain the model each optimization would produce, then feeds the produced (added/modified) nodes into analyze's `RuntimeChecker` and classifies them against the target EP/device rule data. Answers "if I apply this optimization, will its output run on my target?".

```bash
$ winml analyze --model model.onnx --ep qnn --device NPU --check-optim
```

Renders a per-target **OPTIMIZATION OUTPUT SUPPORT** section listing each applicable optimization, its `--enable-*` flag, and the support level (supported / partial / unsupported / unknown) of every operator it would introduce. The probe is target-independent, so it is materialized once and reused across all resolved targets; only the support lookup runs per EP/device. Disabled by default; console-only.

## How it works (universal, no hardcoding)

The probe mirrors `Optimizer.optimize()`: run mandatory shape inference, then walk the pipes in order. For each pipe it computes a baseline output, then re-runs the pipe on a **clone** of the upstream model with only one default-off capability enabled (plus auto-enabled dependencies) and diffs the result:

- Node diff is keyed by **output tensor names** (unique and stable within a valid ONNX graph across input rewiring / attribute changes), yielding removed / added / modified buckets; recurses into subgraphs.
- Initializer diff catches constant-only surgery ops (e.g. `clamp-constant-values`) that modify no nodes.
- Cloning uses `CopyFrom` (not a serialize round-trip) because saving can convert tensors to external data in place, and to avoid the 2 GB protobuf limit.
- Each probe is isolated in try/except so a single failing capability is logged and skipped rather than aborting the run.

The support check correlates produced nodes back to `op_support` results by output-tensor name, and `RuntimeChecker.op_support(node_output_filter=...)` restricts the check to just the produced nodes. No op/tensor/architecture names are hardcoded (CLAUDE.md Cardinal Rule #1).

## Changes

- **New** `src/winml/modelkit/optim/analysis.py` — `analyze_model`, `iter_optimization_outputs`, `CapabilityFinding`, `NodeRef` and diff helpers (shared `_iter_findings` generator)
- `src/winml/modelkit/optim/__init__.py` — export the new public API
- `src/winml/modelkit/commands/optimize.py` — `--check-optim` option + report renderer (with rich-markup escaping of dynamic labels)
- **New** `src/winml/modelkit/analyze/optim_output.py` — bridge that checks produced-operator EP/device support
- `src/winml/modelkit/analyze/core/runtime_checker.py` — `op_support(node_output_filter=...)` to restrict the check to a node subset
- `src/winml/modelkit/commands/analyze.py` — `--check-optim` flag + per-target renderer
- **New/updated tests** — `tests/unit/optim/test_analysis.py`, `tests/unit/analyze/test_optim_output.py`, and CLI coverage in `tests/unit/commands/test_optimize_cli.py` and `tests/unit/analyze/test_static_analyzer_cli.py`
- `docs/commands/optimize.md` and `docs/commands/analyze.md` — flag-table rows + worked examples

## Validation

- `ruff` clean; `mypy -p winml.modelkit` clean; doc/code flag-alignment check ✅
- Targeted `optim` + `analyze` suites pass (hardware/fixture-gated skips and one pre-existing xfail only)

### DIFF
--- a/src/winml/modelkit/analyze/optim_output.py
+++ b/src/winml/modelkit/analyze/optim_output.py
@@ -197,6 +197,11 @@ def _matched_node(runtime: object) -> NodeProto | None:
     matched_nodes = getattr(skeleton, "matched_nodes", None)
     if matched_nodes:
         return cast("NodeProto", matched_nodes[0])
+    logger.debug(
+        "Could not extract matched node from runtime (type=%s); produced operators "
+        "correlated via this runtime will report UNKNOWN support",
+        type(runtime).__name__,
+    )
     return None
 
 
--- a/src/winml/modelkit/analyze/optim_output.py
+++ b/src/winml/modelkit/analyze/optim_output.py
@@ -1,0 +1,239 @@
+# -------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+# --------------------------------------------------------------------------
+"""EP/device support of the operators an optimization would produce.
+
+This bridges the optimizer's dry-run applicability analysis with the analyze
+runtime-support rule data. For each applicable optimization it takes the model
+that optimization would produce, then checks whether the operators the
+optimization *introduces* (its added and modified nodes) are supported on a
+target execution provider and device — answering "if I apply this optimization,
+will its output run on my target?".
+
+The check is universal and architecture-agnostic (CARDINAL RULE #1): the set of
+produced operators is derived entirely from the concrete graph diff carried on
+each :class:`~winml.modelkit.optim.CapabilityFinding`; nothing about specific
+operators or model architectures is hardcoded.
+"""
+
+from __future__ import annotations
+
+import logging
+from collections import Counter
+from dataclasses import dataclass, field
+from typing import TYPE_CHECKING, cast
+
+from .models.onnx_model import ONNXModel
+from .models.support_level import SupportLevel
+
+
+if TYPE_CHECKING:
+    from collections.abc import Iterable
+
+    from onnx import ModelProto, NodeProto
+
+    from ..optim import CapabilityFinding, NodeRef
+    from ..utils.constants import EPName
+
+
+logger = logging.getLogger(__name__)
+
+# Ordering used to summarize a mix of per-operator support levels into a single
+# "worst" level. Higher rank = more concerning for the target.
+_SEVERITY: dict[SupportLevel, int] = {
+    SupportLevel.SUPPORTED: 0,
+    SupportLevel.UNKNOWN: 1,
+    SupportLevel.PARTIAL: 2,
+    SupportLevel.UNSUPPORTED: 3,
+}
+
+
+@dataclass(frozen=True)
+class ProducedOperatorSupport:
+    """Support of a single operator an optimization would introduce.
+
+    Attributes:
+        op_type: The produced operator's op type (e.g. ``"Gemm"``).
+        label: Human-readable node label (op type + name/output).
+        change: Whether the node is ``"added"`` or ``"modified"`` by the
+            optimization.
+        support: Support level of the operator on the target EP/device.
+        reason: Optional reason string from the runtime check.
+    """
+
+    op_type: str
+    label: str
+    change: str
+    support: SupportLevel
+    reason: str | None = None
+
+
+@dataclass
+class OptimizationOutputSupport:
+    """Per-optimization view of its produced operators' target support.
+
+    Attributes:
+        name: Capability name (kebab-case, e.g. ``"matmul-add-fusion"``).
+        enable_flag: CLI flag that turns the optimization on.
+        category: Capability category (e.g. ``"matmul"``).
+        description: Human-readable capability description.
+        pipe_name: Name of the pipe that owns the capability.
+        operators: Per-operator support for every produced (added/modified) node.
+        error: Non-empty when the support check could not be completed.
+    """
+
+    name: str
+    enable_flag: str
+    category: str
+    description: str
+    pipe_name: str
+    operators: list[ProducedOperatorSupport] = field(default_factory=list)
+    error: str | None = None
+
+    @property
+    def worst_support(self) -> SupportLevel:
+        """Most concerning support level across produced operators.
+
+        Returns ``SUPPORTED`` when there are no produced operators to check.
+        """
+        if not self.operators:
+            return SupportLevel.SUPPORTED
+        return max((op.support for op in self.operators), key=lambda level: _SEVERITY[level])
+
+    def support_counts(self) -> dict[SupportLevel, int]:
+        """Return a count of produced operators per support level."""
+        return dict(Counter(op.support for op in self.operators))
+
+
+def _produced_node_refs(
+    finding: CapabilityFinding,
+) -> list[tuple[str, NodeRef]]:
+    """Return ``(change_kind, NodeRef)`` pairs for a finding's produced nodes.
+
+    Produced nodes are those an optimization introduces: newly added nodes and
+    nodes modified in place. Removed nodes are excluded — they no longer exist
+    after the optimization, so their target support is irrelevant.
+    """
+    pairs: list[tuple[str, NodeRef]] = []
+    pairs.extend(("added", ref) for ref in finding.added_nodes)
+    pairs.extend(("modified", ref) for ref in finding.modified_nodes)
+    return pairs
+
+
+def _check_one(
+    finding: CapabilityFinding,
+    produced_model: ModelProto,
+    ep: EPName,
+    device: str,
+    model_path: str,
+) -> OptimizationOutputSupport:
+    """Check target support for the operators one optimization would produce."""
+    from .core.runtime_checker import RuntimeChecker
+
+    result = OptimizationOutputSupport(
+        name=finding.name,
+        enable_flag=finding.enable_flag,
+        category=finding.category,
+        description=finding.description,
+        pipe_name=finding.pipe_name,
+    )
+
+    produced = _produced_node_refs(finding)
+    if not produced:
+        return result
+
+    # Output tensor names uniquely identify produced nodes within the graph.
+    produced_outputs = {out for _, ref in produced for out in ref.outputs}
+
+    try:
+        onnx_model = ONNXModel.from_onnx_model(produced_model, model_path)
+        checker = RuntimeChecker(ep=ep, device=device, model=onnx_model)
+        # Only evaluate the produced nodes; skip the rest of the graph.
+        runtimes = checker.op_support(
+            node_output_filter=produced_outputs,
+            on_node_result=lambda _r: None,
+        )
+    except Exception as exc:  # pragma: no cover - defensive, keeps analyze alive
+        logger.warning(
+            "Could not check produced-operator support for '%s' on %s/%s: %s",
+            finding.name,
+            ep,
+            device,
+            exc,
+        )
+        result.error = str(exc)
+        return result
+
+    # Map each produced output tensor to its support result.
+    support_by_output: dict[str, tuple[SupportLevel, str | None]] = {}
+    for runtime in runtimes:
+        node = _matched_node(runtime)
+        outputs = tuple(node.output) if node is not None else ()
+        classification = runtime.result.classification
+        reason = runtime.result.reason
+        for out in outputs:
+            support_by_output[out] = (classification, reason)
+
+    for change, ref in produced:
+        support, reason = _lookup_support(ref, support_by_output)
+        result.operators.append(
+            ProducedOperatorSupport(
+                op_type=ref.op_type,
+                label=ref.label(),
+                change=change,
+                support=support,
+                reason=reason,
+            )
+        )
+
+    return result
+
+
+def _matched_node(runtime: object) -> NodeProto | None:
+    """Best-effort extraction of the ONNX node behind a ``PatternRuntime``."""
+    pattern_match = getattr(runtime, "pattern_match", None)
+    skeleton = getattr(pattern_match, "skeleton_match_result", None)
+    matched_nodes = getattr(skeleton, "matched_nodes", None)
+    if matched_nodes:
+        return cast("NodeProto", matched_nodes[0])
+    return None
+
+
+def _lookup_support(
+    ref: NodeRef,
+    support_by_output: dict[str, tuple[SupportLevel, str | None]],
+) -> tuple[SupportLevel, str | None]:
+    """Resolve a produced node's support via any of its output tensors."""
+    for out in ref.outputs:
+        if out in support_by_output:
+            return support_by_output[out]
+    return SupportLevel.UNKNOWN, "not evaluated"
+
+
+def check_optimization_output_support(
+    optimization_outputs: Iterable[tuple[CapabilityFinding, ModelProto]],
+    ep: EPName,
+    device: str,
+    model_path: str,
+) -> list[OptimizationOutputSupport]:
+    """Check produced-operator support for applicable optimizations on a target.
+
+    Args:
+        optimization_outputs: ``(finding, produced_model)`` pairs from
+            :func:`winml.modelkit.optim.iter_optimization_outputs`. Materialize
+            the iterator once and reuse it across targets — the dry-run that
+            produces these pairs is target-independent, so only the support
+            lookup needs to run per EP/device.
+        ep: Target execution provider (canonical name).
+        device: Target device string (e.g. ``"NPU"``).
+        model_path: Path to the source model, used for external-data resolution.
+
+    Returns:
+        One :class:`OptimizationOutputSupport` per applicable optimization, in
+        the same order as ``optimization_outputs``.
+    """
+    return [
+        _check_one(finding, produced_model, ep, device, model_path)
+        for finding, produced_model in optimization_outputs
+    ]
--- a/src/winml/modelkit/commands/analyze.py
+++ b/src/winml/modelkit/commands/analyze.py
@@ -914,7 +914,7 @@ def _build_runtime_debug_output_path(model_path: Path, ep_name: str, device_name
     help="Save auto-discovered optimization config to JSON file",
 )
 @click.option(
-    "--check-optim-output/--no-check-optim-output",
+    "--check-optim/--no-check-optim",
     default=False,
     help=(
         "For each optimization that would change the model, check whether the "
@@ -940,7 +940,7 @@ def analyze(
     debug: bool,
     save_node: tuple[str, ...],
     optim_config: Path | None,
-    check_optim_output: bool,
+    check_optim: bool,
 ) -> None:
     r"""Analyze ONNX model for runtime support with live progress.
 
@@ -1194,11 +1194,11 @@ def analyze(
         console = Console(stderr=True)
 
         # Optionally probe which optimizations would change the model and what
-        # operators they would introduce. The dry-run is target-independent, so
+        # operators they would introduce. This probe is target-independent, so
         # materialize it once here and only re-run the (cheap) support lookup per
         # EP/device below. Console-only feature — skipped in quiet mode.
         optim_outputs: list[tuple[Any, Any]] = []
-        if check_optim_output and not quiet:
+        if check_optim and not quiet:
             try:
                 import onnx
 
@@ -1513,7 +1513,7 @@ def analyze(
                         console.print()
 
                     # Optimization output support section (per-EP), when opted in.
-                    if check_optim_output:
+                    if check_optim:
                         from ..analyze.optim_output import check_optimization_output_support
 
                         optim_support = check_optimization_output_support(
--- a/src/winml/modelkit/commands/analyze.py
+++ b/src/winml/modelkit/commands/analyze.py
@@ -24,6 +24,7 @@ import click
 from rich.console import Console
 from rich.live import Live
 from rich.logging import RichHandler
+from rich.markup import escape
 from rich.progress import (
     BarColumn,
     MofNCompleteColumn,
@@ -51,6 +52,7 @@ from ..utils.logging import configure_logging
 
 if TYPE_CHECKING:
     from ..analyze.models.runtime_checks import PatternRuntime
+    from ..analyze.optim_output import OptimizationOutputSupport
 
 
 logger = logging.getLogger(__name__)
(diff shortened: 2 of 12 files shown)


src/winml/modelkit/optim/analysis.py
```diff
@@ -0,0 +1,352 @@
...
+from __future__ import annotations
+
+import logging
+from collections import Counter
+from dataclasses import dataclass, field
+from typing import TYPE_CHECKING, Any
+
+
+if TYPE_CHECKING:
+    import onnx
```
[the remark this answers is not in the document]
xieofxie: Fixed in 61118a8f: the duplicate/inline `import onnx` statements were consolidated into a single top-level `from onnx import AttributeProto, GraphProto, ModelProto, NodeProto` (matching the existing source style in `optimizer.py`/`surgery.py`), so `onnx` is now imported exactly once and only via `import ... from`. CodeQL should clear this on the next scan.

src/winml/modelkit/optim/analysis.py
```diff
@@ -0,0 +1,352 @@
...
+    scope: tuple[Any, ...],
+    table: dict[tuple[Any, ...], tuple[bytes, NodeRef]],
+) -> None:
+    """Populate ``table`` with ``{key: (serialized, NodeRef)}`` for every node.
+
+    Recurses into subgraphs (If/Loop/Scan bodies) so nested rewrites are not
+    missed. Keys are scoped by the containing node to keep subgraph nodes
+    distinct from top-level nodes.
+    """
+    import onnx
```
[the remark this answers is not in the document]
xieofxie: Fixed in 61118a8f: the duplicate/inline `import onnx` statements were consolidated into a single top-level `from onnx import AttributeProto, GraphProto, ModelProto, NodeProto` (matching the existing source style in `optimizer.py`/`surgery.py`), so `onnx` is now imported exactly once and only via `import ... from`. CodeQL should clear this on the next scan.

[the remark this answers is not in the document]
xieofxie: Fixed in 61118a8f: the duplicate/inline `import onnx` statements were consolidated into a single top-level `from onnx import AttributeProto, GraphProto, ModelProto, NodeProto` (matching the existing source style in `optimizer.py`/`surgery.py`), so `onnx` is now imported exactly once and only via `import ... from`. CodeQL should clear this on the next scan.

src/winml/modelkit/optim/analysis.py
```diff
@@ -0,0 +1,352 @@
...
+
+
+def _clone(model: onnx.ModelProto) -> onnx.ModelProto:
+    """Deep-copy a model.
+
+    ``CopyFrom`` is used rather than ``SerializeToString`` round-tripping so
+    that models larger than the 2 GiB protobuf serialization limit can still be
+    cloned in memory.
+    """
+    import onnx
```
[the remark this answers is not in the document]
xieofxie: Fixed in 61118a8f: the duplicate/inline `import onnx` statements were consolidated into a single top-level `from onnx import AttributeProto, GraphProto, ModelProto, NodeProto` (matching the existing source style in `optimizer.py`/`surgery.py`), so `onnx` is now imported exactly once and only via `import ... from`. CodeQL should clear this on the next scan.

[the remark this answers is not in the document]
xieofxie: Fixed in 61118a8f: the duplicate/inline `import onnx` statements were consolidated into a single top-level `from onnx import AttributeProto, GraphProto, ModelProto, NodeProto` (matching the existing source style in `optimizer.py`/`surgery.py`), so `onnx` is now imported exactly once and only via `import ... from`. CodeQL should clear this on the next scan.

tests/unit/optim/test_analysis.py
```diff
@@ -0,0 +1,259 @@
...
+Follows the Cardinal Rules:
+- #1 No hardcoded architectures — models are built generically in-test.
+- #2 Code-generated results — expectations derive from the constructed graphs.
+- #4 No skips — deterministic assertions only (clamp/surgery, graph diffing).
+"""
+
+from __future__ import annotations
+
+import numpy as np
+import onnx
```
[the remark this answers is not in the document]
xieofxie: Fixed in 61118a8f: dropped the bare `import onnx` in the test module and kept a single `from onnx import GraphProto, ModelProto, TensorProto, helper, numpy_helper`, so onnx is imported in one style only.

xieofxie:
should move to analyze command to take advantage of analyzer data

src/winml/modelkit/optim/analysis.py
```diff
@@ -0,0 +1,394 @@
...
+        base_out = _run_pipe(pipe, current, base_config)
+        base_nodes: dict[tuple[Any, ...], tuple[bytes, NodeRef]] = {}
+        _collect_nodes(base_out.graph, (), base_nodes)
+        base_inits = _collect_initializers(base_out)
+
+        probe_caps = [
+            (name, cap)
+            for name, cap in pipe.capabilities.items()
+            if isinstance(cap, BoolCapability) and not cap.default
+        ]
```
DingmaomaoBJTU, line 340: test

DingmaomaoBJTU (commented):
test

src/winml/modelkit/optim/analysis.py
DingmaomaoBJTU, line 340: `_run_pipe` (line 233) clones the model and respects `should_process`, but here you call `pipe.process(_clone(current), probe_config)` directly — duplicating the `should_process` check manually (lines 281-284). Consider reusing `_run_pipe` for consistency, or add a comment explaining why the direct call is preferred (e.g. to avoid double-cloning).

src/winml/modelkit/optim/analysis.py
```diff
@@ -0,0 +1,394 @@
...
+    """Run a pipe on a *clone* of ``model``, respecting ``should_process``.
+
+    Cloning is mandatory: some pipes serialize the model via ``save_onnx`` which
+    can rewrite tensors to external-data references in place. Returning the
+    input unchanged when the pipe opts out keeps the baseline faithful to the
+    real pipeline.
+    """
+    should_process = getattr(pipe, "should_process", None)
+    if callable(should_process) and not should_process(config):
+        return model
```
DingmaomaoBJTU, line 278: `_run_pipe` returns the **input** model unchanged when `should_process` returns False. This means `current` (line 309) may reference the same protobuf object across pipe iterations when the pipe opts out. Since `pipe.process` on a later pipe may mutate the model in place (despite the clone), this could lead to subtle aliasing if a probe's clone shares nested objects with `current`. The `CopyFrom` clone is deep, so this is likely fine — but worth a comment noting the assumption.

src/winml/modelkit/analyze/optim_output.py
```diff
@@ -0,0 +1,239 @@
...
+            exc,
+        )
+        result.error = str(exc)
+        return result
+
+    # Map each produced output tensor to its support result.
+    support_by_output: dict[str, tuple[SupportLevel, str | None]] = {}
+    for runtime in runtimes:
+        node = _matched_node(runtime)
+        outputs = tuple(node.output) if node is not None else ()
```
DingmaomaoBJTU, line 172: `_matched_node` uses a chain of `getattr` to dig into `PatternRuntime` internals (`pattern_match.skeleton_match_result.matched_nodes[0]`). If the internal structure changes, this silently returns `None` and all produced operators show `UNKNOWN`. Consider adding a `logger.debug` when the chain breaks, so a rule-data update doesn't silently degrade output.

src/winml/modelkit/commands/optimize.py
```diff
@@ -151,6 +152,115 @@ def capability_options(func: F) -> F:
...
 
+def _render_check_optim(findings: list[Any], verbose: bool) -> None:
+    """Render the applicability report produced by a ``--check-optim`` probe.
+
+    Args:
+        findings: Applicable ``CapabilityFinding`` objects in pipeline order.
+        verbose: When True, list every affected node instead of a sample.
+    """
+    if not findings:
+        console.print("\n[yellow]No registered optimizations would change this model.[/yellow]")
```
DingmaomaoBJTU, line 164: `_render_check_optim` uses the module-level `console` directly. The parallel renderer in `analyze.py` (`_render_optim_output_support`) accepts `console` as a parameter — worth aligning here for testability.

src/winml/modelkit/commands/analyze.py
```diff
@@ -1105,6 +1193,27 @@ def analyze(
...
+        optim_outputs: list[tuple[Any, Any]] = []
+        if check_optim and not quiet:
+            try:
+                import onnx
+
+                from ..optim import get_all_capabilities, iter_optimization_outputs
+
+                console.print(
+                    "[dim]Probing optimization outputs "
+                    "(this can take a while on large models)…[/dim]"
```
DingmaomaoBJTU, line 1209: The model is loaded via `onnx.load(str(model))` for the optimization probe, but later the analysis loop loads the model independently via `ONNXModel.from_path`. For large models this means two full model loads. Could the optimization probe share the already-loaded model, or is the separation intentional (e.g. because `iter_optimization_outputs` mutates its input)?

src/winml/modelkit/optim/analysis.py
```diff
@@ -0,0 +1,394 @@
...
+
+    This is the shared core behind :func:`analyze_model` and
+    :func:`iter_optimization_outputs`. It walks the pipeline pipe-by-pipe
+    exactly as the real optimizer does, probing each default-off boolean
+    capability in isolation and diffing the result against the pipe baseline.
+
+    ``produced_model`` is the concrete ONNX model that results from enabling the
+    single capability (plus auto-enabled dependencies). It contains the added
+    and modified nodes named by the finding, so downstream consumers can inspect
+    the produced operators directly (e.g. to check their EP/device support).
```
DingmaomaoBJTU, line 297: The outer `infer_shapes(_clone(model))` clones and shape-infers the entire model upfront. For large models this may be expensive. Is the clone necessary here (the docstring says `model` is never modified), or is it just for safety against `infer_shapes` mutating in place?

DingmaomaoBJTU (commented):
## Review Summary

**PR scope**: Adds `--check-optim` to both `winml optimize` and `winml analyze`, with a clean dry-run analysis engine, produced-operator support checking, and good test/doc coverage. Well-structured — new modules (`optim/analysis.py`, `analyze/optim_output.py`) are cleanly separated. Inline comments below.

src/winml/modelkit/optim/analysis.py
xieofxie: Looks like a placeholder/test comment — no code action taken for this one. Let me know if there was something specific you meant to flag on this line.

xieofxie: Good call. I kept the direct `pipe.process` call but added a comment (46f966d2) explaining why it can't reuse `_run_pipe`: the probe has to tell "pipe opts out" (skip and emit no finding) apart from "ran but changed nothing", and it isolates a failing capability in `try/except` so one bad probe can't abort the whole scan. `_run_pipe`'s return-the-input-unchanged contract doesn't express either of those.

src/winml/modelkit/optim/analysis.py
xieofxie: Added a note to `_run_pipe`'s docstring (46f966d2): every pipe run operates on a fresh deep `_clone`, so the shared `current` model is never mutated in place and returning the input object unchanged when the pipe opts out can't alias into a later probe. The `CopyFrom` clone is deep, as you noted.

src/winml/modelkit/analyze/optim_output.py
xieofxie: Done in 46f966d2 — `_matched_node` now emits a `logger.debug` (including the runtime type) when the chain breaks, so a rule-data update that changes `PatternRuntime`'s internals leaves a trace instead of silently degrading every produced operator to UNKNOWN.

src/winml/modelkit/commands/optimize.py
xieofxie: Done in 46f966d2 — `_render_check_optim` now takes `console` as its first parameter, matching `_render_optim_output_support` in analyze.py.

src/winml/modelkit/commands/analyze.py
xieofxie: The separation is intentional. `iter_optimization_outputs` needs a standalone `ModelProto` that it clones and shape-infers internally (and shape inference can mutate large models in place), whereas the analyzer consumes the model *path* and builds its own `ONNXModel`. Sharing one proto across both would risk aliasing between the probe's in-place transforms and the analyzer's state. The probe load also happens once (outside the per-EP loop), not per target, so it's a single extra load rather than one per EP.

src/winml/modelkit/optim/analysis.py
xieofxie: The clone is necessary. `infer_shapes` -> `infer_onnx_shapes` saves large models with `save_as_external_data=True`, which mutates the in-memory model in place (strips `raw_data`, flips `data_location` to EXTERNAL); it re-internalises afterward, but the caller's input would still be observably touched mid-call. Since `analyze_model` / `iter_optimization_outputs` promise never to modify the input model, we clone first. Added a comment to that effect in 46f966d2.

DingmaomaoBJTU (commented):
Second-pass deeper review — additional findings on correctness and performance.

src/winml/modelkit/optim/analysis.py
```diff
@@ -0,0 +1,394 @@
...
+        Returns:
+            ``(op_type, count)`` pairs ordered from most to least frequent.
+        """
+        nodes = {
+            "removed": self.removed_nodes,
+            "added": self.added_nodes,
+            "modified": self.modified_nodes,
+        }[kind]
+        return Counter(n.op_type for n in nodes).most_common()
+
```
DingmaomaoBJTU, line 137: **Node identity assumes immutable output names.** If an optimization renames outputs (e.g. `MatMul` output `mm` -> `mm/MatMulAddFusion`), the same logical node gets different identity keys, causing the diff to report a **remove + add** instead of a **modified** node. This is arguably acceptable (the semantic effect is the same — the optimization changes the graph), but it inflates `removed_nodes` / `added_nodes` counts and could confuse users who expect a 1:1 mapping. Worth documenting this semantics choice.

src/winml/modelkit/commands/analyze.py
```diff
@@ -1105,6 +1193,27 @@ def analyze(
...
+        # EP/device below. Console-only feature — skipped in quiet mode.
+        optim_outputs: list[tuple[Any, Any]] = []
+        if check_optim and not quiet:
+            try:
+                import onnx
+
+                from ..optim import get_all_capabilities, iter_optimization_outputs
+
+                console.print(
+                    "[dim]Probing optimization outputs "
```
DingmaomaoBJTU, line 1208: **Memory concern**: `list(iter_optimization_outputs(...))` materializes every applicable optimization's full `ModelProto` clone into memory at once. For a model with 50 applicable optimizations on a 1 GB model, this could hold ~50 GB. Consider either:
- Processing lazily per EP/device target (call `iter_optimization_outputs` per target), or
- Adding a `logger.info` noting the count before materializing, so users know what's happening.
The current `console.print` message about 'can take a while' is good, but the memory aspect is the real risk.

src/winml/modelkit/analyze/optim_output.py
```diff
@@ -0,0 +1,239 @@
...
+    finding: CapabilityFinding,
+    produced_model: ModelProto,
+    ep: EPName,
+    device: str,
+    model_path: str,
+) -> OptimizationOutputSupport:
+    """Check target support for the operators one optimization would produce."""
+    from .core.runtime_checker import RuntimeChecker
+
+    result = OptimizationOutputSupport(
```
DingmaomaoBJTU, line 134: **Performance**: `_check_one` creates a new `ONNXModel` + `RuntimeChecker` for each finding. For N applicable optimizations this means N model loads and N checker inits. Since the rule data and EP/device context are the same across findings, could the `RuntimeChecker` be created once and reused with different `node_output_filter` sets?

src/winml/modelkit/optim/analysis.py
```diff
@@ -0,0 +1,394 @@
...
+    return removed, added, modified
+
+
+# =============================================================================
+# ANALYSIS DRIVER
+# =============================================================================
+
+
+def _clone(model: ModelProto) -> ModelProto:
+    """Deep-copy a model.
```
DingmaomaoBJTU, line 253: The comment explains `CopyFrom` is used over `SerializeToString` round-tripping for >2 GiB models. Good rationale. However, protobuf `CopyFrom` **does** deep-copy (it's equivalent to `ParseFromString(SerializeToString())` but without the 2 GiB limit). A brief unit test verifying that modifying the original doesn't affect the clone (especially nested subgraph attributes) would increase confidence.

xieofxie: Added `TestClone` in `tests/unit/optim/test_analysis.py` (573a1df2): it clones a model that contains an `If` node with `then`/`else` subgraphs plus an initializer, then mutates the original's top-level node name, a *nested subgraph* node's op_type, and the initializer in place, and asserts the clone stays byte-for-byte identical. That covers the nested-attribute case you flagged.

src/winml/modelkit/analyze/optim_output.py
xieofxie: Reuse isn't possible here because each finding carries a *different* produced model — the whole point of the probe is that every optimization yields its own post-optimization graph. `RuntimeChecker` (and the `RuntimeCheckerQuery` it builds) is bound to a specific `model_proto`, so the checker is inherently per-produced-model. The shared EP/device context is the cheap part; rule-data availability is pre-computed at construction from a couple of (cached) filesystem probes, and the actual rule matching is per-graph work that can't be amortized across distinct graphs. So the N checker inits track N distinct models, not N redundant loads of the same one.

src/winml/modelkit/commands/analyze.py
xieofxie: Went with the `logger.info` option (573a1df2): after materializing I now log the number of applicable optimizations and added a comment documenting the memory-vs-recompute trade-off. The list is materialized deliberately so the per-target support lookup can reuse the produced models across every resolved EP/device without re-running the (expensive) probe per target; making it lazy-per-target would trade this peak memory for re-probing on each EP. In practice only a handful of optimizations apply to a given model, but the count is now visible for the pathological case.

src/winml/modelkit/optim/analysis.py
xieofxie: Agreed, documented in 573a1df2. `_node_identity`'s docstring now spells out that an optimization renaming a node's outputs surfaces as a remove + add pair rather than a single modification, and that this is an intentional graph-delta semantics choice (the diff describes which tensors/nodes appear and disappear, not a logical node-to-node correspondence), so such renames can inflate removed/added counts.

src/winml/modelkit/optim/analysis.py
```diff
@@ -0,0 +1,394 @@
...
+    kebab_defaults = {name: cap.default for name, cap in capabilities.items()}
+
+    # Mandatory pre-stage — mirrors Optimizer.optimize().
+    current = infer_shapes(_clone(model))
+
+    for pipe_class in PIPES:
+        pipe = pipe_class()
+
+        base_config = pipe.build_config(**default_kwargs)
+        base_out = _run_pipe(pipe, current, base_config)
```
DingmaomaoBJTU: **Missing error handling on baseline pipe run.** The probe runs (line ~289) are wrapped in try/except, but this baseline `_run_pipe` call is unprotected. If a pipe fails on its default config, the entire `_iter_findings` iterator crashes with an unhandled exception. Consider wrapping this in try/except and skipping the pipe's probes on failure, similar to the probe error handling below.

src/winml/modelkit/optim/analysis.py
```diff
@@ -0,0 +1,394 @@
...
+    probe_keys = set(probe)
+    removed = [base[k][1] for k in base_keys - probe_keys]
+    added = [probe[k][1] for k in probe_keys - base_keys]
+    modified = [probe[k][1] for k in (base_keys & probe_keys) if base[k][0] != probe[k][0]]
+    return removed, added, modified
+
+
+def _collect_initializers(model: ModelProto) -> dict[str, bytes]:
+    """Return ``{initializer_name: serialized_bytes}`` for the top-level graph."""
+    return {init.name: init.SerializeToString() for init in model.graph.initializer}
```
DingmaomaoBJTU: **False positives with external-data models.** `_collect_initializers` compares initializers via `SerializeToString()`. For models with external data (`data_location=EXTERNAL`), the `TensorProto` serialization includes file path and offset metadata — not just the tensor content. Two identical weights from different external files (or relocated offsets after a pipe's `save_onnx`) would serialize differently, producing false `modified_initializers` in the finding. Consider comparing only shape/dtype/name, or loading external data before comparison.

src/winml/modelkit/optim/analysis.py
```diff
@@ -0,0 +1,394 @@
...
+    """
+    for node in graph.node:
+        key = (scope, _node_identity(node))
+        table[key] = (
+            node.SerializeToString(),
+            NodeRef(node.op_type, node.name, tuple(node.output)),
+        )
+        for attr in node.attribute:
+            if attr.type == AttributeProto.GRAPH:
+                _collect_nodes(attr.g, (*scope, _node_identity(node), attr.name), table)
```
DingmaomaoBJTU: Minor: `_collect_nodes` recurses into subgraphs without a depth limit. Python's default recursion limit (~1000) could be hit by adversarial or deeply-nested models (If/Loop/Scan chains). Low risk in practice, but a depth guard or iterative traversal would be more robust.

DingmaomaoBJTU (commented):
Third-pass review — error handling, edge cases, and correctness deep-dive.

**Verified OK (no issues):**
- `auto_enable_dependencies` returns a new dict, no mutation risk between probe iterations
- `PIPES` ordering is shared between `_iter_findings` and `Optimizer.optimize()` — consistent
- `node_output_filter.intersection(node.output)` handles protobuf repeated fields and empty outputs correctly
- `--check-optim` properly requires `--model` (validated before handler)
- Optimization probe runs once, support check runs per EP/device target — good design for `--ep all`

DingmaomaoBJTU (approved):
LGTM — approving after three review passes. Well-architected feature with clean separation (`analysis.py` for dry-run, `optim_output.py` for EP support checking).

**Non-blocking follow-ups** (can be addressed in later PRs):
- Add try/except around baseline `_run_pipe` call (`analysis.py:280`) for robustness — low risk since default configs rarely fail, but would match the probe's error handling
- `_collect_initializers` may produce false "modified" findings for external-data models (serialization includes file metadata)
- For very large models with many applicable optimizations, `list(iter_optimization_outputs(...))` materializes all `ModelProto` clones in memory — consider a warning or lazy approach

All other areas verified: dependency management, pipeline ordering, filter logic, multi-target reuse, `--model` validation — all clean.

src/winml/modelkit/optim/analysis.py
xieofxie: Rewrote `_collect_nodes` as an iterative explicit-stack traversal (63894bfb), so it no longer consumes a Python frame per nesting level. Interesting side note while adding a test: protobuf itself caps message nesting (`Exceeded upb_DecodeOptions_MaxDepth` at ~100 levels), so a model deep enough to have blown our recursion couldn't be parsed by onnx in the first place — but the iterative version removes our traversal as the limiting factor regardless. Added a multi-level nested-subgraph test to guard it.

src/winml/modelkit/optim/analysis.py
xieofxie: Good catch. Added `_initializer_key` (63894bfb): when a `TensorProto` is external (`data_location == EXTERNAL` or it carries `external_data`), the diff key is now a normalized copy with `external_data` and `data_location` cleared, so relocated offsets / a different sidecar path after a pipe's `save_onnx` no longer read as a `modified` initializer. Inline (`raw_data`) tensors are unchanged. Tests cover both an offset-only change (not modified) and a real shape change (still modified). Note the CLI path loads external data into memory via `onnx.load`, so this mainly hardens the library API for callers that pass external-data models directly.

src/winml/modelkit/optim/analysis.py
xieofxie: Fixed in 63894bfb — the baseline `_run_pipe` call is now wrapped in try/except. If a pipe fails on its default config the scan logs a warning, skips that pipe's probes, and leaves `current` untouched so downstream pipes still see the last good model, mirroring the per-probe error handling below.

DingmaomaoBJTU (approved):
LGTM. The --check-optim feature is well-designed: the dry-run analysis walks the pipeline pipe-by-pipe with per-capability isolation, the graph diff via output-tensor identity is robust, and the produced-model handoff to the analyze bridge (--check-optim on analyze) cleanly separates target-independent probing from EP-specific support checks. Test coverage is thorough (diff helpers, end-to-end analysis, CLI wiring, rendering). A couple minor observations (non-blocking):$([char]10)$([char]10)1. `_matmul_add_model()` is duplicated across test_analysis.py, test_optim_output.py, and test_static_analyzer_cli.py -- consider a shared conftest fixture.$([char]10)2. In analyze.py the `optim_outputs` list is typed as `list[tuple[Any, Any]]` -- using the concrete types `list[tuple[CapabilityFinding, ModelProto]]` would be clearer and catch misuse.

PR #1019, opened 2026-07-02
feat: EP registration + monitoring — subprocess isolation, structured failures, universal op-tracing dispatch, lazy CLI startup

tezheng:
## Summary

This PR completes the EP registration, device-selection, session, and monitoring work while closing the CI and review regressions found during integration.

- **EP discovery and registration:** isolates plugin registration when needed, preserves structured failure attribution, rolls back failed native registrations, enumerates plugins before querying ORT devices, and validates the actual `OrtEpDevice` class before selecting an automatic EP/device pair.
- **Device and build policy:** preserves explicit EP/device choices, retries viable catalog candidates, retains vendor-free CPU/DML fallbacks when hardware probing fails, validates precision/device compatibility through the shared catalog, and restores ONNX content/config cache keys.
- **Compiler and session lifecycle:** restores provider/session options, JIT persistence, QAIRT deferred compilation, shared EP contexts, resolved-device artifact finalization, caller-owned `EPConfig` immutability, and canonical `compile --list` behavior.
- **Monitoring and performance:** keeps aggregate GPU telemetry separate from the selected adapter, fixes live chart labels, hardens QNN CSV/schematic validation and freshness handling, forwards resolved state to composite benchmarks, and restores op-tracing dispatch.
- **Compatibility and CLI:** lazily installs the Transformers/Optimum compatibility shim, restores GenAI plugin registration, isolates import-state tests, and brings CLI coverage into the unit-test matrix.
- **Review cleanup:** addresses or technically resolves all review feedback; the PR currently has no unresolved review threads.

## Validation

- `uv run pytest tests\unit\ -q`: 7451 passed, 77 skipped, 3 xfailed
- `uv run ruff check src\ tests\`: passed
- `uv run mypy -p winml.modelkit`: passed for 409 source files
- Post-final CodeQL cleanup: `uv run pytest tests\unit\sysinfo\ -q` — 89 passed

PR #1239, opened 2026-07-28
feat: restore EP-download progress bar for provider.ensure_ready()

xieofxie:
## Summary

Restores the EP-download progress bar that previously existed in `ep_registry.py` ([PR #788](https://github.com/microsoft/winml-cli/pull/788)) but was dropped during the #1019 EP-discovery refactor, when the `provider.ensure_ready()` call moved into `WinMLCatalogSource._resolve_provider` in `ep_path.py` and lost its progress UX.

## What changed

**`src/winml/modelkit/ep_path.py`** — cherry-picked the progress machinery and wired it into the resolve path:
- `_ep_download_timeout_default()` + `EP_DOWNLOAD_TIMEOUT_SECONDS` — honors `WINMLCLI_EP_DOWNLOAD_TIMEOUT` (default 5 min).
- `_NoopBar` + `_make_progress_bar()` — tqdm bar with a silent fallback when tqdm or `sys.stderr` is unavailable (production installs / spawn workers).
- `_parse_ep_metadata_from_path()` — recovers `(version, package_family_name)` from the MSIX `WindowsApps` install path when the native handle reports them empty.
- `_ensure_provider_ready()` — prints the `[WinML] Installing Execution Provider` notice, drives a live tqdm bar from `ensure_ready_async`'s `on_progress` callback, waits on `on_complete` with a timeout+cancel, and prints success / ❌-failure guidance plus version/PFN.
- `WinMLCatalogSource._resolve_provider` now calls `_ensure_provider_ready(provider)` instead of the silent `provider.ensure_ready()`.

### Note on the sync fast-path
The original helper had a `from windowsml import EpReadyState` sync fast-path for already-`Ready` providers. The current `WinMLCatalogSource` deliberately avoids that coupling (string-based readiness checks) and only invokes this helper on the cold, not-Ready path, so the fast-path was dead code — it's dropped, keeping the helper focused purely on the download+progress flow.

**Tests**
- Added `tests/unit/ep_path/test_ensure_provider_ready.py` — ported the PR #788 coverage (progress bar, stale-callback drop, timeout/cancel, `get_status` error, metadata printing/fallback, failure messages, no-tqdm fallback, timeout-env parsing).
- Updated `tests/unit/ep_path/test_winml_catalog_source.py` fakes to the async download contract.

## Verification
- `ruff check` clean
- `mypy -p winml.modelkit` — no issues (426 files)
- `pytest tests/unit/ep_path/` — 178 passed, 2 skipped (EP-install-gated)

### DIFF
--- a/src/winml/modelkit/ep_path.py
+++ b/src/winml/modelkit/ep_path.py
@@ -46,6 +46,7 @@ import functools
 import logging
 import os
 import platform
+import sys
 from abc import ABC, abstractmethod
 from collections.abc import Callable, Iterable, Iterator, Mapping
 from dataclasses import dataclass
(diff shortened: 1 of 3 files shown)


src/winml/modelkit/ep_path.py
```diff
@@ -775,6 +776,206 @@ def _winml_warn_once(key: str, msg: str, *args: Any) -> None:
...
+    Returns ``("", "")`` when the path is empty or does not match this layout.
+    """
+    import re
+    from itertools import pairwise
+    from pathlib import PurePath
+
+    if not library_path:
+        return "", ""
+
+    parts = PurePath(library_path).parts
```
DingmaomaoBJTU: This should use `PureWindowsPath` instead of `PurePath`. `PurePath` is host-dependent, so on non-Windows hosts this whole `C:\...\WindowsApps\...` string is treated as one path component and the `WindowsApps` lookup never matches. Since `library_path` is a Windows install path regardless of where the parser/test is running, parsing it with `PureWindowsPath` keeps the metadata fallback and the new Windows-path tests portable.

src/winml/modelkit/ep_path.py
```diff
@@ -775,6 +776,206 @@ def _winml_warn_once(key: str, msg: str, *args: Any) -> None:
...
+
+    def _on_progress(fraction: float) -> None:
+        # Native ops may fire a stale on_progress after on_complete; once done
+        # is set the main thread owns bar.n (forces it to 100 and closes the
+        # bar), so silently drop late callbacks instead of clobbering 100 with
+        # an earlier fraction or writing to a closed bar.
+        if done.is_set():
+            return
+        bar.n = max(0, min(100, int(fraction * 100)))
+        bar.refresh()
```
DingmaomaoBJTU, line 1002: The no-op fallback only protects progress-bar construction; rendering can still abort the EP download. If stderr is closed or starts failing after `tqdm(...)` succeeds, this `refresh()` raises from the native progress callback, and the unguarded `bar.refresh()`/`bar.close()` calls below can also turn a successful provider install into a discovery failure. Can we make progress rendering best-effort for the whole helper, for example by wrapping refresh/close and degrading to `_NoopBar` on the first I/O failure?

DingmaomaoBJTU (commented):
Found two inline issues while reviewing the EP download progress restore.

src/winml/modelkit/ep_path.py
xieofxie: Good catch — switched to PureWindowsPath in c48709a2 so the WindowsApps lookup and the Windows-path tests work regardless of the host OS.

src/winml/modelkit/ep_path.py
xieofxie: Agreed. Added a _SafeProgressBar wrapper in c48709a2 that makes rendering best-effort: the n setter, refresh(), and close() all catch I/O errors and drop the underlying bar on first failure, so a mid-download stderr failure degrades to a no-op instead of aborting the install. Covered by a new test that fails refresh/close and asserts readiness still completes.

src/winml/modelkit/ep_path.py
```diff
@@ -775,6 +776,260 @@ def _winml_warn_once(key: str, msg: str, *args: Any) -> None:
...
+    cancelled and ``TimeoutError`` is raised.
+    """
+    import threading
+
+    # Lazy-import to keep ep_path import cheap (rich pulls in pygments etc.);
+    # this branch only runs on the cold "EP needs download" path.
+    from .utils.console import get_console
+
+    console = get_console()
+    console.print(f"[WinML] Installing Execution Provider: [bold]{provider.name}[/bold]")
```
DingmaomaoBJTU: `_SafeProgressBar` covers tqdm refresh/close now, but the console output path is still unguarded. If stderr is closed or otherwise unwritable, these `console.print()` calls can raise before/after a successful async install, and `resolve()` will treat that as a provider iteration failure and skip the EP. Since the goal is for progress/status rendering failures not to block provider readiness, can we route these install/failure/success metadata prints through the same best-effort/no-op-on-I/O-error behavior and add a test with a failing console sink?

DingmaomaoBJTU (changes requested):
One review issue remains: console output failures can still abort the EP readiness path, so I am not approving yet.

src/winml/modelkit/ep_path.py
xieofxie: Good catch. Routed all the install/failure/success/metadata prints in _ensure_provider_ready through a new _safe_console_print helper (f4480d65) that catches I/O errors and logs at debug, mirroring _SafeProgressBar. So a closed/unwritable stderr before or after a successful async install can no longer raise and drop the EP in resolve(). Added test_ensure_provider_ready_survives_console_sink_failure with a console sink whose print() always raises, asserting readiness still completes.

DingmaomaoBJTU (approved):
LGTM. Well-structured async progress-driven EP download with solid error handling (timeout, I/O resilience, missing tqdm fallback) and comprehensive test coverage.

PR #1240, opened 2026-07-28
Add e2e regressions for recent fixes

DingmaomaoBJTU:
## Summary
- Add inspect e2e coverage for default third-party warning suppression, with an unsuppressed control run that proves the host emits at least one target warning-noise pattern before asserting default suppression.
- Add analyze e2e coverage for real default auto target selection, while keeping the patched exact-binding regression in unit selection-matrix coverage.
- Add perf and compile subprocess e2e coverage for native session cleanup before process exit, gated on an explicit policy-supported QNN NPU/GPU target.
- Leave the QNN CPU bridge auto-EP policy scenario to the existing controlled resolver unit test instead of an unreliable host-dependent e2e.

## Validation
- `uv tool run ruff check --fix tests\e2e\require_ep.py tests\e2e\test_inspect_e2e.py tests\e2e\test_perf_e2e.py tests\e2e\test_compile_e2e.py`
- `uv run --python C:\Users\qiowu\AppData\Roaming\uv\python\cpython-3.11.14-windows-x86_64-none\python.exe pytest tests\e2e\test_inspect_e2e.py::TestInspectWarningNoise tests\e2e\test_perf_e2e.py::TestPerfONNXDirect::test_benchmark_qnn_process_exit_releases_native_session tests\e2e\test_compile_e2e.py::TestProcessExitCleanup::test_compile_validation_process_exits_cleanly_after_releasing_session tests\unit\session\test_ep_device.py::test_resolve_auto_ep_for_explicit_cpu_skips_policy_unsupported_candidates -m 'e2e or not e2e' -q`

### DIFF
--- a/tests/e2e/require_ep.py
+++ b/tests/e2e/require_ep.py
@@ -170,7 +170,7 @@ def require_qnn_accelerator_device() -> tuple[str, str]:
         if device in {"npu", "gpu"} and device in registered_devices:
             return provider, device
 
-    pytest.skip(f"EP not available on a policy-supported accelerator device: {provider}")
+    return pytest.skip(f"EP not available on a policy-supported accelerator device: {provider}")
 
 
 def require_device(device: str) -> None:
--- a/tests/e2e/require_ep.py
+++ b/tests/e2e/require_ep.py
@@ -144,35 +144,6 @@ def require_ep(ep: str, *, device: str | None = None) -> str:
     return provider
 
 
-def require_qnn_accelerator_device() -> tuple[str, str]:
-    """Skip unless QNN is available on a policy-supported accelerator device.
-
-    Returns:
-        ``("QNNExecutionProvider", device)`` where ``device`` is ``"npu"`` or
-        ``"gpu"``, following the shared policy order.
-    """
-    from winml.modelkit.ep_path import EP_CATALOG
-    from winml.modelkit.utils import normalize_ep_name
-    from winml.modelkit.utils.constants import EP_SUPPORTED_DEVICES
-
-    provider = normalize_ep_name("qnn")
-    if provider is None:
-        pytest.skip("Unknown EP: 'qnn'")
-
-    try:
-        if not EP_CATALOG.is_compatible(provider):
-            pytest.skip(f"EP is not compatible with this host: {provider}")
-    except RuntimeError as e:
-        pytest.skip(f"Unable to determine EP compatibility for {provider}: {e}")
-
-    registered_devices = _registered_device_types(provider)
-    for device in EP_SUPPORTED_DEVICES[provider]:
-        if device in {"npu", "gpu"} and device in registered_devices:
-            return provider, device
-
-    return pytest.skip(f"EP not available on a policy-supported accelerator device: {provider}")
-
-
 def require_device(device: str) -> None:
     """Skip the current test unless any registered EP exposes ``device``."""
     from winml.modelkit.session import WinMLEPRegistry
--- a/tests/e2e/require_ep.py
+++ b/tests/e2e/require_ep.py
@@ -144,6 +144,35 @@ def require_ep(ep: str, *, device: str | None = None) -> str:
     return provider
 
 
+def require_qnn_accelerator_device() -> tuple[str, str]:
+    """Skip unless QNN is available on a policy-supported accelerator device.
+
+    Returns:
+        ``("QNNExecutionProvider", device)`` where ``device`` is ``"npu"`` or
+        ``"gpu"``, following the shared policy order.
+    """
+    from winml.modelkit.ep_path import EP_CATALOG
+    from winml.modelkit.utils import normalize_ep_name
+    from winml.modelkit.utils.constants import EP_SUPPORTED_DEVICES
+
+    provider = normalize_ep_name("qnn")
+    if provider is None:
+        pytest.skip("Unknown EP: 'qnn'")
+
+    try:
+        if not EP_CATALOG.is_compatible(provider):
+            pytest.skip(f"EP is not compatible with this host: {provider}")
+    except RuntimeError as e:
+        pytest.skip(f"Unable to determine EP compatibility for {provider}: {e}")
+
+    registered_devices = _registered_device_types(provider)
+    for device in EP_SUPPORTED_DEVICES[provider]:
+        if device in {"npu", "gpu"} and device in registered_devices:
+            return provider, device
+
+    pytest.skip(f"EP not available on a policy-supported accelerator device: {provider}")
+
+
 def require_device(device: str) -> None:
     """Skip the current test unless any registered EP exposes ``device``."""
     from winml.modelkit.session import WinMLEPRegistry
--- a/tests/e2e/test_analyze_e2e.py
+++ b/tests/e2e/test_analyze_e2e.py
@@ -47,7 +47,7 @@ from click.testing import CliRunner
 from tests.e2e.require_ep import require_ep
 from winml.modelkit.commands.analyze import analyze
 from winml.modelkit.utils.constants import EP_ALIASES as _EP_ALIASES
-from winml.modelkit.utils.constants import SUPPORTED_EPS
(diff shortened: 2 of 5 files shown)


tests/e2e/test_analyze_e2e.py
```diff
@@ -377,6 +377,58 @@ def test_default_auto_selects_single_ep_when_ep_omitted(
...
+                ep_name
+                for ep_name, devices in EP_SUPPORTED_DEVICES.items()
+                if {"npu", "gpu"}.issubset(devices)
+            ),
+            None,
+        )
+        if multi_device_ep is None:
+            pytest.skip("No EP supports both NPU and GPU in EP_SUPPORTED_DEVICES")
+
+        monkeypatch.setattr(
```
xieofxie: should we treat patch test as e2e ? do it need mark?

DingmaomaoBJTU: Removed the monkeypatched auto/auto selection case from e2e. That path is already covered in the unit selection matrix (`TestAnalyzeEPDeviceSelectionMatrix::test_auto_ep_auto_device_uses_device_priority_before_resolved_device_fallback`), and the e2e file now keeps only the real CLI runs gated on actual local EP availability.

xieofxie (commented):
Found four regression-coverage gaps where the new e2e tests can pass without exercising the corresponding fix, or fail for an unsupported host binding.

tests/e2e/test_inspect_e2e.py
```diff
@@ -197,6 +224,47 @@ def test_hierarchy_flag_accepted_without_model(self):
...
 
+class TestInspectWarningNoise:
+    """Default inspect output hides third-party warning chatter."""
+
+    def test_default_table_hides_common_third_party_warning_noise(
+        self, monkeypatch: pytest.MonkeyPatch
+    ) -> None:
+        monkeypatch.delenv("WINMLCLI_SHOW_ALL_WARNINGS", raising=False)
+
+        proc = _run_winml_cli_subprocess(["inspect", "--model-type", "bert"])
```
xieofxie: This does not guarantee that any of the listed warnings would be emitted without the suppression fix. `--model-type bert` avoids the network/processor paths, and the native bridge warning is host-dependent, so both new absence-only tests can pass unchanged on a quiet host even if suppression is removed. Please add an unsuppressed control run (`WINMLCLI_SHOW_ALL_WARNINGS=1`) that confirms at least one target pattern is present before asserting the default run hides it (or skip when the host cannot exercise any warning source).

tests/e2e/test_perf_e2e.py
```diff
@@ -593,6 +610,61 @@ def test_benchmark_ep_device_npu(self, ep: str, tmp_path: Path, model_arg: str):
...
+        """A successful perf CLI subprocess exits cleanly after releasing ORT sessions."""
+        output_file = tmp_path / "perf_cpu_process_exit.json"
+
+        proc = _run_winml_cli_subprocess(
+            [
+                "perf",
+                "-m",
+                str(onnx_model_path),
+                "--device",
+                "cpu",
```
xieofxie: Selecting CPU here prevents this test from exercising the QNN teardown failure fixed by #1230: after #1231, auto-EP resolution explicitly filters QNN's CPU bridge row. A normal CPU/OpenVINO session can therefore exit cleanly even if `PerfBenchmark.close()` is removed. Please gate on an available policy-supported QNN NPU/GPU target and pass both `--ep qnn` and that `--device`, skipping when such a target is unavailable.

tests/e2e/test_perf_e2e.py
```diff
@@ -593,6 +610,61 @@ def test_benchmark_ep_device_npu(self, ep: str, tmp_path: Path, model_arg: str):
...
+        assert data["benchmark_info"]["device"] == "cpu"
+        assert data["latency_ms"]["mean"] > 0
+
+    def test_auto_ep_for_cpu_device_skips_policy_unsupported_bridge_provider(
+        self,
+        tmp_path: Path,
+        model_arg: str,
+    ) -> None:
+        """When QNN is present, --device cpu must not resolve to QNN's CPU bridge row."""
+        qnn_provider = require_ep("qnn", device="cpu")
```
xieofxie: `require_ep(..., device="cpu")` proves only that QNN registered a CPU bridge; it does not prove QNN is vendor-compatible or would be selected. Before the fix, the resolver checked compatibility and considered OpenVINO/CPU before QNN/CPU, so on a non-Qualcomm or hybrid host this test can select another EP and satisfy every assertion without exercising the policy-filter regression. Please require that QNN is compatible and is the first viable pre-policy CPU candidate, or leave this scenario to the existing controlled-registry unit test.

tests/e2e/test_compile_e2e.py
```diff
@@ -470,6 +489,41 @@ def test_default_runs_validation(
...
+
+@pytest.mark.e2e
+class TestProcessExitCleanup:
+    def test_compile_validation_process_exits_cleanly_after_releasing_session(
+        self,
+        simple_matmul_onnx: Path,
+        tmp_path: Path,
+    ) -> None:
+        """Compile+validate in a subprocess exits 0 after native sessions are released."""
+        ep = next((candidate for candidate in EPCONTEXT_EPS if is_host(candidate)), None)
```
xieofxie: `is_host()` returns true when any device row registers. QNN can therefore win this `next()` with only its CPU bridge available, even though QNN/CPU is excluded by `EP_SUPPORTED_DEVICES`; the subprocess then fails auto-device resolution instead of skipping. Falling back to OpenVINO also does not reproduce the QNN teardown crash from #1230. Please gate an explicit policy-supported QNN NPU/GPU target with `require_ep(...)` and pass that device to the subprocess.

tests/e2e/test_inspect_e2e.py
DingmaomaoBJTU: Added an unsuppressed control run for both inspect warning-noise tests. Each test now first runs the same real CLI subprocess with `WINMLCLI_SHOW_ALL_WARNINGS=1`, requires at least one target warning-noise pattern to be present, and skips if this host cannot exercise any warning source before asserting the default run suppresses it.

tests/e2e/test_perf_e2e.py
DingmaomaoBJTU: Changed the perf process-exit test to use an explicit QNN accelerator target. It now gates on a policy-supported QNN NPU/GPU device via `require_qnn_accelerator_device()` and invokes `perf --ep qnn --device <device>` so the subprocess exercises the QNN session teardown path instead of CPU/OpenVINO.

tests/e2e/test_perf_e2e.py
DingmaomaoBJTU: Removed this host-dependent e2e case. The CPU bridge policy regression is covered by the controlled resolver unit test `tests/unit/session/test_ep_device.py::test_resolve_auto_ep_for_explicit_cpu_skips_policy_unsupported_candidates`, which can force the QNN/CPU candidate conditions reliably.

tests/e2e/test_compile_e2e.py
DingmaomaoBJTU: Changed the compile process-exit test to use the same explicit QNN accelerator gate as perf. It now skips unless QNN is compatible and registered on policy-supported NPU/GPU, then runs `compile --ep qnn --device <device> --validate` in the subprocess.

tests/e2e/require_ep.py
```diff
@@ -144,6 +144,35 @@ def require_ep(ep: str, *, device: str | None = None) -> str:
     return provider
 
 
+def require_qnn_accelerator_device() -> tuple[str, str]:
```
xieofxie: we already have require_ep, we should not dup implementation. Maybe just use it and use device=npu for related tests

DingmaomaoBJTU: Reworked this to reuse the existing gate. The custom `require_qnn_accelerator_device()` helper is removed, and the perf/compile teardown e2e tests now call `require_ep("qnn", device="npu")` directly before invoking the subprocess with `--ep qnn --device npu`.

xieofxie (approved):

Issue #1233, opened 2026-07-27
winml perf: post-result Windows 0xC0000005 in Microsoft.Windows.AI.MachineLearning.dll teardown


ssss141414:
## Summary

On Windows, `winml perf` can complete every warmup and timed iteration, print and persist valid results, and then terminate with `0xC0000005` during native teardown. Windows Error Reporting identifies `Microsoft.Windows.AI.MachineLearning.dll` from `windowsml==2.0.300` as the faulting module.

This reproduces on bundled CPU with a recipe-free current-main OWLv2 build and with both checked-in candidate precision recipes, so it is not caused by OWLv2 export input binding or a recipe field. It blocks an otherwise passing model-support Goal L1 because process exit remains authoritative.

## Environment

- Windows, Intel Core i9-10900X
- winml-cli current main: `3a2b2f7db5e70bb1f9039dd9037697ef92b2bc2a`
- `windowsml==2.0.300`
- Python `3.11.14`
- bundled `CPUExecutionProvider`
- checkpoint: `google/owlv2-base-patch16`

The tested commit contains both:

- #1220 / `eef88afb75fc3b534fd93841e9a899b691924e2a` (`windowsml.EpCatalog` registration and `atexit` `catalog.close()`)
- #1223 / `57c2cdf2381cc3a55f1ef065d247d838b3c6a0db` (concurrent native-stderr drain)

#1223 prevents pipe deadlock but does not affect this teardown access violation.

## Exact reproduction

Build without any recipe:

```powershell
uv run winml build -m google/owlv2-base-patch16 -o temp/baseline_owlv2 --ep cpu --device cpu --no-analyze --no-optimize --no-quant --no-compile --rebuild
uv run winml perf -m temp/baseline_owlv2/model.onnx --ep cpu --device cpu
```

The same failure occurs for independently built CPU fp32 and fp16 artifacts:

```powershell
uv run winml perf -m <fp32-model.onnx> --ep cpu --device cpu
uv run winml perf -m <fp16-model.onnx> --ep cpu --device cpu
```

Observed on three runs:

| Artifact | Work completed before fault | Native exit |
|---|---|---|
| recipe-free fp32 baseline | 10 warmups + 100/100 timed iterations; Avg 4154.85 ms; result printed | `-1073741819` / `0xC0000005` |
[...]

PR #1242, opened 2026-07-28
fix(config): skip winml quantization for EPs that quantize internally

KayMKM:
## Problem

`tests/e2e/test_eval_e2e.py` does not fail on an AMD Ryzen AI host — it **kills the pytest process on its very first test**, so none of its 46 cases run:

```
$ pytest tests/e2e/test_eval_e2e.py::TestEvalPerTask::test_image_classification -m e2e
collected 1 item
tests/e2e/test_eval_e2e.py::TestEvalPerTask::test_image_classification
EXIT=-1073740791          # 0xC0000409, native fail-fast
```

The same reproduces straight from the CLI, with no test harness involved:

```powershell
winml eval -m google/vit-base-patch16-224 --task image-classification --streaming --samples 10
```

```
[W PartitionPass.cpp] xir::Op{... type = dequantize-linear} is partitioned to CPU as :
    doesn't supported by target [AMD_AIE2P_4x8_CMC_Overlay].
[W PartitionPass.cpp] xir::Op{... type = :IsNaN} is partitioned to CPU as : ...
[F attrs_imp.cpp:38] Check failed: attrs_.count(key) > 0 Attrs doesn't contain attribute data
*** Check failure stack trace: ***
```

## Root cause

`_AUTO_PRECISION` in `config/precision.py` is keyed on the **device only**:

```python
_AUTO_PRECISION: dict[str, str] = {
    "npu": "w8a16",
    "gpu": "fp32",
    "cpu": "fp32",
}
```

On an AMD host the auto-selected NPU EP is VitisAI, so `--precision auto` produces a winml `w8a16` QDQ graph. VitisAI quantizes to XINT8 **inside the EP** and is not meant to consume that graph: it partitions the quantize/dequantize and `IsNaN` nodes back to CPU, then aborts inside xir. The abort is a native fail-fast, so the process dies rather than the build reporting an error.

This policy was already established — but only inside the eval harness. `scripts/e2e_eval/run_eval.py` carried its own `_EPS_SKIP_WINML_QUANT` and worked around the problem by passing `--no-quant` to `winml config` / `winml build` (see #1226). The product itself still handed every AMD NPU user a `w8a16` graph.

Verification that the unquantized track is the fix, same model and host:

| precision | result |
| --- | --- |
| `w8a16` (the auto default) | xir abort, `0xC0000409` |
| unquantized | `accuracy 1.0`, results written |

## Change

- `utils/constants.py`: add `EPS_WITH_INTERNAL_QUANT` as the single source of truth for "this EP runs its own quantization pipeline". `run_eval.py` now consumes it instead of defining its own copy.
- `config/precision.py`: `resolve_precision` sets `PrecisionPolicy.skip_quantization` when the resolved EP matches, and leaves `precision` at `"auto"` — winml makes no precision choice here, the EP does. No fake precision value is invented.
- `config/build.py`: both policy consumers (`_apply_device_precision_policy`, `resolve_quant_compile_config`) honor the flag first and apply exactly what `--no-quant` does (`quant = None`).
- Warn at `WARNING` level so the deviation is visible to end users:

  ```
  [WARNING winml.modelkit.config.precision] EP 'VitisAIExecutionProvider' quantizes
  internally, so winml is skipping its own quantization stage (equivalent to
  --no-quant) instead of applying the 'w8a16' default for device 'npu'. This EP
  therefore behaves differently from the others: the artifact stays unquantized and
  the accelerator applies its own scheme at load time. Pass --precision explicitly
  to force a winml quantization pass.
  ```

- `tests/e2e/test_eval_e2e.py`: drop the 13 per-test `require_not_ep("vitisai")` guards. They had been added one at a time as individual models hit the abort; `test_image_classification` was never guarded, so the module took the whole process down on AMD regardless. These tests now run for real.

An explicit `--precision` is unaffected — only `auto` is redirected.

| EP | `--precision auto` | change |
| --- | --- | --- |
| VitisAI | skip quantization | was `w8a16` |
| QNN | `w8a16` | unchanged |
| no `--ep` on a QNN host | `w8a16` | unchanged |
| VitisAI + explicit `--precision w8a16` | `w8a16` | user override still honored |

## Verification

On an AMD Ryzen AI NPU host:

- `pytest tests/e2e/test_eval_e2e.py -m e2e` → **41 passed, 5 skipped, 0 failed** (was: process killed on test 1, 0 cases completed). Zero VitisAI-related skips; the 5 remaining are `require_ep("qnn")` ×3 and `require_not_ep("migraphx")` ×2.
- `pytest tests/unit` → 7897 passed, 77 skipped, 3 xfailed.
- `winml config -m microsoft/resnet-50 --device npu --ep vitisai` → warning shown, generated config has `"quant": null`.
- `ruff check` / `ruff format --check` clean.

## Notes for follow-up (not in this PR)

- Switching VitisAI to the float graph makes its `vaip-ort-api.cpp` emit one `E ... Invalid NodeIndex` line **per graph node, twice** (808 lines for a 404-node ViT; 98.6% of the full e2e log). Inference is correct and the tests pass — it is pure log noise from the EP, but it buries real test output. Worth an upstream report to AMD.
- The e2e run still exits non-zero: after `41 passed, 5 skipped` is printed, the process hits `0xC0000005` in `Microsoft.Windows.AI.MachineLearning.dll` teardown. That is #1233, reproduces on Intel CPU EP too, and is unrelated to this change.

### DIFF
--- a/scripts/e2e_eval/run_eval.py
+++ b/scripts/e2e_eval/run_eval.py
@@ -44,6 +44,7 @@ from __future__ import annotations
 
 import argparse
 import contextlib
+import functools
 import hashlib
 import json
 import logging
@@ -115,14 +116,49 @@ _NPU_FALLBACK_PRECISIONS: tuple[str, ...] = ("w8a8", "w8a16")
 # ``--no-quant`` cannot override.
 
 
-def _should_skip_winml_quant(ep: str | None) -> bool:
-    """True if the eval harness should run this EP on the unquantized model."""
+@functools.cache
+def _deduce_ep_for_device(device: str) -> str | None:
+    """Canonical EP a concrete device resolves to on this host, or ``None``.
+
+    Mirrors what ``winml config`` / ``winml build`` do internally when ``--ep``
+    is omitted, so harness-side policy decisions see the same EP the product
+    will. ``"auto"`` is deliberately excluded: the product resolves it through
+    its own device detection, and the harness never forces a precision for it,
+    so the product's auto-precision policy already applies.
+    """
+    if not device or device.lower() == "auto":
+        return None
+    # Lazy import: keeps ``scripts/e2e_eval`` cheap to load (winml.modelkit
+    # transitively imports onnxruntime).
+    from winml.modelkit.session import default_ep_for_device
+
+    return default_ep_for_device(device.lower())
+
+
+def _effective_ep(ep: str | None, device: str | None) -> str | None:
+    """Canonical EP this run will actually target: explicit ``--ep``, else deduced."""
+    from winml.modelkit.utils.constants import normalize_ep_name
+
+    if ep:
+        return normalize_ep_name(ep)
+    return _deduce_ep_for_device(device or "auto")
+
+
+def _should_skip_winml_quant(ep: str | None, device: str | None = None) -> bool:
+    """True if the eval harness should run this EP on the unquantized model.
+
+    ``ep`` is ``None`` whenever only ``--device`` was pinned, so the effective EP
+    has to be deduced from the device before deciding. Without that,
+    ``run_eval.py --device npu`` on an AMD-only host would keep expanding
+    quantized jobs and forcing ``--precision w8a16`` on VitisAI -- exactly the
+    QDQ graph this policy exists to avoid.
+    """
     # Lazy import: keeps ``scripts/e2e_eval`` cheap to load (winml.modelkit
     # transitively imports onnxruntime) and matches the existing in-function
     # import pattern used elsewhere in this script.
-    from winml.modelkit.utils.constants import EPS_WITH_INTERNAL_QUANT, normalize_ep_name
+    from winml.modelkit.utils.constants import EPS_WITH_INTERNAL_QUANT
 
-    return normalize_ep_name(ep) in EPS_WITH_INTERNAL_QUANT
+    return _effective_ep(ep, device) in EPS_WITH_INTERNAL_QUANT
 
 
 def _resolve_precision(device: str, explicit: str | None, ep: str | None = None) -> str | None:
@@ -144,11 +180,12 @@ def _resolve_precision(device: str, explicit: str | None, ep: str | None = None)
 
     Otherwise an explicit per-model precision always takes precedence.
     """
-    if _should_skip_winml_quant(ep):
+    if _should_skip_winml_quant(ep, device):
         if explicit:
             safe_print(
-                f"  [precision] Ignoring explicit precision={explicit!r} for EP {ep!r}: "
-                "this EP is run on the unquantized variant (--no-quant)."
+                f"  [precision] Ignoring explicit precision={explicit!r} for EP "
+                f"{_effective_ep(ep, device)!r}: this EP is run on the unquantized "
+                "variant (--no-quant)."
             )
         return None
     if explicit:
@@ -634,7 +671,7 @@ def _run_build(
     # written with quant=None up-front; otherwise on NPU the config command
     # would still apply its default precision (w8a16) and we'd be relying on
     # --no-quant at build time alone to override it.
-    if _should_skip_winml_quant(ep):
+    if _should_skip_winml_quant(ep, device):
         config_args += ["--no-quant"]
 
     config_proc = _run_subprocess(config_args, timeout)
@@ -688,7 +725,7 @@ def _run_build(
         # Mirror the --no-quant passed to winml config above so the build
         # stage also skips QDQ regardless of what the config carries (defence
         # in depth; see _should_skip_winml_quant for the rationale).
-        if _should_skip_winml_quant(ep):
+        if _should_skip_winml_quant(ep, device):
             build_args += ["--no-quant"]
 
         build_proc = _run_subprocess(build_args, timeout)
@@ -1722,7 +1759,7 @@ def run_model(
             args += ["--task", entry.task]
         if ep:
             args += ["--ep", ep]
-        if _should_skip_winml_quant(ep):
+        if _should_skip_winml_quant(ep, device):
             args += ["--no-quant"]
         args += ["--iterations", "10", "--warmup", "2"]
         if trace:
@@ -2398,7 +2435,7 @@ def _build_jobs(
     # would produce duplicate artifacts under distinct precision slugs, and an
     # authored quantized recipe would hand the EP a QDQ graph its compiler is
     # not expected to consume.  Both are suppressed.
-    skip_quant = _should_skip_winml_quant(ep)
+    skip_quant = _should_skip_winml_quant(ep, device)
     expand_npu_quant = npu and not skip_quant
     jobs: list[EvalJob] = []
     for entry in entries:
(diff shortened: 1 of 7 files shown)


src/winml/modelkit/config/precision.py
```diff
@@ -410,16 +422,36 @@ def resolve_precision(
         )
 
     # Resolve "auto" precision for the resolved device
+    skip_quantization = False
     if resolved_precision == "auto":
-        resolved_precision = _AUTO_PRECISION[resolved_device]
-
-        # GPU + LLM: warn about w4a16 recommendation
-        if resolved_device == "gpu" and task in _LLM_TASKS:
+        if ep in EPS_WITH_INTERNAL_QUANT:
```
DingmaomaoBJTU: This still misses the common `--device npu` / omitted `--ep` path. In that case `_resolve_policy_target()` currently returns `("npu", None)`, so `ep` is `None` here and auto precision falls through to `w8a16`; only after that does the compile-provider fallback call `default_ep_for_device(resolved_device)`, which can resolve to VitisAI on an AMD-only host. That means `winml config -m ... --device npu` (and direct `resolve_quant_compile_config(device="npu")`) can still produce a winml QDQ graph for VitisAI. Please resolve/check the default EP before this auto-precision decision and add coverage for concrete device + omitted EP.

KayMKM: Thanks — you were right, and I confirmed the gap on an AMD Ryzen AI host before fixing it:

```
winml config -m microsoft/resnet-50 --task image-classification --device npu
-> quant: weight_type=uint8 activation_type=uint16 mode=static
```

Exactly as you described: `_resolve_policy_target` short-circuits for a concrete device and returns `("npu", None)`, so `resolve_precision` saw `ep=None`, fell through to the `w8a16` NPU default, and only deduced the EP afterwards when computing `compile_provider`.

Fixed in cb5fe9b9 by resolving the effective EP once, up front, and deriving both the auto-precision decision and `compile_provider` from it so the two can never disagree:

```python
effective_ep = ep if ep else default_ep_for_device(resolved_device)
```

Same command now emits the skip-quantization warning and writes `quant: null`.

Added the coverage you asked for, all on the concrete-device / omitted-EP path:

- `test_concrete_device_with_omitted_ep_on_an_amd_only_host` — skip-quant + `compile_provider == "vitisai"`
- `test_concrete_device_with_omitted_ep_on_a_qnn_host` — still `w8a16`, so the change stays scoped
- `test_quant_compile_config_skips_quant_for_deduced_internal_quant_ep` — the `resolve_quant_compile_config(device="npu")` entry point you called out

Also moved the typing-only `EPNameOrAlias` import into a `TYPE_CHECKING` block for the CodeQL alert (it is referenced only from the quoted `cast()` that ruff's TC006 requires).

Re-verified: full unit suite 7900 passed / 0 failed; e2e smoke on the AMD NPU 2 passed / 1 skipped / 0 failed.

scripts/e2e_eval/run_eval.py
```diff
@@ -104,29 +104,25 @@
...
-# Entries are canonical ``EPName`` values (the ``*ExecutionProvider`` form);
-# user-facing aliases like ``vitisai`` are normalised via
-# ``normalize_ep_name`` in :func:`_should_skip_winml_quant` so each EP only
-# needs to be listed once.
-_EPS_SKIP_WINML_QUANT = frozenset({"VitisAIExecutionProvider"})
+# rather than running winml's QDQ pass on top.  The EP list itself is the
+# product-level policy constant ``EPS_WITH_INTERNAL_QUANT`` -- e.g. VitisAI /
+# AMD Ryzen AI quantizes internally (XINT8) and aborts inside xir when handed a
+# winml-produced QDQ graph.  ``winml config`` / ``winml build`` already fall
+# back to the unquantized track for these EPs on their own; the harness still
```
DingmaomaoBJTU, line 112: This is only true when the runner was invoked with an explicit VitisAI `--ep`. If someone runs the harness on an AMD NPU host with `--device npu` and leaves `--ep` omitted, `_should_skip_winml_quant(ep)` sees `None`, `_build_jobs()` still expands quantized NPU jobs, and `_resolve_precision()` returns `w8a16`. `_run_build()` then passes that as an explicit `--precision`, so the product-side auto-precision skip no longer applies and `--no-quant` is not injected either. That leaves `run_eval.py --device npu ...` able to hand VitisAI the same QDQ graph this PR is trying to avoid. Please deduce the effective EP for the harness before `_should_skip_winml_quant()` / `_build_jobs()` make their decisions, or avoid forcing an explicit NPU precision while the EP is still unknown.

KayMKM: Good catch again — same class of gap, and the harness side is worse because it forwards an explicit `--precision`, which suppresses the product-side policy from the previous commit.

Confirmed on an AMD Ryzen AI host before fixing:

```
_should_skip_winml_quant(None, "npu")  ->  False
_resolve_precision("npu", None)        ->  "w8a16"
```

Fixed in 2c8f1aba by deducing the effective EP the same way the product does, before any policy decision:

```python
_deduce_ep_for_device(device)   # default_ep_for_device on a concrete device
_effective_ep(ep, device)       # explicit --ep, else the deduced one
```

`_should_skip_winml_quant` now takes the device and all five call sites pass it (`_resolve_precision`, both `_run_build` injection points, `run_model`'s perf path, and `_build_jobs`). Same host after the fix:

```
_should_skip_winml_quant(None, "npu")  ->  True
_resolve_precision("npu", None)        ->  None
```

`--device auto` deliberately does **not** deduce: the harness forces no precision there, so the product resolves the device itself and its own auto-precision policy already applies. Covered by `test_auto_device_defers_to_the_product`.

New coverage for the pinned-device path, both directions so the change stays scoped:

- `test_skips_quant_when_device_deduces_to_vitisai` / `test_keeps_quant_when_device_deduces_to_qnn`
- `test_resolve_precision_drops_w8a16_for_deduced_vitisai`
- `test_build_jobs_drops_quantized_fanout_for_deduced_vitisai` — the `_build_jobs()` expansion you called out, collapses to one unquantized job
- `test_build_jobs_still_expands_for_deduced_qnn` — w8a8 + w8a16 fan-out preserved

One thing worth flagging: adding the deduction made five *existing* tests depend on whichever EP the host registers — they passed on a QNN box and failed on this AMD one. Pinned it with an autouse fixture so the suite stays deterministic everywhere, and the new tests patch it explicitly.

Full unit suite: 7906 passed / 0 failed.

DingmaomaoBJTU (approved):
LGTM — solid fix. Centralizing the internal-quant EP policy into `EPS_WITH_INTERNAL_QUANT` in constants.py and deducing the effective EP before the auto-precision decision is the right approach. The `skip_quantization` flag on `PrecisionPolicy` makes the caller contract explicit. Test coverage is comprehensive (mocked host deduction for AMD vs QNN, build.py integration, e2e skip removal). Minor nit: the `_deduce_ep_for_device.cache_clear()` calls in test teardown could be wrapped in `try/finally` to avoid stale cache if an assertion fails mid-test, but the `autouse` fixture already handles the main path. Approved.

PR #1243, opened 2026-07-28
e2e_eval: record the precision the build actually applied

KayMKM:
## Problem

`run_eval` wrote `EvalJob.precision` — the **declared** precision — into `eval_result.json`. That value is `None` for any `winml config` fallback job that pins no precision, but such a build is not precision-less: `winml config` resolves `auto` to the device default, which on NPU is `w8a16`.

So a `--device npu` (pre-#1163) or `--device auto --ep <npu ep>` run produced a **w8a16 QDQ model** while the result recorded no precision at all, and the result dir carried no `__<precision>` suffix. The downstream coverage report maps a missing precision to the `unquantized` label, so those quantized runs were published as unquantized baselines — models that never had an unquantized NPU run showed a PASS "unquantized" row.

Verified against the published data: `openai__clip-vit-base-patch32__zero-shot-image-classification` (no suffix, no `precision` field) has `onnx_size_bytes` = 178,466,841 ≈ 2 × the model's w8a16 artifact (89.6 MB), vs 253.9 MB for its fp16 build — i.e. the "unquantized" cell is a quantized run.

## Change

- `_run_build` reports the effective precision back to the caller: the pinned `--precision` flag when there is one, otherwise read from the generated build config's `quant` section (`weight_type`/`activation_type` bit widths -> `w{x}a{y}`, `mode="fp16"` -> `fp16`, `mode="rtn"` -> `int{bits}`). Reading the config back is what `winml config` actually decided, so the harness never has to re-implement the CLI's precision policy.
- Only what the config states is reported: a config with no quant stage yields no precision (unchanged from today), since "winml skipped quantization" is not a claim about the graph's dtype.
- `_run_recipe_build` reports its authored variant precision.
- `main()` writes that effective precision into `eval_result.json` (including the accuracy-backfill merge, which rebuilds the model).

The job's **declared** precision still drives the result-dir slug and the log label, so folder naming, `--continue` resume and the existing data layout are unchanged.

## Verification

Ran the harness on `prajjwal1/bert-tiny` (perf only):

| run | result dir | recorded `precision` | `onnx_size_bytes` |
| --- | --- | --- | --- |
| `--device cpu` | `…__feature-extraction` | *(absent — config has no quant stage)* | 17,507,493 |
| `--device auto --ep openvino` | `…__feature-extraction` | **`w8a16`** | 8,465,338 |
| `--device npu --ep openvino` | `…__feature-extraction__w8a8` | `w8a8` (pinned) | 4,489,977 |
| `--device npu --ep openvino` | `…__feature-extraction__w8a16` | `w8a16` (pinned) | 4,489,977 |

The second row is exactly the previously-broken case: same untagged folder, but the generated config is `weight_type=uint8, activation_type=uint16` and the artifact is half the size. Before this change that file carried no `precision` at all.

Recipe path (`microsoft/resnet-18`, which has `fp16` + `w8a16` recipes) records `fp16` / `w8a16` from the variant, as it already did.

Note the real config uses `quant.mode = "static"`, not `"qdq"`, which is why the mapping keys off the weight/activation bit widths rather than the mode.

`uv run --no-sync pytest tests/unit/eval` -> 540 passed.

### DIFF
--- a/scripts/e2e_eval/run_eval.py
+++ b/scripts/e2e_eval/run_eval.py
@@ -159,6 +159,47 @@ def _resolve_precision(device: str, explicit: str | None, ep: str | None = None)
     return _DEFAULT_PRECISION_NPU if device == "npu" else None
 
 
+# Quantization type -> bit width, used to rebuild the ``w{x}a{y}`` label from a
+# generated build config's quant section.
+_QUANT_TYPE_BITS = {"uint8": 8, "int8": 8, "uint16": 16, "int16": 16}
+
+
+def _precision_from_build_config(config_path: Path) -> str | None:
+    """Read back the precision ``winml config`` resolved into a build config.
+
+    The harness omits ``--precision`` whenever it has no explicit value (CPU/GPU,
+    ``--device auto``, or a recipe-less model with no pinned precision), and
+    ``winml config`` then resolves ``auto`` against the target device (NPU ->
+    w8a16, CPU/GPU -> fp32). The generated config's ``quant`` section is the only
+    faithful record of that decision, so recording it keeps a quantized build
+    from being written out as "no precision" -- which downstream reports read as
+    "unquantized".
+
+    Returns None when the quant section has an unrecognised shape.
+    """
+    try:
+        cfg = json.loads(config_path.read_text(encoding="utf-8"))
+    except (OSError, json.JSONDecodeError):
+        return None
+    quant = cfg.get("quant")
+    if quant is None:
+        # No quant stage and no fp16 conversion: the graph stays fp32.
+        return "fp32"
+    if not isinstance(quant, dict):
+        return None
+    mode = quant.get("mode")
+    if mode == "fp16":
+        return "fp16"
+    if mode == "rtn":
+        bits = quant.get("rtn_bits")
+        return f"int{bits}" if bits else None
+    weight_bits = _QUANT_TYPE_BITS.get(quant.get("weight_type"))
+    activation_bits = _QUANT_TYPE_BITS.get(quant.get("activation_type"))
+    if weight_bits and activation_bits:
+        return f"w{weight_bits}a{activation_bits}"
+    return None
+
+
 def _load_timeout_skip_set() -> set[tuple[str, str]]:
     """Load the timeout skip list as a set of (hf_id, task) tuples."""
     if not TIMEOUT_SKIP_LIST_PATH.exists():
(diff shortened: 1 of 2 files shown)


scripts/e2e_eval/run_eval.py
```diff
@@ -2970,6 +3026,11 @@ def main() -> None:
             # {success, onnx_paths, stage, proc}; build is shared by perf + eval.
             build_result, recipe_meta, trust = _build_for_job(job, args, model_dir)
 
+            # A fallback job with no pinned precision still builds at the device
+            # default (NPU -> w8a16), so record what was built, not what the job
+            # declared -- an empty precision reads as "unquantized" downstream.
+            recorded_precision = build_result.get("precision") or precision
```
DingmaomaoBJTU: This fallback can reintroduce a declared precision that the build deliberately did not apply. For skip-quant EPs like VitisAI, `_resolve_precision()` drops even an explicit registry precision and `_run_build()` returns `precision=None` after generating/building with `--no-quant`; falling back to `job.precision` here would still stamp that ignored value into `eval_result.json`. Since explicit pinned builds already return their precision from `_run_build()`/`_run_recipe_build()`, can this use the build-reported value as the source of truth instead of falling back after it reports `None`?

KayMKM: Good catch -- fixed in a72869f6. The build-reported value is now the sole source of truth: `recorded_precision = build_result.get("precision")`, no fallback.

For the VitisAI case that chain is now `_resolve_precision` -> None (explicit registry precision dropped with the existing warning) -> `_run_build` builds with `--no-quant` -> generated config has no quant stage -> reported precision is None -> `eval_result.json` carries no `precision` key. The declared value still drives the result-dir slug and the log label only, which keeps folder naming and `--continue` resume unchanged.

Pinned builds are unaffected since `_run_build` reports the flag it actually passed and `_run_recipe_build` reports its variant precision. The one place that lost a value is the pre-exported `composite_onnx` path, where nothing is built at all -- that early return now reports the caller's resolved precision explicitly rather than relying on the fallback.

Added `TestBuildForJobPrecision::test_skip_quant_ep_drops_explicit_entry_precision` to pin the behaviour.

scripts/e2e_eval/run_eval.py
```diff
@@ -3058,6 +3123,10 @@ def main() -> None:
             # build_eval_result would store, so it's spliced in directly.
             result = dict(backfill_existing)
             result["accuracy"] = accuracy_result
+            # The backfill rebuilt the model, so stamp the precision it used —
+            # results written before the build reported it carry none at all.
+            if recorded_precision is not None:
```
DingmaomaoBJTU: `None` is an authoritative build result here for skip-quant/unquantized builds. With this guard, accuracy backfill preserves any stale `precision` already present in `backfill_existing` (for example a perf-only result written before this fix with VitisAI + declared `w8a16`), so the merged result can still claim the ignored precision. Please clear `result["precision"]` when `recorded_precision is None` instead of leaving the old value in place.

DingmaomaoBJTU (approved):
LGTM. Fixes precision reporting so auto-resolved quantization (e.g. NPU -> w8a16) is faithfully recorded instead of appearing as 'no precision'. Clean extraction of _merge_backfill_result, proper stale-precision cleanup, and comprehensive test coverage.

PR #1248, opened 2026-07-29
Fix Qwen3 genai bundle component builds

DingmaomaoBJTU:
## Summary
- Preserve quantization for raw graphs that intentionally skip ORT optimization.
- Add a build-only path for HF/composite builds so genai bundle assembly consumes ONNX artifacts without creating runtime sessions.
- Use the build-only component path in genai bundle orchestration and cover the Qwen3 regression cases.

## Validation
- `uv run ruff check src\winml\modelkit\build\common.py src\winml\modelkit\commands\build.py src\winml\modelkit\config\build.py src\winml\modelkit\models\auto.py src\winml\modelkit\models\winml\composite_model.py src\winml\modelkit\models\winml\genai_bundle.py src\winml\modelkit\models\hf\qwen3\qwen_transformer_only.py tests\unit\build\test_hf.py tests\unit\build\test_onnx.py tests\unit\commands\test_build.py tests\unit\models\auto\test_from_pretrained_ep.py tests\unit\models\qwen3\test_qwen3_modeling.py tests\unit\models\winml\test_genai_bundle_orchestrator.py`
- `uv run pytest tests\unit\build\test_hf.py tests\unit\build\test_onnx.py tests\unit\commands\test_build.py tests\unit\models\auto\test_from_pretrained_ep.py tests\unit\models\qwen3\test_qwen3_modeling.py tests\unit\models\winml\test_genai_bundle_orchestrator.py`
- x64 cmd full rebuild of `Qwen/Qwen3-0.6B` with `--export-type optimized --ep qnn --device npu --rebuild` exits `0` and produces matching bundle hashes.

### DIFF
--- a/src/winml/modelkit/commands/build.py
+++ b/src/winml/modelkit/commands/build.py
@@ -1040,6 +1040,14 @@ def build(
         # to honor the requested policy. fp16/fp32 clear quant; npu/int8 etc set it.
         if cli_utils.is_cli_provided(ctx, "device") or cli_utils.is_cli_provided(ctx, "precision"):
             from ..compiler.configs import WinMLCompileConfig
+            from ..onnx import is_quantized_onnx
+
+            is_pre_quantized_onnx_input = (
+                model_input is not None
+                and model_input.kind is ModelInputKind.ONNX_FILE
+                and model is not None
+                and is_quantized_onnx(Path(model))
+            )
 
             def _patch_device(cfg: WinMLBuildConfig) -> None:
                 from ..config import resolve_quant_compile_config
@@ -1047,7 +1055,7 @@ def build(
                 resolved_quant, _ = resolve_quant_compile_config(
                     device=device, precision=precision, ep=ep_value
                 )
-                if not quant or resolved_quant is None or (cfg.skip_optimize and cfg.quant is None):
+                if not quant or resolved_quant is None or is_pre_quantized_onnx_input:
                     cfg.quant = None
                 elif cfg.quant is None:
                     # Populate calibration identifiers from the loader/model
--- a/src/winml/modelkit/commands/build.py
+++ b/src/winml/modelkit/commands/build.py
@@ -1047,7 +1047,7 @@ def build(
                 resolved_quant, _ = resolve_quant_compile_config(
                     device=device, precision=precision, ep=ep_value
                 )
-                if cfg.skip_optimize or not quant or resolved_quant is None:
+                if not quant or resolved_quant is None or (cfg.skip_optimize and cfg.quant is None):
                     cfg.quant = None
                 elif cfg.quant is None:
                     # Populate calibration identifiers from the loader/model
@@ -1641,8 +1641,6 @@ def _run_optimize_stage(
         show_io_first: If True, show I/O tensors at the start of the stage
             (used in ONNX mode where there is no export stage).
         skip_optimize: When True, skip the ORT graph-optimization pass.
-            Used for pre-quantized models (QDQ or QOperator format) whose
-            integer ops have no kernel on the host EP.
 
     Returns:
         Tuple of (current_path, opt_elapsed).
@@ -1796,10 +1794,6 @@ def _run_quantize_stage(
     """
     from ..quant import quantize_onnx
     from ..utils.console import StageLive
-
-    if config.skip_optimize:
-        config.quant = None
-        return current_path
 
     if config.quant is None:
         # ``generate_onnx_build_config`` and ``ensure_pre_quantized_stamped``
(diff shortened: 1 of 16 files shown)


src/winml/modelkit/models/hf/qwen3/qwen_transformer_only.py
```diff
@@ -333,6 +333,7 @@ def outputs(self) -> dict[str, dict[int, str]]:
     # Pure graph (no post-export RMSNorm fusion / matmul-add fusion): the default
     # WinMLOptimizationConfig() leaves every fusion flag off.
     optim=WinMLOptimizationConfig(),
+    skip_optimize=True,
```
xieofxie, line 336: `skip_optimize=True` is lost when this registered config is consumed. `generate_hf_build_config()` goes through `_assemble_config()`, which reconstructs `WinMLBuildConfig` while copying only `optim`, `quant`, and `compile`, so the generated Qwen config falls back to `skip_optimize=False` and still runs ORT optimization. Please propagate this field through assembly and cover the generated config in a test, rather than only asserting the registry value.

src/winml/modelkit/commands/build.py
```diff
@@ -1047,7 +1047,7 @@ def _patch_device(cfg: WinMLBuildConfig) -> None:
                 resolved_quant, _ = resolve_quant_compile_config(
                     device=device, precision=precision, ep=ep_value
                 )
-                if cfg.skip_optimize or not quant or resolved_quant is None:
+                if not quant or resolved_quant is None or (cfg.skip_optimize and cfg.quant is None):
```
xieofxie: `skip_optimize` no longer establishes that the model is already quantized. With a raw model config where `skip_optimize=True` and `quant=None`, an explicit INT8/NPU policy can populate `resolved_quant`, but this branch clears it and silently leaves the model unquantized. Please preserve `quant=None` only when the ONNX input is actually detected as pre-quantized instead of inferring that state from `skip_optimize`.

xieofxie (changes requested):
Two correctness issues need to be addressed before merge: the Qwen skip-optimization setting is discarded during config assembly, and skip_optimize incorrectly suppresses explicit quantization.

KayMKM:
Heads-up on the `src/winml/modelkit/ep_path.py` part of this PR (the removal of
`_release_winml_catalog` and its `atexit.register(...)`). It is not mentioned in
the Summary and `ep_path.py` / `test_winml_catalog_source.py` are not in the
Validation lists, so flagging it in case it was incidental.

**It does not stop the native catalog release from running at shutdown.**

`_get_catalog()` is `@functools.cache`d, so the catalog object stays strongly
referenced for the life of the process and is only released when the interpreter
clears module dicts during finalization. At that point:

```
windowsml/__init__.py:439  EpCatalog.__del__
windowsml/__init__.py:428  EpCatalog.close()
                           -> WinMLEpCatalogRelease   (the call that fails)
```

So removing the `atexit` hook only moves that call *later* into finalization,
where the process is further torn down.

### Measured

Full `tests/e2e/test_perf_e2e.py` on the OV agent hardware, same machine, same
command, only the `ep_path` behaviour differing:

| variant | tests | process exit code | faulting module / code / offset |
|---|---|---|---|
| current `main` behaviour (atexit -> close) | 35 passed, 15 skipped | `-1073741819` | `Microsoft.Windows.AI.MachineLearning.dll` 2.0.300.41724 / `0xc0000005` / `0x26f5` |
| this PR's `ep_path` change | 35 passed, 15 skipped | `-1073741819` | identical |

Every test passes in both cases; the process then dies during shutdown, 1.0s
(baseline) / 1.4s (this PR) after the last test. Note the exit code reaches
Azure DevOps as `PowerShell exited with code '1'` after WerFault gets involved,
which makes it look like a test failure rather than a crash.

The fault site itself, disassembled at RVA `0x26f5`:

```asm
+0x026f2  mov  rax, qword ptr [rdx]         ; load vtable   -- this one is fine
+0x026f5  mov  rax, qword ptr [rax + 0x10]  ; vtable[2] = IUnknown::Release   <-- AV
+0x026f9  jmp  <Release thunk>
```

A `ComPtr`-style release: the object's first qword still reads back, but the
vtable pointer taken from it is already invalid, i.e. the release runs after the
state/module that owns it is gone.

The same call also has a second failure mode: it can block forever instead of
faulting. Captured with `py-spy --native` on a CI process that had finished every
test and then sat there until the job timed out:

```
Py_FinalizeEx -> atexit -> EpCatalog.close()
  -> WinMLEpCatalogRelease -> WinMLAsyncClose -> SleepConditionVariableSRW
```

Which of the two you get depends on how far native teardown has progressed, so
neither removing one caller nor wrapping it in a timeout covers both.

### Suggestion

If the intent is to stop the release at shutdown, clearing the handle is what
actually achieves it, because `close()` starts with `if self._handle:`:

```python
catalog._handle = None
```

That disarms `atexit` *and* `__del__` in one place. The handle is process-scoped,
so the OS reclaims it at exit.

Also worth noting: `test_get_catalog_does_not_register_atexit_cleanup` now
asserts `registered == []`, which locks in "no cleanup hook" without providing a
replacement, so a later fix in this area would have to revisit that assertion.

Happy to share the raw logs / WER records if useful.

DingmaomaoBJTU:
Handled in 749f024e by registering an atexit cleanup that only disarms the cached WinML catalog handle (`catalog._handle = None`) instead of calling `close()`. The lifecycle test now asserts the cleanup is registered, clears the handle, and does not invoke `close()`. I also ran the ep_path unit coverage and a live `_get_catalog()` exit smoke on this Windows machine.

xieofxie (approved):
Re-reviewed the latest head. The registered `skip_optimize` value now survives config assembly, quantization is based on actual ONNX graph state rather than `skip_optimize`, and regression tests cover both paths. No remaining issues found.

PR #1249, opened 2026-07-29
Fix QNN detail tracing and compile EP options

xieofxie:
## Summary

- keep system hardware monitoring behind the explicit `--monitor` flag while allowing EP op tracing to run independently
- add repeatable `--ep-options KEY=VALUE` support to `winml compile`, with CLI values overriding matching config values
- improve QAIRT SDK discovery for direct installation roots
- resolve the architecture-matched QNN optrace reader DLL and pass its path to `qnn-profile-viewer`
- parse the generated QNN HTP analysis summary rather than the viewer output prefix
- add regression coverage and compile command documentation

### DIFF
--- a/src/winml/modelkit/commands/perf.py
+++ b/src/winml/modelkit/commands/perf.py
@@ -1117,11 +1117,9 @@ class PerfBenchmark:
         The EP monitor is integrated into ``session.perf()`` so op-tracing
         observes the user's actual benchmark iterations.
 
-        HWMonitor (system-wide CPU/RAM/NPU metrics) is engaged when available
-        AND either ``--monitor`` was set or HW data is otherwise needed. When
-        HWMonitor is unavailable but op-tracing is still requested, the run
-        proceeds with the EP monitor only — op-tracing is the headline goal
-        and must not be blocked by missing HW telemetry.
+        HWMonitor (system-wide CPU/RAM/NPU metrics) is engaged only when
+        ``--monitor`` was set. Op-tracing still uses the EP monitor required to
+        collect its profiling artifacts, without enabling hardware telemetry.
         """
         from ..session.monitor.hw_monitor import HWMonitor
 
@@ -1142,10 +1140,9 @@ class PerfBenchmark:
             output_dir=output_dir,
         )
 
-        # HWMonitor is best-effort: required only for the live-chart UI on
-        # --monitor. When it's unavailable but op-tracing is requested, run
-        # without HW telemetry rather than degrading op-tracing to a no-op.
-        hw_available = HWMonitor.is_available()
+        # Keep system telemetry under explicit --monitor control. The EP monitor
+        # remains active independently when op-tracing needs profiling data.
+        hw_available = self.config.monitor and HWMonitor.is_available()
         if self.config.monitor and not hw_available:
             Console(stderr=True).print(
                 "[yellow]Warning:[/yellow] HWMonitor unavailable on this system. "
@@ -1191,7 +1188,8 @@ class PerfBenchmark:
             if ep_dict:  # NullEPMonitor returns {}, real monitors return data
                 self._hw_metrics["ep_proof"] = ep_dict
         else:
-            # HW unavailable: run with EP monitor only (op-tracing path).
+            # No --monitor (or HWMonitor unavailable): run with the EP monitor
+            # only so op-tracing and proof-of-execution still work.
             with session.perf(warmup=self.config.warmup, monitor=ep_monitor) as ctx:
                 _run_simple_loop(
                     session,
--- a/src/winml/modelkit/session/monitor/qnn/viewer.py
+++ b/src/winml/modelkit/session/monitor/qnn/viewer.py
@@ -66,11 +66,11 @@ def find_qnn_sdk() -> Path | None:
     for base_path in _COMMON_SDK_PATHS:
         if not base_path.is_dir():
             continue
-        if (base_path / "bin").is_dir():
-            return base_path
         for child in sorted(base_path.iterdir(), reverse=True):
             if child.is_dir() and (child / "bin").is_dir():
                 return child
+        if (base_path / "bin").is_dir():
+            return base_path
 
     return None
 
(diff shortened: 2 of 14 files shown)


src/winml/modelkit/session/monitor/qnn/viewer.py
```diff
@@ -64,6 +66,8 @@ def find_qnn_sdk() -> Path | None:
     for base_path in _COMMON_SDK_PATHS:
         if not base_path.is_dir():
             continue
+        if (base_path / "bin").is_dir():
```
DingmaomaoBJTU: The early-return for `(base_path / 'bin').is_dir()` changes existing semantics: previously if both a versioned child AND the base_path itself had `bin/`, the versioned child would win (latest version, sorted descending). Now base_path wins unconditionally. Consider placing this check *after* the versioned-dir loop as a fallback, so versioned installs still take priority over a flat/stale top-level `bin/`.

src/winml/modelkit/session/monitor/qnn/viewer.py
```diff
@@ -96,6 +100,18 @@ def _find_viewer_exe(sdk_root: Path | None = None) -> Path | None:
     return None
 
 
+def _find_optrace_reader(viewer: Path) -> Path | None:
+    """Locate the optrace reader matching the selected viewer architecture."""
+    bin_dir = viewer.parent
+    if bin_dir.parent.name == "bin":
+        sdk_root = bin_dir.parent.parent
```
DingmaomaoBJTU, line 108: The arch-matching logic assumes `lib/` mirrors `bin/` layout (`lib/<arch>/QnnHtpOptraceProfilingReader.dll`). Edge case: if the viewer falls back to `bin/qnn-profile-viewer.exe` (no arch subdirectory), `bin_dir.parent.name` is the SDK root name (not `'bin'`), so it takes the `else` branch and checks `sdk_root / 'lib' / _OPTRACE_READER_NAME`. This is likely correct for flat layouts but a clarifying comment would help future readers.

src/winml/modelkit/session/session.py
```diff
@@ -205,6 +205,12 @@ def _build_session_options(
         if provider_options is not None
         else _build_provider_options(ep_device, ep_config, ep_monitor)
     )
+    logger.info(
+        "Building session options for ep=%s device=%s with provider_options=%s",
+        ep_device.ep.arg0,
```
DingmaomaoBJTU, line 210: The `options` dict logged at INFO may contain sensitive values (e.g. auth tokens for remote compilation endpoints). Consider logging only the keys at INFO and the full dict at DEBUG.

src/winml/modelkit/session/monitor/qnn_monitor.py
```diff
@@ -681,31 +681,31 @@ def _try_qhas(
         if qhas_override is not None:
             # Offline path: caller supplied the QHAS JSON; parse directly.
             if not qhas_override.is_file():
-                logger.debug("QNNMonitor: qhas_override %s is not a file", qhas_override)
+                logger.info("QNNMonitor: qhas_override %s is not a file", qhas_override)
```
DingmaomaoBJTU, line 684: nit: Bulk `debug` -> `info` upgrade means these messages now appear in the default INFO stream for all users, even those who never use QNN. If the goal is QNN diagnosability, consider keeping at DEBUG and surfacing via `--verbose`.

src/winml/modelkit/commands/perf.py
```diff
@@ -1142,10 +1140,9 @@ def _run_benchmark_monitored(self) -> PerfStats:
             output_dir=output_dir,
         )
 
-        # HWMonitor is best-effort: required only for the live-chart UI on
-        # --monitor. When it's unavailable but op-tracing is requested, run
-        # without HW telemetry rather than degrading op-tracing to a no-op.
-        hw_available = HWMonitor.is_available()
+        # Keep system telemetry under explicit --monitor control. The EP monitor
+        # remains active independently when op-tracing needs profiling data.
+        hw_available = self.config.monitor and HWMonitor.is_available()
```
DingmaomaoBJTU, line 1145: Clean fix decoupling HW monitoring from op-tracing. nit: `hw_available = self.config.monitor and HWMonitor.is_available()` followed by `if self.config.monitor and not hw_available` is slightly indirect — it means is_available() was False. Reads fine, just noting for posterity.

DingmaomaoBJTU (commented):
Good changes overall. The --ep-options extension to compile is well-tested, the --reader DLL path fix is the correct approach for the newer QHAS SDK layout, and decoupling HW monitoring from op-tracing is the right semantic. A few inline observations below (none blocking).

src/winml/modelkit/session/monitor/qnn_monitor.py
xieofxie: currently if it is info, -v will show it. by default not. debug is for -vv

src/winml/modelkit/session/monitor/qnn/viewer.py
xieofxie: Moved the flat-root check after the versioned-child scan and added a regression test preserving versioned SDK precedence.

src/winml/modelkit/session/monitor/qnn/viewer.py
xieofxie: Added comments distinguishing the architecture-specific mirrored bin/lib layout from the flat bin/lib fallback.

src/winml/modelkit/session/session.py
xieofxie: Changed INFO logging to include only sorted provider-option keys, with the full options retained at DEBUG, and added coverage ensuring values are absent from INFO logs.

src/winml/modelkit/session/monitor/qnn_monitor.py
xieofxie: Keeping these diagnostics at INFO intentionally: default output remains silent, `-v` enables INFO, and `-vv` enables DEBUG.

src/winml/modelkit/commands/perf.py
xieofxie: Agreed this is a minor readability tradeoff. Keeping the short-circuit expression because it avoids probing HWMonitor unless `--monitor` was explicitly requested, with tests covering both paths.

DingmaomaoBJTU (approved):
LGTM. Already reviewed with inline comments earlier (no blocking issues). The --ep-options extension, --reader DLL path fix, and HW monitor decoupling are all correct.

PR #1252, opened 2026-07-30
Auto-compile raw ONNX models for QNN detail tracing

xieofxie:
## Summary

- detect raw ONNX inputs used with QNN detail op-tracing
- automatically run EPContext compilation with the required `optrace` profiling options
- preserve explicit `--no-compile` and `--skip-build` opt-outs
- restrict automatic compilation to QNN targets and document the behavior

## Validation

- `uv run pytest tests/unit/commands/test_perf_optracing.py tests/unit/models/auto/test_auto_onnx.py -q`
- `uv run pytest tests/unit/commands/test_perf_cli.py tests/unit/commands/test_perf_save_footer.py -q`


src/winml/modelkit/commands/perf.py
```diff
@@ -2960,10 +2962,49 @@ def perf(
...
 
+    compile_ep_options = None
+    from ..session import short_ep_name
+
+    qnn_tracing_target = (
+        short_ep_name(ep_name) == "qnn"
+        if ep_name is not None
+        else device.lower() in ("auto", "npu")
+    )
+    if op_tracing == "detail" and is_onnx and qnn_tracing_target:
```
DingmaomaoBJTU, line 3042: The auto-compile block uses `is_onnx` from line 2769 (`model_input.kind is ModelInputKind.ONNX_FILE`), but `is_onnx` is redefined at line 3045 (`model_path.suffix.lower() == '.onnx'`). Both should agree for `.onnx` files, but the redundant redefinition at 3045 is now dead code for any path that reaches the auto-compile block. Consider either (a) removing the later reassignment and using the `model_input.kind` check throughout, or (b) adding a comment that the redefinition is intentional for the non-auto-compile code path.

src/winml/modelkit/commands/perf.py
```diff
@@ -2960,10 +2962,49 @@ def perf(
...
+        from ..onnx import is_compiled_onnx
+
+        if not is_compiled_onnx(model_path):
+            no_compile_source = ctx.get_parameter_source("no_compile")
```
DingmaomaoBJTU, line 3046: `ctx.get_parameter_source('no_compile')` — the Click parameter name for `--compile/--no-compile` flags is typically the positive form (`compile`), not `no_compile`. Verify this returns the correct source; if the parameter is registered as `compile` then this check would always return `DEFAULT` and the `--no-compile` guard would never fire. The test mocks `BenchmarkConfig` before the check runs so this wouldn't be caught by the test.

src/winml/modelkit/commands/perf.py
```diff
@@ -2960,10 +2962,49 @@ def perf(
...
+                )
+            if skip_build and skip_build_source is not click.core.ParameterSource.DEFAULT:
+                raise click.UsageError(
+                    "--op-tracing detail requires a compiled EPContext model; "
+                    "--skip-build prevents automatic compilation."
+                )
+
+            no_compile = False
+            skip_build = False
+            compile_ep_options = {
```
DingmaomaoBJTU, line 3061: `profiling_file_path` is set to a path derived from `output` (the JSON report path). If the user passes a relative `-o result.json`, `output.with_name(...)` resolves relative to cwd but QNN writes relative to the process working directory at compile time, which could be a temp build dir. The `.resolve()` call here handles that — good.

src/winml/modelkit/models/auto.py
```diff
@@ -215,6 +217,8 @@ def from_onnx(
             override=config,
             no_compile=no_compile,
         )
+        if compile_provider_options and config.compile is not None:
```
DingmaomaoBJTU: This guard checks `config.compile is not None` but if `compile_provider_options` is provided and `config.compile` is `None` (e.g. `--no-compile` was set upstream but `compile_ep_options` leaked through), the options would silently be dropped. The CLI code already prevents this combination, but a defensive `else` branch with a warning would make this more robust against future callers.

src/winml/modelkit/commands/perf.py
```diff
@@ -2960,10 +2962,49 @@ def perf(
...
         output = generate_output_path(hf_model, submodel=submodel)
+    model_path = Path(hf_model)
 
     # Refuse to clobber an existing report unless the user opted in.
     cli_utils.guard_output(output, overwrite)
 
+    compile_ep_options = None
+    from ..session import short_ep_name
+
+    qnn_tracing_target = (
```
DingmaomaoBJTU: The `qnn_tracing_target` heuristic falls through to `device.lower() in ('auto', 'npu')` when `--ep` is omitted. On a host with only DML (GPU) registered, `--device auto` would still trigger the auto-compile path, which would fail at compile time because QNN isn't available. Consider gating on whether QNN is actually registered (e.g. via `default_ep_for_device`) rather than assuming NPU/auto always means QNN. Non-blocking for this iteration since detail tracing is QNN-specific anyway.

DingmaomaoBJTU (commented):
Good feature — auto-compiling for QNN detail op-tracing is the right UX. Clear error messages for conflicting flags. A few observations below.

src/winml/modelkit/commands/perf.py
xieofxie: Removed the redundant suffix-based reassignment. The command now uses the existing ModelInputKind classification throughout.

src/winml/modelkit/models/auto.py
xieofxie: Changed this to fail explicitly with ValueError when compile-only provider options are supplied without a compile config, and added coverage for that case.

src/winml/modelkit/commands/perf.py
xieofxie: Auto/NPU requests now require QNNMonitor.is_available() before entering precompilation, so a host without QNN does not attempt a QNN compile. Added a regression test for the auto-device case.

src/winml/modelkit/commands/perf.py
xieofxie: Verified that cli_utils.compile_option explicitly registers the Click parameter as `no_compile`, so `get_parameter_source("no_compile")` is correct. The explicit `--no-compile` test exercises this guard before BenchmarkConfig construction.

src/winml/modelkit/commands/perf.py
xieofxie: Confirmed; the compile profiling path remains absolute so QNN writes it consistently regardless of the build working directory.

src/winml/modelkit/commands/perf.py
```diff
@@ -2960,10 +2962,50 @@ def perf(
...
 
+    compile_ep_options = None
+    from ..session import short_ep_name
+
+    if ep_name is not None:
+        qnn_tracing_target = short_ep_name(ep_name) == "qnn"
+    else:
+        from ..session.monitor.qnn_monitor import QNNMonitor
+
+        qnn_tracing_target = device.lower() in ("auto", "npu") and QNNMonitor.is_available()
```
DingmaomaoBJTU, line 3041: `device.lower()` — is `device` guaranteed non-None here? The Click parameter has a default (`"auto"`), so this should be safe. But if future refactoring makes it optional-at-this-point, this raises `AttributeError`. A defensive `(device or "auto").lower()` is cheap insurance.

src/winml/modelkit/commands/perf.py
```diff
@@ -2960,10 +2962,50 @@ def perf(
...
+                    "--skip-build prevents automatic compilation."
+                )
+
+            no_compile = False
+            skip_build = False
+            compile_ep_options = {
+                **(ep_provider_options or {}),
+                "profiling_level": "optrace",
+                "profiling_file_path": str(
+                    output.with_name(f"{output.stem}_compile_optrace.csv").resolve()
```
DingmaomaoBJTU, line 3065: The `profiling_file_path` is derived from `output.with_name(...)` — keeps the compile trace next to the benchmark report. One edge case: if `output` has no stem (user passes `-o .json`), the result is `_compile_optrace.csv`. Unlikely but worth noting.

src/winml/modelkit/models/auto.py
```diff
@@ -215,6 +217,10 @@ def from_onnx(
             override=config,
             no_compile=no_compile,
         )
+        if compile_provider_options:
+            if config.compile is None:
+                raise ValueError("compile_provider_options requires compilation to be enabled.")
+            config.compile.ep_config.provider_options.update(compile_provider_options)
```
DingmaomaoBJTU: `.update()` mutates `config.compile.ep_config.provider_options` in-place. If `generate_onnx_build_config` ever returns a cached/shared config object, this mutation would leak. Currently safe (configs are freshly generated), but a non-mutating `{**existing, **compile_provider_options}` would avoid the latent risk.

src/winml/modelkit/commands/perf.py
```diff
@@ -2960,10 +2962,50 @@ def perf(
...
+    if ep_name is not None:
+        qnn_tracing_target = short_ep_name(ep_name) == "qnn"
+    else:
+        from ..session.monitor.qnn_monitor import QNNMonitor
+
+        qnn_tracing_target = device.lower() in ("auto", "npu") and QNNMonitor.is_available()
+    if op_tracing == "detail" and is_onnx and qnn_tracing_target:
+        from ..onnx import is_compiled_onnx
+
+        if not is_compiled_onnx(model_path):
```
DingmaomaoBJTU, line 3045: If the model is already an EPContext binary, the auto-compile block is skipped and `compile_ep_options` stays `None` — avoids double-compiling. The profiling options only inject into the fresh compilation. Good.

DingmaomaoBJTU (commented):
Well-scoped feature that neatly closes the gap for `--op-tracing detail` on raw ONNX models. The auto-compile detection is clean — errors early when user-provided flags conflict, and the HWMonitor decoupling is a sensible simplification.

Tests are thorough: covers the auto-compile happy path, both conflicting-flag rejections, non-QNN EP fallthrough, and the `--monitor` independence.

**HWMonitor change (line ~1147):** `hw_available = self.config.monitor and HWMonitor.is_available()` removes the implicit HWMonitor activation that op-tracing used to trigger. Now `--monitor` is the single control point for system telemetry — easier to reason about.

src/winml/modelkit/commands/perf.py
xieofxie: Confirmed. Existing EPContext models bypass this block, so compile-only profiling options are injected only when a fresh EPContext compilation is required.

src/winml/modelkit/commands/perf.py
xieofxie: The Click option is required at this point and has a string default of `auto`, while the function parameter is typed as `str`, so `None` is not a valid state here. I kept the direct `.lower()` so an invalid programmatic call fails rather than being silently normalized.

src/winml/modelkit/models/auto.py
xieofxie: Changed this to assign a newly merged provider-options dictionary instead of mutating the generated config dictionary in place.

src/winml/modelkit/commands/perf.py
xieofxie: For a dotfile such as `.json`, pathlib treats the full filename as the stem, so this produces `.json_compile_optrace.csv` rather than an empty-stem filename. No special case is needed.

DingmaomaoBJTU (approved):
LGTM. Auto-compile logic is clean, error handling for conflicting flags is solid, HWMonitor decoupling simplifies the control flow, and tests cover all the relevant branches. Previous inline comments are non-blocking suggestions.

PR #1254, opened 2026-07-30
fix(cli): show correct short help for `sys`, untruncate `build`/`quantize`

xieofxie:
## Problem

`winml --help` showed a garbage description for `sys`:

```
sys   Register ``dll_path`` in a fresh subprocess; yield the ``to_dict
```

`_parse_click_help` extracted the **first decorated function's** docstring from each command module. In `sys.py` the first decorated function is an internal `@contextlib.contextmanager` helper, not the `@click.command`, so its docstring leaked into the command listing — the very first screen a new user sees.

Two other rows (`build`, `quantize`) also truncated mid-sentence at an 80-column width.

## Fix

- **`cli.py`** — `_parse_click_help` now only reads the function carrying a `@click.command` / `@click.group` decorator, and prefers an explicit `short_help=` decorator argument when present. Still AST-only (no module import), so startup stays fast. Added `_command_decorator` and `_short_help_kwarg` helpers.
- **`build.py` / `quantize.py`** — added concise `short_help=` so their rows no longer truncate.

## Result

```
build     Build a WinML-optimized ONNX model from HuggingFace or ONNX.
quantize  Quantize an ONNX model (QDQ, RTN weight-only, or FP16).
sys       Display system information for WinML CLI export.
```

## Tests

Added `TestParseClickHelp` in `tests/cli/test_help_cli.py`:
- `sys` regression test — asserts the real command docstring and that `dll_path`/`subprocess` never leak.
- Skips a decorated non-Click helper preceding the command.
- Prefers an explicit `short_help=` over the docstring.

All 55 tests in `test_help_cli.py` pass; `ruff check` clean.


src/winml/modelkit/cli.py
```diff
@@ -138,23 +138,65 @@ def _print_banner(version: str, *, _console: Console | None = None) -> None:
...
+
+    Matches both ``@click.command`` and bare ``@command`` forms, whether or
+    not they are called (``@click.command`` vs ``@click.command("name")``).
+    Returns ``None`` when the function carries no Click command decorator.
+    """
+    for decorator in node.decorator_list:
+        target = decorator.func if isinstance(decorator, ast.Call) else decorator
+        name = getattr(target, "attr", None) or getattr(target, "id", None)
+        if name in {"command", "group"}:
+            return decorator
```
DingmaomaoBJTU, line 168: `getattr(target, 'attr', None) or getattr(target, 'id', None)` — if `target` is an `ast.Attribute` with `attr='command'`, this works. But for a nested attribute like `@my_module.click.command`, `target.attr` would be `'command'` (correct). For a bare `@command` import alias, `target` would be `ast.Name` and you'd hit `getattr(target, 'id', None)`. Both cases look correct. One edge case: `@click.group()` decorated subcommand groups — the check handles this via `name in {'command', 'group'}`, good.

src/winml/modelkit/cli.py
```diff
@@ -138,23 +138,65 @@ def _print_banner(version: str, *, _console: Console | None = None) -> None:
...
+        if not isinstance(node, ast.FunctionDef):
+            continue
+        decorator = _command_decorator(node)
+        if decorator is None:
+            continue
+        short_help = _short_help_kwarg(decorator)
+        if short_help:
+            return short_help.split("\n")[0].strip()
+        docstring = ast.get_docstring(node)
+        if docstring:
```
DingmaomaoBJTU, line 214: The early `return ''` here means if the Click command function has no docstring AND no `short_help`, we stop searching. This is correct — there's exactly one Click command per module by convention — but worth a brief comment explaining why we don't continue to the next function.

src/winml/modelkit/commands/build.py
```diff
@@ -710,7 +710,7 @@ def _maybe_build_genai_bundle(
 # =============================================================================
 
 
-@click.command("build")
+@click.command("build", short_help="Build a WinML-optimized ONNX model from HuggingFace or ONNX.")
```
DingmaomaoBJTU, line 713: nit: The `short_help` here is 73 chars which could get truncated in very narrow terminals (Click truncates to terminal width minus indentation). The docstring first line presumably was even longer — so this is an improvement. Just noting.

DingmaomaoBJTU (commented):
Clean fix — the root cause is clear (`sys.py`'s `@contextlib.contextmanager isolated_ep_register` was the first decorated function, so its docstring leaked as the command's short help). The AST-based approach correctly targets only `@click.command`/`@click.group`. A couple of minor inline notes.

src/winml/modelkit/cli.py
xieofxie: Thanks — yes, both forms are covered: st.Attribute (click.command, and nested like .b.command) via .attr, and a bare imported alias (st.Name) via .id. group is matched too for the group case.

src/winml/modelkit/cli.py
xieofxie: Good call — added a comment (968f6420) noting that a module holds exactly one Click command by convention, so we stop once it's found rather than falling through to an unrelated helper's docstring.

src/winml/modelkit/commands/build.py
xieofxie: The final short_help is 60 chars (uild) / 55 chars (quantize), both under the 66-char cap at 80 cols (width − 6 − longest name), so neither truncates. The 73 was the original docstring first line. Thanks for double-checking.

src/winml/modelkit/cli.py
```diff
@@ -138,23 +138,68 @@ def _print_banner(version: str, *, _console: Console | None = None) -> None:
...
+    Returns ``None`` when the function carries no Click command decorator.
+    """
+    for decorator in node.decorator_list:
+        target = decorator.func if isinstance(decorator, ast.Call) else decorator
+        name = getattr(target, "attr", None) or getattr(target, "id", None)
+        if name in {"command", "group"}:
+            return decorator
+    return None
+
+
```
DingmaomaoBJTU, line 171: `name in {"command", "group"}` — this matches any decorator named `command` or `group`, not just Click's. In practice this is safe (CLI command modules always use Click), but a stricter check like verifying `target` is an `ast.Attribute` with `value.id == "click"` would eliminate false positives from e.g. a hypothetical `@typer.command`. Very low risk as-is.

src/winml/modelkit/cli.py
```diff
@@ -138,23 +138,68 @@ def _print_banner(version: str, *, _console: Console | None = None) -> None:
...
+    explicit ``short_help=`` decorator argument when present, otherwise the
+    first line of the function docstring (which Click uses as short help).
+
+    Only the Click command function is considered, so unrelated decorated
+    helpers in the module (e.g. a ``@contextlib.contextmanager``) never leak
+    their docstring into the command listing.
     """
     try:
         tree = ast.parse(path.read_text(encoding="utf-8"))
     except (SyntaxError, OSError):
```
DingmaomaoBJTU, line 199: Good: once the Click command function is found, the function returns immediately (either with `short_help`, docstring first line, or `""`). This prevents fallthrough to a subsequent decorated helper — exactly the bug this PR fixes.

src/winml/modelkit/commands/build.py
DingmaomaoBJTU, line 713: Explicit `short_help=` is the right approach here. The `build` command's docstring is long and Click's auto-truncation was cutting it to something unhelpful. This gives full control over what appears in `winml --help`.

DingmaomaoBJTU (commented):
Clean fix. The root cause was clear — `_parse_click_help` grabbed the first decorated function's docstring, which for `sys.py` was a `@contextlib.contextmanager` helper instead of the Click command. The new AST-based `_command_decorator` filter is precise and handles both `@click.command` and bare `@command` forms.

Explicit `short_help=` on `build` and `quantize` is the right call since their docstrings are multi-paragraph and Click's auto-truncation cut them off.

Tests are solid — regression test on actual `sys.py`, synthetic tests for decorator ordering and `short_help=` precedence.

src/winml/modelkit/cli.py
```diff
@@ -138,23 +138,68 @@ def _print_banner(version: str, *, _console: Console | None = None) -> None:
...
+    Matches both ``@click.command`` and bare ``@command`` forms, whether or
+    not they are called (``@click.command`` vs ``@click.command("name")``).
+    Returns ``None`` when the function carries no Click command decorator.
+    """
+    for decorator in node.decorator_list:
+        target = decorator.func if isinstance(decorator, ast.Call) else decorator
+        name = getattr(target, "attr", None) or getattr(target, "id", None)
+        if name in {"command", "group"}:
+            return decorator
+    return None
```
DingmaomaoBJTU, line 169: Handles all four decorator forms correctly:
- `@click.command` (bare `ast.Attribute`)
- `@click.command()` (`ast.Call` wrapping `ast.Attribute`)
- `@command` (bare `ast.Name`)
- `@command("name")` (`ast.Call` wrapping `ast.Name`)

The `getattr(target, 'attr', None) or getattr(target, 'id', None)` chain extracts the name regardless of form. Clean.

src/winml/modelkit/cli.py
```diff
@@ -138,23 +138,68 @@ def _print_banner(version: str, *, _console: Console | None = None) -> None:
...
+
+
+def _short_help_kwarg(decorator: ast.expr) -> str | None:
+    """Return an explicit ``short_help="..."`` argument on *decorator*, if any."""
+    if isinstance(decorator, ast.Call):
+        for keyword in decorator.keywords:
```
DingmaomaoBJTU, line 175: Only `ast.FunctionDef` is checked, not `ast.AsyncFunctionDef`. That's correct — Click commands are always sync functions — but worth a comment if async decorated functions are ever added to a command module (they'd be silently skipped, which is the desired behavior).

tests/cli/test_help_cli.py
```diff
@@ -192,6 +204,70 @@ def test_enabled_command_has_help_text(self, cmd: str) -> None:
...
+            textwrap.dedent(
+                '''
+                import contextlib
+                import click
+
+                @contextlib.contextmanager
+                def _helper():
+                    """Internal helper docstring that must not be used."""
+                    yield
+
```
DingmaomaoBJTU, line 243: Good regression test: directly asserts on the actual `sys.py` module rather than a synthetic fixture. If someone adds another decorated helper above the Click command in `sys.py`, this test catches it immediately.

tests/cli/test_help_cli.py
```diff
@@ -192,6 +204,70 @@ def test_enabled_command_has_help_text(self, cmd: str) -> None:
...
+                @click.command("fake", short_help="Concise summary.")
+                def fake():
+                    """A much longer docstring line that should be ignored."""
+                '''
+            ),
+            encoding="utf-8",
+        )
+        assert _parse_click_help(module) == "Concise summary."
+
+
```
DingmaomaoBJTU, line 314: Missing edge case test: a module with a Click command that has **no docstring and no `short_help`** should return `""`. Currently tested implicitly by the early `return ""` at line 183, but an explicit test would lock the contract.

DingmaomaoBJTU (commented):
Second pass — checked edge cases in AST parsing, decorator form variations, and multi-command modules. All solid.

src/winml/modelkit/cli.py
xieofxie: Good point — tightened in 1b1d777f. The attribute form now matches only when the owner is the click name (@click.command/@click.group), plus the bare @command/@group alias from rom click import .... A @typer.command (owner 	yper) no longer matches, and I added a regression test for it.

src/winml/modelkit/cli.py
xieofxie: Exactly — that immediate return is the crux of the fix. Covered by the sys regression test in TestParseClickHelp.

src/winml/modelkit/commands/build.py
xieofxie: Thanks — agreed. Explicit short_help keeps winml --help stable regardless of docstring length or terminal width.

DingmaomaoBJTU (approved):
LGTM. AST-based Click command detection is correct and well-tested. All inline comments are non-blocking.

src/winml/modelkit/cli.py
xieofxie: Note the decorator matching was tightened after this comment (1b1d777f): the attribute form now also requires the owner to be the click name, so @typer.command can't false-positive. All four Click forms you list are still handled.

src/winml/modelkit/cli.py
xieofxie: Added a comment (75ee3c62) noting ast.AsyncFunctionDef is intentionally skipped since Click commands are always sync.

tests/cli/test_help_cli.py
xieofxie: Thanks — that was the intent: assert against the real sys.py so any future helper added above the command is caught.

tests/cli/test_help_cli.py
xieofxie: Good call — added test_command_without_docstring_or_short_help_returns_empty (75ee3c62) to lock the contract explicitly, with a trailing helper to prove there's no fallthrough.

DingmaomaoBJTU (approved):

PR #1255, opened 2026-07-30
Consolidate development dependencies

xieofxie:
## Summary

- merge the legacy `dev` optional extra into the default `dev` dependency group
- update development setup documentation to use locked synchronization
- update Azure pipeline editable installs to consume the dependency group
- keep the lockfile metadata aligned without unrelated package churn

### DIFF
--- a/.pipelines/templates/e2e-test-jobs.yml
+++ b/.pipelines/templates/e2e-test-jobs.yml
@@ -157,14 +157,11 @@ jobs:
         displayName: 'Authenticate pip with Azure Artifacts'
 
       # NOTE: This is the single source of truth for installing project deps
-      # into the venv. Every later step uses `uv run --no-sync` to *skip*
-      # uv's lockfile sync and rely on what was installed here. Do NOT remove
-      # this step assuming `uv run` will install on demand -- it won't, with
-      # `--no-sync` -- and do NOT drop `--no-sync` below assuming it's a
-      # no-op: without it, `uv run` would re-resolve from uv.lock and may
-      # silently replace this editable install.
+      # into the venv. Every later step uses `uv run --no-sync` and relies on
+      # the environment created here. Do not remove this step while those
+      # commands use `--no-sync`.
       - script: |
-          uv pip install -e .[dev]
+          uv pip install -e . --group dev
         workingDirectory: $(Build.SourcesDirectory)
         displayName: 'Install dependencies'
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -16,7 +16,7 @@ See the [README](./README.md#getting-started) for prerequisites and installation
 ```bash
 git clone https://github.com/microsoft/winml-cli.git
 cd winml-cli
-uv sync --extra dev
+uv sync --locked
 uv run pre-commit install
 ```
 
--- a/docs/getting-started/installation.md
+++ b/docs/getting-started/installation.md
@@ -27,7 +27,7 @@ uv pip install winml-cli
 
     ```powershell
     uv python install cpython-3.11-windows-x86_64-none
-    uv sync --extra dev --python cpython-3.11-windows-x86_64-none
+    uv sync --locked --python cpython-3.11-windows-x86_64-none
     ```
 
     The x64 interpreter runs under Windows emulation on Arm64 hardware; `winml sys` still reports the Arm64 machine, and the NPU/GPU/CPU providers work normally.
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -84,25 +84,6 @@ dependencies = [
   "windowsml==2.0.300 ; sys_platform == 'win32'",
 ]
 
-optional-dependencies.dev = [
-  "ipykernel>=6.30.1",
-  "ipywidgets>=8.1.7",
-  "jupyter>=1.1.1",
-  "markdown-it-py>=3",
-  "matplotlib>=3.10",
-  "mkdocs-jupyter>=0.25",
-  "mkdocs-material>=9.5",
-  "mypy>=1.18",
-  "nbconvert>=7.16",
-  "pymdown-extensions>=10.7",
-  "pytest>=8.4",
-  "pytest-cov>=7",
-  "pytest-timeout>=2.3",
-  "ruff>=0.15",
-  "seaborn>=0.13",
-  "timm>=1.0.20",
-  "types-colorama>=0.4.15.20250801",
-]
 optional-dependencies.audio = [ "soundfile>=0.13" ]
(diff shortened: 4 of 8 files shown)


pyproject.toml
```diff
@@ -167,12 +148,25 @@ torchvision = [
 
 [dependency-groups]
```
DingmaomaoBJTU, line 149: The `dev` group consolidates all previous `optional-dependencies.dev` entries plus the existing `dependency-groups.dev` entries. Looks correct — no packages lost. Note that `timm>=1.0.20` (previously in `[dev]` extra) is dropped because `timm>=1.0.22` already exists in the main `dependencies` list. Good cleanup.

CONTRIBUTING.md
```diff
@@ -16,7 +16,7 @@ See the [README](./README.md#getting-started) for prerequisites and installation
 ```bash
 git clone https://github.com/microsoft/winml-cli.git
 cd winml-cli
-uv sync --extra dev
+uv sync --locked
```
DingmaomaoBJTU, line 19: `uv sync --locked` installs the full default group (which includes `dev`) — this works because `dependency-groups.dev` is now the default dev group. Just confirming: does `uv sync --locked` also install the main package in editable mode? If not, contributors may need `uv pip install -e .` separately. (uv's default behavior is editable install for workspace packages, so this should be fine.)

DingmaomaoBJTU (commented):
Clean migration from `optional-dependencies.dev` to `dependency-groups.dev`. LGTM overall.

**Verified:**
- CI pipelines updated from `uv pip install -e .[dev]` to `uv pip install -e . --group dev`
- Docs updated from `uv sync --extra dev` to `uv sync --locked`
- `timm` duplicate removed (was in both `dependencies` and `dev`)
- Lock file correctly reflects the new structure

pyproject.toml
xieofxie: Agreed. The main dependency already requires the newer `timm>=1.0.22`, so retaining the weaker development-only constraint would be redundant.

CONTRIBUTING.md
xieofxie: Confirmed: `uv sync --locked` installs the workspace project in editable mode by default and creates `.venv\Scripts\winml.exe`; a separate editable install is not needed.

pyproject.toml
```diff
@@ -167,12 +148,25 @@ torchvision = [
...
+  "ipykernel>=6.30.1",
+  "ipywidgets>=8.1.7",
   "jupyter>=1.1.1",
+  "markdown-it-py>=3",
+  "matplotlib>=3.10",
+  "mkdocs-jupyter>=0.25",
+  "mkdocs-material>=9.5",
+  "mypy>=1.18",
+  "nbconvert>=7.16",
   "pandas-stubs>=3.0.0.260204",
```
DingmaomaoBJTU, line 160: `pytest-timeout>=2.4.0` — the old `optional-dependencies.dev` had `>=2.3` while the old `dependency-groups.dev` had `>=2.4.0`. The merged version correctly keeps the stricter constraint. Good.

.pipelines/templates/e2e-test-jobs.yml
```diff
@@ -157,14 +157,11 @@ jobs:
...
-      # this step assuming `uv run` will install on demand -- it won't, with
-      # `--no-sync` -- and do NOT drop `--no-sync` below assuming it's a
-      # no-op: without it, `uv run` would re-resolve from uv.lock and may
-      # silently replace this editable install.
+      # into the venv. Every later step uses `uv run --no-sync` and relies on
+      # the environment created here. Do not remove this step while those
+      # commands use `--no-sync`.
       - script: |
-          uv pip install -e .[dev]
+          uv pip install -e . --group dev
```
DingmaomaoBJTU, line 164: `uv pip install -e . --group dev` — correct migration from `-e .[dev]` (PEP 621 extra) to `--group dev` (PEP 735 dependency group). The `--group` flag explicitly selects the group when using `uv pip install` (unlike `uv sync` which includes groups by default).

DingmaomaoBJTU (commented):
Second pass — full package audit done. All 17 packages from `optional-dependencies.dev` are present in the new `dependency-groups.dev` (merged with the 11 that were already there). `timm` correctly dropped (already a runtime dep). `pytest-timeout` version correctly takes the stricter `>=2.4.0` from the old dep-groups over `>=2.3` from the old extras.

`uv sync --locked` is the right replacement: uv installs all dependency groups by default (unlike extras which required `--extra`), so the `dev` group is included automatically. `--locked` ensures the lockfile isn't modified.

LGTM.

CONTRIBUTING.md
DingmaomaoBJTU, line 19: `uv sync --locked` installs all dependency groups (including `dev`) by default and asserts the lockfile is up-to-date. This is semantically equivalent to the old `uv sync --extra dev` now that dev deps have moved from extras to a dependency group. Correct.

docs/getting-started/installation.md
```diff
@@ -27,7 +27,7 @@ uv pip install winml-cli
 
     ```powershell
     uv python install cpython-3.11-windows-x86_64-none
-    uv sync --extra dev --python cpython-3.11-windows-x86_64-none
+    uv sync --locked --python cpython-3.11-windows-x86_64-none
```
DingmaomaoBJTU, line 30: The x64 cross-platform install path also updated. `--locked` preserves the lockfile assertion while `--python` selects the target interpreter. Consistent with the other doc changes.

.pipelines/templates/e2e-test-jobs.yml
xieofxie: Agreed. The pipeline now selects the PEP 735 development group explicitly while preserving the editable project install.

docs/getting-started/installation.md
xieofxie: Agreed. The interpreter selection remains explicit for Windows on Arm while dependency installation now uses the locked development group.

CONTRIBUTING.md
xieofxie: Confirmed. The default development group plus `--locked` provides the intended contributor setup without selecting an extra.

pyproject.toml
xieofxie: Agreed. The consolidated group intentionally retains the stricter `pytest-timeout>=2.4.0` constraint.

DingmaomaoBJTU (approved):
LGTM. Clean migration from optional-dependencies to dependency-groups. All packages accounted for, CI and docs updated consistently.

DingmaomaoBJTU (approved):

PR #1256, opened 2026-07-30
feat: Add EP and device targeting to optimize

xieofxie:
## Summary

- add \--ep\ and \--device\ options to resolve provider-aware optimization targets
- bind ORT graph optimization to the resolved EP device while preserving CPU as the default
- report node-count increases accurately and cover target forwarding/provider binding


DingmaomaoBJTU (commented):
Good feature — makes `--check-optim` target-aware so it can skip capabilities irrelevant to the target EP and run the ORT optimization session with the resolved provider. Well-tested.

src/winml/modelkit/optim/analysis.py
```diff
@@ -340,6 +342,14 @@ def _iter_findings(
         ]
 
         for cap_name, cap in probe_caps:
+            ep_device = optimizer_kwargs.get("ep_device")
+            if ep_device is not None and cap.ep_constraint is not None:
+                from ..utils.constants import normalize_ep_name
+
```
DingmaomaoBJTU: The `ep_constraint` filter uses `normalize_ep_name(name) == target_ep` but `cap.ep_constraint` contains short names like `("DML",)`. Does `normalize_ep_name("DML")` return `"DmlExecutionProvider"`? And is `ep_device.device.ep_name` the full provider name (`"DmlExecutionProvider"`)? If so, this comparison works. But if either side uses inconsistent casing/format, applicable capabilities could be silently skipped. Worth a quick debug log when filtering occurs.

src/winml/modelkit/optim/analysis.py
```diff
@@ -305,6 +306,7 @@ def _iter_findings(
 
     # Baseline kwargs = every capability at its default value.
     default_kwargs = {cap.python_name: cap.default for cap in capabilities.values()}
+    default_kwargs.update(optimizer_kwargs)
     kebab_defaults = {name: cap.default for name, cap in capabilities.items()}
 
     # Mandatory pre-stage — mirrors Optimizer.optimize(). The clone is required:
```
DingmaomaoBJTU, line 327: `default_kwargs.update(optimizer_kwargs)` merges `ep_device` into the **default** kwargs dict passed to every `pipe.build_config()` call. This means the baseline pipe config also gets `ep_device` — which is correct (the baseline should use the same EP as the probes for a fair comparison). Good.

src/winml/modelkit/commands/optimize.py
```diff
@@ -529,6 +595,13 @@ def optimize(
     if verbose:
         optimizer_kwargs["verbose"] = True
 
+    if ep_device is not None:
+        assert target is not None
+        optimizer_kwargs["ep_device"] = ep_device
+        console.print(
```
DingmaomaoBJTU, line 616: `assert target is not None` — using `assert` for runtime validation means this check disappears under `python -O`. Since this is in a CLI code path (not a hot loop), consider `if target is None: raise` or just rely on the earlier `if ep is not None or device is not None` guard. Low risk since the condition is logically unreachable, but asserts in production code are fragile.

src/winml/modelkit/optim/pipes/graph.py
```diff
@@ -562,15 +577,31 @@ def process(self, model: onnx.ModelProto, config: ORTGraphPipeConfig) -> onnx.Mo
...
+                from ...session import lookup_device_spec
+
+                spec = lookup_device_spec(
+                    config.ep_device.device.ep_name,
+                    config.ep_device.device.device_type.lower(),
+                )
+                provider_options = dict(spec.default_provider_options) if spec else {}
+                sess_opts.add_provider_for_devices(
+                    [config.ep_device.device.ort_handle],
+                    provider_options,
```
DingmaomaoBJTU, line 597: When `config.ep_device is not None`, the session is created without explicitly specifying providers (`ort.InferenceSession(str(input_file), sess_opts)`). The EP is bound via `sess_opts.add_provider_for_devices`. Is this guaranteed to make ORT use that provider for optimization? If ORT falls back to CPU when the provider isn't available, optimizations may silently not apply. Consider verifying the session's active providers after creation, or adding a debug log.

src/winml/modelkit/commands/optimize.py
```diff
@@ -543,10 +616,21 @@ def optimize(
 
         # Report results
         optimized_nodes = len(optimized_model.graph.node)
-        reduction = (1 - optimized_nodes / original_nodes) * 100 if original_nodes else 0
+        node_change = (
+            abs(optimized_nodes - original_nodes) / original_nodes * 100 if original_nodes else 0
+        )
 
         console.print(f"\n[bold green]Success![/bold green] Model optimized: {output}")
```
DingmaomaoBJTU, line 638: Node change reporting now handles increases:
`
node_change = abs(optimized_nodes - original_nodes) / original_nodes * 100
`
Good improvement. Minor: when `optimized_nodes == original_nodes`, `change_label = "change"` with `0.0% change` — reads slightly odd. Maybe skip the percentage line entirely when there's no change?

src/winml/modelkit/utils/cli.py
```diff
@@ -440,7 +443,11 @@ def ep_option(
     if optional_message:
```
DingmaomaoBJTU, line 448: `include_cuda` defaults to `False` for backward compat. Only `optimize` passes `include_cuda=True`. Makes sense — CUDA EP is relevant for graph optimization but typically excluded from other commands that focus on edge hardware.

src/winml/modelkit/optim/analysis.py
xieofxie: Normalized both the resolved provider name and each constraint before comparison, and added a debug log whenever a capability is filtered for the target EP.

src/winml/modelkit/optim/analysis.py
xieofxie: Agreed. The baseline and probes intentionally receive the same resolved EP device so the diff isolates only the capability change.

src/winml/modelkit/commands/optimize.py
xieofxie: Replaced the production assert with an explicit runtime error for the logically unreachable missing-target state.

src/winml/modelkit/optim/pipes/graph.py
xieofxie: The optimization session now checks get_providers() and fails explicitly if the requested normalized provider is not active, while logging the active provider list at debug level.

src/winml/modelkit/commands/optimize.py
xieofxie: Updated the unchanged case to render `Nodes: X -> X (no change)` instead of `0.0% change`.

src/winml/modelkit/utils/cli.py
xieofxie: Agreed. CUDA remains opt-in on the shared decorator and is enabled only for the optimize command.

DingmaomaoBJTU (commented):
Second pass. Found an important **merge interaction with PR #1257** in `_iter_findings` — see inline comment on `analysis.py`.

Also verified: `default_kwargs.update(optimizer_kwargs)` forwards `ep_device` to ALL pipes via `build_config(**kwargs)`. Safe because other pipes accept `**kwargs` and ignore unknown keys. The provider verification after ORT session creation (graph.py) addresses my first-pass concern about silent EP fallback.

src/winml/modelkit/optim/analysis.py
```diff
@@ -340,6 +342,14 @@ def _iter_findings(
         ]
 
         for cap_name, cap in probe_caps:
+            ep_device = optimizer_kwargs.get("ep_device")
+            if ep_device is not None and cap.ep_constraint is not None:
+                from ..utils.constants import normalize_ep_name
+
+                target_ep = ep_device.device.ep_name
+                if not any(normalize_ep_name(name) == target_ep for name in cap.ep_constraint):
```
DingmaomaoBJTU: **⚠️ Cross-PR interaction with #1257:** The `continue` after this check exits the loop body *before* the point where PR #1257 wraps it in `try/finally { complete_probe(cap_name) }`. When both PRs merge, EP-filtered capabilities bypass `complete_probe`, so the progress bar stalls at `N-k/N`.

**Fix (in whichever merges second):** move this check inside the `try` block, or call `complete_probe(cap_name)` before the `continue`.

src/winml/modelkit/commands/optimize.py
```diff
@@ -543,10 +616,21 @@ def optimize(
...
-        node_info = f"Nodes: {original_nodes} -> {optimized_nodes} ({reduction:.1f}% reduction)"
+        if optimized_nodes < original_nodes:
+            change_label = "reduction"
```
DingmaomaoBJTU, line 640: Node change reporting now handles increase, reduction, and no-change cases cleanly. Displaying `(no change)` when node count is equal is better than `0.0% change`.

src/winml/modelkit/optim/pipes/graph.py
```diff
@@ -562,15 +577,31 @@ def process(self, model: onnx.ModelProto, config: ORTGraphPipeConfig) -> onnx.Mo
...
                 )
 
+            if config.ep_device is not None:
+                from ...session import lookup_device_spec
+
+                spec = lookup_device_spec(
+                    config.ep_device.device.ep_name,
+                    config.ep_device.device.device_type.lower(),
+                )
+                provider_options = dict(spec.default_provider_options) if spec else {}
```
DingmaomaoBJTU, line 594: When `lookup_device_spec` returns `None` (unknown EP), `provider_options` defaults to `{}`. Safe — ORT still binds the provider, just without custom options. Consider a debug log noting the empty-options fallback for troubleshooting.

DingmaomaoBJTU (approved):
LGTM. EP/device targeting is well-implemented with proper provider verification. The merge interaction with PR #1257 (EP constraint continue bypasses complete_probe) needs a one-line fix in whichever merges second — non-blocking for approval.

src/winml/modelkit/optim/analysis.py
xieofxie: Merged main and moved EP filtering inside the callback-protected try/finally, so every filtered probe now calls complete_probe. Added regression assertions covering start and completion for both compatible and filtered capabilities.

src/winml/modelkit/commands/optimize.py
xieofxie: Confirmed and retained the explicit `(no change)` output while resolving the merge.

src/winml/modelkit/optim/pipes/graph.py
xieofxie: Added a debug log when no device specification is found and empty provider options are used, with targeted coverage for the fallback.

DingmaomaoBJTU (approved):
LGTM. Re-reviewed the current head after the latest main merges. The prior cross-PR progress issue is resolved: EP-constrained skips now occur inside the try/finally, so complete_probe always runs, with regression coverage for started/completed callbacks. Provider fallback validation and target propagation remain correct.

PR #1257, opened 2026-07-30
Show optimization capability probe progress

xieofxie:
## Summary

- replace the indeterminate `--check-optim` spinner with a determinate progress bar
- display the capability currently being checked
- track each unique capability once, including skipped and failed probes
- add analyzer callback and CLI coverage for probe progress

image

### DIFF
--- a/src/winml/modelkit/commands/optimize.py
+++ b/src/winml/modelkit/commands/optimize.py
@@ -28,6 +28,7 @@ from typing import TYPE_CHECKING, Any, TypeVar
 import click
 from rich.console import Console
 from rich.markup import escape
+from rich.progress import BarColumn, MofNCompleteColumn, Progress, TextColumn, TimeElapsedColumn
 
 from ..onnx import load_onnx, save_onnx
 from ..utils import cli as cli_utils
@@ -256,8 +257,20 @@ def _run_check_optim(model: Path, all_caps: dict[str, Any], verbose: bool) -> No
         f"[bold]Probing {probe_count} optimization capabilities...[/bold] "
         "[dim](this can take a while on large models)[/dim]"
     )
-    with console.status("[bold]Analyzing...[/bold]", spinner="dots"):
-        findings = analyze_model(onnx_model, all_caps)
+    with Progress(
+        TextColumn("[bold]Checking[/bold] [cyan]{task.description}[/cyan]"),
+        BarColumn(),
+        MofNCompleteColumn(),
+        TimeElapsedColumn(),
+        console=console,
+    ) as progress:
+        task_id = progress.add_task("preparing capabilities", total=probe_count)
+        findings = analyze_model(
+            onnx_model,
+            all_caps,
+            on_probe_start=lambda name: progress.update(task_id, description=name),
+            on_probe_complete=lambda _: progress.advance(task_id),
+        )
 
     _render_check_optim(console, findings, verbose)
 
--- a/src/winml/modelkit/optim/analysis.py
+++ b/src/winml/modelkit/optim/analysis.py
@@ -32,7 +32,7 @@ from onnx import AttributeProto, GraphProto, ModelProto, NodeProto, TensorProto
 
 
 if TYPE_CHECKING:
-    from collections.abc import Iterator
+    from collections.abc import Callable, Iterator
 
     from .registry import CapabilityDef
 
@@ -283,6 +283,8 @@ def _run_pipe(pipe: Any, model: ModelProto, config: Any) -> ModelProto:
 def _iter_findings(
     model: ModelProto,
     capabilities: dict[str, CapabilityDef],
+    on_probe_start: Callable[[str], None] | None = None,
+    on_probe_complete: Callable[[str], None] | None = None,
 ) -> Iterator[tuple[CapabilityFinding, ModelProto]]:
     """Yield ``(finding, produced_model)`` for every applicable optimization.
 
@@ -303,6 +305,18 @@ def _iter_findings(
     from .pipes import PIPES
     from .registry import BoolCapability, auto_enable_dependencies
 
+    remaining_probes = Counter(
+        name
+        for pipe_class in PIPES
+        for name, cap in pipe_class.capabilities.items()
+        if isinstance(cap, BoolCapability) and not cap.default
+    )
+
+    def complete_probe(cap_name: str) -> None:
+        remaining_probes[cap_name] -= 1
+        if remaining_probes[cap_name] == 0 and on_probe_complete is not None:
+            on_probe_complete(cap_name)
+
     # Baseline kwargs = every capability at its default value.
     default_kwargs = {cap.python_name: cap.default for cap in capabilities.values()}
     kebab_defaults = {name: cap.default for name, cap in capabilities.items()}
@@ -315,6 +329,11 @@ def _iter_findings(
 
     for pipe_class in PIPES:
         pipe = pipe_class()
+        probe_caps = [
+            (name, cap)
+            for name, cap in pipe.capabilities.items()
+            if isinstance(cap, BoolCapability) and not cap.default
+        ]
 
         base_config = pipe.build_config(**default_kwargs)
         try:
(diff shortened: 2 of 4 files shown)


src/winml/modelkit/optim/analysis.py
```diff
@@ -303,6 +305,18 @@ def _iter_findings(
...
 
+    remaining_probes = Counter(
+        name
+        for pipe_class in PIPES
+        for name, cap in pipe_class.capabilities.items()
+        if isinstance(cap, BoolCapability) and not cap.default
+    )
+
+    def complete_probe(cap_name: str) -> None:
+        remaining_probes[cap_name] -= 1
```
DingmaomaoBJTU, line 316: `remaining_probes` is a `Counter` counting how many times each capability name appears across all pipes. But each capability name should appear in exactly one pipe (capabilities are registered per-pipe). So `remaining_probes[cap_name]` should always be 1. Is the `Counter` + decrement pattern future-proofing for a scenario where the same capability could span multiple pipes? If not, a simpler `set` of completed names would suffice. Either way, the logic is correct.

src/winml/modelkit/optim/analysis.py
```diff
@@ -328,76 +347,76 @@ def _iter_findings(
...
+                # Enable only this capability (plus its dependencies) on top of
+                # the all-defaults configuration.
+                kebab = dict(kebab_defaults)
+                kebab[cap_name] = True
+                kebab = auto_enable_dependencies(kebab, capabilities)
+                probe_kwargs = {
+                    capabilities[name].python_name: value
+                    for name, value in kebab.items()
+                    if name in capabilities
+                }
```
DingmaomaoBJTU, line 370: Good: the `finally` block ensures `complete_probe(cap_name)` fires even when the inner code uses `continue` (e.g. when `should_process` returns False). This prevents the progress bar from stalling on skipped capabilities. Clean pattern.

src/winml/modelkit/commands/optimize.py
```diff
@@ -256,8 +257,20 @@ def _run_check_optim(model: Path, all_caps: dict[str, Any], verbose: bool) -> No
...
     )
-    with console.status("[bold]Analyzing...[/bold]", spinner="dots"):
-        findings = analyze_model(onnx_model, all_caps)
+    with Progress(
+        TextColumn("[bold]Checking[/bold] [cyan]{task.description}[/cyan]"),
+        BarColumn(),
+        MofNCompleteColumn(),
+        TimeElapsedColumn(),
+        console=console,
+    ) as progress:
```
DingmaomaoBJTU, line 266: The progress bar's `total=probe_count` matches the `on_probe_complete` callback count because `analyze_model` fires `on_probe_complete` once per boolean capability that's off by default. The `advance` happens in `on_probe_complete`, so skipped and failed probes also advance the bar — correct UX (users want to see progress regardless of result).

src/winml/modelkit/optim/analysis.py
```diff
@@ -328,76 +347,76 @@ def _iter_findings(
                 getattr(pipe, "name", pipe_class.__name__),
                 exc,
             )
+            for cap_name, _ in probe_caps:
+                complete_probe(cap_name)
             continue
         base_nodes: dict[tuple[Any, ...], tuple[bytes, NodeRef]] = {}
         _collect_nodes(base_out.graph, (), base_nodes)
```
DingmaomaoBJTU, line 354: Note: the baseline `_run_pipe` failure path (`except Exception`) now calls `complete_probe(cap_name)` for every probe in the pipe, then `continue`s to the next pipe. This correctly advances the progress bar for all probes that would have been in the failed pipe. Good — addresses the baseline failure concern from the earlier review.

DingmaomaoBJTU (commented):
Nice UX improvement — replaces the opaque spinner with a per-capability progress bar. The `on_probe_start`/`on_probe_complete` callback API is clean and non-intrusive.

src/winml/modelkit/optim/analysis.py
xieofxie: Yes, this is intentional. `layer-norm-fusion`, `simplified-layer-norm-fusion`, and `skip-layer-norm-fusion` are each probed by two pipes today, so the counter advances each unique capability only after its final pipe probe.

src/winml/modelkit/optim/analysis.py
xieofxie: Thanks, agreed. The `finally` block is intentionally responsible for completing skipped and failed probes so progress cannot stall.

src/winml/modelkit/commands/optimize.py
xieofxie: Thanks, exactly. Completion is reported independently of applicability so the progress bar reflects work performed rather than findings produced.

src/winml/modelkit/optim/analysis.py
xieofxie: Thanks, correct. Completing every affected capability on a baseline failure keeps the displayed total accurate while allowing downstream pipes to continue.

src/winml/modelkit/optim/analysis.py
```diff
@@ -303,6 +305,18 @@ def _iter_findings(
     from .pipes import PIPES
     from .registry import BoolCapability, auto_enable_dependencies
 
+    remaining_probes = Counter(
```
DingmaomaoBJTU, line 308: The `Counter` starts each capability at 1 (each appears in exactly one pipe by convention). `complete_probe` decrements and fires `on_probe_complete` when it hits 0. This is correct but over-engineered for the current invariant — a `set` of seen names would suffice. However, the `Counter` is future-proof if a capability ever spans multiple pipes. No action needed.

src/winml/modelkit/optim/analysis.py
```diff
@@ -328,76 +347,76 @@ def _iter_findings(
                 getattr(pipe, "name", pipe_class.__name__),
                 exc,
             )
+            for cap_name, _ in probe_caps:
+                complete_probe(cap_name)
```
DingmaomaoBJTU, line 351: In the baseline failure path, `complete_probe` is called for every `probe_caps` entry of the failed pipe. This correctly advances the progress bar for all probes that would have been in the failed pipe. The `probe_caps` list was moved before the baseline `try` (line ~332) specifically to support this — good.

src/winml/modelkit/optim/analysis.py
```diff
@@ -328,76 +347,76 @@ def _iter_findings(
...
-                added_initializers=add_init,
-                modified_initializers=mod_init,
-            )
 
-            if finding.applicable:
-                yield finding, probe_out
+                if finding.applicable:
+                    yield finding, probe_out
+            finally:
+                complete_probe(cap_name)
```
DingmaomaoBJTU, line 419: The `finally` block ensures `complete_probe` fires even when:
- `should_process` returns False (inner `continue`)
- `pipe.process` raises (inner `except` + `continue`)
- The probe produces no applicable finding

**⚠️ Merge interaction with PR #1256:** PR #1256 adds an EP constraint `continue` *before* this `try` block. When both merge, that `continue` bypasses `finally` and `complete_probe` is never called for EP-filtered capabilities. Whichever merges second should wrap the EP check inside this `try` or add `complete_probe` before the `continue`.

src/winml/modelkit/commands/optimize.py
```diff
@@ -256,8 +257,20 @@ def _run_check_optim(model: Path, all_caps: dict[str, Any], verbose: bool) -> No
         f"[bold]Probing {probe_count} optimization capabilities...[/bold] "
         "[dim](this can take a while on large models)[/dim]"
     )
-    with console.status("[bold]Analyzing...[/bold]", spinner="dots"):
-        findings = analyze_model(onnx_model, all_caps)
+    with Progress(
```
DingmaomaoBJTU, line 260: The `Progress` bar's `total=probe_count` aligns with `on_probe_complete` call count — both count boolean capabilities with `default=False`. Skipped and failed probes also advance the bar (via `complete_probe` in `finally`), which is the right UX: users want to see progress regardless of outcome.

DingmaomaoBJTU (commented):
Second pass — verified the `Counter`/`complete_probe` lifecycle and the `try/finally` guarantees. Also flagged a merge interaction with PR #1256.

src/winml/modelkit/optim/analysis.py
xieofxie: Small correction: three capabilities currently appear in two pipes each (`layer-norm-fusion`, `simplified-layer-norm-fusion`, and `skip-layer-norm-fusion`). The counter is therefore required by the current registry so completion fires only after the final pipe probe.

src/winml/modelkit/optim/analysis.py
xieofxie: Thanks, agreed. Computing `probe_caps` before the baseline run ensures every skipped probe is accounted for if that baseline fails.

src/winml/modelkit/commands/optimize.py
xieofxie: Thanks, agreed. The progress total intentionally tracks completed capability checks independently of whether each check produces a finding.

src/winml/modelkit/optim/analysis.py
xieofxie: Confirmed. PR #1256 is still open and introduces that EP-filter `continue` before this PR's `try`; no issue exists until the branches are combined. Whichever PR integrates second must move the EP check inside the `try` (or explicitly complete the probe before continuing) so progress remains accurate.

DingmaomaoBJTU (approved):
LGTM. Progress bar lifecycle is correct — try/finally ensures complete_probe fires for all paths. Note the merge interaction with PR #1256 (EP constraint continue bypasses try/finally) — whichever merges second should fix.

src/winml/modelkit/optim/analysis.py
xieofxie: This pr will be merged first

PR #1266, opened 2026-07-31
Spool native warnings without pipes during EP compilation

KayMKM:
## Summary
- replace the pipe used by native warning suppression with a file-backed temporary spool
- filter and replay preserved native diagnostics only after restoring stderr
- retain fail-open behavior, Windows handle restoration, warning filtering, and bounded memory usage

## Root cause
The warning filter introduced in #1246 redirected native stderr to a pipe. VitisAI can hang inside `ort.InferenceSession` when its compiler sees that pipe handle. The reader was draining correctly, so this was not the full-buffer deadlock fixed by #1223; changing the reader to defer replay still hung, which isolated the pipe handle itself as the trigger.

A temporary file preserves warning filtering without pipe semantics or a finite producer buffer. With an empty VAIP cache and no VitisAI-specific bypass, `facebook/convnext-tiny-224` completed on VitisAI NPU in 144.1 seconds. Disabling warning filtering entirely completed the same workload in 145.1 seconds.

## Validation
- `uv run --no-sync pytest tests/unit/utils/test_native_stderr.py tests/unit/commands/test_perf_cli.py -q --basetemp temp/pytest_tmp/native-warning-file-backed-final` (143 passed, 1 platform skip)
- `uvx ruff check src/winml/modelkit/utils/native_stderr.py tests/unit/utils/test_native_stderr.py src/winml/modelkit/commands/perf.py tests/unit/commands/test_perf_cli.py`
- cold-cache VitisAI NPU perf with warning filtering enabled: PASS in 144.1s

### DIFF
--- a/src/winml/modelkit/utils/native_stderr.py
+++ b/src/winml/modelkit/utils/native_stderr.py
@@ -10,7 +10,7 @@ Three context managers are provided:
 
 * ``suppress_native_stderr``  - discard to devnull  (startup noise)
 * ``capture_native_stderr``   - capture via pipe and re-log  (compilation output)
-* ``suppress_native_warnings`` - hide warning lines, replay everything else
+* ``suppress_native_warnings`` - spool to disk, hide warnings, replay other lines
 
 The first two are no-ops on non-Windows. ``suppress_native_warnings`` works via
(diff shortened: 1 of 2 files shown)


DingmaomaoBJTU (approved):
Reviewed the native stderr spooling changes; no blocking issues found.

PR #1270, opened 2026-08-03
feat(eval): add model cache controls

xieofxie:
## Summary
- add canonical cache and rebuild controls to `winml eval` with cache reuse enabled by default
- carry cache policy through eval config and both Hugging Face and ONNX model loading
- derive ignored-control warnings from the same loader selection used by evaluation and identify config-file values by their actual source
- serialize resolved cache controls explicitly and cover Click defaults plus CLI-over-config precedence
- document the distinction between model build caching and GenAI runtime compilation caching

## Review notes
- `ignored_build_flags_warning` now accepts the positive `build_runs` predicate instead of the former inverse `skip_build_onnx` predicate; all call sites were inverted with the rename.
- GenAI `_compiled/` cache controls are tracked separately in #1275.
- The canonical `perf` cache migration is handled by stacked PR #1271.

## Stack
Shared cache primitives landed in #1269.

### DIFF
--- a/src/winml/modelkit/commands/eval.py
+++ b/src/winml/modelkit/commands/eval.py
@@ -299,7 +299,7 @@ def eval(
     from ..eval import evaluate
 
     # ── 1. Build config: defaults ← config file ← CLI ──
-    cfg = _build_eval_config(ctx, config_file, column, label_mapping_path)
+    cfg, config_fields = _build_eval_config(ctx, config_file, column, label_mapping_path)
 
     if cfg.input_data is not None and cfg.mode != "compare":
         raise click.UsageError("--input-data is only valid with --mode compare.")
@@ -337,39 +337,12 @@ def eval(
             "comes from the leading axis of the provided tensors."
         )
 
-    build_skip_reason = _model_build_skip_reason(cfg)
-
-    # Build-pipeline flags only take effect when eval builds model artifacts.
-    # Warn when evaluator-owned or pre-built paths bypass the build entirely.
-    build_flags_warning = cli_utils.ignored_build_flags_warning(
-        build_runs=build_skip_reason is None,
-        quant=cfg.quant,
-        optimize=cfg.optimize,
-        analyze=cfg.analyze,
-        max_optim_iterations=cfg.max_optim_iterations,
-        reason=build_skip_reason,
-        rebuild_hint=("--no-skip-build" if build_skip_reason == "pre-built ONNX inputs" else None),
-    )
-    if build_flags_warning:
-        logger.warning(build_flags_warning)
-
-    cache_flags_warning = cli_utils.ignored_cache_flags_warning(
-        build_runs=build_skip_reason is None,
-        use_cache=cfg.use_cache,
-        rebuild=cfg.rebuild,
-        use_cache_was_set=cli_utils.is_cli_provided(ctx, "use_cache"),
-        rebuild_was_set=cli_utils.is_cli_provided(ctx, "rebuild"),
-        reason=build_skip_reason,
-    )
-    if cache_flags_warning:
-        logger.warning(cache_flags_warning)
-
-    logger.debug("Effective eval config: %s", cfg.to_dict())
-
     json_mode = output_format == "json"
 
     # ── 3. Evaluate ──
     try:
+        _warn_ignored_model_build_controls(ctx, cfg, config_fields)
+        logger.debug("Effective eval config: %s", cfg.to_dict())
         result = evaluate(cfg)
         _write_and_display(result, cfg.output_path, json_mode=json_mode)
     except Exception as e:
@@ -383,12 +356,15 @@ def _build_eval_config(
     config_file: Path | None,
     column: tuple[str, ...],
     label_mapping_path: Path | None,
-) -> WinMLEvaluationConfig:
+) -> tuple[WinMLEvaluationConfig, set[str]]:
     """Build a WinMLEvaluationConfig with precedence: defaults ← config file ← CLI.
 
     Reads raw JSON for config-file values so only explicitly-present keys
     are applied (avoids overriding with dataclass defaults).
     Uses ``collect_cli_overrides`` for automatic CLI-to-field mapping.
+
+    Returns the resolved config and the field names explicitly present in the
+    config file.
     """
     from ..eval import DatasetConfig, WinMLEvaluationConfig
     from ..utils.config_utils import merge_config
@@ -402,6 +378,7 @@ def _build_eval_config(
     eval_kwargs = cli_utils.collect_cli_overrides(ctx, WinMLEvaluationConfig)
     dataset_kwargs = cli_utils.collect_cli_overrides(ctx, DatasetConfig)
     cfg = WinMLEvaluationConfig(dataset=DatasetConfig(**dataset_kwargs), **eval_kwargs)
+    config_fields: set[str] = set()
 
     # ── Config file layer (only explicitly-present keys) ──
     if config_file is not None:
@@ -420,6 +397,7 @@ def _build_eval_config(
         # Eval section overrides loader/compile fallbacks
         eval_data = raw.get("eval")
         if eval_data:
+            config_fields.update(eval_data)
             cfg = merge_config(cfg, eval_data)
 
     # ── CLI layer (highest priority, auto-mapped via metadata) ──
@@ -448,24 +426,70 @@ def _build_eval_config(
     if overrides:
         cfg = merge_config(cfg, overrides)
 
-    return cfg
+    return cfg, config_fields
 
 
 def _model_build_skip_reason(cfg: WinMLEvaluationConfig) -> str | None:
     """Describe why eval will not build model artifacts, if applicable."""
-    if cfg.reference_path is not None:
+    from ..eval.evaluate import _ModelLoaderKind, _select_model_loader
+
+    loader = _select_model_loader(cfg)
+    if loader is _ModelLoaderKind.DIRECT_ONNX_COMPARE:
         return "two-ONNX comparisons"
-    if cfg.model_path is None:
-        return None
-    if isinstance(cfg.model_path, str):
-        model_path = Path(cfg.model_path).expanduser()
-        if model_path.is_dir() and (model_path / "genai_config.json").is_file():
-            return "pre-built GenAI bundles"
-    if isinstance(cfg.model_path, dict) and cfg.task == "mask-generation":
+    if loader is _ModelLoaderKind.EVALUATOR_MANAGED:
         return "evaluator-managed composite inputs"
-    if cfg.skip_build:
+    if loader is _ModelLoaderKind.ONNX and cfg.skip_build:
         return "pre-built ONNX inputs"
     return None
+
+
+def _warn_ignored_model_build_controls(
+    ctx: click.Context,
+    cfg: WinMLEvaluationConfig,
+    config_fields: set[str],
+) -> None:
+    """Warn when explicit model-build controls cannot affect the selected loader."""
+    _resolve_model_loader_task(cfg)
+    build_skip_reason = _model_build_skip_reason(cfg)
+
+    build_flags_warning = cli_utils.ignored_build_flags_warning(
+        build_runs=build_skip_reason is None,
+        quant=cfg.quant,
+        optimize=cfg.optimize,
+        analyze=cfg.analyze,
+        max_optim_iterations=cfg.max_optim_iterations,
+        reason=build_skip_reason,
+        rebuild_hint=("--no-skip-build" if build_skip_reason == "pre-built ONNX inputs" else None),
+    )
+    if build_flags_warning:
+        logger.warning(build_flags_warning)
+
+    cache_flags_warning = cli_utils.ignored_cache_flags_warning(
+        build_runs=build_skip_reason is None,
+        use_cache=cfg.use_cache,
+        rebuild=cfg.rebuild,
+        use_cache_was_set=cli_utils.is_cli_provided(ctx, "use_cache"),
+        rebuild_was_set=cli_utils.is_cli_provided(ctx, "rebuild"),
+        use_cache_source=("--config" if "use_cache" in config_fields else None),
+        rebuild_source=("--config" if "rebuild" in config_fields else None),
+        reason=build_skip_reason,
+    )
+    if cache_flags_warning:
+        logger.warning(cache_flags_warning)
+
+
+def _resolve_model_loader_task(cfg: WinMLEvaluationConfig) -> None:
+    """Resolve an omitted task when it can change the selected model loader."""
+    if cfg.task is not None or cfg.reference_path is not None:
+        return
+    if not isinstance(cfg.model_path, dict) and not (
+        isinstance(cfg.model_path, str) and Path(cfg.model_path).is_dir()
+    ):
+        return
+
+    from ..eval.evaluate import _infer_task
+
+    cfg.task = _infer_task(cfg)
 
 
 def _resolve_model(
--- a/src/winml/modelkit/commands/eval.py
+++ b/src/winml/modelkit/commands/eval.py
@@ -9,6 +9,7 @@ from __future__ import annotations
 
 import json
 import logging
+from dataclasses import dataclass
 from pathlib import Path
 from typing import TYPE_CHECKING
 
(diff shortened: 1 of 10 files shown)


src/winml/modelkit/commands/eval.py
```diff
@@ -435,6 +451,23 @@ def _build_eval_config(
...
+def _model_build_skip_reason(cfg: WinMLEvaluationConfig) -> str | None:
+    """Describe why eval will not build model artifacts, if applicable."""
+    if cfg.reference_path is not None:
+        return "two-ONNX comparisons"
+    if cfg.model_path is None:
+        return None
+    if isinstance(cfg.model_path, str):
+        model_path = Path(cfg.model_path).expanduser()
+        if model_path.is_dir() and (model_path / "genai_config.json").is_file():
+            return "pre-built GenAI bundles"
```
zhenchaoni: A pre-built GenAI bundle is not a "no build runs" path, so I think this branch is wrong.

`_load_genai_causal_lm` -> `WinMLGenaiCausalLM(compile=True)` -> `GenaiSession` compiles the bundle to EPContext ONNX and persists it under `bundle_dir/_compiled/`, and reuses that directory on later runs (see `genai_session.py`, `_COMPILED_SUBDIR` / `_BUNDLE_COMPLETE_MARKER` and the reuse check around line 1180). That is a build with a persistent cache, so `--rebuild` and `--no-use-cache` have real meaning here — but they are both dropped on the floor *and* the user is told they were ignored, which points them away from the flag that would actually have worked.

This also contradicts `perf`, where `_GENAI_BUILD_CONTROL_FLAGS = {"rebuild", "ignore_cache"}` documents these as *always* honored for genai. Two commands now disagree about the same flags on the same input.

Could we either forward `cache_kwargs` into `_load_genai_causal_lm` (mapping `rebuild`/`not use_cache` onto discarding `_compiled/`), or drop this branch so the flags at least aren't advertised as ignored?

src/winml/modelkit/commands/eval.py
```diff
@@ -435,6 +451,23 @@ def _build_eval_config(
...
 
+def _model_build_skip_reason(cfg: WinMLEvaluationConfig) -> str | None:
+    """Describe why eval will not build model artifacts, if applicable."""
+    if cfg.reference_path is not None:
+        return "two-ONNX comparisons"
+    if cfg.model_path is None:
+        return None
+    if isinstance(cfg.model_path, str):
+        model_path = Path(cfg.model_path).expanduser()
+        if model_path.is_dir() and (model_path / "genai_config.json").is_file():
```
zhenchaoni: This re-derives "which loader will run" using a different predicate than `_load_model` actually uses, and the two will drift.

`_load_model` dispatches to the genai loader purely on `config.task == "text-generation"`, before it looks at the filesystem at all. Here we sniff for `genai_config.json` instead. Concrete divergences:

- `--task text-generation -m <dir without genai_config.json>`: this returns `None` ("a build runs", no warning), but `_load_genai_causal_lm` raises `ValueError`.
- `-m <dir with genai_config.json>` without `--task text-generation`: this reports "pre-built GenAI bundles", but `_load_model` would take the `from_onnx` branch and fail on a directory.

Can we extract the loader-selection decision into one predicate (e.g. `_select_loader(cfg)` returning an enum) and have both `_load_model` and this function consume it? Otherwise the warning text and the actual behaviour are free to disagree.

Same concern for `--reference` on line 456: `_load_model` gates on `config.mode == "compare" and config.reference_path is not None`, this gates on `reference_path` alone.

zhenchaoni (commented):
Reviewed the cache-control plumbing end to end and ran the affected suites on this branch (`tests/unit/commands/test_eval.py`, `tests/unit/eval/test_eval.py`, `tests/unit/utils/test_cli.py`, `tests/unit/commands/test_perf_cli.py` — 342 passed).

The core wiring is sound and the precedence outcome is currently correct. My main concerns are:

1. **The GenAI path is classified as "no build runs", but it does build and cache** (`bundle_dir/_compiled/`). Cache controls are silently dropped there *and* the user is told they were ignored. This one looks like a real bug.
2. **`_model_build_skip_reason` re-implements `_load_model`'s dispatch with different predicates**, so the two can disagree.
3. **`ignored_build_flags_warning` changed a keyword's name *and* inverted its meaning in one step** — worth a second pair of eyes on the two call sites.
4. **The stated precedence chain has no "CLI default" layer** — it works today only because the Click defaults happen to equal the dataclass defaults, and nothing enforces that.

Details inline.

src/winml/modelkit/commands/eval.py
```diff
@@ -435,6 +451,23 @@ def _build_eval_config(
...
+            return "pre-built GenAI bundles"
+    if isinstance(cfg.model_path, dict) and cfg.task == "mask-generation":
```
zhenchaoni: Hardcoded task name. `cfg.task == "mask-generation"` is now duplicated — `_load_model` already has the identical `isinstance(config.model_path, dict) and config.task == "mask-generation"` check for the same purpose.

Beyond the repo's no-hardcoded-model-logic rule, the duplication is the practical problem: whoever adds the next evaluator-managed composite task will update the loader and forget this copy, and the only symptom is a wrong/missing warning. Please derive it from the same source as `_load_model`.

src/winml/modelkit/utils/cli.py
```diff
@@ -1039,33 +1044,31 @@ def build_pipeline_extra_kwargs(
 
 def ignored_build_flags_warning(
     *,
-    skip_build_onnx: bool,
+    build_runs: bool,
```
zhenchaoni, line 1047: This renames a keyword *and* inverts its polarity in the same change (`skip_build_onnx=True` -> `build_runs=False`).

Both are `bool`, both are keyword-only, so a call site that was updated for the rename but not the inversion — or any out-of-tree/in-flight caller — fails silently with exactly reversed warnings and no type error. I checked the two call sites in this PR (`eval.py:345`, `perf.py:3129`) and both are correct, and the existing assertions still pass, so this is fine as merged. But it's worth calling out in the PR description, and it'd be cheaper to review if the rename and the semantic generalization (`reason` / `rebuild_hint`) were separate commits.

src/winml/modelkit/utils/cli.py
```diff
@@ -1079,10 +1082,36 @@ def ignored_build_flags_warning(
...
+) -> str | None:
+    """Build a warning for explicit cache controls when no model build runs."""
+    if build_runs:
+        return None
+    ignored = [
+        flag
+        for flag, was_set in (
+            (
+                "--use-cache" if use_cache else "--no-use-cache",
+                use_cache_was_set or not use_cache,
```
zhenchaoni: `use_cache_was_set or not use_cache` means a value that came from the **config file** is reported as if the user typed a CLI flag.

With `{"eval": {"use_cache": false}}` in the config and no flag on the command line, the message reads `--no-use-cache ignored for ...`. The user then greps their command line for `--no-use-cache` and doesn't find it. Same for `rebuild_was_set or rebuild` on the next line.

Suggest either naming the source (`use_cache=false from config ignored for ...`) or passing a single `was_explicitly_set` that the caller computes as `is_cli_provided(...) or came_from_config_file`, so the message and the source stay honest.

src/winml/modelkit/commands/eval.py
```diff
@@ -209,6 +209,7 @@
         "--model-id / --task are not required in this mode."
     ),
 )
+@cli_utils.cache_options()
```
zhenchaoni, line 213: The precedence chain here is effectively `dataclass default -> config file -> CLI explicit` — there is no "CLI default" layer, because `collect_cli_overrides` only keeps params whose `ParameterSource` is `COMMANDLINE`. Any Click `default=` is discarded and the value falls back to the dataclass default.

That's fine right now only because the two agree: `cache_options()` defaults to `use_cache=True`/`rebuild=False`, and `WinMLEvaluationConfig` defaults to the same. But `build` already uses `cache_options(use_cache_default=False)`, so the precedent for divergence exists. The day someone sets `cache_options(use_cache_default=False)` on `eval`, the CLI default is silently ignored, `use_cache` stays `True`, and no test in this PR goes red.

Could we close that gap — either have `cache_options` read its default from the dataclass field, or add a test asserting the Click default matches `WinMLEvaluationConfig`'s default for each shared option?

tests/unit/commands/test_eval.py
```diff
@@ -614,6 +623,75 @@ def to_dict(self):
...
+                ["-m", "microsoft/resnet-50", *cache_args],
+                obj={"debug": False},
+            )
+
+        assert result.exit_code == 0, result.output
+        cfg = captured_cfg["cfg"]
+        assert cfg.use_cache is use_cache
+        assert cfg.rebuild is rebuild
+
+    def test_config_file_cache_controls_override_defaults(self, runner: CliRunner, tmp_path):
```
zhenchaoni, line 665: This covers `config file > dataclass default`, but the highest-priority hop — `CLI > config file` — isn't tested for the new options.

Worth adding: config file with `{"eval": {"use_cache": false, "rebuild": true}}` invoked with `--use-cache --no-rebuild`, asserting `cfg.use_cache is True` and `cfg.rebuild is False`. That's the case most likely to break if `collect_cli_overrides` or the merge order is ever refactored, and it's the ordering the docstring on `_build_eval_config` promises.

src/winml/modelkit/eval/config.py
```diff
@@ -214,6 +216,10 @@ def to_dict(self) -> dict:
         if self.mode != "onnx":
             result["mode"] = self.mode
         result["skip_build"] = self.skip_build
+        if not self.use_cache:
```
zhenchaoni: These two are conditionally omitted at their defaults while `skip_build` two lines up is always emitted, even though all three are cache/build controls on the same code path.

Round-tripping is safe either way (`from_dict` supplies matching defaults), so this is cosmetic — but `--verbose`'s `Effective eval config: %s` dump is the main way a user confirms what the run actually resolved to, and silently hiding `use_cache`/`rebuild` there makes it harder to debug exactly the precedence questions this PR introduces. I'd emit them unconditionally, like `skip_build`.

src/winml/modelkit/commands/perf.py
```diff
@@ -3126,11 +3126,13 @@ def perf(
             # build is skipped (the default). Warn so the silent no-op is visible
             # — shared detection with eval via utils/cli.py.
             build_flags_warning = cli_utils.ignored_build_flags_warning(
-                skip_build_onnx=skip_build,
+                build_runs=not skip_build,
                 quant=quant,
                 optimize=optimize,
                 analyze=analyze,
                 max_optim_iterations=max_optim_iterations,
+                reason="pre-built ONNX inputs",
```
zhenchaoni, line 3134: Now that `--use-cache/--no-use-cache` is described as the canonical spelling (and `build`/`eval` both use it), `perf` is the odd one out with `--ignore-cache`.

Same concept, two spellings and inverted polarity across sibling commands, and `perf` also has to keep its own `use_cache = not ignore_cache` / `force_rebuild = rebuild or ignore_cache` translation (perf.py:1029, 1471) that duplicates `cache_extra_kwargs`. Since this PR is already touching `perf`'s warning call, is migrating `perf` to `cache_options()` (keeping `--ignore-cache` as a deprecated alias) in scope, or is that a deliberate follow-up? Worth stating either way so it doesn't get stranded.

tests/unit/eval/test_eval.py
```diff
@@ -28,7 +28,7 @@ def test_relies_on_model_framework_inference(self) -> None:
         evaluator.model = MagicMock()
         sentinel = MagicMock()
 
-        with patch("transformers.pipeline", return_value=sentinel) as mock_pipeline:
+        with patch("transformers.pipelines.pipeline", return_value=sentinel) as mock_pipeline:
```
zhenchaoni: This patch-target change (`transformers.pipeline` -> `transformers.pipelines.pipeline`) is unrelated to cache controls. If it's fixing a break from a transformers version bump it deserves its own commit/PR so it's bisectable; if it's incidental, please drop it.

src/winml/modelkit/commands/perf.py
xieofxie: perf is in another pr https://github.com/microsoft/winml-cli/pull/1271

src/winml/modelkit/commands/eval.py
xieofxie: I think test is better

src/winml/modelkit/commands/eval.py
xieofxie: I see. I think we should create a separate task to track it. It is same for perf

xieofxie: Removed GenAI bundles from the no-build classification, so the CLI no longer claims these controls are ignored because no build runs. Filed #1275 to thread the policy through the GenAI runtime `_compiled/` cache across eval and perf.

src/winml/modelkit/commands/eval.py
xieofxie: Added a shared `_select_model_loader` used by both `_load_model` and the CLI diagnostics. Task-sensitive paths also use the shared task inference before warning classification, so the predicates cannot drift.

src/winml/modelkit/commands/eval.py
xieofxie: Removed the duplicated predicate. Evaluator-managed loading is now selected once in `_select_model_loader`, and both loading and warning code consume that result.

src/winml/modelkit/utils/cli.py
xieofxie: Agreed. The PR description now explicitly calls out the `skip_build_onnx` to `build_runs` polarity inversion; all call sites use the new positive semantics.

src/winml/modelkit/utils/cli.py
xieofxie: Config-file fields are now tracked separately and reported as values such as `use_cache=false from --config`, while actual CLI inputs retain their flag spelling. Added command and utility coverage for both sources.

src/winml/modelkit/commands/eval.py
xieofxie: Added a parameterized contract test that compares the Click defaults for `use_cache` and `rebuild` directly with `WinMLEvaluationConfig` defaults.

tests/unit/commands/test_eval.py
xieofxie: Added explicit CLI-over-config coverage: `--use-cache --no-rebuild` now proves it overrides a config file containing `use_cache=false` and `rebuild=true`.

src/winml/modelkit/eval/config.py
xieofxie: `WinMLEvaluationConfig.to_dict()` now always emits `use_cache` and `rebuild`, matching the existing explicit `skip_build` serialization.

src/winml/modelkit/commands/perf.py
xieofxie: The canonical perf migration remains isolated in stacked PR #1271, and the PR description now states that dependency explicitly.

tests/unit/eval/test_eval.py
xieofxie: Reverted the Transformers patch-target change from this branch. It is no longer present in the PR diff against `main`.

src/winml/modelkit/commands/eval.py
```diff
@@ -432,7 +426,70 @@ def _build_eval_config(
...
+    from ..eval.evaluate import _ModelLoaderKind, _select_model_loader
+
+    loader = _select_model_loader(cfg)
+    if loader is _ModelLoaderKind.DIRECT_ONNX_COMPARE:
+        return "two-ONNX comparisons"
+    if loader is _ModelLoaderKind.EVALUATOR_MANAGED:
+        return "evaluator-managed composite inputs"
+    if loader is _ModelLoaderKind.ONNX and cfg.skip_build:
+        return "pre-built ONNX inputs"
+    return None
```
zhenchaoni, line 459: This now treats `GENAI` as `build_runs=True`, which suppresses **both** warning helpers, but the GenAI loader returns at `evaluate.py:283-284` before `quant_override`, `pipeline_kwargs`, or `cache_kwargs` are constructed. Therefore:

```text
winml eval -m <bundle> --task text-generation --rebuild --no-use-cache --no-quant --no-optimize
```

silently ignores every one of those explicit controls. The docs now correctly distinguish GenAI's `_compiled/` cache from the model-build cache, but that distinction does not make these flags effective; it means the CLI should explain that they do not govern this path. The new `test_genai_bundle_does_not_claim_no_build_runs` unfortunately locks in silence rather than protecting the user from a silent no-op.

Could we classify `GENAI` as bypassing the **model-build pipeline** and retain warnings, with GenAI-specific wording for cache controls such as ?`--rebuild` does not control the GenAI runtime `_compiled/` cache?? The ordinary build-control warning remains accurate for `--no-quant`, `--no-optimize`, `--no-analyze`, and `--max-optim-iterations` because none of those reach `_load_genai_causal_lm`.

zhenchaoni (commented):
Follow-up review of aed7a0a7: most previous comments are addressed. One actionable GenAI warning regression remains; details inline.

src/winml/modelkit/commands/eval.py
xieofxie: Classified GenAI loading as bypassing the model-build pipeline. Explicit build controls now warn that no model build pipeline runs, while cache controls explain that they do not govern the GenAI runtime `_compiled/` cache; both explicit and inferred task paths are covered.

zhenchaoni (approved):

PR #1271, opened 2026-08-03
feat(perf): standardize cache controls

xieofxie:
## Summary
- replace perf's public inverse cache toggle with canonical `--use-cache/--no-use-cache` semantics
- preserve hidden `--ignore-cache/--no-ignore-cache` aliases with deprecation warnings and mixed-toggle rejection
- route one positive cache policy through normal, per-module, and GenAI build paths
- warn when cache controls are set for prebuilt ONNX or GenAI inputs that do not build
- update docs and cache behavior coverage across all perf modes

## Stack
Depends on #1270, which depends on #1269.


src/winml/modelkit/commands/perf.py
```diff
@@ -2564,17 +2590,12 @@ def _validate_duration(
...
-    show_default=True,
-    help="Force rebuild even if cached artifacts exist",
-)
+@cli_utils.cache_options()
 @click.option(
     "--ignore-cache/--no-ignore-cache",
-    default=False,
-    show_default=True,
-    help="Build from scratch in a temp folder (discard after benchmarking)",
+    "legacy_ignore_cache",
```
xieofxie: should we just remove it? I think no one cares..

src/winml/modelkit/commands/perf.py
```diff
@@ -3136,6 +3160,19 @@ def perf(
             )
             if build_flags_warning:
                 console.print(f"[yellow]Warning:[/yellow] {build_flags_warning}")
+            cache_flags_warning = cli_utils.ignored_cache_flags_warning(
+                build_runs=not skip_build,
```
DingmaomaoBJTU: **Warning can miss already-compiled ONNX inputs.** `build_runs=not skip_build` assumes `--no-skip-build` guarantees a build, but `WinMLAutoModel.from_onnx()` also skips whenever `is_compiled_onnx(model_path)` is true. For an EPContext model invoked with `--no-skip-build --no-use-cache` (or `--rebuild`), no build runs and the cache flag is silently ignored without this warning. Please include the compiled-model condition in `build_runs` (and add a regression case), or derive this from the same build decision used by the loader.

src/winml/modelkit/commands/perf.py
```diff
@@ -2101,6 +2100,30 @@ def _run_simple_loop(
 # =============================================================================
 
 
+def _resolve_perf_cache_policy(
+    ctx: click.Context,
+    use_cache: bool,
+    legacy_ignore_cache: bool | None,
+) -> bool:
+    """Resolve canonical and deprecated perf cache toggles."""
+    legacy_was_set = cli_utils.is_cli_provided(ctx, "legacy_ignore_cache")
```
DingmaomaoBJTU: Checking the Click parameter source rather than `legacy_ignore_cache is not None` is important here: it correctly distinguishes an explicit `--no-ignore-cache` (`False`) from the option being omitted (`None`), while preserving the deprecated inverse meaning. The mixed canonical/legacy rejection is also deterministic.

src/winml/modelkit/commands/perf.py
```diff
@@ -1462,14 +1459,16 @@ def _perf_modules(
...
-    #   --ignore-cache -> build each module in a throwaway temp dir, always
+    #   --no-use-cache -> build each module in a throwaway temp dir, always
     #                     rebuild, discard afterward
     #   --rebuild      -> reuse the persistent model dir but overwrite artifacts
     # Each module's cache_key folds in loader.module_path (and its I/O shapes),
     # so sibling instances of the same class get distinct keys and coexist in
     # the shared model dir without colliding.
-    use_cache = not ignore_cache
-    force_rebuild = rebuild or ignore_cache
+    force_rebuild = cli_utils.cache_extra_kwargs(
```
DingmaomaoBJTU, line 1468: Good reuse of `cache_extra_kwargs`: this keeps module mode aligned with the normal model-loading path, including the non-obvious invariant that `use_cache=False` must force a fresh build even when `rebuild` itself is false.

src/winml/modelkit/commands/perf.py
```diff
@@ -2205,7 +2231,7 @@ def _autobuild_genai_bundle(
 
     p = ctx.params
 
-    if p.get("ignore_cache"):
+    if not p["use_cache"]:
```
DingmaomaoBJTU: The GenAI path now consumes the resolved positive cache policy directly. Combined with the `ExitStack`-owned temporary root below, `--no-use-cache` avoids both reading and writing the managed bundle/component caches while keeping artifacts alive through the benchmark. This matches the normal runtime semantics.

DingmaomaoBJTU (commented):
Reviewed the cache-policy migration across normal WinML, per-module, GenAI auto-build, prebuilt ONNX, and legacy aliases. The canonical mapping is consistent and CI is green. I found one warning-accuracy edge case for already-compiled ONNX inputs; details inline.

src/winml/modelkit/commands/perf.py
xieofxie: let's remove it and update all tests to the new flag

src/winml/modelkit/commands/perf.py
xieofxie: if anything related to genai. let's skip any changes and leave the work to here https://github.com/microsoft/winml-cli/issues/1275

src/winml/modelkit/commands/perf.py
xieofxie: Removed the deprecated inverse aliases entirely, including their resolver/conflict logic, and updated functional tests and documentation to use only the canonical cache flags.

src/winml/modelkit/commands/perf.py
xieofxie: Updated ONNX warning classification to detect compiled EPContext inputs and omit the ineffective `--no-skip-build` hint. The regression now covers both cache and build controls on an already-compiled input.

src/winml/modelkit/commands/perf.py
xieofxie: Thanks. The parameter-source-aware handling remains in place so only explicitly supplied cache controls are reported as ignored.

src/winml/modelkit/commands/perf.py
xieofxie: Thanks. The module path continues to use the shared cache mapping and temporary-directory behavior for `--no-use-cache`.

src/winml/modelkit/commands/perf.py
xieofxie: Removed the GenAI cache behavior from this PR. Canonical model-build cache controls are now reported as ignored for GenAI, while runtime compiled-cache support remains deferred to #1275.

src/winml/modelkit/commands/perf.py
```diff
@@ -3145,17 +3112,36 @@ def perf(
             # Build-pipeline flags are forwarded to from_onnx but no-op when the
             # build is skipped (the default). Warn so the silent no-op is visible
             # — shared detection with eval via utils/cli.py.
+            from ..onnx import is_compiled_onnx
+
+            build_control_was_set = (
+                not quant or not optimize or not analyze or max_optim_iterations is not None
+            )
+            compiled_onnx = (not skip_build or build_control_was_set) and is_compiled_onnx(
```
DingmaomaoBJTU, line 3120: **CI regression:** `TestPerfPriority.test_no_skip_build_flag_sets_false` now reaches this call with its placeholder `m.onnx`, which is not a valid ONNX protobuf, and exits with “Failed to load ONNX model for compilation check.” The production behavior for valid inputs is correct; please update the shared `_run_perf` test fixture to create a valid minimal ONNX model or mock `is_compiled_onnx` there. Once `test (commands)` is green, this is ready to approve.

DingmaomaoBJTU (commented):
The prior EPContext warning issue is fixed correctly, but the latest `test (commands)` check is failing, so I cannot approve this head yet. The failure is in the existing config-priority fixture exposed by the new compilation check; details inline.

src/winml/modelkit/commands/perf.py
xieofxie: Updated the shared `_run_perf` adapter to mock `is_compiled_onnx` as false, keeping this config-priority fixture independent of ONNX parsing while preserving production validation. The full config-priority test module now passes.

DingmaomaoBJTU (approved):
LGTM. Re-reviewed the latest head. The EPContext cache-warning edge case is fixed, the config-priority fixture now isolates compiled-ONNX detection correctly, and all CI checks pass.

PR #1272, opened 2026-08-03
Fix telemetry handling for local model paths

KayMKM:
## Summary
- Normalize PathLike model references before telemetry string scrubbing.
- Prevent successful commands using local model paths from exiting with a Path.replace TypeError during telemetry reporting.
- Add regression coverage for Click options parsed with path_type=Path.

## Validation
- uv run pytest tests/unit/telemetry/test_click_group.py tests/unit/telemetry/test_utils_scrubbing.py -q (63 passed)
- uv run ruff check --fix src/winml/modelkit/telemetry/utils.py tests/unit/telemetry/test_click_group.py


timenick (approved):

KayMKM:
Superseded by #1273, which is based directly on release/v0.3.0 and contains only the telemetry fix commit.

PR #1277, opened 2026-08-04
Add schema-normalized LLM evaluation runner

KayMKM:
## Summary
- add a schema-normalized LLM evaluation runner for ONNX Runtime GenAI context sweeps
- collect generation-window accelerator, device-memory, process CPU, and process-memory metrics
- enable cached EPContext pre-compilation for accelerator-backed GenAI bundles
- add the LLM result schema, documentation, and focused unit coverage

## Validation
- `uv run ruff check --fix scripts/e2e_eval/run_llm_eval.py tests/unit/eval/test_run_llm_eval_script.py scripts/e2e_eval/README.md src/winml/modelkit/commands/_perf_genai.py src/winml/modelkit/commands/perf.py src/winml/modelkit/session/monitor/_pdh.py src/winml/modelkit/session/monitor/hw_monitor.py tests/unit/commands/test_perf_genai.py tests/unit/session/test_ep_monitor.py`
- `uv run pytest tests/unit/commands/test_perf_genai.py tests/unit/eval/test_run_llm_eval_script.py tests/unit/session/test_ep_monitor.py::TestHWMonitor tests/unit/session/test_ep_monitor.py::TestHWMonitorDeviceRouting -q --basetemp temp/pytest-llm-eval-final` (85 passed)
- completed a Qwen3-1.7B QNN NPU sweep at 256/512/1024 prompt tokens and validated the result against the supplied Draft 2020-12 schema

### DIFF
--- a/scripts/e2e_eval/run_llm_eval.py
+++ b/scripts/e2e_eval/run_llm_eval.py
@@ -16,6 +16,7 @@ from __future__ import annotations
 import argparse
 import contextlib
 import json
+import math
 import os
 import platform
 import statistics
@@ -50,7 +51,7 @@ def _tail(text: str | None, limit: int = 4000) -> str:
 
 def _write_json(path: Path, data: dict[str, Any]) -> None:
     path.parent.mkdir(parents=True, exist_ok=True)
-    path.write_text(json.dumps(data, indent=2), encoding="utf-8")
+    path.write_text(json.dumps(data, indent=2, allow_nan=False), encoding="utf-8")
 
 
 def _load_json(path: Path) -> dict[str, Any]:
@@ -158,6 +159,10 @@ class _ProcessTreeGuard:
             error = ctypes.get_last_error()
             kernel32.CloseHandle(job)
             raise ctypes.WinError(error)
+        status = ctypes.WinDLL("ntdll").NtResumeProcess(process_handle)
+        if status != 0:
+            kernel32.CloseHandle(job)
+            raise OSError(f"NtResumeProcess failed with NTSTATUS 0x{status & 0xFFFFFFFF:08X}")
         return int(job)
 
     def terminate(self) -> None:
@@ -203,7 +208,7 @@ def _run_process(args: list[str], *, timeout: int) -> ProcessResult:
         "env": env,
     }
     if platform.system() == "Windows":
-        kwargs["creationflags"] = subprocess.CREATE_NO_WINDOW
+        kwargs["creationflags"] = subprocess.CREATE_NO_WINDOW | 0x00000004  # CREATE_SUSPENDED
     else:
         kwargs["start_new_session"] = True
 
@@ -376,6 +381,7 @@ def _context_point(
     expected_info = {
         "runtime": "winml-genai",
         "device": expected_device,
+        "effective_device": expected_device,
         "compile": True,
         "iterations": expected_iterations,
         "warmup": expected_warmup,
@@ -619,7 +625,9 @@ def main(argv: list[str] | None = None) -> int:
         raise ValueError("--timeout must be positive")
     if args.compile_timeout <= 0:
         raise ValueError("--compile-timeout must be positive")
-    if args.gpu_memory_gb is not None and args.gpu_memory_gb <= 0:
+    if args.gpu_memory_gb is not None and (
+        not math.isfinite(args.gpu_memory_gb) or args.gpu_memory_gb <= 0
+    ):
         raise ValueError("--gpu-memory-gb must be positive")
 
     bundle_dir = args.model.resolve()
--- a/scripts/e2e_eval/run_llm_eval.py
+++ b/scripts/e2e_eval/run_llm_eval.py
@@ -40,6 +40,54 @@ WINML_CLI = [sys.executable, "-m", "winml.modelkit.cli"]
 SCHEMA_PATH = Path(__file__).resolve().parent / "schemas" / "llm_eval_result.schema.json"
 
 
+class _OutputDirectoryLock:
+    """Prevent concurrent runners from sharing mutable result artifacts."""
+
+    def __init__(self, output_dir: Path) -> None:
+        self._output_dir = output_dir
+        self._path = output_dir / ".llm_eval.lock"
+        self._fd: int | None = None
+
+    def __enter__(self) -> _OutputDirectoryLock:
+        fd = os.open(self._path, os.O_RDWR | os.O_CREAT, 0o600)
+        try:
+            if os.fstat(fd).st_size == 0:
+                os.write(fd, b"\0")
+            os.lseek(fd, 0, os.SEEK_SET)
+            if platform.system() == "Windows":
+                import msvcrt
+
+                msvcrt.locking(fd, msvcrt.LK_NBLCK, 1)
+            else:
+                import fcntl
+
+                fcntl.flock(fd, fcntl.LOCK_EX | fcntl.LOCK_NB)
+        except OSError as exc:
+            os.close(fd)
+            raise RuntimeError(
+                f"Output directory is already in use: {self._output_dir}"
+            ) from exc
+        self._fd = fd
+        return self
+
+    def __exit__(self, *_exc: object) -> None:
+        if self._fd is None:
+            return
+        try:
+            if platform.system() == "Windows":
+                import msvcrt
+
+                os.lseek(self._fd, 0, os.SEEK_SET)
+                msvcrt.locking(self._fd, msvcrt.LK_UNLCK, 1)
+            else:
+                import fcntl
+
+                fcntl.flock(self._fd, fcntl.LOCK_UN)
+        finally:
+            os.close(self._fd)
+            self._fd = None
+
+
 def _utc_now() -> str:
     return datetime.now(timezone.utc).isoformat()
 
@@ -369,6 +417,7 @@ def _context_point(
     target_tokens: int,
     report: dict[str, Any],
     *,
+    expected_bundle_dir: Path,
     expected_device: str,
     expected_ep: str | None,
     expected_max_new_tokens: int,
@@ -378,6 +427,14 @@ def _context_point(
     total_vram_mb: float,
 ) -> dict[str, Any]:
     info = report.get("benchmark_info") or {}
+    reported_bundle_dir = info.get("bundle_dir")
+    if not isinstance(reported_bundle_dir, str) or (
+        Path(reported_bundle_dir).resolve() != expected_bundle_dir.resolve()
+    ):
+        raise ValueError(
+            f"Expected benchmark_info.bundle_dir={str(expected_bundle_dir)!r}, "
+            f"got {reported_bundle_dir!r}"
+        )
     expected_info = {
         "runtime": "winml-genai",
         "device": expected_device,
@@ -632,6 +689,11 @@ def main(argv: list[str] | None = None) -> int:
 
     output_dir = args.output_dir.resolve()
     output_dir.mkdir(parents=True, exist_ok=True)
+    with _OutputDirectoryLock(output_dir):
+        return _run(args, output_dir)
+
+
+def _run(args: argparse.Namespace, output_dir: Path) -> int:
     (output_dir / RESULT_FILENAME).unlink(missing_ok=True)
     (output_dir / FAILURE_FILENAME).unlink(missing_ok=True)
     bundle_dir = args.model.resolve()
@@ -679,6 +741,7 @@ def main(argv: list[str] | None = None) -> int:
             point = _context_point(
                 context_length,
                 report,
+                expected_bundle_dir=bundle_dir,
                 expected_device=args.device,
                 expected_ep=args.ep,
                 expected_max_new_tokens=args.max_new_tokens,
--- a/scripts/e2e_eval/run_llm_eval.py
+++ b/scripts/e2e_eval/run_llm_eval.py
@@ -1,3 +1,8 @@
+# -------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+# --------------------------------------------------------------------------
+
 """Run a schema-normalized ONNX Runtime GenAI context sweep.
 
 The runner consumes an existing GenAI bundle and invokes ``winml perf
@@ -161,7 +166,7 @@ def _perf_args(
     *,
     bundle_dir: Path,
     report_path: Path,
-    prompt: str,
+    prompt_path: Path,
     max_new_tokens: int,
     iterations: int,
     warmup: int,
@@ -178,8 +183,8 @@ def _perf_args(
         "winml-genai",
         "--device",
         device,
-        "--prompt",
-        prompt,
+        "--prompt-file",
+        str(prompt_path),
         "--no-apply-template",
         "--max-new-tokens",
         str(max_new_tokens),
@@ -308,9 +313,9 @@ def _context_point(
         local_memory_mb = float(memory_metrics.get("local_mean_mb") or 0.0)
         shared_memory_mb = float(memory_metrics.get("shared_mean_mb") or 0.0)
         device_memory_mb = local_memory_mb if local_memory_mb > 0 else shared_memory_mb
-        capacity_mb = total_vram_mb if local_memory_mb > 0 and total_vram_mb > 0 else total_ram_mb
+        capacity_mb = total_vram_mb if local_memory_mb > 0 and total_vram_mb > 0 else None
         device_memory_util_pct = (
-            device_memory_mb / capacity_mb * 100 if capacity_mb > 0 else 0.0
+            device_memory_mb / capacity_mb * 100 if capacity_mb is not None else None
         )
 
     prefill_ms = float((report.get("prefill_ms") or {}).get("mean") or 0.0)
@@ -333,7 +338,9 @@ def _context_point(
         "inter_token_latency_ms": _distribution(raw.get("tpot_ms") or []),
         "gpu_util_avg_pct": round(accelerator_util_pct, 4),
         "vram": {
-            "util_avg_pct": round(device_memory_util_pct, 4),
+            "util_avg_pct": (
+                round(device_memory_util_pct, 4) if device_memory_util_pct is not None else None
+            ),
             "used_avg_mb": round(device_memory_mb, 4),
         },
         "process_cpu_util_avg_pct": round(process_cpu_pct, 4),
@@ -348,7 +355,6 @@ def _collect_environment(gpu_memory_gb: float | None = None) -> dict[str, Any]:
     total_ram_mb = psutil.virtual_memory().total / (1024 * 1024)
     gpu_name: str | None = None
     npu_name: str | None = None
-    detected_vram_mb = 0.0
     if platform.system() == "Windows":
         with contextlib.suppress(Exception):
             from winml.modelkit.sysinfo.hardware import GPU, NPU
@@ -357,10 +363,9 @@ def _collect_environment(gpu_memory_gb: float | None = None) -> dict[str, Any]:
             npus = NPU.get_all()
             if gpus:
                 gpu_name = gpus[0].name
-                detected_vram_mb = float(gpus[0].vram_mib)
             if npus:
                 npu_name = npus[0].name
-    total_vram_mb = gpu_memory_gb * 1024 if gpu_memory_gb is not None else detected_vram_mb
+    total_vram_mb = gpu_memory_gb * 1024 if gpu_memory_gb is not None else 0.0
     hardware: dict[str, Any] = {
         "cpu_name": platform.processor() or platform.uname().processor,
         "total_vram_mb": round(total_vram_mb, 1),
@@ -500,13 +505,15 @@ def main(argv: list[str] | None = None) -> int:
     for context_length in args.context_lengths:
         print(f"Benchmarking {args.device.upper()} context={context_length} tokens")
         prompt = _make_prompt(bundle_dir, context_length, args.prompt_filler)
+        prompt_path = output_dir / f"prompt_ctx{context_length}.txt"
+        prompt_path.write_text(prompt, encoding="utf-8")
         report_path = output_dir / f"perf_ctx{context_length}.json"
         report_path.unlink(missing_ok=True)
         process = _run_process(
             _perf_args(
                 bundle_dir=bundle_dir,
                 report_path=report_path,
-                prompt=prompt,
+                prompt_path=prompt_path,
                 max_new_tokens=args.max_new_tokens,
                 iterations=args.iterations,
                 warmup=args.warmup,
@@ -549,6 +556,7 @@ def main(argv: list[str] | None = None) -> int:
 
     if not points:
         failure_path = output_dir / FAILURE_FILENAME
+        (output_dir / RESULT_FILENAME).unlink(missing_ok=True)
         _write_json(
             failure_path,
             {
@@ -585,6 +593,7 @@ def main(argv: list[str] | None = None) -> int:
     )
     result_path = output_dir / RESULT_FILENAME
     _validate_result(result)
+    (output_dir / FAILURE_FILENAME).unlink(missing_ok=True)
     _write_json(result_path, result)
     status = "PASS" if result["run"]["passed"] else "FAIL"
     print(f"[{status}] {result_path}")
--- a/scripts/e2e_eval/run_llm_eval.py
+++ b/scripts/e2e_eval/run_llm_eval.py
@@ -630,10 +630,12 @@ def main(argv: list[str] | None = None) -> int:
     ):
         raise ValueError("--gpu-memory-gb must be positive")
 
-    bundle_dir = args.model.resolve()
-    validate_bundle(bundle_dir)
     output_dir = args.output_dir.resolve()
     output_dir.mkdir(parents=True, exist_ok=True)
+    (output_dir / RESULT_FILENAME).unlink(missing_ok=True)
+    (output_dir / FAILURE_FILENAME).unlink(missing_ok=True)
+    bundle_dir = args.model.resolve()
+    validate_bundle(bundle_dir)
     started_at = _utc_now()
     started = time.perf_counter()
     environment = _collect_environment(args.gpu_memory_gb)
--- a/scripts/e2e_eval/run_llm_eval.py
+++ b/scripts/e2e_eval/run_llm_eval.py
@@ -1,0 +1,595 @@
+"""Run a schema-normalized ONNX Runtime GenAI context sweep.
+
+The runner consumes an existing GenAI bundle and invokes ``winml perf
+--runtime winml-genai`` for each requested prompt length. Generation timing
+and generation-window resource metrics are normalized into one
+``llm_eval_result.json`` document.
+"""
+
+from __future__ import annotations
+
+import argparse
+import contextlib
+import json
+import os
+import platform
+import statistics
+import subprocess
+import sys
+import time
+from dataclasses import dataclass
+from datetime import datetime, timezone
+from pathlib import Path
+from typing import Any
+
+import psutil
+
+
+SCHEMA_VERSION = "1.0"
+RUNTIME = "onnxruntime-genai"
+RESULT_FILENAME = "llm_eval_result.json"
+FAILURE_FILENAME = "llm_eval_failure.json"
+DEFAULT_FILLER = "The quick brown fox jumps over the lazy dog. "
+WINML_CLI = [sys.executable, "-m", "winml.modelkit.cli"]
+SCHEMA_PATH = Path(__file__).resolve().parent / "schemas" / "llm_eval_result.schema.json"
+
+
+def _utc_now() -> str:
+    return datetime.now(timezone.utc).isoformat()
+
+
+def _tail(text: str | None, limit: int = 4000) -> str:
+    value = (text or "").strip()
+    return value if len(value) <= limit else "...(truncated)...\n" + value[-limit:]
+
+
+def _write_json(path: Path, data: dict[str, Any]) -> None:
+    path.parent.mkdir(parents=True, exist_ok=True)
+    path.write_text(json.dumps(data, indent=2), encoding="utf-8")
+
+
+def _load_json(path: Path) -> dict[str, Any]:
+    data = json.loads(path.read_text(encoding="utf-8"))
+    if not isinstance(data, dict):
+        raise TypeError(f"Expected a JSON object in {path}")
+    return data
+
+
+@dataclass
+class ProcessResult:
+    """Captured subprocess output and timing."""
+
+    args: list[str]
+    exit_code: int
+    elapsed_s: float
+    stdout: str
+    stderr: str
+    timed_out: bool
+
+    @property
+    def command(self) -> str:
+        """Return the subprocess arguments as a reproducible command line."""
+        return subprocess.list2cmdline(self.args)
+
+
+def _kill_process_tree(pid: int) -> None:
+    try:
+        root = psutil.Process(pid)
+    except psutil.Error:
+        return
+    processes = [root]
+    with contextlib.suppress(psutil.Error):
+        processes.extend(root.children(recursive=True))
+    for process in reversed(processes):
+        with contextlib.suppress(psutil.Error):
+            process.kill()
+    with contextlib.suppress(Exception):
+        psutil.wait_procs(processes, timeout=5)
+
+
+def _run_process(args: list[str], *, timeout: int) -> ProcessResult:
+    env = {**os.environ, "PYTHONIOENCODING": "utf-8", "PYTHONUTF8": "1"}
+    kwargs: dict[str, Any] = {
+        "stdout": subprocess.PIPE,
+        "stderr": subprocess.PIPE,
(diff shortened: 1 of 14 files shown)


scripts/e2e_eval/run_llm_eval.py
```diff
@@ -0,0 +1,595 @@
...
+        *WINML_CLI,
+        "perf",
+        "-m",
+        str(bundle_dir),
+        "--runtime",
+        "winml-genai",
+        "--device",
+        device,
+        "--prompt",
+        prompt,
```
xieofxie: This breaks the long-context cases this runner is meant to support on Windows. The entire decoded prompt is placed on the child process command line, whose maximum length is 32,767 characters. A realistic 16K/32K-token prompt will exceed that and `CreateProcess` will fail with `WinError 206` before `perf` starts. Please pass the prompt through stdin or a file-backed `perf` option instead of argv.

src/winml/modelkit/commands/_perf_genai.py
```diff
@@ -392,8 +400,34 @@ def run(self) -> GenaiBenchmarkResult:
...
+        result = self._aggregate(samples)
+        result.hw_monitor = hw_metrics
+        return result
+
+    def _monitor_device(self) -> str:
+        """Return the concrete device whose adapter counters should be sampled."""
+        device = (self._config.device or "").lower()
+        if device in ("cpu", "gpu", "npu"):
+            return device
+        return self._session_device() or "auto"
```
xieofxie: For the default `device="config"` path with no EP override, `_session_device()` returns `None`, so this falls back to `"auto"`. `HWMonitor` then probes NPU before GPU regardless of the bundle's routing. On a machine with both devices, a DML/GPU-configured bundle will therefore report unrelated NPU utilization. Please resolve the effective device/adapter from the bundle/session routing, or omit adapter metrics when it cannot be determined rather than selecting an arbitrary accelerator.

scripts/e2e_eval/run_llm_eval.py
```diff
@@ -0,0 +1,595 @@
...
+            )
+        adapter_metrics = hw_monitor.get("adapter") or hw_monitor.get(device_kind) or {}
+        if int(adapter_metrics.get("sample_count") or 0) <= 0:
+            raise ValueError(f"hw_monitor collected no {expected_device} samples")
+        accelerator_util_pct = float(adapter_metrics.get("mean_pct") or 0.0)
+        memory_metrics = hw_monitor.get("device_memory") or {}
+        local_memory_mb = float(memory_metrics.get("local_mean_mb") or 0.0)
+        shared_memory_mb = float(memory_metrics.get("shared_mean_mb") or 0.0)
+        device_memory_mb = local_memory_mb if local_memory_mb > 0 else shared_memory_mb
+        capacity_mb = total_vram_mb if local_memory_mb > 0 and total_vram_mb > 0 else total_ram_mb
```
xieofxie: `total_vram_mb` comes from `GPU.get_all()[0]` (or the manual override), but `local_memory_mb` belongs to the adapter LUID selected by `HWMonitor`. This gives invalid percentages for an NPU with local memory and for multi-GPU systems where inference uses a different adapter. Please obtain capacity for the reported `adapter_luid`, and leave the percentage unavailable when that capacity cannot be matched.

scripts/e2e_eval/run_llm_eval.py
```diff
@@ -0,0 +1,595 @@
...
+        reported_ep = str((report.get("benchmark_info") or {}).get("ep") or "").lower()
+        if reported_ep and reported_ep != "config":
+            reported_eps.add(reported_ep)
+        points.append(point)
+        print(
+            f"  decode={point['tokens_per_second']:.2f} tok/s "
+            f"ttft={point['ttft_s']:.3f}s total={point['total_elapsed_s']:.3f}s"
+        )
+
+    if not points:
```
xieofxie, line 765: Reusing an output directory can leave mutually contradictory artifacts: this all-failed path writes `llm_eval_failure.json` without removing an older `llm_eval_result.json`, while a later successful run leaves the old failure file in place. Downstream consumers can ingest stale success data after a failed rerun. Please unlink the opposite artifact before publishing either outcome.

xieofxie (commented):
Found four correctness issues; details are inline.

scripts/e2e_eval/run_llm_eval.py
KayMKM: Fixed in 868b2e28. The runner now writes each generated prompt to a UTF-8 file and invokes `winml perf --prompt-file <path>`, so prompt contents no longer consume Windows argv space. `winml perf` now supports the file-backed option, rejects an explicit `--prompt` combined with it, and has CLI coverage for both cases.

src/winml/modelkit/commands/_perf_genai.py
KayMKM: Fixed in 868b2e28. `GenaiSession` now derives `effective_device` from the post-override bundle routing. Config-mode DML resolves to GPU; multi-device/unknown routing remains ambiguous. `GenaiPerfBenchmark` uses that resolved device and falls back to CPU/RAM-only monitoring when no unique adapter can be identified, rather than probing NPU first.

scripts/e2e_eval/run_llm_eval.py
KayMKM: Fixed in 868b2e28. The runner no longer takes capacity from `GPU.get_all()[0]`. Capacity-based utilization is calculated only when the caller supplies the matching `--gpu-memory-gb`; otherwise `vram.util_avg_pct` is `null` while the measured MB value is retained. The schema and tests now represent that unavailable state explicitly.

scripts/e2e_eval/run_llm_eval.py
KayMKM: Fixed in 868b2e28. Before publishing an all-failed run the runner removes any stale `llm_eval_result.json`; before publishing a successful/partial result it removes any stale `llm_eval_failure.json`. Both transitions have focused tests.

src/winml/modelkit/commands/perf.py
```diff
@@ -2362,17 +2361,30 @@ def _run_genai_runtime(ctx: click.Context, *, console: Console, json_mode: bool)
...
             ep = resolve_genai_ep(device)
 
+        prompt = p["prompt"]
+        prompt_file: Path | None = p.get("prompt_file")
+        if prompt_file is not None:
+            if cli_utils.is_cli_provided(ctx, "prompt"):
+                raise click.UsageError("--prompt and --prompt-file are mutually exclusive.")
+            try:
+                prompt = prompt_file.read_text(encoding="utf-8")
+            except OSError as exc:
```
xieofxie: `read_text(encoding="utf-8")` can also raise `UnicodeDecodeError`, which is not an `OSError`. An ANSI or corrupted prompt file therefore escapes Click's normal error handling and prints a traceback. Please catch `UnicodeError` (or `UnicodeDecodeError`) here and surface the same `ClickException`.

src/winml/modelkit/commands/_perf_genai.py
```diff
@@ -392,8 +400,35 @@ def run(self) -> GenaiBenchmarkResult:
...
+            samples = [self._time_one_generation(session, gen_config) for _ in range(total_runs)]
+
+        result = self._aggregate(samples)
+        result.hw_monitor = hw_metrics
+        return result
+
+    def _monitor_device(self) -> str:
+        """Return the concrete device whose adapter counters should be sampled."""
+        device = (self._config.device or "").lower()
+        if device in ("cpu", "gpu", "npu"):
```
xieofxie: This still selects the requested device when the EP override was a no-op. For example, `--device npu --ep qnn` on a flat/all-CPU bundle leaves routing unchanged and `session.effective_device` correctly resolves from the bundle, but this branch monitors an unrelated NPU. Please treat the loaded session's effective device as authoritative (or fail/collect CPU-only when it is unknown), even when a concrete device was requested.

src/winml/modelkit/session/genai_session.py
```diff
@@ -1834,6 +1848,78 @@ def _plugin_eps(session_options: object) -> Iterator[str]:
                             discovered.setdefault(ep, None)
         return tuple(discovered)
 
+    def _resolve_effective_device(self, cfg: dict[str, Any]) -> str | None:
+        """Resolve the single device targeted by the effective bundle config."""
+        if self._ep_override is not None and self._override_effective:
+            if self._ep_override == "CPUExecutionProvider":
+                return "cpu"
+            if self._device in ("npu", "gpu", "cpu"):
```
xieofxie: `self._device` is not necessarily the device in the effective config. With `--ep openvino` on an existing OpenVINO/GPU bundle, `_session_device()` supplies the primary NPU default while `_apply_ep_override()` preserves the shipped `device_type: GPU`; this reports/monitors NPU although inference remains on GPU. It also permits contradictions such as DML + NPU. Please infer the device from the post-override config first and validate any explicit fallback against the EP's supported devices.

src/winml/modelkit/session/genai_session.py
```diff
@@ -1834,6 +1848,78 @@ def _plugin_eps(session_options: object) -> Iterator[str]:
...
+                if not isinstance(entry, dict):
+                    continue
+                for name, options in entry.items():
+                    canonical = normalize_ep_name(str(name))
+                    if canonical == "CPUExecutionProvider":
+                        continue
+                    if canonical not in EP_SUPPORTED_DEVICES:
+                        return None
+                    supported = EP_SUPPORTED_DEVICES[cast("EPName", canonical)]
+                    requested = (
```
xieofxie, line 1954: This only recognizes `device_type`, so the QNN HTP bundles generated by this repo (`backend_path: QnnHtp.dll`) remain ambiguous because QNN advertises both NPU and GPU support. Consequently default config-mode `winml perf --runtime winml-genai --monitor` collects CPU/RAM only for the project's Qwen3 NPU bundle. Please resolve provider options through shared/catalog metadata (rather than hardcoding a QNN name here) so provider-specific routing can identify the effective device.

scripts/e2e_eval/run_llm_eval.py
```diff
@@ -0,0 +1,604 @@
...
+        stdout, stderr = process.communicate(timeout=timeout)
+        exit_code = process.returncode
+    except subprocess.TimeoutExpired:
+        timed_out = True
+        _kill_process_tree(process.pid)
+        try:
+            stdout, stderr = process.communicate(timeout=5)
+        except subprocess.TimeoutExpired:
+            process.kill()
+            stdout, stderr = process.communicate()
```
xieofxie: The timeout path can still block indefinitely here. If the root process has already exited while a descendant retains the stdout/stderr handles, killing the root is ineffective and this unbounded `communicate()` waits until that descendant exits. That defeats `--timeout`, especially around native compile-worker failures. Please terminate through a Windows Job Object/POSIX process group and keep the final pipe drain bounded.

scripts/e2e_eval/run_llm_eval.py
```diff
@@ -0,0 +1,604 @@
...
+    return _load_json(config_path)
+
+
+def _make_prompt(bundle_dir: Path, target_tokens: int, filler: str) -> str:
+    from transformers import AutoTokenizer
+
+    tokenizer = AutoTokenizer.from_pretrained(str(bundle_dir), local_files_only=True)
+    repeats = max(16, target_tokens // 4)
+    token_ids: list[int] = []
+    while len(token_ids) < target_tokens:
```
xieofxie, line 333: `--prompt-filler ""` (or any filler that tokenizes to zero IDs) makes this loop run forever because doubling the repeat count never grows `token_ids`. This occurs before `_run_process`, so the configured timeout cannot recover. Please reject a zero-token filler or detect lack of progress before looping.

scripts/e2e_eval/run_llm_eval.py
```diff
@@ -0,0 +1,604 @@
...
+        "apply_template": False,
+        "monitor": True,
+    }
+    for key, expected in expected_info.items():
+        if info.get(key) != expected:
+            raise ValueError(f"Expected benchmark_info.{key}={expected!r}, got {info.get(key)!r}")
+
+    reported_ep = str(info.get("ep") or "").lower()
+    if expected_ep:
+        normalized_ep = expected_ep.lower().removesuffix("executionprovider")
```
xieofxie: This is not the CLI's canonical EP normalization. For example, `--ep nv_tensorrt_rtx` is valid, but `GenaiSession.effective_ep` reports the preferred alias `nvtensorrtrtx`, so every successful context is rejected here. Please compare canonical names using the shared `normalize_ep_name` path (after handling any source suffix) rather than lowercasing/removing a suffix.

scripts/e2e_eval/run_llm_eval.py
```diff
@@ -0,0 +1,604 @@
...
+        if reported_ep not in {expected_ep.lower(), normalized_ep}:
+            raise ValueError(f"Expected execution provider {expected_ep!r}, got {info.get('ep')!r}")
+
+    prompt_tokens = int(info.get("prompt_tokens") or 0)
+    if prompt_tokens != target_tokens:
+        raise ValueError(
+            f"Target context was {target_tokens} tokens, but perf measured {prompt_tokens}"
+        )
+    generated_tokens = int(info.get("generated_tokens") or 0)
+    if generated_tokens != expected_max_new_tokens:
```
xieofxie: `max_new_tokens` is a soft upper bound: `generate_timed()` stops when ORT GenAI signals EOS, so a valid run may produce fewer tokens. Requiring exact equality discards every context for models that terminate early and writes an all-failure artifact. Please accept a positive generated count up to the requested maximum, or explicitly configure generation to require a fixed output length.

xieofxie (commented):
The prior four findings are addressed. I found eight additional correctness issues in the latest head; details are inline.

src/winml/modelkit/commands/perf.py
KayMKM: Fixed in b763b1ec. Prompt-file decoding now catches `UnicodeError` alongside `OSError` and converts both to the same `ClickException`. Added an invalid-UTF-8 CLI test that verifies no traceback escapes.

src/winml/modelkit/commands/_perf_genai.py
KayMKM: Fixed in b763b1ec. `_monitor_device()` now treats the loaded session `effective_device` as authoritative for every request, including concrete requested devices. A no-op NPU/QNN override on a CPU bundle therefore monitors CPU/RAM only.

src/winml/modelkit/session/genai_session.py
KayMKM: Fixed in b763b1ec. Effective-device resolution now inspects the post-override config first. An existing OpenVINO `device_type: GPU` remains GPU even if the request fallback says NPU, and requested fallbacks are only used when supported by the effective EP. DML+NPU is covered as GPU config routing.

src/winml/modelkit/session/genai_session.py
KayMKM: Fixed in b763b1ec. `EPDeviceSpec` now carries data-only provider-option routing hints, and GenAI routing matches options generically against that catalog. The repo QNN HTP bundle (`QnnHtp.dll`, including absolute paths) resolves to NPU without a QNN branch in `GenaiSession`; GPU/CPU QNN backend hints live in the same catalog.

scripts/e2e_eval/run_llm_eval.py
KayMKM: Fixed in b763b1ec. The runner now assigns the child immediately to a Windows Job Object with `KILL_ON_JOB_CLOSE`; POSIX uses the existing new session as a process group. Timeout terminates the whole group, the follow-up pipe drain is bounded to five seconds, and the last-resort path closes pipes without another unbounded `communicate()`. A real regression test covers an exited parent whose descendant retains stdout.

scripts/e2e_eval/run_llm_eval.py
KayMKM: Fixed in b763b1ec. `_make_prompt()` rejects a filler that tokenizes to zero IDs and also detects a non-growing token count inside the expansion loop. Added a zero-token filler regression test.

scripts/e2e_eval/run_llm_eval.py
KayMKM: Fixed in b763b1ec. Expected source-qualified EPs and reported aliases are now compared through shared `normalize_ep_name`. The regression test covers `nv_tensorrt_rtx@catalog` against the preferred reported alias `nvtensorrtrtx`.

scripts/e2e_eval/run_llm_eval.py
KayMKM: Fixed in b763b1ec. Generated token count is now accepted when it is positive and no greater than `max_new_tokens`, allowing valid early-EOS generations. Zero and over-limit counts remain rejected and are covered by tests.

scripts/e2e_eval/run_llm_eval.py
```diff
@@ -0,0 +1,724 @@
...
+        "env": env,
+    }
+    if platform.system() == "Windows":
+        kwargs["creationflags"] = subprocess.CREATE_NO_WINDOW
+    else:
+        kwargs["start_new_session"] = True
+
+    started = time.perf_counter()
+    process = subprocess.Popen(args, **kwargs)  # noqa: S603
+    tree = _ProcessTreeGuard(process)
```
xieofxie: If Job Object creation or `AssignProcessToJobObject` fails, this constructor raises before execution enters the cleanup `try/finally`, leaving the newly spawned benchmark process running unmanaged. Assignment can legitimately fail under restrictive/incompatible parent-job policies. Please wrap guard construction in cleanup that terminates, drains, and waits for `process` (or fall back to the psutil tree guard) before propagating the setup error.

src/winml/modelkit/session/genai_session.py
```diff
@@ -1834,6 +1848,103 @@ def _plugin_eps(session_options: object) -> Iterator[str]:
...
+    def _resolve_effective_device(self, cfg: dict[str, Any]) -> str | None:
+        """Resolve the single device targeted by the effective bundle config."""
+        configured_device = self._device_from_config(cfg)
+        if configured_device is not None:
+            return configured_device
+        if self._ep_override is not None and self._override_effective:
+            if self._ep_override == "CPUExecutionProvider":
+                return "cpu"
+            supported = EP_SUPPORTED_DEVICES[self._ep_override]
+            if self._device in supported:
```
xieofxie: This turns unresolved routing into a concrete device claim. For a multi-device EP whose effective options are empty or unrecognized (for example QNN without a known backend hint, or an existing OpenVINO stage without `device_type`), `_device_from_config()` returns `None`, but this substitutes the requested device even though the effective config does not prove that target. Monitoring can again sample an unrelated adapter. Please retain `None` for multi-device EPs unless the post-override provider options unambiguously identify the device.

scripts/e2e_eval/run_llm_eval.py
```diff
@@ -0,0 +1,724 @@
...
+    cpu_metrics = hw_monitor.get("cpu") or {}
+    ram_metrics = hw_monitor.get("ram") or {}
+    process_cpu_pct = float(cpu_metrics.get("process_mean_pct") or 0.0)
+    process_memory_mb = float(ram_metrics.get("mean_mb") or 0.0)
+    if int(cpu_metrics.get("sample_count") or 0) <= 0:
+        raise ValueError("hw_monitor collected no process CPU samples")
+    if process_memory_mb <= 0:
+        raise ValueError("hw_monitor collected no process memory samples")
+
+    if expected_device == "cpu":
```
xieofxie, line 493: The CPU path does not verify the monitor's resolved device. `--device cpu --ep dml` is accepted and the effective config runs on GPU; `perf` reports `device_kind="gpu"`, but this branch discards those metrics, records zero accelerator usage, and emits a passing CPU result. Please require `device_kind` to be absent/CPU here, or reject incompatible EP/device pairs before launching the benchmark.

xieofxie (commented):
The previous findings are addressed. I found three additional correctness issues in the latest revision; details are inline.

scripts/e2e_eval/run_llm_eval.py
KayMKM: Fixed in 54b625ea. Guard construction is now wrapped immediately after `Popen`; if Job Object setup/assignment fails, the existing psutil tree cleanup runs and pipe draining is bounded to five seconds before the setup error is re-raised. Added a regression test for this setup-failure path.

src/winml/modelkit/session/genai_session.py
KayMKM: Fixed in 54b625ea. The requested device is no longer used as evidence for a multi-device EP. When post-override options do not resolve through `device_type` or catalog hints, `effective_device` remains `None`; only single-device EPs retain an unambiguous fallback. Added an unresolved QNN-options test.

scripts/e2e_eval/run_llm_eval.py
KayMKM: Fixed in 54b625ea. The CPU result path now rejects any monitor `device_kind` other than absent/CPU, so an effective DML/GPU route cannot be emitted as a passing CPU result with zero accelerator metrics. Added focused coverage.

scripts/e2e_eval/run_llm_eval.py
```diff
@@ -0,0 +1,739 @@
...
+        kwargs["creationflags"] = subprocess.CREATE_NO_WINDOW
+    else:
+        kwargs["start_new_session"] = True
+
+    started = time.perf_counter()
+    process = subprocess.Popen(args, **kwargs)  # noqa: S603
+    try:
+        tree = _ProcessTreeGuard(process)
+    except Exception:
+        _kill_process_tree(process.pid)
```
xieofxie, line 268: This cleanup still has a race on Windows: the process begins running before Job assignment, so it can spawn a child and exit before assignment fails. At that point `psutil.Process(process.pid)` no longer exists, and `_kill_process_tree` cannot discover or terminate the orphan that inherited the pipes. Please create the child suspended, assign it to the Job, and only then resume it (or use an equivalent atomic Job-assignment mechanism).

src/winml/modelkit/commands/_perf_genai.py
```diff
@@ -392,8 +400,32 @@ def run(self) -> GenaiBenchmarkResult:
...
+            samples = [self._time_one_generation(session, gen_config) for _ in range(total_runs)]
+
+        result = self._aggregate(samples)
+        result.hw_monitor = hw_metrics
+        return result
+
+    def _monitor_device(self) -> str:
+        """Return the concrete device whose adapter counters should be sampled."""
+        effective = getattr(self._session, "effective_device", None)
+        return effective if effective in ("cpu", "gpu", "npu") else "cpu"
```
xieofxie: `None` here means the effective routing is unresolved, not that it is CPU. Collapsing it to `"cpu"` can hide accelerator execution: for example, `--device cpu` applies a CPU override that is a no-op on a flat mixed/accelerator config, `_device_from_config()` remains ambiguous, and monitoring reports no adapter. The evaluator then accepts the empty `device_kind` as CPU-only. Please preserve an explicit unresolved state and reject concrete-device evaluation when the effective device cannot be proven.

scripts/e2e_eval/run_llm_eval.py
```diff
@@ -0,0 +1,739 @@
...
+    args = parse_args(argv)
+    if any(length <= 0 for length in args.context_lengths):
+        raise ValueError("--context-lengths values must be positive")
+    if args.iterations <= 0 or args.warmup < 0 or args.max_new_tokens <= 0:
+        raise ValueError("iterations and max-new-tokens must be positive; warmup must be >= 0")
+    if args.timeout <= 0:
+        raise ValueError("--timeout must be positive")
+    if args.compile_timeout <= 0:
+        raise ValueError("--compile-timeout must be positive")
+    if args.gpu_memory_gb is not None and args.gpu_memory_gb <= 0:
```
xieofxie: `nan` and `inf` both pass this positivity check. They then flow into `gpu_memory_gb`/`total_vram_mb` and `json.dumps` emits non-standard `NaN` or `Infinity`, which strict JSON consumers reject despite the file claiming a JSON schema. Please require `math.isfinite(args.gpu_memory_gb)` and preferably serialize results with `allow_nan=False` as a final guard.

xieofxie (commented):
The latest three fixes are substantially improved, but three correctness gaps remain; details are inline.

scripts/e2e_eval/run_llm_eval.py
KayMKM: Fixed in bb1d8d2c. Windows children are now created with `CREATE_SUSPENDED`, assigned to the kill-on-close Job, and resumed only after assignment succeeds. If setup fails, the still-suspended process is cleaned up, so it cannot spawn an orphan before Job ownership is established.

src/winml/modelkit/commands/_perf_genai.py
KayMKM: Fixed in bb1d8d2c. Unresolved routing now remains explicit: `_monitor_device()` returns `None`, process-only monitoring uses CPU counters internally, and `benchmark_info.effective_device` is serialized as `null`. The evaluator requires this field to match the requested concrete device, so unresolved accelerator execution cannot pass as CPU.

scripts/e2e_eval/run_llm_eval.py
KayMKM: Fixed in bb1d8d2c. `--gpu-memory-gb` now rejects all non-finite values via `math.isfinite`, and `_write_json` uses `allow_nan=False` as the final serialization guard. Tests cover NaN, positive/negative infinity, and direct non-finite serialization.

src/winml/modelkit/commands/_perf_genai.py
```diff
@@ -392,8 +402,32 @@ def run(self) -> GenaiBenchmarkResult:
             self._config.iterations,
             self._config.max_new_tokens,
         )
-        samples = [self._time_one_generation(session, gen_config) for _ in range(total_runs)]
-        return self._aggregate(samples)
+        hw_metrics: dict[str, Any] | None = None
+        if self._config.monitor and HWMonitor.is_available():
+            monitor_device = self._monitor_device()
+            ep_name = normalize_ep_name(self._config.ep) if self._config.ep is not None else None
```
xieofxie: This still passes the *requested* EP to LUID resolution even when that override did not take effect. On a multi-GPU system, a flat OpenVINO/GPU bundle invoked with a device-selected DML override can continue running on its configured OpenVINO adapter while `HWMonitor` resolves the DML GPU's LUID. `effective_device` is still `"gpu"`, so the evaluator accepts the unrelated metrics. Please derive the EP/adapter hint from the loaded effective configuration, and omit or reject accelerator metrics unless a unique effective adapter can be proven.

xieofxie (commented):
The latest fixes address the prior findings. I found one remaining adapter-selection issue; details are inline.

src/winml/modelkit/commands/_perf_genai.py
KayMKM: Fixed in cf58d023. `GenaiSession` now records the unique non-CPU EP from the post-override effective config, and `GenaiPerfBenchmark` uses that value as the sole adapter-resolution hint instead of the requested EP. If the effective config has multiple/unknown hardware EPs, accelerator monitoring is omitted (CPU/RAM-only) and concrete-device evaluation rejects the unresolved result. Tests cover a no-op DML request over an OpenVINO/GPU effective config and ambiguous effective EP routing.

xieofxie (commented):
The effective-EP fix resolves the previous thread. I found two remaining correctness issues; details are inline.

src/winml/modelkit/session/genai_session.py
```diff
@@ -1834,6 +1856,138 @@ def _plugin_eps(session_options: object) -> Iterator[str]:
...
+        for options in session_options:
+            if not isinstance(options, dict):
+                continue
+            provider_options = options.get("provider_options", [])
+            if not isinstance(provider_options, list):
+                continue
+            for entry in provider_options:
+                if not isinstance(entry, dict):
+                    continue
+                for name in entry:
```
xieofxie, line 1899: A unique EP/device pair still does not prove a unique adapter. This extraction discards selectors in the effective provider options (for example `device_id`) and collapses multiple stages using the same EP. On a dual-GPU system, DML stages configured for GPU 1 still become only `(DmlExecutionProvider, gpu)`, so `resolve_adapter_luid()` can choose the first DML GPU and monitor GPU 0 while inference runs on GPU 1; the evaluator accepts it because the device kind matches. Please preserve a concrete adapter selector/LUID across all effective hardware stages and enable accelerator telemetry only when they resolve to one adapter.

scripts/e2e_eval/run_llm_eval.py
```diff
@@ -0,0 +1,747 @@
...
+    if args.timeout <= 0:
+        raise ValueError("--timeout must be positive")
+    if args.compile_timeout <= 0:
+        raise ValueError("--compile-timeout must be positive")
+    if args.gpu_memory_gb is not None and (
+        not math.isfinite(args.gpu_memory_gb) or args.gpu_memory_gb <= 0
+    ):
+        raise ValueError("--gpu-memory-gb must be positive")
+
+    bundle_dir = args.model.resolve()
```
xieofxie: Canonical artifacts are only cleared in the terminal success/no-points branches. If a reused output directory already contains a passing `llm_eval_result.json`, then `_make_prompt()`, process startup, or final schema validation raises, the stale success remains available for downstream ingestion. Please clear both canonical artifacts when the new run starts, or publish from a run-scoped temporary location via atomic replacement only after completion.

src/winml/modelkit/session/genai_session.py
KayMKM: Fixed conservatively in 466024ad. `resolve_adapter_luid` now gathers all ORT LUIDs matching the effective EP/device and returns one only when the physical adapter is unique. Multiple matching adapters are explicitly unresolved and do not fall back to the first PDH GPU, so concrete-device evaluation rejects the telemetry. I intentionally did not add EP-specific `device_id` interpretation in this PR; those selector schemas need dedicated per-EP support rather than guessing.

scripts/e2e_eval/run_llm_eval.py
KayMKM: Fixed in 466024ad. Both canonical result artifacts are now removed immediately after the output directory is established, before bundle validation, prompt construction, process startup, or schema validation. A regression test starts with both stale files and verifies they are gone when bundle validation raises.

src/winml/modelkit/sysinfo/pdh_adapters.py
```diff
@@ -342,12 +345,21 @@ def _resolve_via_ort(kind: str, ep_name: EPName | None) -> str | None:
...
         logger.debug(
-            "Resolved %s LUID via ORT (ep=%s, ep_name=%s): %s",
+            "Candidate %s LUID via ORT (ep=%s, ep_name=%s): %s",
             kind.upper(),
             ep_dev.ep_name,
             ep_name,
             formatted,
         )
-        return formatted
+        candidates.add(formatted)
```
xieofxie, line 338: Uniqueness is currently determined only after filtering candidates through PDH. If ORT reports two matching physical LUIDs but PDH exposes counters for only one, the hidden candidate is dropped and this returns the visible adapter, even though inference may bind to either one. Please establish uniqueness across all parseable matching ORT LUIDs first; only after proving a single ORT adapter should PDH visibility decide whether that adapter is monitorable.

scripts/e2e_eval/run_llm_eval.py
```diff
@@ -0,0 +1,749 @@
...
+    if args.timeout <= 0:
+        raise ValueError("--timeout must be positive")
+    if args.compile_timeout <= 0:
+        raise ValueError("--compile-timeout must be positive")
+    if args.gpu_memory_gb is not None and (
+        not math.isfinite(args.gpu_memory_gb) or args.gpu_memory_gb <= 0
+    ):
+        raise ValueError("--gpu-memory-gb must be positive")
+
+    output_dir = args.output_dir.resolve()
```
xieofxie, line 690: Concurrent runs sharing an output directory are not isolated: both delete the same canonical artifacts and use identical `prompt_ctx<N>.txt` / `perf_ctx<N>.json` paths. One process can therefore read and validate the other run's report; `_context_point()` also does not verify `benchmark_info.bundle_dir`, so runs with the same device/EP/settings but different models can silently publish mixed results. Please use an exclusive output-directory lock or per-run temporary directory with atomic promotion, and validate the reported bundle identity.

xieofxie (commented):
The latest fixes address the prior threads. I found two remaining correctness issues; details are inline.

src/winml/modelkit/sysinfo/pdh_adapters.py
xieofxie: Fixed in 0cc095c9. Verified the updated implementation and regression coverage.

scripts/e2e_eval/run_llm_eval.py
xieofxie: Fixed in 0cc095c9. Verified the updated implementation and regression coverage.

xieofxie (approved):
Reviewed the latest head. The previously identified correctness issues are fixed, all review threads are resolved, and CI is green.

PR #1280, opened 2026-08-05
Fix FP16 conversion for initializer-backed graph outputs

DingmaomaoBJTU:
## Summary

- handle safe top-level FP32 graph outputs supplied directly by dense initializers during FP16 conversion
- preserve exact FP32 initializer/output values when `keep_io_types=True`
- convert the backing initializer data when producing pure FP16 I/O
- reject shared, nested, overridable, unloaded-external-data, or generated-name-collision cases before ORT mutates the model

## Root cause

ORT's FP16 converter assumes every graph output has a node producer. For an output supplied directly by an initializer:

- `keep_io_types=True` creates an orphan `graph_output_castN` input, and WinML topological sorting fails with `Graph is not a DAG`
- `keep_io_types=False` changes the graph-output declaration to FLOAT16 but leaves the backing initializer FLOAT, causing strict type inference / ORT session creation to fail

The fix captures only provably safe top-level initializer outputs. It removes the exact orphan Cast for preserved FP32 I/O, or converts the resident initializer data for pure FP16. Ambiguous mixed-precision semantics fail closed.

## Validation

- `pytest tests/unit/optim/test_fp16.py -q` — 27 passed
- related optimization/quantization suite — 672 passed, 16 skipped, 1 xfailed
- `ruff check src/winml/modelkit/quant/fp16.py tests/unit/optim/test_fp16.py`
- `mypy -p winml.modelkit` — 429 source files clean
- public `winml quantize --precision fp16` reproduction on keen_hominy now succeeds; the generated model creates an ORT session with all 7 outputs, while the initializer-backed `landmarks_2d` output remains FP32 under the default keep-I/O policy

## Scope

This intentionally does not guess semantics for initializer outputs that are shared with graph consumers, captured by nested graphs, exposed as overridable graph inputs, or backed by unloaded external data. Those cases return an explicit error before conversion rather than publishing an invalid model.

### DIFF
--- a/src/winml/modelkit/quant/fp16.py
+++ b/src/winml/modelkit/quant/fp16.py
@@ -4,22 +4,251 @@
 # --------------------------------------------------------------------------
 """FP16 conversion utility for ONNX models.
 
-Provides a single entry point for FP32→FP16 model conversion, used by
+Provides a single entry point for FP32->FP16 model conversion, used by
 the quantizer's ``mode="fp16"`` path.
 """
 
 from __future__ import annotations
 
 import logging
-from typing import TYPE_CHECKING
+from dataclasses import dataclass
+from math import prod
+from typing import TYPE_CHECKING, cast
 
 from google.protobuf.message import EncodeError
 
 
 if TYPE_CHECKING:
-    from onnx import ModelProto
+    from onnx import GraphProto, ModelProto, TensorProto
 
 logger = logging.getLogger(__name__)
+
+
+@dataclass(frozen=True)
+class _InitializerOutput:
+    """A safe top-level graph output supplied directly by an initializer."""
+
+    name: str
+    output_index: int
+
+
+def _tensor_data_is_loaded(initializer: TensorProto) -> bool:
+    """Whether a FLOAT tensor carries resident data rather than only a sidecar ref."""
+    if initializer.raw_data or initializer.float_data:
+        return True
+    return prod(initializer.dims) == 0
+
+
+def _all_tensor_names(model: ModelProto) -> set[str]:
+    """Collect tensor names across scopes relevant to ORT's global name maps."""
+    names: set[str] = set()
+    for graph in [model.graph, *_iter_nested_graphs(model)]:
+        names.update(value.name for value in graph.input)
+        names.update(value.name for value in graph.output)
+        names.update(value.name for value in graph.value_info)
+        names.update(initializer.name for initializer in graph.initializer)
+        names.update(sparse.values.name for sparse in graph.sparse_initializer)
+        names.update(name for node in graph.node for name in (*node.input, *node.output) if name)
+    return names
+
+
+def _all_node_names(model: ModelProto) -> set[str]:
+    """Collect node names across scopes because ORT reserves generated names globally."""
+    return {
+        node.name
+        for graph in [model.graph, *_iter_nested_graphs(model)]
+        for node in graph.node
+        if node.name
+    }
+
+
+def _iter_nested_graphs(model: ModelProto) -> list[GraphProto]:
+    """Return nested attribute graphs without relying on tensor-name scope."""
+    from onnx import AttributeProto
+
+    nested: list[GraphProto] = []
+    pending = [model.graph]
+    while pending:
+        graph = pending.pop()
+        for node in graph.node:
+            for attribute in node.attribute:
+                if attribute.type == AttributeProto.GRAPH:
+                    nested.append(attribute.g)
+                    pending.append(attribute.g)
+                elif attribute.type == AttributeProto.GRAPHS:
+                    nested.extend(attribute.graphs)
+                    pending.extend(attribute.graphs)
+    return nested
+
+
+def _reject_nested_initializer_outputs(model: ModelProto) -> None:
+    """Reject nested initializer outputs whose lexical semantics are ambiguous."""
+    from onnx import TensorProto
+
+    for graph in _iter_nested_graphs(model):
+        produced = {name for node in graph.node for name in node.output if name}
+        output_names = {output.name for output in graph.output}
+        matches = [
+            initializer.name
+            for initializer in graph.initializer
+            if initializer.name in output_names
+            and initializer.name not in produced
+            and initializer.data_type == TensorProto.FLOAT
+        ]
+        if matches:
+            msg = (
+                "FP16 conversion cannot safely normalize initializer-backed outputs "
+                f"inside nested graphs: {', '.join(matches)}."
+            )
+            raise RuntimeError(msg)
+
+
+def _all_node_inputs(model: ModelProto) -> set[str]:
+    """Collect node input names across top-level and nested graphs."""
+    inputs = {name for node in model.graph.node for name in node.input if name}
+    for graph in _iter_nested_graphs(model):
+        inputs.update(name for node in graph.node for name in node.input if name)
+    return inputs
+
+
+def _capture_safe_initializer_outputs(
+    model: ModelProto,
+    *,
+    keep_io_types: bool,
+) -> list[_InitializerOutput]:
+    """Capture safe direct initializer outputs or fail before ORT mutates the model.
+
+    This fix intentionally covers only top-level, non-overridable, no-consumer
+    dense FLOAT outputs. Shared, nested, or overridable initializer outputs need
+    lexical-scope and mixed-precision semantics that ORT's converter does not
+    model safely; fail explicitly instead of publishing an invalid graph.
+    """
+    from onnx import TensorProto
+    from onnx.external_data_helper import uses_external_data
+
+    if not hasattr(model.graph, "input") or not hasattr(model.graph, "output"):
+        return []
+
+    _reject_nested_initializer_outputs(model)
+    produced = {name for node in model.graph.node for name in node.output if name}
+    consumed = _all_node_inputs(model)
+    graph_inputs = {value.name for value in model.graph.input}
+    initializers = {initializer.name: initializer for initializer in model.graph.initializer}
+    tensor_names = _all_tensor_names(model)
+    node_names = _all_node_names(model)
+
+    captured: list[_InitializerOutput] = []
+    for output_index, output in enumerate(model.graph.output):
+        initializer = initializers.get(output.name)
+        if (
+            initializer is None
+            or output.name in produced
+            or initializer.data_type != TensorProto.FLOAT
+        ):
+            continue
+
+        if output.name in graph_inputs:
+            msg = (
+                f"Initializer-backed output '{output.name}' is also a graph input; "
+                "FP16 conversion cannot preserve overridable-initializer semantics."
+            )
+            raise RuntimeError(msg)
+        if output.name in consumed:
+            msg = (
+                f"Initializer-backed output '{output.name}' has internal consumers; "
+                "FP16 conversion cannot safely infer their mixed-precision semantics."
+            )
+            raise RuntimeError(msg)
+        if uses_external_data(initializer) and not _tensor_data_is_loaded(initializer):
+            msg = (
+                f"Initializer-backed output '{output.name}' uses unloaded external data; "
+                "load external weights before FP16 conversion."
+            )
+            raise RuntimeError(msg)
+
+        if keep_io_types:
+            generated_tensor = f"graph_output_cast_{output_index}"
+            generated_node = f"graph_output_cast{output_index}"
+            collisions = []
+            if generated_tensor in tensor_names:
+                collisions.append(generated_tensor)
+            if generated_node in node_names:
+                collisions.append(generated_node)
+            if collisions:
+                msg = (
+                    f"FP16 conversion cannot safely allocate ORT graph-output Cast names for "
+                    f"'{output.name}'; existing names collide: {', '.join(collisions)}."
+                )
+                raise RuntimeError(msg)
+
+        captured.append(_InitializerOutput(output.name, output_index))
+    return captured
+
+
+def _convert_output_initializer_to_fp16(model: ModelProto, name: str) -> None:
+    """Convert a captured output's resident FLOAT initializer in place."""
+    from onnx import TensorProto
+    from onnx.external_data_helper import uses_external_data
+    from onnxruntime.transformers.float16 import convert_tensor_float_to_float16
+
+    initializer = next(value for value in model.graph.initializer if value.name == name)
+    if uses_external_data(initializer):
+        del initializer.external_data[:]
+        initializer.data_location = TensorProto.DEFAULT
+    initializer.CopyFrom(cast("TensorProto", convert_tensor_float_to_float16(initializer)))
+
+
+def _internalize_output_initializer(model: ModelProto, name: str) -> None:
+    """Drop stale external metadata after resident bytes were loaded."""
+    from onnx import TensorProto
+    from onnx.external_data_helper import uses_external_data
+
+    initializer = next(value for value in model.graph.initializer if value.name == name)
+    if uses_external_data(initializer):
+        del initializer.external_data[:]
+        initializer.data_location = TensorProto.DEFAULT
+
+
+def _remove_orphan_output_casts(
+    model: ModelProto,
+    captured: list[_InitializerOutput],
+) -> None:
+    """Remove ORT output Casts whose inputs cannot have producers by construction."""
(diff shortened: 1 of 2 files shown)


src/winml/modelkit/quant/fp16.py
```diff
@@ -4,24 +4,253 @@
...
+        ):
+            continue
+
+        if output.name in graph_inputs:
+            msg = (
+                f"Initializer-backed output '{output.name}' is also a graph input; "
+                "FP16 conversion cannot preserve overridable-initializer semantics."
+            )
+            raise RuntimeError(msg)
+        if output.name in consumed:
```
xieofxie: This rejects valid pure-FP16 conversions when `keep_io_types=False`. An initializer may be both a graph output and an input to an FP16-capable node such as `Add`; ORT converts the initializer and its consumers consistently, producing a checker-valid runnable model. Could this guard reject only cases where the initializer must remain FP32 (for example because of a blocked consumer), rather than rejecting every internal consumer?

src/winml/modelkit/quant/fp16.py
```diff
@@ -4,24 +4,253 @@
...
+            if initializer.name in output_names
+            and initializer.name not in produced
+            and initializer.data_type == TensorProto.FLOAT
+        ]
+        if matches:
+            msg = (
+                "FP16 conversion cannot safely normalize initializer-backed outputs "
+                f"inside nested graphs: {', '.join(matches)}."
+            )
+            raise RuntimeError(msg)
```
xieofxie, line 484: The blanket nested-graph rejection also blocks valid conversions. For example, an `If` branch can return a local FLOAT initializer that is also consumed by an FP16-capable `Identity`; ORT converts both consistently and the resulting model passes ONNX checking and inference. Can this be narrowed to nested initializer outputs that would actually remain FP32, based on consumer/block-list handling or post-conversion validation on a clone?

src/winml/modelkit/quant/fp16.py
```diff
@@ -31,34 +382,38 @@ def convert_to_fp16(
...
+    the output declaration but not its initializer, so convert that initializer
+    explicitly.
     """
     from onnx import TensorProto
     from onnxruntime.transformers.float16 import convert_float_to_float16
 
-    # Skip if model is already FP16 (check floating-point initializer dtypes)
     fp32_types = {TensorProto.FLOAT, TensorProto.DOUBLE, TensorProto.BFLOAT16}
-    initializers = model.graph.initializer
+    initializers = [
```
xieofxie: This shortcut can return an otherwise FP32 model unchanged merely because its floating initializers happen to be FP16. For example, a model with FP32 top-level `Identity` I/O and FP16 initializer-backed `If` branches satisfies this test, although upstream ORT would convert the top-level graph to FP16. Could we remove the shortcut or make it verify all convertible graph I/O, value metadata, nodes, and initializers across scopes before declaring the model already FP16?

src/winml/modelkit/quant/fp16.py
```diff
@@ -79,19 +434,32 @@ def convert_to_fp16(
...
+                item.has_consumers
+                and _initializer_data_type(converted, item.name) == TensorProto.FLOAT16
+            ):
+                _internalize_output_initializer(converted, item.name)
+            elif not item.has_consumers:
+                _convert_output_initializer_to_fp16(converted, item.name)
+
+    from onnxruntime.transformers.onnx_model import OnnxModel
+
+    OnnxModel.graph_topological_sort(converted.graph)
```
xieofxie: This unconditional ORT topological sort rejects valid graphs that consume sparse initializers because its dependency setup does not include `graph.sparse_initializer` names. A checker-valid FP32 `Reshape` using a sparse INT64 shape initializer converts with upstream ORT but fails here as "Graph is not a DAG." Could we use a sorter that treats sparse initializers as available values, or avoid this sorter when sparse initializers are present?

src/winml/modelkit/quant/fp16.py
DingmaomaoBJTU: Fixed in f49f48bd. Pure-FP16 conversion no longer rejects initializer-backed outputs solely because they have FP16-capable internal consumers. The fix now allows ORT-consistent conversion for cases like Add, while keep_io_types/block-list mismatch cases are still rejected through clone conversion plus post-conversion initializer/output dtype validation.

src/winml/modelkit/quant/fp16.py
DingmaomaoBJTU: Fixed in f49f48bd. The blanket nested-graph rejection was removed. Nested direct initializer-backed outputs now convert on a safe clone and are only rejected when the converted output declaration and backing initializer dtype diverge; the If/Identity valid case and blocked/mismatch cases are covered by regression tests.

src/winml/modelkit/quant/fp16.py
DingmaomaoBJTU: Fixed in 12e257b4. Removed the initializer-only already-FP16 shortcut so FP32 graph I/O/nodes still go through ORT conversion even when floating initializers are FP16. Added `test_fp16_initializers_do_not_skip_fp32_graph_conversion`; the existing already-FP16 tests now verify preserved FP16 graph structure without requiring an early return.

src/winml/modelkit/quant/fp16.py
DingmaomaoBJTU: Fixed in 12e257b4. Replaced the unconditional ORT sorter call with a local topological sort that treats dense initializers, sparse initializer value names, and graph inputs as available dependencies. Added a checker-valid sparse-shape `Reshape` regression test covering this case.

src/winml/modelkit/quant/fp16.py
```diff
@@ -31,34 +419,26 @@ def convert_to_fp16(
...
-    # Skip if model is already FP16 (check floating-point initializer dtypes)
-    fp32_types = {TensorProto.FLOAT, TensorProto.DOUBLE, TensorProto.BFLOAT16}
-    initializers = model.graph.initializer
-    if initializers:
-        float_inits = [t for t in initializers if t.data_type in fp32_types | {TensorProto.FLOAT16}]
-        if float_inits and all(t.data_type == TensorProto.FLOAT16 for t in float_inits):
-            logger.info("Model is already FP16 — skipping conversion.")
-            return model
-
+    _reject_duplicate_float_initializer_names(model)
```
xieofxie: This preflight scans every nested graph even when its parent node is excluded by `op_block_list`. Upstream ORT does not traverse subgraphs beneath blocked nodes, so a valid `If` whose branches each define a local FLOAT initializer named `same` converts successfully with `op_block_list=["If"]`, while this check rejects it. Could the duplicate-name traversal mirror ORT and skip subgraphs of blocked nodes, or only handle duplicate-name failures that the converter actually encounters?

xieofxie (commented):
One correctness issue remains in duplicate-initializer validation for subgraphs beneath blocked operators.

src/winml/modelkit/quant/fp16.py
DingmaomaoBJTU: Fixed in 3416cabc. The duplicate-name preflight now mirrors ORT's graph traversal for the effective `op_block_list`: it scans only graphs that ORT will visit and skips subgraphs under blocked nodes. Added `test_duplicate_initializers_under_blocked_if_are_not_rejected` for the blocked `If` case.

src/winml/modelkit/quant/fp16.py
```diff
@@ -79,19 +483,30 @@ def convert_to_fp16(
...
+            if (
+                item.has_consumers
+                and _initializer_data_type(converted, item.name) == TensorProto.FLOAT16
+            ):
+                _internalize_output_initializer(converted, item.name)
+            elif not item.has_consumers:
+                _convert_output_initializer_to_fp16(converted, item.name)
+
+    _graph_topological_sort(converted.graph)
+    _validate_initializer_output_types(converted)
```
xieofxie, line 3062: This validation still rejects valid unconsumed nested initializer outputs because the repair above only converts captured top-level initializers. For example, direct FLOAT initializer outputs in `If` branches are declared FLOAT16 by ORT while their initializer data remains FLOAT, so this throws even though converting those nested initializers yields a checker-valid model. Could we capture and convert direct initializer outputs in every graph that ORT traverses?

src/winml/modelkit/quant/fp16.py
```diff
@@ -4,24 +4,436 @@
...
+    because ORT rewrites consumer inputs to the generated output-cast alias.
+    """
+    from onnx import TensorProto
+    from onnx.external_data_helper import uses_external_data
+
+    if not hasattr(model.graph, "input") or not hasattr(model.graph, "output"):
+        return []
+
+    produced = {name for node in model.graph.node for name in node.output if name}
+    consumed = _all_node_inputs(model)
```
xieofxie: This global input-name set is not lexical-scope aware and includes subgraphs that ORT skips due to `op_block_list`. A top-level output initializer named `shared` is therefore treated as consumed when blocked `If` branches contain local, shadowing values named `shared`, causing a valid `keep_io_types=True` conversion to be rejected. Could consumer detection resolve names per graph scope and traverse only subgraphs ORT will visit for the supplied block list?

xieofxie (commented):
Two correctness gaps remain around nested initializer outputs and lexical consumer scope.

src/winml/modelkit/quant/fp16.py
DingmaomaoBJTU: Fixed in 99e1fa28/697ecb71. Direct FLOAT initializer-backed outputs are now captured across ORT-traversed graphs, and unconsumed nested outputs are repaired by updating the nested output ValueInfo and backing initializer together before validation.

src/winml/modelkit/quant/fp16.py
DingmaomaoBJTU: Fixed in 99e1fa28/697ecb71. Consumer detection now follows ONNX lexical scope and ORT traversal for the supplied op_block_list, so blocked branch-local shadows are not treated as top-level consumers. Added safeguards for ORT keep_io_types global name mapping and blocked free-capture cases as well.

src/winml/modelkit/quant/fp16.py
```diff
@@ -4,24 +4,685 @@
...
+        for output_index, output in enumerate(getattr(graph, "output", [])):
+            initializer = initializers.get(output.name)
+            if (
+                initializer is None
+                or output.name in produced
+                or initializer.data_type != TensorProto.FLOAT
+            ):
+                continue
+
+            if output.name in graph_inputs:
```
xieofxie, line 1858: This rejection should not be unconditional. With `keep_io_types=False`, an initializer that is also a graph input and output can have its declaration, default value, and consumers converted together. Upstream ORT converts that valid graph to FP16 and it runs with both the default and an FP16 caller override, whereas this now raises. Could this guard apply only when `keep_io_types=True` and allow the pure-FP16 repair path otherwise?

src/winml/modelkit/quant/fp16.py
```diff
@@ -4,24 +4,685 @@
...
+        _internalize_output_initializer(model.graph, item.name)
+    retained = [value for value in model.graph.value_info if value.name not in orphan_inputs]
+    del model.graph.value_info[:]
+    model.graph.value_info.extend(retained)
+
+
+def _graph_topological_sort(graph: GraphProto) -> None:
+    """Topologically sort nodes while treating dense and sparse initializers as inputs."""
+    deps = {initializer.name for initializer in getattr(graph, "initializer", [])}
+    deps.update(sparse.values.name for sparse in getattr(graph, "sparse_initializer", []))
```
xieofxie, line 2852: Treating sparse initializers as available fixes sorting, but it exposes a type mismatch that ORT conversion does not repair: sparse FLOAT values remain FP32 while consumers and outputs become FP16. The returned model then fails ORT loading (`tensor(float16)` does not match `tensor(float)`). Could we convert/cast sparse FLOAT initializers consistently and validate them, or reject this case before mutation rather than returning an unusable model?

xieofxie (commented):
Two conversion regressions remain for overridable initializer outputs and sparse FLOAT initializers.

src/winml/modelkit/quant/fp16.py
DingmaomaoBJTU: Kept this rejected in 908f2d25 after testing the pure-FP16 alias path. The upstream-converted model is checker-valid, but on Windows it native-crashes ORT's default optimizer when loading; disabling graph optimizations loads it. Because the value is simultaneously graph input, graph output, and initializer, ONNX SSA does not give us a safe repair that preserves the public output name plus override-default semantics. The wrapper now keeps rejecting before mutation rather than returning a model that can crash ORT.

src/winml/modelkit/quant/fp16.py
DingmaomaoBJTU: Fixed in 908f2d25. Sparse FLOAT values are now repaired after ORT conversion when every lexical use is FP16; matching sparse_tensor metadata is updated, keep_io_types and mixed FP16/FP32 cases are rejected before mutation, tensor_type sparse metadata is preflight-rejected before ORT shape inference, and direct sparse outputs are covered by regression tests.

src/winml/modelkit/quant/fp16.py
```diff
@@ -4,24 +4,857 @@
...
+                    f"Initializer-backed output '{output.name}' has internal consumers; "
+                    "FP16 conversion cannot safely preserve keep_io_types semantics."
+                )
+                raise RuntimeError(msg)
+            if (
+                keep_io_types
+                and graph_index == 0
+                and not has_consumers
+                and any(
+                    _descendant_has_node_reference(child, output.name, blocked_ops)
```
xieofxie, line 1876: This collision check only looks for nested node references, so it misses nested graph input declarations that shadow the preserved top-level output name. For example, a top-level initializer-backed output named `same` plus a traversed `Loop` body with a local state input also named `same` returns a model whose Loop state/output types diverge; ORT rejects it with `Tensor element type mismatch`. Could this account for traversed nested input/value-info declarations too, or reject/repair those shadowing declarations before conversion?

xieofxie (commented):
One correctness issue remains in keep-I/O handling for nested graph input shadowing.

src/winml/modelkit/quant/fp16.py
DingmaomaoBJTU: Fixed in 44aba031. The keep_io_types scope-safety preflight now treats ORT-traversed nested graph input/value_info declarations as global mapping references in addition to node input/output references, so local Loop state/input or value_info names that shadow a preserved top-level initializer-backed output are rejected before mutation. Added regression coverage for both nested Loop input shadowing and nested value_info shadowing.

src/winml/modelkit/quant/fp16.py
```diff
@@ -4,24 +4,875 @@
...
+        value_name == name
+        for node in getattr(graph, "node", [])
+        for value_name in (*node.input, *node.output)
+    )
+
+
+def _graph_keep_io_mapping_declares_name(graph: GraphProto, name: str) -> bool:
+    """Whether graph-level declarations can be hit by ORT's global keep-I/O map."""
+    return any(value.name == name for value in getattr(graph, "input", [])) or any(
+        value.name == name for value in getattr(graph, "value_info", [])
```
xieofxie: `value_info` only annotates an existing value; it does not introduce or shadow a graph-local binding. Treating an annotation alone as a declaration rejects valid `keep_io_types=True` models even when no nested node references that name. The Loop/value-info case added here converts to a checker-valid, strictly inferred, runnable model if this annotation-only check is omitted. Could shadow detection be limited to actual bindings/references such as graph inputs, initializers, sparse initializers, and node inputs/outputs?

xieofxie (commented):
One correctness issue remains in nested keep-I/O collision detection.

src/winml/modelkit/quant/fp16.py
DingmaomaoBJTU: Fixed in 605a62c0. `value_info` is now treated only as an annotation, not as a lexical binding, so free captures and initializer-backed outputs are resolved against the actual graph inputs/initializers/node outputs. Added executable regressions for both the free-capture and initializer-output annotation cases, including caller-model mutation checks.

xieofxie (approved):
All previously identified correctness issues have been addressed. The latest changes look good to merge.

PR #1281, opened 2026-08-05
Enable cross-EP recipe evaluation

ssss141414:
## Why

The E2E runner previously discovered recipes only from the legacy model-level directory. Recipes authored for a specific execution provider/device under `<model>/<ep>/<device>/` were therefore ignored, and there was no safe way to materialize a shared recipe set for another target before a run.

### Target-specific recipe discovery

- Resolve recipes from `examples/recipes/<model>/<ep>/<device>/` for the requested target.
- Fall back to the legacy `examples/recipes/<model>/` layout only when the target directory does not exist.
- Resolve automatic EP/device axes before recipe lookup so discovery uses the actual evaluation target.
- Recognize `fp16`, `fp32`, `w8a16`, and `w8a8` recipe variants.
- Support both composite recipe naming conventions:
  - `<task>_<precision>_config_<role>.json`
  - `<task>_<precision>_<role>_config.json`

### Safe recipe materialization

Adds:

```text
--copy-recipes-from EP DEVICE
```

Before jobs are built, the runner copies missing recipe configs from each model's source target into the requested concrete target directory. Existing target files are never overwritten, so EP-specific recipes remain authoritative. Automatic destination axes are rejected because copied recipes must use the same concrete target that discovery reads.

For example, `--ep qnn --device npu --copy-recipes-from cpu cpu` copies missing CPU recipes into `<model>/qnn/npu/`, then evaluates those target-local copies for the models selected by the normal registry or single-model arguments.

### Registry coverage

- Adds the 13 exact model/task pairs that were missing from `models_all.json`, at priority `P3`.
- Keeps the 30 existing model/task pairs at their merge-base priorities: 6 remain `P2` and 24 remain `P3`.
- Uses the existing required `Top200` group; this PR does not add a dedicated batch group or selector for the 43-pair hardware-validation set.

### Timeout cleanup resilience

Windows process-tree cleanup is best-effort when a timed-out child exits concurrently or cannot be enumerated/killed due to `psutil.NoSuchProcess` / `psutil.AccessDenied`. This prevents one cleanup race from terminating the remaining run. The existing 600-second timeout policy is unchanged.

## Usage

Run a selected model on QNN NPU using CPU recipes as the source:

```powershell
uv run python scripts/e2e_eval/run_eval.py `
  --hf-model microsoft/resnet-50 `
  --ep qnn --device npu `
  --copy-recipes-from cpu cpu `
  --eval-type both `
  --output-dir eval_results/qnn_npu `
  --continue
```

Use the existing registry filters or a custom registry when running a larger model set. Separate output directories are required because result slugs do not include EP/device.

## Output and reporting

Each command writes one `eval_result.json` per `(model, task, precision)` under its output directory. The runner records raw perf and accuracy facts; cross-EP compatibility, latency, and baseline-aware accuracy reports are generated separately in the ModelKitArtifacts repository.

## Validation

```text
uv run pytest tests/unit/eval/test_recipes.py tests/unit/eval/test_run_eval_script.py tests/unit/test_build_registry.py -q
```

Hardware validation on commit `b7a7b893`:

- QNN NPU: 83/83 jobs completed and produced unique result files.
- QNN GPU: 83/83 jobs completed and produced unique result files.
- Verified target-directory precedence, legacy fallback, FP32/composite parsing, and non-overwriting copy behavior.

Archived report data is tracked separately in gim-home/ModelKitArtifacts#196.


scripts/e2e_eval/run_eval.py
```diff
@@ -2496,7 +2504,9 @@ def _build_jobs(
     jobs: list[EvalJob] = []
     for entry in entries:
         variants = (
-            discover_recipe_variants(recipes_dir, entry.hf_id, entry.task)
+            discover_recipe_variants(
+                recipes_dir, entry.hf_id, entry.task, ep=ep, device=device
```
KayMKM: When only one target axis is supplied (for example, `--device cpu` while EP remains `auto`, or vice versa), `_build_jobs` forwards the raw `auto` value into recipe discovery. This misses authored recipes under concrete target directories such as `<model>/cpu/cpu` and silently falls back to a generated config, dropping recipe-defined inputs and evaluation settings. Could we resolve automatic EP/device values using the runtime target-selection policy before recipe lookup and cover both single-axis CLI forms?

ssss141414: Fixed in `541be416`. `_build_jobs` now resolves `(ep or "auto", device or "auto")` through the runtime `resolve_device(EPDeviceTarget(...))` policy before recipe discovery. The resolved concrete EP/device pair is also reused for NPU fan-out and skip-quant decisions, so all job-expansion policy sees the same target. Added parameterized coverage for both single-axis forms (`--device cpu` with automatic EP and `--ep qnn --device auto`) and verified the concrete pair passed to `discover_recipe_variants`. Validation: `163 passed` across `test_recipes.py` and `test_run_eval_script.py`.

scripts/e2e_eval/testsets/models_curated.json
```diff
@@ -31,5 +31,49 @@
   {"hf_id": "openai/clip-vit-large-patch14-336", "group": "ISV", "priority": "P1"},
   {"hf_id": "timm/mobilenetv3_small_100.lamb_in1k", "group": "ISV", "priority": "P1"},
   {"hf_id": "PaddlePaddle/PP-OCRv6_medium_det_onnx/inference.onnx", "task": "image-to-text", "group": "ISV", "priority": "P1"},
-  {"hf_id": "PaddlePaddle/PP-OCRv6_medium_rec_onnx/inference.onnx", "task": "image-to-text", "group": "ISV", "priority": "P1"}
+  {"hf_id": "PaddlePaddle/PP-OCRv6_medium_rec_onnx/inference.onnx", "task": "image-to-text", "group": "ISV", "priority": "P1"},
+
+  {"hf_id": "Alibaba-NLP/gte-reranker-modernbert-base", "task": "text-classification", "group": "Cross EP", "priority": "P2"},
```
KayMKM: These entries do not add an independent batch with the current registry schema: `group` is a single string, and the curated merge overwrites the existing group for each `(hf_id, task)` key. Regenerating `models_all.json` therefore moves 30 existing entries out of `Top200` (`412 -> 382`), so `--group Top200` silently stops selecting them. If Cross EP is intended to be an additional batch, could we represent that membership separately (for example, with tags or multi-group support) while preserving `Top200` membership? If this is an intentional migration, please document and test the change in set membership.

scripts/e2e_eval/run_eval.py
```diff
@@ -476,30 +484,38 @@ def _kill_process_tree(pid: int) -> None:
...
+    except ImportError:
+        psutil = None
 
-        parent = psutil.Process(pid)
+    if psutil is not None:
+        try:
+            parent = psutil.Process(pid)
+        except (psutil.NoSuchProcess, psutil.AccessDenied):
+            return
         children = parent.children(recursive=True)
```
KayMKM: `parent.children(recursive=True)` can itself raise `psutil.NoSuchProcess` when the process exits after `psutil.Process(pid)` succeeds, or `psutil.AccessDenied` when permissions change. Because this call is outside the protected blocks, the exception escapes from the normal `TimeoutExpired` and Ctrl+C cleanup paths and can abort the batch before the remaining cleanup/result handling runs. I reproduced this by making `parent.children` raise `NoSuchProcess`. Could we catch `psutil.Error` around child enumeration, continue with parent/platform fallback cleanup, and add a regression test for this race?

scripts/e2e_eval/run_eval.py
```diff
@@ -2961,6 +2998,31 @@ def main() -> None:
     # w8a16); off-NPU builds only the non-quantized recipe variants, else a
     # single winml-config fallback (variant=None).
     recipes_dir = None if args.no_recipes else args.recipes_dir
+    if args.copy_recipes_from:
+        if recipes_dir is None:
+            parser_error = "--copy-recipes-from cannot be used with --no-recipes"
+            raise ValueError(parser_error)
+        if not args.ep or args.device == "auto":
```
KayMKM: `--ep auto` is truthy, so this check accepts it even though the error says the target must be explicit. The copy then writes recipes under `auto/<device>`, while `_build_jobs()` resolves the target to a concrete EP and looks under (for example) `qnn/npu`; the copied recipes are silently ignored and the run falls back to generated configs. I reproduced this as a copy to `auto/npu` followed by zero variants discovered for `QNNExecutionProvider/npu`. Could we either reject `args.ep.casefold() == "auto"` here or resolve the target once before copying and reuse that concrete pair for job construction?

scripts/e2e_eval/testsets/models_curated.json
ssss141414: Fixed in 5a842f08. Added backward-compatible additional group memberships, restored the 30 original Top200 primary memberships, regenerated models_all.json, and added a 43 Cross EP / 30 Top200 overlap regression test.

scripts/e2e_eval/run_eval.py
ssss141414: Fixed in 5a842f08. Child enumeration now catches psutil.Error and falls through to the platform tree-kill fallback. Added a NoSuchProcess race regression test.

scripts/e2e_eval/run_eval.py
ssss141414: Fixed in 5a842f08. Recipe copying now rejects auto for either destination axis and requires a concrete EP/device pair. Added rejection and concrete-target tests.

scripts/e2e_eval/testsets/models_curated.json
```diff
@@ -31,5 +31,49 @@
   {"hf_id": "openai/clip-vit-large-patch14-336", "group": "ISV", "priority": "P1"},
   {"hf_id": "timm/mobilenetv3_small_100.lamb_in1k", "group": "ISV", "priority": "P1"},
   {"hf_id": "PaddlePaddle/PP-OCRv6_medium_det_onnx/inference.onnx", "task": "image-to-text", "group": "ISV", "priority": "P1"},
-  {"hf_id": "PaddlePaddle/PP-OCRv6_medium_rec_onnx/inference.onnx", "task": "image-to-text", "group": "ISV", "priority": "P1"}
+  {"hf_id": "PaddlePaddle/PP-OCRv6_medium_rec_onnx/inference.onnx", "task": "image-to-text", "group": "ISV", "priority": "P1"},
+
+  {"hf_id": "Alibaba-NLP/gte-reranker-modernbert-base", "task": "text-classification", "group": "Top200", "priority": "P3"},
+  {"hf_id": "Helsinki-NLP/opus-mt-en-ru", "task": "translation", "group": "Top200", "priority": "P3"},
```
KayMKM: This blanket P3 assignment also downgrades six entries that already existed as Top200/P2 at the PR merge base: both Helsinki translation entries, `cross-encoder/ms-marco-MiniLM-L4-v2`, `cross-encoder/ms-marco-MiniLM-L6-v2`, `facebook/bart-large-mnli`, and `mixedbread-ai/mxbai-rerank-xsmall-v1`. Since the default runner excludes P3, those six models silently disappear from the existing default P0/P1/P2 coverage. Only 13 of these 43 `(model, task)` pairs are actually new to `models_all.json`; could we keep the existing priorities for the other 30 and assign P3 only to the newly added entries? If removing these six from default coverage is intentional, please call that out explicitly.

scripts/e2e_eval/README.md
```diff
@@ -85,6 +86,11 @@ uv run python scripts/e2e_eval/run_eval.py --hf-model microsoft/resnet-50
 # Ignore recipes and build every model via winml config
 uv run python scripts/e2e_eval/run_eval.py --no-recipes --priority P0
 
+# Materialize CPU recipes for QNN NPU, then run the Cross EP P2 batch
+uv run python scripts/e2e_eval/run_eval.py --group "Cross EP" --priority P2 \
```
KayMKM: The `Cross EP` group has been removed from both curated and generated registries, so this documented command now exits with `No models matched the filters.` I reproduced that against this head. Please remove or replace this example and update the PR description as well, which still advertises a `Cross EP / P2` batch. Replacing it with `--group Top200 --priority P3` would not preserve the old scope: that currently selects 234 entries rather than these 43.

scripts/e2e_eval/testsets/models_curated.json
ssss141414: Fixed in 7f3d5840. Compared all 43 target pairs against the PR merge base: all 30 existing entries now preserve their original priorities with zero mismatches (the six listed entries are restored to P2), while all 13 genuinely new pairs remain P3. Added a regression test that locks the exact six-P2 / 37-P3 split in both registries.

scripts/e2e_eval/README.md
ssss141414: Fixed in 7f3d5840. Removed the stale Cross EP command from the README. Also rewrote the PR description to remove the dedicated batch/P2 claims and both invalid batch commands; it now states that no dedicated selector exists and documents generic concrete-target recipe copying instead.

KayMKM (approved):

PR #1282, opened 2026-08-05
feat: Add native Hugging Face PyTorch eval runtime

xieofxie:
## Summary
- add `winml eval --runtime winml|pytorch`, defaulting to the existing WinML export and evaluation path
- expose `evaluate(config, pytorch_model=model)` so callers can pass an existing PyTorch model through the current model-loader and evaluator abstraction
- infer tokenizer and processor assets from an explicit `model_id` or `model.config._name_or_path`, while retaining model-ID loading as the CLI convenience path
- preserve task-selected model loading and checkpoint dtype with exact CPU/CUDA pipeline placement
- adapt native causal language models to the existing text-generation evaluator contract for perplexity evaluation
- put caller-supplied modules in evaluation mode without moving or reloading them
- identify the selected runtime in console and JSON reports
- reject WinML-only options, compare/reference/input-data modes, composite ONNX models, and GenAI bundles for the PyTorch runtime while leaving task compatibility to evaluator contracts
- include focused unit coverage plus CPU and CUDA-gated E2E coverage

## Documentation
- document runtime selection, supplied-model evaluation, text-generation adaptation, supported inputs, device mapping, and incompatible options


src/winml/modelkit/commands/eval.py
```diff
@@ -115,6 +115,14 @@
         "Ignored for pre-built ONNX inputs."
     ),
 )
+@click.option(
+    "--export/--no-export",
```
xieofxie: need to discuss a better name

src/winml/modelkit/eval/evaluate.py
```diff
@@ -279,6 +335,20 @@ def _load_model(
     from ..session import EPDeviceTarget, WinMLEPRegistry, resolve_device
     from ..utils import cli as cli_utils
 
+    if not config.export_model:
```
xieofxie: how about using _select_model_loader to make it aligned?

src/winml/modelkit/commands/eval.py
xieofxie: Two alternatives to consider: (1) `--backend onnx|pytorch`, which matches the config/report terminology; or (2) `--runtime onnx|pytorch`, which emphasizes the execution path. I have left `--export/--no-export` unchanged for now pending agreement.

src/winml/modelkit/loader/__init__.py
```diff
@@ -53,6 +53,8 @@
     "HF_TASK_DEFAULTS",
     "KNOWN_TASKS",
     "TASK_SYNONYM_EXTENSIONS",
+    "NativeDevice",
```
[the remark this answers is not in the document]
xieofxie: Fixed in b5bbd98e by importing the native symbols explicitly from `.native`. That module keeps Torch/Hugging Face loading inside function calls, so the names are statically defined for CodeQL without adding heavy import-time work.

src/winml/modelkit/loader/__init__.py
```diff
@@ -53,6 +53,8 @@
...
+    "NativeHFModel",
```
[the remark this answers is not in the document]
xieofxie: Fixed in b5bbd98e by importing the native symbols explicitly from `.native`. That module keeps Torch/Hugging Face loading inside function calls, so the names are statically defined for CodeQL without adding heavy import-time work.

src/winml/modelkit/loader/__init__.py
```diff
@@ -61,11 +63,13 @@
     "get_task_abbrev",
     "load_hf_config",
     "load_hf_model",
+    "load_native_hf_model",
```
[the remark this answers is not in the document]
xieofxie: Fixed in b5bbd98e by importing the native symbols explicitly from `.native`. That module keeps Torch/Hugging Face loading inside function calls, so the names are statically defined for CodeQL without adding heavy import-time work.

src/winml/modelkit/loader/__init__.py
```diff
@@ -61,11 +63,13 @@
...
     "normalize_task",
     "resolve_composite",
     "resolve_hf_model_class",
     "resolve_hf_onnx_path",
     "resolve_loader_config",
+    "resolve_native_device",
```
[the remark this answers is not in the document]
xieofxie: Fixed in b5bbd98e by importing the native symbols explicitly from `.native`. That module keeps Torch/Hugging Face loading inside function calls, so the names are statically defined for CodeQL without adding heavy import-time work.

src/winml/modelkit/eval/evaluate.py
xieofxie: Fixed in b5bbd98e. `_ModelLoaderKind` now includes `NATIVE`, `_select_model_loader()` selects it first for the PyTorch backend, and `_load_model()` dispatches through that shared selector. CLI build-bypass diagnostics also recognize the native loader kind.

src/winml/modelkit/commands/eval.py
xieofxie: I think runtime is better which is already used in perf and we could have `winml | pytorch` here

src/winml/modelkit/eval/evaluate.py
```diff
@@ -31,6 +33,7 @@
 
 
 class _ModelLoaderKind(Enum):
+    NATIVE = auto()
```
xieofxie: maybe change to PYTORCH follow runtime

xieofxie: Implemented in d8925454. The loader kind is now `_ModelLoaderKind.PYTORCH`, matching `--runtime pytorch` throughout loader selection and build-bypass handling.

zhenchaoni:
I think this PR is adding too much runtime-specific logic instead of reusing the existing evaluator abstraction.

Evaluators already accept a model object in their constructor and invoke its forward path. `run_eval.py` demonstrates this by constructing/adapting the PyTorch model first and then passing that object to the evaluator. The new path should follow the same design:

1. Add an optional PyTorch model-instance parameter to the public evaluation API.
2. If supplied, pass that instance directly to the existing evaluator.
3. Perform any required adaptation, such as wrapping a causal LM with `HFCausalLM`, before evaluator construction.
4. Keep `model_id` loading as a separate convenience path rather than making it mandatory.

Specifically, please reconsider:

- `evaluate.py:341`: this always loads from `model_id`, so an existing model instance cannot be evaluated.
- `hf.py:279`: selecting `config.architectures[0]` introduces architecture-specific behavior and can override the correct task-selected model class.
- `evaluate.py:169`: `supports_native` duplicates evaluator capability knowledge and rejects combinations, such as PyTorch text generation, that the existing evaluator path already supports.

**Suggested direction:** simplify this PR to expose the model-instance interface and route it through the current evaluator flow. Remove architecture introspection and the hardcoded runtime/evaluator support matrix; compatibility should be determined by the evaluator's existing model contract.

xieofxie:
Implemented the suggested evaluator-first design in a28fbe06.

- `evaluate(config, pytorch_model=model)` now accepts an existing PyTorch model and passes that exact instance to the selected evaluator without reloading or relocating it.
- Explicit `config.model_id` overrides processor/tokenizer resolution; otherwise it is inferred from `model.config._name_or_path`, with a clear error when neither exists.
- Model-ID loading remains the CLI/convenience path and now uses the task-resolved class with `torch_dtype="auto"`; the `config.architectures[0]` override was removed.
- The duplicated `supports_native` flags and capability check were removed. Direct evaluators place inputs on the model's pipeline device, preserving exact CUDA ordinals.
- PyTorch text generation remains intentionally unsupported for this PR, and mask generation remains on its required composite-ONNX path.

The default WinML path is unchanged. Focused unit regressions, strict mypy, and the real CPU Hugging Face E2E pass; the CUDA E2E remains hardware-gated.

src/winml/modelkit/eval/config.py
```diff
@@ -215,14 +244,25 @@ def to_dict(self) -> dict:
...
+            result["use_cache"] = self.use_cache
+            result["rebuild"] = self.rebuild
+        if self.trust_remote_code:
+            result["trust_remote_code"] = True
         return result
 
     @classmethod
     def from_dict(cls, data: dict) -> WinMLEvaluationConfig:
         """Create from dictionary, ignoring unknown fields."""
+        legacy_runtime_fields = {"backend", "export_model"}.intersection(data)
```
xieofxie: remove them. "backend", "export_model" never exist before

xieofxie: Removed in 5f51940d. The `backend` and `export_model` compatibility handling, private exception, and associated tests are gone because those keys were never released. I also audited the superseded design and removed the unused `NativeHFModel.config` and `.task` return metadata; no other obsolete runtime-name compatibility remains.

zhenchaoni:
Thanks for the update. The model-instance interface and the `config.architectures[0]` issue are now addressed. I still see two remaining concerns:

1. `evaluate.py:167` explicitly rejects `text-generation` and `mask-generation`. This is still a hardcoded capability matrix, just in a smaller form, and it blocks valid evaluator flows. The existing baseline adapts causal language models with `HFCausalLM` and passes them to `WinMLTextGenerationEvaluator`, whose contract only requires the expected model methods. Please remove these centralized task-name bans and perform any required model adaptation in the loading/adaptation layer before passing the model to the evaluator.

2. `evaluate.py:646` passes a caller-supplied PyTorch model directly to the evaluator without calling `.eval()`. `torch.no_grad()` does not disable dropout or batch normalization training behavior, so this can produce stochastic or incorrect evaluation results. Please put supplied models into evaluation mode before constructing the evaluator, and add a test using a real model or module that verifies its training state is disabled rather than relying only on a fake evaluator.

With these changes, the implementation would better preserve the existing model/evaluator separation without duplicating capability knowledge in the evaluation entry point.

xieofxie:
Addressed in 3f21b0f3. Supplied models now go through `_load_model`, so `_select_model_loader` remains the shared runtime dispatch instead of bypassing it. The centralized `text-generation` and `mask-generation` bans are removed; native causal LMs are adapted in the native loading layer to the existing `HFCausalLM`/`WinMLTextGenerationEvaluator` contract, while other compatibility remains evaluator-owned. Caller-supplied modules are put into `.eval()` before evaluator construction without being moved or reloaded. Coverage includes a real `nn.Module` training-state assertion and a real CPU tiny-GPT2 perplexity E2E. Ruff, strict mypy, 270 broader eval tests, and both native CPU E2Es pass; CUDA remains hardware-gated.

zhenchaoni (approved):

PR #1284, opened 2026-08-05
Feat/text to image eval primitives: add prompt-driven evaluation primitives (PromptDataset + CLIPScore + Recall@K + MRR)

AChenreddy24:
This PR has the prerequisites needed to support stable diffusion models. In order to do it cleanly, SD work will be done in 2 PRs where we keep this PR general enough while adding the necessary components for SD work. 

What's in here:

1. [PromptDataset](vscode-file://vscode-app/c:/Users/achenreddy/AppData/Local/Programs/Microsoft%20VS%20Code/5264f2156c/resources/app/out/vs/code/electron-browser/workbench/workbench.html) in [datasets/prompt_dataset.py](vscode-file://vscode-app/c:/Users/achenreddy/AppData/Local/Programs/Microsoft%20VS%20Code/5264f2156c/resources/app/out/vs/code/electron-browser/workbench/workbench.html) : a task-agnostic prompt corpus. Loaders for JSONL, HuggingFace datasets, and in-memory lists. Has [filter()](vscode-file://vscode-app/c:/Users/achenreddy/AppData/Local/Programs/Microsoft%20VS%20Code/5264f2156c/resources/app/out/vs/code/electron-browser/workbench/workbench.html) and [sample()](vscode-file://vscode-app/c:/Users/achenreddy/AppData/Local/Programs/Microsoft%20VS%20Code/5264f2156c/resources/app/out/vs/code/electron-browser/workbench/workbench.html), both routed through a single [_derive()](vscode-file://vscode-app/c:/Users/achenreddy/AppData/Local/Programs/Microsoft%20VS%20Code/5264f2156c/resources/app/out/vs/code/electron-browser/workbench/workbench.html) hook so future derived-dataset ops (deduplicate, group_by, whatever) are one method call each. Uses [type(self)](vscode-file://vscode-app/c:/Users/achenreddy/AppData/Local/Programs/Microsoft%20VS%20Code/5264f2156c/resources/app/out/vs/code/electron-browser/workbench/workbench.html) so subclasses get instances of themselves back.
2. [CLIPScoreMetric](vscode-file://vscode-app/c:/Users/achenreddy/AppData/Local/Programs/Microsoft%20VS%20Code/5264f2156c/resources/app/out/vs/code/electron-browser/workbench/workbench.html) in [eval/metrics/clip_score.py](vscode-file://vscode-app/c:/Users/achenreddy/AppData/Local/Programs/Microsoft%20VS%20Code/5264f2156c/resources/app/out/vs/code/electron-browser/workbench/workbench.html) : text-image alignment via [weight * max(0, cos(text, image))](vscode-file://vscode-app/c:/Users/achenreddy/AppData/Local/Programs/Microsoft%20VS%20Code/5264f2156c/resources/app/out/vs/code/electron-browser/workbench/workbench.html) (Hessel et al. 2021). Metric only handles the scoring; the caller runs whichever CLIP variant they want and passes in the embeddings.
3. [RecallAtKMetric](vscode-file://vscode-app/c:/Users/achenreddy/AppData/Local/Programs/Microsoft%20VS%20Code/5264f2156c/resources/app/out/vs/code/electron-browser/workbench/workbench.html) + [MeanReciprocalRankMetric](vscode-file://vscode-app/c:/Users/achenreddy/AppData/Local/Programs/Microsoft%20VS%20Code/5264f2156c/resources/app/out/vs/code/electron-browser/workbench/workbench.html) in [eval/metrics/{recall_at_k,mean_reciprocal_rank}.py](vscode-file://vscode-app/c:/Users/achenreddy/AppData/Local/Programs/Microsoft%20VS%20Code/5264f2156c/resources/app/out/vs/code/electron-browser/workbench/workbench.html) : retrieval metrics with the same input shape, so both can be fed from a single [(ranked, gt)](vscode-file://vscode-app/c:/Users/achenreddy/AppData/Local/Programs/Microsoft%20VS%20Code/5264f2156c/resources/app/out/vs/code/electron-browser/workbench/workbench.html) stream. Supports single- and multi-relevant ground truth.
While I was working on the evaluation components, I noticed [WinMLImageFeatureExtractionEvaluator](vscode-file://vscode-app/c:/Users/achenreddy/AppData/Local/Programs/Microsoft%20VS%20Code/5264f2156c/resources/app/out/vs/code/electron-browser/workbench/workbench.html) already computes the full cosine similarity matrix and ranks neighbours for the k-NN classifier, but only reports classification accuracy. The SSL / embedding-quality literature (DINO, DINOv2, MoCo, MAE) reports Recall@K and MRR -those are the direct measures; k-NN accuracy is a proxy. So this PR also has it report recall_at_{1,5,10} and [mrr](vscode-file://vscode-app/c:/Users/achenreddy/AppData/Local/Programs/Microsoft%20VS%20Code/5264f2156c/resources/app/out/vs/code/electron-browser/workbench/workbench.html) alongside knn_top{1,5}_accuracy. Same similarity matrix, no meaningful extra compute, backward-compatible.

One non-trivial base-class change: widened [BaseTaskDataset.model_name](vscode-file://vscode-app/c:/Users/achenreddy/AppData/Local/Programs/Microsoft%20VS%20Code/5264f2156c/resources/app/out/vs/code/electron-browser/workbench/workbench.html) from [str](vscode-file://vscode-app/c:/Users/achenreddy/AppData/Local/Programs/Microsoft%20VS%20Code/5264f2156c/resources/app/out/vs/code/electron-browser/workbench/workbench.html) to [str | None](vscode-file://vscode-app/c:/Users/achenreddy/AppData/Local/Programs/Microsoft%20VS%20Code/5264f2156c/resources/app/out/vs/code/electron-browser/workbench/workbench.html). [PromptDataset](vscode-file://vscode-app/c:/Users/achenreddy/AppData/Local/Programs/Microsoft%20VS%20Code/5264f2156c/resources/app/out/vs/code/electron-browser/workbench/workbench.html) is task-agnostic so it doesn't have a model to bind to. All 7 existing subclasses still receive a [str](vscode-file://vscode-app/c:/Users/achenreddy/AppData/Local/Programs/Microsoft%20VS%20Code/5264f2156c/resources/app/out/vs/code/electron-browser/workbench/workbench.html) from their callers, I ran their tests to confirm nothing behaves differently.

Things I deliberately left out:

1. No WinMLTextToImageEvaluator base. [WinMLEvaluator](vscode-file://vscode-app/c:/Users/achenreddy/AppData/Local/Programs/Microsoft%20VS%20Code/5264f2156c/resources/app/out/vs/code/electron-browser/workbench/workbench.html) is fine to subclass directly, and adding a general base with only SD as its consumer would be speculative.
2. No from_csv / [from_dict](vscode-file://vscode-app/c:/Users/achenreddy/AppData/Local/Programs/Microsoft%20VS%20Code/5264f2156c/resources/app/out/vs/code/electron-browser/workbench/workbench.html) loaders on [PromptDataset](vscode-file://vscode-app/c:/Users/achenreddy/AppData/Local/Programs/Microsoft%20VS%20Code/5264f2156c/resources/app/out/vs/code/electron-browser/workbench/workbench.html). No models that would need this yet, easy to add when one shows up.
No shared cosine kernel across metrics. Each metric implements its own math (matches what every other metric here does).

Testing:
115 new tests. 773 pass in [tests/unit/{datasets,eval}/](vscode-file://vscode-app/c:/Users/achenreddy/AppData/Local/Programs/Microsoft%20VS%20Code/5264f2156c/resources/app/out/vs/code/electron-browser/workbench/workbench.html) (131 datasets + 642 eval). Broader unit sweep (~2,440 tests across build/quant/commands/inference/models/loader) has no new failures relative to [upstream/main](vscode-file://vscode-app/c:/Users/achenreddy/AppData/Local/Programs/Microsoft%20VS%20Code/5264f2156c/resources/app/out/vs/code/electron-browser/workbench/workbench.html). CI-parity ([tests/regression/](vscode-file://vscode-app/c:/Users/achenreddy/AppData/Local/Programs/Microsoft%20VS%20Code/5264f2156c/resources/app/out/vs/code/electron-browser/workbench/workbench.html) + [tests/cli/](vscode-file://vscode-app/c:/Users/achenreddy/AppData/Local/Programs/Microsoft%20VS%20Code/5264f2156c/resources/app/out/vs/code/electron-browser/workbench/workbench.html) with the CI [-m](vscode-file://vscode-app/c:/Users/achenreddy/AppData/Local/Programs/Microsoft%20VS%20Code/5264f2156c/resources/app/out/vs/code/electron-browser/workbench/workbench.html) filter) is bit-identical to [upstream/main](vscode-file://vscode-app/c:/Users/achenreddy/AppData/Local/Programs/Microsoft%20VS%20Code/5264f2156c/resources/app/out/vs/code/electron-browser/workbench/workbench.html). Ruff check + format clean on all 13 touched files.

### DIFF
--- a/src/winml/modelkit/datasets/__init__.py
+++ b/src/winml/modelkit/datasets/__init__.py
@@ -24,6 +24,7 @@ from .keypoint_detection import KeypointDetectionDataset
 from .mask_generation import MaskGenerationDataset
 from .object_detection import DEFAULT_OBJECT_DETECTION_SIZE, ObjectDetectionDataset
 from .processor_utils import get_image_processor_config
+from .prompt_dataset import PromptDataset, PromptRecord
 from .random_dataset import RandomDataset
 from .text import TextDataset
 
--- a/src/winml/modelkit/datasets/base.py
+++ b/src/winml/modelkit/datasets/base.py
@@ -25,7 +25,10 @@ class BaseTaskDataset(ABC):
     properties are readonly to ensure dataset immutability and thread safety.
 
     Attributes:
-        model_name: HuggingFace model identifier or local model path
+        model_name: HuggingFace model identifier or local model path.
+            Optional — task-agnostic data sources (e.g. prompt corpora used
+            by generative-model evaluators) may leave this ``None``. Most
+            calibration and task-oriented subclasses require it.
         dataset_name: Dataset identifier (HF dataset or local path)
         data_split: Dataset split to use (e.g., 'train', 'validation', 'test')
     """
@@ -35,7 +38,7 @@ class BaseTaskDataset(ABC):
 
     def __init__(
         self,
-        model_name: str,
+        model_name: str | None = None,
         dataset_name: str | None = None,
         max_samples: int | None = None,
         data_split: str | None = None,
@@ -44,7 +47,10 @@ class BaseTaskDataset(ABC):
         """Initialize dataset with readonly properties.
 
         Args:
-            model_name: HuggingFace model identifier or path
+            model_name: HuggingFace model identifier or path. Optional —
+                task-agnostic subclasses may pass ``None``. Task-oriented
+                subclasses that need a tokenizer/processor should validate
+                its presence themselves.
             dataset_name: Dataset name (uses DEFAULT_DATASET if None)
             max_samples: Maximum number of samples (None = use all)
             data_split: Dataset split (None = let subclass decide)
@@ -78,8 +84,12 @@ class BaseTaskDataset(ABC):
 
     # Readonly properties
     @property
-    def model_name(self) -> str:
-        """Get the model name (readonly)."""
+    def model_name(self) -> str | None:
+        """Get the model name (readonly).
+
+        ``None`` for task-agnostic subclasses that are not bound to a
+        specific model.
+        """
         return self._model_name
 
     @property
@@ -91,7 +101,6 @@ class BaseTaskDataset(ABC):
     def data_split(self) -> str | None:
         """Get the dataset split (readonly)."""
         return self._data_split
-
 
     def __len__(self) -> int:
         """Return the number of samples in the dataset."""
--- a/src/winml/modelkit/datasets/prompt_dataset.py
+++ b/src/winml/modelkit/datasets/prompt_dataset.py
@@ -1,0 +1,362 @@
+# -------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+# --------------------------------------------------------------------------
+"""Prompt-corpus dataset for evaluators that iterate over text prompts.
+
+``PromptDataset`` is a task-agnostic data source: unlike the calibration-
+oriented ``TextDataset`` (which yields tokenized tensors bound to a
+specific model) or the image-primary ``MaskGenerationDataset`` (which
+carries a prompt column alongside images), ``PromptDataset``'s primary
+yield is a bare prompt record. It is intended for evaluators that consume
+prompts to produce something and score the output against a reference
+(e.g. zero-shot classification with textual labels, VLM caption/QA
+evaluation, retrieval, future generative-image workflows).
+
+Records
+-------
+
+Each item yielded by ``PromptDataset`` is a plain ``dict`` with these keys::
+
+    {
+        "prompt":            str,                    # required
+        "negative_prompt":   str | None,             # optional
+        "reference_text":    str | None,             # optional
+        "reference_image":   str | None,             # optional (path or URL)
+        "metadata":          dict[str, Any],         # optional, arbitrary
+    }
+
+The dict shape (rather than a dataclass) matches the convention used by
+sibling datasets in this package. A parallel :class:`PromptRecord`
+dataclass is provided for callers that prefer a typed representation
+during construction.
+
+Loading
+-------
+
+Three constructors are supported::
+
+    PromptDataset.from_list([{"prompt": "..."}, ...])
+    PromptDataset.from_jsonl("path/to/prompts.jsonl")
+    PromptDataset.from_hf("dataset/repo", split="test", prompt_col="prompt")
+
+``from_hf`` also accepts a pre-loaded ``datasets.Dataset`` object in
+place of the name, which makes it straightforward to test without hitting
+the network.
+"""
+
+from __future__ import annotations
+
+import json
+import logging
+from dataclasses import dataclass, field
+from pathlib import Path
+from typing import TYPE_CHECKING, Any
+
+from .base import BaseTaskDataset
+
+
+if TYPE_CHECKING:
+    from collections.abc import Iterable
+
+logger = logging.getLogger(__name__)
+
+
+# ---------------------------------------------------------------------------
+# Record type
+# ---------------------------------------------------------------------------
+
+
+@dataclass(frozen=True)
+class PromptRecord:
+    """Typed record for a single prompt-corpus entry.
+
+    Callers may construct :class:`PromptDataset` from a list of
+    :class:`PromptRecord` instances (via :meth:`PromptDataset.from_list`)
+    or from plain dicts \u2014 they are coerced to the same internal shape.
+    """
+
+    prompt: str
+    negative_prompt: str | None = None
+    reference_text: str | None = None
+    reference_image: str | None = None
+    metadata: dict[str, Any] = field(default_factory=dict)
+
+    def to_dict(self) -> dict[str, Any]:
+        """Return the record as a dict with all keys present."""
+        return {
+            "prompt": self.prompt,
+            "negative_prompt": self.negative_prompt,
+            "reference_text": self.reference_text,
+            "reference_image": self.reference_image,
+            "metadata": dict(self.metadata),
+        }
+
+
+_ALLOWED_KEYS = frozenset(
+    {"prompt", "negative_prompt", "reference_text", "reference_image", "metadata"}
+)
+
+
+def _coerce_record(raw: Any) -> dict[str, Any]:
+    """Coerce a raw input (dict or PromptRecord) into the canonical dict shape.
+
+    Raises:
+        ValueError: If ``prompt`` is missing / empty, or if unknown keys
+            are present at the top level.
+        TypeError: If ``raw`` is not a dict or PromptRecord.
+    """
+    if isinstance(raw, PromptRecord):
+        return raw.to_dict()
+    if not isinstance(raw, dict):
+        raise TypeError(
+            f"PromptDataset record must be a dict or PromptRecord, got {type(raw).__name__}"
+        )
+
+    unknown = set(raw) - _ALLOWED_KEYS
+    if unknown:
+        raise ValueError(
+            f"PromptDataset record has unknown keys {sorted(unknown)}; "
+            f"allowed keys are {sorted(_ALLOWED_KEYS)}",
+        )
+
+    prompt = raw.get("prompt")
+    if not isinstance(prompt, str) or not prompt:
+        raise ValueError(
+            "PromptDataset record requires a non-empty string 'prompt' field",
+        )
+
+    metadata = raw.get("metadata", {}) or {}
+    if not isinstance(metadata, dict):
+        raise TypeError(
+            f"PromptDataset record 'metadata' must be a dict, got {type(metadata).__name__}"
+        )
+
+    return {
+        "prompt": prompt,
+        "negative_prompt": raw.get("negative_prompt"),
+        "reference_text": raw.get("reference_text"),
+        "reference_image": raw.get("reference_image"),
+        "metadata": dict(metadata),
+    }
+
+
+# ---------------------------------------------------------------------------
+# Dataset
+# ---------------------------------------------------------------------------
+
+
+class PromptDataset(BaseTaskDataset):
+    """Prompt-corpus dataset for evaluators.
+
+    Not intended as a calibration dataset \u2014 use :class:`TextDataset` for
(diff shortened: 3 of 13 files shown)


src/winml/modelkit/eval/metrics/recall_at_k.py
```diff
@@ -0,0 +1,132 @@
...
+            relevant: set[int] = {int(ground_truth)}
+        else:
+            relevant = {int(x) for x in np.asarray(list(ground_truth)).ravel()}
+        if not relevant:
+            raise ValueError("ground_truth must contain at least one relevant ID")
+
+        total_relevant = len(relevant)
+        for k in self._k_values:
+            top_k = ranked[:k]
+            hits = sum(1 for item in top_k if int(item) in relevant)
```
DingmaomaoBJTU, line 107: This counts every occurrence of a relevant label in `top_k`, so `Recall@K` can exceed 1.0 when the ranked list contains duplicates. That happens with the new image-feature evaluator because it passes `ranked_labels`; for example, five same-class neighbors in the top 5 would report `recall_at_5 = 5.0` for a scalar ground truth. Please count unique retrieved relevant IDs (or treat scalar ground truth as a boolean hit) and add a duplicate-label regression test.

src/winml/modelkit/datasets/__init__.py
```diff
@@ -24,6 +24,7 @@
 from .mask_generation import MaskGenerationDataset
 from .object_detection import DEFAULT_OBJECT_DETECTION_SIZE, ObjectDetectionDataset
 from .processor_utils import get_image_processor_config
+from .prompt_dataset import PromptDataset, PromptRecord
```
DingmaomaoBJTU, line 27: Since these are being exposed from the package root, please also add `PromptDataset` and `PromptRecord` to `__all__` below. Otherwise `from winml.modelkit.datasets import *` and code/docs that rely on the declared public API won't include the new prompt dataset primitives, unlike the other dataset classes exported here.

DingmaomaoBJTU (commented):
Looking at this as the foundation for a future Stable Diffusion/text-to-image evaluator: can you show how these primitives are intended to be exercised through winml eval today? For example, is there a command/config that loads a prompt dataset and reports CLIPScore / Recall@K / MRR, or are these library-only primitives until the follow-up SD evaluator PR wires them into the eval registry?

src/winml/modelkit/datasets/prompt_dataset.py
```diff
@@ -0,0 +1,466 @@
...
+        TypeError: If ``raw`` is not a dict or PromptRecord.
+    """
+    if isinstance(raw, PromptRecord):
+        return raw.to_dict()
+    if not isinstance(raw, dict):
+        raise TypeError(
+            f"PromptDataset record must be a dict or PromptRecord, got {type(raw).__name__}"
+        )
+
+    unknown = set(raw) - _ALLOWED_KEYS
```
DingmaomaoBJTU, line 117: This strict unknown-key check may make real SD prompt corpora harder to use than necessary. Prompt sets often carry extra fields such as category, style, seed, source, or difficulty; requiring callers to pre-pack all of those into metadata means common JSONL/HF prompt datasets will fail before the evaluator can use them. Could we either collect unknown top-level keys into metadata by default, or expose a non-strict mode for evaluator-facing prompt datasets?

src/winml/modelkit/datasets/prompt_dataset.py
```diff
@@ -0,0 +1,466 @@
...
+                    records.append(json.loads(line))
+                except json.JSONDecodeError as e:
+                    raise ValueError(
+                        f"PromptDataset JSONL parse error at {path}:{line_no}: {e.msg}",
+                    ) from e
+        kwargs.setdefault("dataset_name", str(path))
+        return cls(records, **kwargs)
+
+    @classmethod
+    def from_hf(
```
DingmaomaoBJTU, line 259: For the future SD evaluator, I think we should be careful not to create a parallel dataset-loading path that bypasses WinMLEvaluator.prepare_data() / DatasetConfig. The existing eval flow already owns path, 
ame, split, samples, shuffle, streaming, 
evision, default datasets, and --schema validation. Could the eventual text-to-image evaluator load via the existing DatasetConfig path and normalize rows into prompt records there, so this stays testable through winml eval instead of only through direct PromptDataset.from_hf/from_jsonl calls?

src/winml/modelkit/datasets/prompt_dataset.py
```diff
@@ -0,0 +1,466 @@
...
+        "metadata": dict(metadata),
+    }
+
+
+# ---------------------------------------------------------------------------
+# Dataset
+# ---------------------------------------------------------------------------
+
+
+class PromptDataset(BaseTaskDataset):
```
DingmaomaoBJTU, line 150: Could we narrow this abstraction so PromptDataset is primarily a row normalizer / typed view rather than the owner of dataset loading? The eval stack already has DatasetConfig + task schema + WinMLEvaluator.prepare_data() to own path, 
ame, split, samples, shuffle, streaming, 
evision, defaults, and CLI schema validation. For a future text-to-image evaluator, I think the main path should load through that existing lifecycle, then normalize rows into prompt records via something like PromptDataset.from_rows() / rom_dataset(). rom_jsonl() / rom_hf() can still exist as library conveniences, but they shouldn't become the winml eval integration path unless they deliberately mirror all of DatasetConfig semantics.

DingmaomaoBJTU (commented):
One broader design question: why do we need PromptDataset as a first-class dataset abstraction here, instead of keeping prompt handling as a text-to-image evaluator normalization step over the existing DatasetConfig / HF dataset flow? I can see value in a canonical PromptRecord shape, but I want to understand the intended boundary before we build more evaluator code on top of it.

src/winml/modelkit/datasets/prompt_dataset.py
DingmaomaoBJTU, line 150: Stepping back from the future SD evaluator use case: do we need a first-class PromptDataset abstraction yet? A text-to-image evaluator could already load a regular HF datasets.Dataset through the existing DatasetConfig / WinMLEvaluator.prepare_data() path, then read prompt, 
egative_prompt, reference columns, and metadata via columns_mapping. That keeps split/sampling/shuffle/streaming/schema behavior aligned with the rest of winml eval. Unless we expect multiple prompt-driven evaluators to share this immediately, a smaller PromptRecord/row-normalization helper may be enough for now and would avoid introducing a parallel dataset abstraction.

zhenchaoni (commented):
Reviewing the new prompt-driven evaluation primitives for correctness and alignment with the existing evaluator architecture.

src/winml/modelkit/eval/metrics/recall_at_k.py
zhenchaoni, line 107: `ranked_predictions` may contain duplicate labels, as it does in the image feature evaluator. Counting every occurrence means `[0, 0, 0]` with ground truth `0` produces Recall@K = 3.0. Please count the unique overlap, or use `any(...)` for scalar ground truth, and add a duplicate-label test.

src/winml/modelkit/datasets/prompt_dataset.py
```diff
@@ -0,0 +1,466 @@
...
+            data_split: Optional split label, purely informational.
+            **kwargs: Forwarded to the base class.
+
+        Raises:
+            ValueError: If any record fails validation or if ``records``
+                is empty.
+        """
+        # Coerce and validate before the base class initialises so that
+        # _initialize can just assign the list.
+        self._raw_records = list(records)
```
zhenchaoni, line 195: `max_samples` is applied only after `list(records)` and full validation, so a large or streaming dataset is still consumed entirely. Please limit the iterable before materializing it, and use the appropriate `select`/`take` path for Hugging Face datasets.

src/winml/modelkit/eval/image_feature_extraction_evaluator.py
```diff
@@ -95,8 +108,62 @@ def compute(self) -> dict[str, Any]:
...
+        from .metrics.mean_reciprocal_rank import MeanReciprocalRankMetric
+        from .metrics.recall_at_k import RecallAtKMetric
+
+        # L2-normalise with an eps floor to guard degenerate embeddings.
+        # Matches KNNAccuracyMetric so the ranking is bit-identical.
+        norms = np.linalg.norm(embeddings, axis=1, keepdims=True)
+        norms = np.maximum(norms, 1e-9)
+        normalized = embeddings / norms
+
+        similarity = normalized @ normalized.T
```
zhenchaoni, line 142: `KNNAccuracyMetric.compute()` already normalizes the embeddings and builds an N×N similarity matrix. Repeating that work here, followed by a full sort, significantly increases cost for large datasets. Could the shared ranking/similarity computation live in the metrics layer and be reused by kNN, Recall@K, and MRR?

src/winml/modelkit/datasets/__init__.py
zhenchaoni, line 27: Could we wire these primitives into the evaluator flow in this PR? `PromptDataset` and `CLIPScoreMetric` are currently only exported; no evaluator or task registration selects them or runs inference through the model abstraction, so `winml eval` cannot use this new functionality. If this PR is intentionally primitives-only, that scope should be made explicit and the evaluator integration tracked separately.

src/winml/modelkit/datasets/base.py
```diff
@@ -35,7 +38,7 @@ class BaseTaskDataset(ABC):
 
     def __init__(
         self,
```
zhenchaoni, line 40: Should `PromptDataset` inherit from `BaseTaskDataset` if doing so requires making `model_name` optional for every task dataset? This weakens the base contract and shifts validation to all existing subclasses. A model-agnostic dataset abstraction, or a separate prompt dataset base, may preserve the current task-dataset invariant more safely.

PR #1289, opened 2026-08-07
Report QNN detail fallback reasons

DingmaomaoBJTU:
## Summary

- add a machine-readable `fallback_reason` when QNN detail op tracing degrades to basic CSV
- distinguish missing QNN logs, schematics, SDK tools, viewer failures, missing QHAS output, and QHAS parse failures
- preserve the existing `basic_fallback` exit-0 contract while showing reason-specific CLI guidance
- keep the existing `run_qhas_viewer() -> Path | None` API through a compatibility wrapper

## Problem

Every QHAS failure previously collapsed into `status="basic_fallback"` with no reason. The CLI always suggested setting `QNN_SDK_ROOT`, even when the SDK was already configured and the actual issue was a missing optrace schematic on a precompiled EPContext model. Machine consumers could not distinguish causes either.

## Behavior

The op-trace JSON now includes an additive `fallback_reason` field with stable values:

- `qnn_log_missing`
- `schematic_missing`
- `sdk_missing`
- `viewer_failed`
- `qhas_output_missing`
- `qhas_parse_failed`

Viewer preparation, subprocess, output inspection, SDK discovery, QNN-log metadata, and offline-QHAS metadata failures remain degraded detail results when usable CSV data exists rather than escaping as `parse_failed`.

## Validation

- `python -m pytest tests/unit/session/monitor -q`: 243 passed
- `python -m pytest tests/unit/commands/test_perf_optracing.py -q`: 62 passed, 1 hardware-gated skip
- Ruff check and format check on all changed files
- mypy on all four changed production modules
- live QNN/NPU validation with `QNN_SDK_ROOT` configured and a precompiled keen_hominy EPContext lacking a schematic:
  - exit 0
  - status `basic_fallback`
  - fallback reason `schematic_missing`
  - 207 basic operators retained
  - CLI recommends rerunning from raw ONNX
  - CLI does not incorrectly suggest `QNN_SDK_ROOT`


src/winml/modelkit/session/monitor/qnn_monitor.py
```diff
@@ -682,37 +694,58 @@ def _try_qhas(
...
+                logger.info("QNNMonitor: QNN SDK discovery failed: %s", exc)
+                return None, None, None, TraceFallbackReason.SDK_MISSING
             if sdk_root is None:
                 logger.info("QNNMonitor: QNN SDK not located; skipping QHAS")
-                return None, None, None
+                return None, None, None, TraceFallbackReason.SDK_MISSING
             schematic = self._publish_schematic(schematic)
             if schematic is None:
-                return None, None, None
+                return None, None, None, TraceFallbackReason.SCHEMATIC_MISSING
```
xieofxie: Could we distinguish publication failure from a genuinely missing schematic here? `_publish_schematic()` also returns `None` when `copy2`/`os.link` raises (for example, on a filesystem without hard-link support), so this reports `schematic_missing` and recommends rerunning from raw ONNX even though the source schematic exists. A separate publish/I/O fallback reason, or falling back to the original schematic path, would keep the diagnostic and recovery action accurate.

src/winml/modelkit/session/monitor/op_metrics.py
```diff
@@ -145,15 +157,26 @@ class OpTraceResult:
     # Status of the trace. See :data:`TraceStatus` for the closed set of
     # legal values; static type checkers enforce the alias.
     status: TraceStatus = "ok"
+    # Populated when status == "basic_fallback".
+    fallback_reason: TraceFallbackReason | None = None
```
xieofxie: Please place `fallback_reason` after the existing `error` field to preserve the public positional constructor. `OpTraceResult` is exported publicly, and callers using the previous positional order (`status, error`) will now bind their error string to `fallback_reason`; for `parse_failed`, `__post_init__` then raises `ValueError`. A positional compatibility test would also cover this API contract.

xieofxie (approved):

PR #1295, opened 2026-08-09
Key EPContext cache by compile identity

DingmaomaoBJTU:
## Summary

- replace direct `WinMLSession` mtime-only EPContext reuse with a strict compile identity
- compile cache misses into immutable UUID generations selected by an atomic marker
- serialize same-identity compilation across threads/processes while isolating different identities
- hash source ONNX/external-data bytes, EP DLL, hardware, effective options, embed mode, ORT version, and generated artifacts
- publish stable CLI outputs atomically with content-addressed binary references while retaining the legacy stable binary alias

## Problem

`WinMLSession.compile()` previously reused `<model>_<device>_ctx.onnx` whenever it was newer than the source ONNX. The cache key omitted the EP/source/version, hardware, provider/session options, embed mode, ORT version, external-data sidecars, and context binary integrity. Different configurations could therefore reuse or overwrite one shared EPContext artifact.

## Behavior

- legacy or malformed caches without an identity marker rebuild once
- exact identities reuse an immutable generation
- source/external data, EP DLL, hardware, provider/session options, embed mode, or ORT changes produce a different cache namespace
- source changes observed while waiting or compiling are retried and never published under a stale identity
- custom opaque `SessionOptions` and uncertain source identity use unique non-cacheable generations
- marker/artifact failures fail closed without discarding a successfully compiled generation
- missing, modified, or timestamp-preserving substituted context binaries invalidate reuse
- compiler finalization consumes the exact session generation and never scans stale siblings after raw fallback
- public ONNX output remains stable, references an immutable content-addressed binary, and still publishes the documented stable binary alias

## Validation

- affected unit suites: 125 passed, 6 existing skips
- focused cache/concurrency suite: 16 passed
- Ruff check and format check on all changed Python files
- mypy on all changed production modules
- real QNN/NPU keen_hominy compile:
  - first compile produced one identity marker and immutable generation
  - second compile logged `Using cached EPContext`
  - second compile reduced total time from about 13.0s to 1.75s
  - public `model_qnn_ctx.onnx` references `model_qnn_ctx_qnn.<sha256>.bin`
  - legacy `model_qnn_ctx_qnn.bin` alias remains present with matching size

`uv lock --offline --check` could not resolve uncached Linux `onnx` metadata on this ARM64 host. The lockfile TOML parses and the added unit test verifies that every direct project dependency is represented in both the editable package dependencies and `requires-dist` metadata; CI runs on the supported AMD64 environment.


src/winml/modelkit/session/session.py
```diff
@@ -551,6 +537,321 @@ def compile(self) -> None:
...
+        cache_identity: dict[str, object] | None,
+    ) -> Path | None:
+        """Reuse or compile one EPContext while the caller holds its file lock."""
+        if cache_identity is not None:
+            cached_generation = self._epcontext_cached_generation(cache_path, cache_identity)
+            if cached_generation is not None:
+                logger.info("Using cached EPContext: %s", cached_generation)
+                return cached_generation
+
+        generation_path = cache_path.with_name(
```
xieofxie, line 710: What is the retention policy for these immutable generations? Every identity change creates a new cache path and UUID generation, but no successful old generation, marker, lock, or sidecar is removed. EPContext binaries can be large, so ordinary model/options/ORT updates create an unbounded disk cache. Please add bounded cleanup/GC, or simplify this to a bounded cache slot while preserving locking and atomic publication.

src/winml/modelkit/session/session.py
```diff
@@ -551,6 +537,321 @@ def compile(self) -> None:
...
+                provider_options=self._provider_options,
+            )
+            model_compiler = ort.ModelCompiler(
+                so,
+                str(self._onnx_path),
+                embed_compiled_data_into_model=self._embed_context,
+            )
+            with _suppress_native_output(compile_log):
+                model_compiler.compile_to_file(str(generation_path))
+        except Exception as exc:
```
xieofxie, line 729: Please clean up the private generation on this failure path. A provider may write a large external context binary before `compile_to_file()` raises; because every retry uses a new UUID, those partial ONNX/sidecar artifacts otherwise remain orphaned.

src/winml/modelkit/session/session.py
```diff
@@ -551,6 +537,321 @@ def compile(self) -> None:
...
+            "ep_dll": dll_fingerprint,
+            "device": self._device,
+            "hardware": {
+                "vendor_id": hardware.vendor_id,
+                "device_id": hardware.device_id,
+                "name": _optional_text(self._ep_device.device.hardware_name),
+                "driver_version": _optional_text(self._ep_device.device.driver_version),
+                "compiler_version": _optional_text(self._ep_device.device.compiler_version),
+            },
+            "provider_options": dict(sorted(self._provider_options.items())),
```
xieofxie, line 958: String equality is insufficient for provider options that refer to compiler input files: replacing such a file in place leaves the option value unchanged and may reuse an EPContext compiled from old contents. Please define an architecture-agnostic way to fingerprint file-backed compile dependencies, or narrow the claim that this identity covers all effective options.

xieofxie (commented):
The strict compile identity and per-identity locking solve a real correctness problem. However, the PR currently grows that fix into an unbounded immutable-generation/content-addressed publication system. I think the core direction is right, but the storage lifecycle should be bounded and the identity contract completed before taking on this maintenance surface.

xieofxie (commented):
The follow-up fixes cleanup after `compile_to_file()` failure and adds coverage for absolute file-backed options. Retention is still incomplete, though, and the new pruning path introduces a race with runtime session construction. I would not consider the cache lifecycle resolved until these cases are addressed.

src/winml/modelkit/session/session.py
```diff
@@ -551,6 +539,441 @@ def compile(self) -> None:
...
+                    cache_identity,
+                )
+            except (OSError, TypeError, ValueError) as exc:
+                logger.warning(
+                    "Compiled EPContext but could not write cache marker %s: %s",
+                    self._epcontext_cache_marker_path(cache_path),
+                    exc,
+                )
+            else:
+                self._prune_epcontext_cache(cache_path)
```
xieofxie, line 761: This only bounds generations after a marker is written successfully. The non-cacheable path used for opaque `SessionOptions` (`cache_identity is None`) and the marker-write failure path both return a usable UUID generation without a marker, so repeated sessions still accumulate large artifacts indefinitely. Please track and remove markerless generations after their runtime session is constructed/reset, or include them in bounded cleanup.

src/winml/modelkit/session/session.py
```diff
@@ -551,6 +539,441 @@ def compile(self) -> None:
...
+                )
+            )
+        entries.sort(key=lambda entry: (entry[0], entry[1], entry[2]), reverse=True)
+        for is_current, *_unused, marker_path, _cache_path, generation_path, lock_path in entries[
+            keep_count:
+        ]:
+            if is_current:
+                continue
+            if generation_path is not None:
+                self._discard_epcontext_generation(generation_path)
```
xieofxie: This can delete a generation another caller is about to open. A cache-hit caller returns the generation and releases its identity lock in `_compile_epcontext_with_stable_source`; before it reaches `ort.InferenceSession(...)`, a different-identity compile can prune the file here. Please pin the selected generation through runtime session construction (or use an equivalent lease); acquiring the old lock only during deletion is insufficient unless the cache-hit caller holds it through the open.

src/winml/modelkit/session/session.py
```diff
@@ -551,6 +539,441 @@ def compile(self) -> None:
...
+            return None
+        try:
+            raw_path = Path(value).expanduser()
+        except (RuntimeError, ValueError):
+            return None
+        candidates: tuple[Path, ...]
+        if raw_path.is_absolute():
+            candidates = (raw_path,)
+        else:
+            candidates = (self._onnx_path.parent / raw_path, Path.cwd() / raw_path)
```
xieofxie: This may fingerprint a different file from the one ORT consumes. Relative option values are passed to ORT unchanged, but this code guesses model-directory-first and then CWD. If both locations contain that name and the EP resolves it from CWD/search paths, changes to the consumed file will not invalidate the cache. Please canonicalize file-valued options before both fingerprinting and passing them to ORT, or require absolute paths rather than guessing resolution order.

xieofxie (commented):
The cache-hit lease now covers `InferenceSession` construction, which resolves the original prune-before-open race. However, the markerless lifecycle currently deletes the active model path immediately, and the new canonicalization/pruning logic has correctness gaps. These should be fixed before considering the lifecycle complete.

src/winml/modelkit/session/session.py
```diff
@@ -550,6 +623,499 @@ def compile(self) -> None:
         self._session = session
         self._running_model_path = model_path
         self._state = SessionState.COMPILED
+        if prepared_model.markerless and model_path != self._onnx_path:
+            self._markerless_epcontext_generations[model_path] = prepared_model.lock_path
+            self._cleanup_markerless_epcontext_generations()
```
xieofxie: This immediately deletes the markerless EPContext that was just stored in `_running_model_path`. The current ORT session may continue working after loading it, but `perf()` later resets and reopens `active_model_path` to apply/restore monitor options, so custom `SessionOptions` or marker-write-failure sessions will deterministically try to open a deleted file. Please retain markerless artifacts for as long as this session can reopen `_running_model_path`, then clean them on final reset/destruction (or regenerate from the source before reopening).

src/winml/modelkit/session/session.py
```diff
@@ -550,6 +623,499 @@ def compile(self) -> None:
...
+                prepared = self._prepare_epcontext_model(
+                    cache_path,
+                    compile_log,
+                    locked_identity,
+                )
+                if prepared is None:
+                    continue
+                try:
+                    final_identity = self._epcontext_cache_identity()
+                except (OSError, ValueError):
```
xieofxie, line 679: If marker publication failed, `prepared` can contain a markerless generation here. When this final identity calculation raises, `continue` drops that generation without returning it for session-owned cleanup or discarding it. Please discard any unpublished markerless generation before every post-preparation `continue` path.

src/winml/modelkit/session/session.py
```diff
@@ -550,6 +623,499 @@ def compile(self) -> None:
...
+                    continue
+                cache_name = marker_path.name.removesuffix(marker_suffix)
+                if not cache_name.startswith(marker_prefix) or not cache_name.endswith(
+                    cache_name_suffix
+                ):
+                    continue
+                cache_path = marker_path.with_name(cache_name)
+                generation_path: Path | None = None
+                try:
+                    recorded = json.loads(marker_path.read_text(encoding="utf-8"))
```
xieofxie: The marker contents are read before acquiring this entry's lease, then the marker is deleted later without revalidation. Another process can replace the marker with a newly compiled generation between the read and lease acquisition; this pruner then deletes the old snapshot and removes the new marker, orphaning the newly published generation. Please reread and validate the marker after acquiring the lease and prune that locked version only.

src/winml/modelkit/session/session.py
```diff
@@ -550,6 +623,499 @@ def compile(self) -> None:
...
+            option_path = self._resolve_option_file(value)
+            if option_path is not None:
+                fingerprints[key] = self._file_fingerprint(option_path)
+        return fingerprints
+
+    def _canonicalize_option_files(self, options: dict[str, str]) -> dict[str, str]:
+        """Resolve existing file-valued options once before hashing and ORT binding."""
+        canonicalized = dict(options)
+        for key, value in options.items():
+            option_path = self._resolve_option_file(value)
```
xieofxie, line 997: This changes every option whose value happens to match a CWD file, even when the option is not file-valued. For example, if CWD contains a file named `default`, QNN `htp_performance_mode=default` is rewritten to that absolute path and ORT receives an invalid mode. Please canonicalize only values explicitly identified as file-backed; existence alone cannot determine an option's semantics in an architecture-agnostic way.

xieofxie (commented):
Quick re-review: no new high-confidence findings in this delta. The markerless EPContext is now retained through `perf()` reopen and finalized before reset, unpublished generations are cleaned on final-identity failure, pruning revalidates the marker while holding its lease, and file canonicalization is limited to explicitly declared provider-option keys. The previously raised lifecycle concerns appear resolved.

xieofxie (approved):
The previously raised cache lifecycle, pruning, markerless artifact, and file-backed option concerns are resolved. No remaining high-confidence blockers from my review.

PR #1298, opened 2026-08-10
perf(optim): accelerate capability checks

xieofxie:
## Summary

- cache the unchanged ONNX Runtime input model across isolated capability probes
- infer shapes and match rewrite patterns once, then reuse the results for each rewrite capability
- avoid redundant model clones and full initializer serialization while preserving semantic diffs
- suppress false initializer changes caused only by ONNX storage metadata normalization

On the 2.8 GB WAN model used to investigate this, `--check-optim` decreased from 49:15 to 11:56 (about 76% faster).


src/winml/modelkit/optim/analysis.py
```diff
@@ -353,6 +377,7 @@ def complete_probe(cap_name: str) -> None:
         base_nodes: dict[tuple[Any, ...], tuple[bytes, NodeRef]] = {}
         _collect_nodes(base_out.graph, (), base_nodes)
         base_inits = _collect_initializers(base_out)
+        prepared_probe_model = pipe.prepare_analysis_model(current)
```
DingmaomaoBJTU: 这里的 preparation 已经脱离了原来的 per-probe 异常隔离：`prepare_analysis_model()` 一旦因 shape inference、全量 pattern matching 或缓存模型写入失败，就会直接终止整个 analysis，后续 pipe 也不会再检查。另外，成功 preparation 后如果 probe 抛出未覆盖的异常或 lazy iterator 被提前关闭，`finish_analysis()` 也不会执行，`ORTGraphPipe` 的临时目录会一直保留到对象被回收。能否把 preparation/探测生命周期包在 pipe-level `try/finally` 中，并在 preparation 失败时保持原有的跳过/降级行为？

src/winml/modelkit/optim/analysis.py
```diff
@@ -206,41 +206,65 @@ def _diff_nodes(
...
+
+    ignored_fields = {"data_location", "external_data"}
+    for descriptor in TensorProto.DESCRIPTOR.fields:
+        if descriptor.name in ignored_fields:
+            continue
+
+        base_value = getattr(base, descriptor.name)
+        probe_value = getattr(probe, descriptor.name)
+        if descriptor.is_repeated:
+            if list(base_value) != list(probe_value):
```
DingmaomaoBJTU: 这里把 repeated scalar field 转成 Python list 后比较，会把 `float_data`/`double_data` 中的 `-0.0` 和 `+0.0` 视为相同，因此 initializer 的真实变化会被漏报；signed zero 对 `Reciprocal` 等运算具有不同语义。建议对浮点 repeated fields 做位级比较，只忽略 `data_location`/`external_data` 这些存储元数据。

DingmaomaoBJTU (changes requested):
发现两处会影响 capability analysis 正确性或故障隔离的回归，建议修复后再 approve。

src/winml/modelkit/optim/analysis.py
xieofxie: 已修复。现在 pipe preparation 和 probes 都包含在 pipe-level try/finally 中；preparation 失败会记录 warning 并回退到原来的逐 capability 隔离探测路径，iterator 提前关闭时也会执行 finish_analysis()。新增了 preparation 失败降级和 lazy iterator close 清理测试。

src/winml/modelkit/optim/analysis.py
xieofxie: 已修复。float_data 和 double_data 现在通过对应宽度的字节表示做位级比较，不再使用 Python 浮点值相等语义；新增了 float/double 的 -0.0 到 +0.0 变化测试。

DingmaomaoBJTU (approved):
最新修改已解决此前两处 blocking issues：analysis preparation 失败时会回退到隔离 probe，并通过 finally 保证资源清理；initializer repeated float/double fields 使用位级比较，可正确识别 signed-zero 变化。未发现新的阻塞问题。

src/winml/modelkit/optim/analysis.py
```diff
@@ -356,87 +389,112 @@
         base_nodes: dict[tuple[Any, ...], tuple[bytes, NodeRef]] = {}
         _collect_nodes(base_out.graph, (), base_nodes)
         base_inits = _collect_initializers(base_out)
-
-        for cap_name, cap in probe_caps:
-            if on_probe_start is not None:
-                on_probe_start(cap_name)
+        prepared_probe_model = current
```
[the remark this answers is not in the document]
xieofxie: Fixed in 265632ba. The variable is now declared without the redundant initial value and is assigned only when preparation succeeds.

src/winml/modelkit/optim/analysis.py
```diff
@@ -206,41 +207,73 @@ def _diff_nodes(
...
+    for descriptor in TensorProto.DESCRIPTOR.fields:
+        if descriptor.name in ignored_fields:
+            continue
 
-def _collect_initializers(model: ModelProto) -> dict[str, bytes]:
-    """Return ``{initializer_name: signature_bytes}`` for the top-level graph."""
-    return {init.name: _initializer_key(init) for init in model.graph.initializer}
+        base_value = getattr(base, descriptor.name)
+        probe_value = getattr(probe, descriptor.name)
+        if descriptor.is_repeated:
```
DingmaomaoBJTU: `FieldDescriptor.is_repeated` is not available in protobuf 4/5, while ONNX 1.18 declares `protobuf>=4.25.1`. On those supported versions, any initializer comparison that passes the protobuf equality fast path can raise `AttributeError` here and abort capability analysis. Please use the stable descriptor API (`descriptor.label == descriptor.LABEL_REPEATED`) instead and add coverage for the comparison path.

DingmaomaoBJTU (changes requested):
The latest initialization cleanup is safe, but the refreshed full-PR review found a protobuf compatibility regression that can break initializer diffing on supported environments.

src/winml/modelkit/optim/analysis.py
xieofxie: Fixed in e9910650. Rather than relying on either is_repeated (missing in protobuf 4/5) or label (missing in protobuf 7), initializer comparison no longer depends on repeated-field descriptor properties. Float fields retain bit-exact comparison, other scalar/repeated fields use container equality, and message presence uses has_presence. Added repeated integer-field coverage; the initializer suite passes on protobuf 7.35.1.

src/winml/modelkit/optim/analysis.py
```diff
@@ -206,41 +207,71 @@ def _diff_nodes(
...
+    would allocate and copy all model weights for every capability probe.
     """
-    if init.data_location == TensorProto.EXTERNAL or len(init.external_data):
-        normalized = TensorProto()
-        normalized.CopyFrom(init)
-        normalized.ClearField("external_data")
-        normalized.ClearField("data_location")
-        return normalized.SerializeToString()
-    return init.SerializeToString()
+    if base == probe:
```
DingmaomaoBJTU: This fast path can still bypass the signed-zero fix. With protobuf 4/5 and PROTOCOL_BUFFERS_PYTHON_IMPLEMENTATION=python, protobuf message equality treats repeated -0.0 and +0.0 as equal even though their serialized bits differ, so _initializers_equal() returns here before reaching the bitwise float_data/double_data checks. Please perform those float-field bitwise checks before the protobuf equality fast path (and cover the pure-Python implementation) so behavior does not depend on the selected protobuf backend.

DingmaomaoBJTU (changes requested):
The descriptor API compatibility issue is fixed, but signed-zero detection still fails under protobuf's supported pure-Python implementation because the protobuf equality fast path runs before the bitwise comparison.

src/winml/modelkit/optim/analysis.py
xieofxie: Fixed in e170e01e. float_data and double_data are now compared bitwise before protobuf message equality, so the fast path cannot hide signed-zero changes. Added a subprocess regression with PROTOCOL_BUFFERS_PYTHON_IMPLEMENTATION=python.

src/winml/modelkit/optim/analysis.py
```diff
@@ -206,41 +207,71 @@ def _diff_nodes(
...
-    if init.data_location == TensorProto.EXTERNAL or len(init.external_data):
-        normalized = TensorProto()
-        normalized.CopyFrom(init)
-        normalized.ClearField("external_data")
-        normalized.ClearField("data_location")
-        return normalized.SerializeToString()
-    return init.SerializeToString()
+    float_field_types = {"float_data": "f", "double_data": "d"}
+    for field_name, typecode in float_field_types.items():
+        if array(typecode, getattr(base, field_name)).tobytes() != array(
```
DingmaomaoBJTU, line 236: This now materializes full copies of both float_data and double_data for every initializer comparison, even when the protos are identical. Since the same baseline initializers are compared across many capability probes, valid typed-field models can incur repeated O(weight-size) allocations and a major slowdown, which conflicts with this PR's performance goal. Could we preserve the zero-copy protobuf equality fast path for backends where equality is bitwise-correct, and special-case the pure-Python backend (or otherwise detect signed-zero differences without copying every payload on every probe)?

DingmaomaoBJTU (changes requested):
The pure-Python signed-zero issue is fixed, but the fix introduces a substantial hot-path regression for typed-field initializers by copying float payloads before the protobuf equality fast path on every probe.

src/winml/modelkit/optim/analysis.py
xieofxie: Fixed in bfd3de38. Native protobuf backends now return immediately on message equality without materializing typed fields; only the pure-Python backend performs the bitwise float check when equality succeeds. Coverage verifies both the pure-Python signed-zero case and that the native equality path never calls array().

src/winml/modelkit/optim/analysis.py
```diff
@@ -206,41 +209,75 @@ def _diff_nodes(
...
     """
-    if init.data_location == TensorProto.EXTERNAL or len(init.external_data):
-        normalized = TensorProto()
-        normalized.CopyFrom(init)
-        normalized.ClearField("external_data")
-        normalized.ClearField("data_location")
-        return normalized.SerializeToString()
-    return init.SerializeToString()
+    protobuf_equal = base == probe
+    if protobuf_equal and not _PURE_PYTHON_PROTOBUF:
```
DingmaomaoBJTU: This condition treats both upb and cpp as bitwise-safe. The cpp backend's message equality uses MessageDifferencer numeric float comparison, where -0.0 == +0.0, so it can still return here and miss a real initializer change. Please restrict the zero-copy equality fast path to api_implementation.Type() == 'upb' and retain the bitwise float/double comparison for both python and cpp backends.

DingmaomaoBJTU (changes requested):
The native fast path restores performance for upb, but it currently also applies to the cpp protobuf backend, whose equality is not bitwise-safe for signed zero.

src/winml/modelkit/optim/analysis.py
xieofxie: Fixed in 714958a3 (integrated and pushed in 58ef8ae0). The zero-copy equality fast path is now restricted to api_implementation.Type() == "upb"; both cpp and python backends perform bitwise float/double comparison. Added coverage that cpp mode inspects typed payloads while upb still avoids copies.

DingmaomaoBJTU (approved):
Re-reviewed the latest head after the main merge. The protobuf equality fast path is now correctly limited to upb, while python and cpp retain bitwise signed-zero comparison. The prior correctness and performance concerns are resolved, and no new blocking issues were found.

PR #1301, opened 2026-08-11
Add routed affine and positive Exp scale folding

DingmaomaoBJTU:
## Summary

- extend `conv-channel-affine-folding` across nested static channel `Split` routes and parallel non-overlapping channel `Slice` routes
- add opt-in `exp-positive-scale-folding` for finite, strictly positive constant scales after `Exp`
- make the rewrites fail closed for dynamic or ambiguous routes, observed intermediates, custom domains, unloaded external data, definition collisions, and cyclic graphs

The implementation is generic ONNX graph rewriting. It contains no model-name, architecture, or execution-provider special cases.

## CLI

The complete optimization now runs in one invocation:

```powershell
winml optimize -m model.onnx -o model_opt.onnx `
  --enable-gather-slice-to-split-fusion `
  --enable-conv-channel-affine-folding `
  --enable-exp-positive-scale-folding
```

`ORTGraphPipe` restores eligible `Split` nodes first. `AlgebraicRewritePipe` then folds routed Conv affine operations and positive post-`Exp` scales.

For `Exp(x) * C`, the new capability requires floating-point `C` to be immutable, finite, strictly positive, and statically broadcastable. It either combines `log(C)` with an existing safe pre-`Exp` constant bias or replaces the post-`Exp` `Mul` with a pre-`Exp` `Add(log(C))` while preserving element order through static shape-only views.

## Validation

- `93 passed` in `tests/unit/optim/pipes/test_pipe_algebraic.py`
- `640 passed, 16 skipped, 1 xfailed` in `tests/unit/optim`
- Ruff check and format check pass
- independent review completed after graph-safety hardening

Generated-graph coverage includes CLI composition, nested `Split`, parallel `Slice`, broadcast and dtype behavior, float32/float64 constants, idempotence, custom domains, overridable initializers, external data, shared/captured/output tensors, malformed definitions, cycles, and route-depth limits.

## Real-model evidence

A single CLI invocation on the motivating source model produced:

- nodes: `195 -> 181`
- both static channel `Split` nodes restored
- seven routed Conv affine nodes and the positive post-`Exp` scale `Mul` removed
- exact eight-output public signature
- compact spatial log-bias initializer retained
- all eight CPU outputs bit-exact to the previously confirmed C1 graph on the fixed input

The graph-equivalent C1 candidate was previously confirmed on QNN NPU with 20 warmups and 200 measured iterations across four alternating AB/BA pairs:

- baseline mean p50: `27.167 ms`
- candidate mean p50: `21.786 ms`
- mean gain: `5.381 ms` / `19.808%`
- paired Student-t 95% CI: `[5.261, 5.502] ms`
- target hits: `4/4` at or below `25 ms`

Quality evidence for that candidate passed canonical tensor replay and 40/40 local-smoke renders (mean PSNR `89.340 dB`, minimum `83.508 dB`). The final FFHQ quality gate remains pending.


xieofxie:
could you add a anyone can understand example of exp-positive-scale-folding to prove its numeric correctness?

DingmaomaoBJTU:
Addressed in the latest commits by adding an explicit numeric example and by documenting the algebraic assumptions in the tests.

For `exp-positive-scale-folding`, the optimized graph rewrites this shape of computation:

```text
biased = x + bias
original = Exp(biased) * scale
```

into the log-domain equivalent:

```text
folded_bias = bias + log(scale)
rewritten = Exp(x + folded_bias)
```

The identity is elementwise, so broadcasting-safe constants work the same way:

```text
Exp(x + bias) * scale
= Exp(x + bias) * Exp(log(scale))        because scale > 0
= Exp((x + bias) + log(scale))           because Exp(a) * Exp(b) = Exp(a + b)
= Exp(x + (bias + log(scale)))
```

The new test `TestExpPositiveScaleFolding.test_simple_numeric_example_matches_log_domain_identity` uses small readable values:

```text
x         = [0, 2]
bias      = [1, -1]
log_scale = [2, 0.5]
scale     = Exp(log_scale)

original[0] = Exp(0 + 1) * Exp(2)   = Exp(3)
rewritten[0] = Exp(0 + 1 + 2)       = Exp(3)

original[1] = Exp(2 - 1) * Exp(0.5) = Exp(1.5)
rewritten[1] = Exp(2 - 1 + 0.5)     = Exp(1.5)
```

The test verifies both sides with ONNX Runtime, and also checks that the optimizer actually removed the post-`Exp` `Mul` and replaced the pre-`Exp` bias with `[3, -0.5]`.

The implementation is guarded so the rewrite only happens when the scale is a finite floating-point constant, strictly positive, broadcast-compatible with the `Exp` output, and the computed `bias + log(scale)` remains finite. If any of those conditions fail, the graph is left unchanged.

For the other algebraic rewrites in the same update:

- `conv-channel-affine-folding` folds per-channel affine branches after `Conv`. For each output channel `c`, if `Conv_c(x) = W_c * x + b_c` and the branch applies `Conv_c(x) * s_c + o_c`, the folded parameters are `W'_c = s_c * W_c` and `b'_c = s_c * b_c + o_c`, so `Conv'_c(x) = s_c * (W_c * x + b_c) + o_c`, which is identical.
- The fallback sibling `Slice -> Split` rewrite only handles contiguous, non-overlapping, step-1 static slices that cover the full input extent along one axis. A `Split` with those segment lengths returns the same tensor intervals as the original sibling `Slice` nodes, preserving output names.

Validation run on the branch:

```text
python -m ruff check src\winml\modelkit\optim\pipes\algebraic.py tests\unit\optim\pipes\test_pipe_algebraic.py
python -m pytest tests\unit\optim\pipes\test_pipe_algebraic.py::TestStaticSplitToSlice tests\unit\optim\pipes\test_pipe_algebraic.py::TestConvChannelAffineFolding tests\unit\optim\pipes\test_pipe_algebraic.py::TestExpPositiveScaleFolding -q
python -m mypy -p winml.modelkit
```

CI is also green on the latest push.

src/winml/modelkit/optim/pipes/algebraic.py
```diff
@@ -490,6 +789,246 @@ def _channel_preserving_view_output(
...
+    )
+
+
+def _fold_exp_positive_scales(
+    model: onnx.ModelProto,
+    allocator: _NameAllocator,
+) -> None:
+    """Fold eligible positive post-Exp constants into existing input biases."""
+    index = _GraphIndex.build(model)
+    for add in list(model.graph.node):
```
xieofxie: **High:** Rebuilding `graph.node` inside this loop invalidates the `NodeProto` references held by the original `list(model.graph.node)` snapshot. After the first successful fold, later candidates are discovered from stale nodes: mutating them does not update the live graph, while the corresponding live `Mul` can still be removed. With two eligible Exp chains this can produce a wrong result or an undefined graph output. Please restart candidate discovery after each mutation (or resolve candidates through stable tensor/node identities), and add a two-chain validity/equivalence test.

xieofxie (commented):
Found one graph-corruption bug and two additional correctness/resource issues in the new Exp scale folding path.

src/winml/modelkit/optim/pipes/algebraic.py
```diff
@@ -490,6 +789,246 @@ def _channel_preserving_view_output(
...
+        return None
+    output_node, mul, scale, output_shape = post_exp
+    input_shape = _static_shape(index, exp.input[0])
+    if (
+        input_shape is None
+        or any(dimension <= 0 for dimension in (*input_shape, *output_shape))
+        or np.prod(input_shape, dtype=np.int64) != np.prod(output_shape, dtype=np.int64)
+    ):
+        return None
+    broadcast_scale = np.asarray(np.broadcast_to(scale, output_shape))
```
xieofxie: **Medium:** This fully materializes the broadcasted constant and then serializes it as the new initializer. A scalar scale on a large activation therefore turns into an activation-sized initializer (for example, `[1,1024,65536]` creates roughly 256 MiB of float32 data). Please preserve a compact broadcastable shape where possible, derive a compact equivalent through view routes, or decline the fold when expansion would be excessive.

src/winml/modelkit/optim/pipes/algebraic.py
```diff
@@ -490,6 +789,246 @@ def _channel_preserving_view_output(
...
+    log_scale = np.asarray(np.log(broadcast_scale).reshape(input_shape), dtype=scale.dtype)
```
xieofxie: **Medium:** This rewrite is not semantics-preserving at floating-point overflow/underflow boundaries. For float32 `x = 89` and `scale = 1e-10`, `Exp(x) * scale` evaluates to `inf`, while `Exp(x + log(scale))` is finite (about `4.49e28`). Unless this optimization is explicitly approximate-math, please retain the `Mul` without range proof; otherwise gate/document the relaxed semantics and test boundary behavior.

src/winml/modelkit/optim/pipes/algebraic.py
DingmaomaoBJTU: Fixed in `562f96eb`. `_fold_exp_positive_scales` now mutates at most one candidate per pass and rebuilds `_GraphIndex` before discovering the next one, so later folds are resolved from live `graph.node` references instead of a stale snapshot. I added `test_multiple_exp_chains_are_folded_from_live_graph_nodes`, which reproduces the two-chain case and verifies the rewritten model is still valid and ONNX Runtime-equivalent for both outputs.

src/winml/modelkit/optim/pipes/algebraic.py
DingmaomaoBJTU: Fixed in `562f96eb`. The insert path no longer materializes `broadcast_to(scale, output_shape)` just to serialize it. `_compact_log_scale_for_input` preserves the original compact broadcastable shape when it already broadcasts to the `Exp` input shape, and only reshapes to the input shape when the scale already has that exact element count. The new scalar-scale regression test asserts the generated `log_scale` initializer remains scalar instead of expanding to the activation shape.

src/winml/modelkit/optim/pipes/algebraic.py
DingmaomaoBJTU: Addressed in `562f96eb` by making the relaxed floating-point boundary behavior explicit for this opt-in capability and covering it with a regression test. The capability description now says it uses relaxed floating-point overflow semantics, and `test_float32_boundary_behavior_is_relaxed_when_enabled` covers the `float32 x=89, scale=1e-10` case where the original graph overflows while the log-domain rewrite remains finite.

src/winml/modelkit/optim/pipes/algebraic.py
```diff
@@ -490,6 +799,277 @@ def _channel_preserving_view_output(
...
+        log_scale=log_scale,
+    )
+
+
+def _compact_log_scale_for_input(
+    scale: np.ndarray,
+    input_shape: tuple[int, ...],
+) -> np.ndarray | None:
+    """Return a log-scale initializer without expanding broadcast-only dimensions."""
+    if _shape_broadcasts_to(scale.shape, input_shape):
```
xieofxie: **High:** Checking only whether the post-view scale broadcasts to the pre-`Exp` shape does not prove that it maps to the same flattened elements across the view route. For example, reshape an Exp input from `(2,2,3)` to `(3,2,2)` and apply a `(1,2,1)` scale: this shape also broadcasts to `(2,2,3)`, but moving it before the reshape changes the flattened scale pattern (the original alternates groups differently than the rewrite). Please verify broadcast equivalence through every view, expand the post-view scale and reshape it back when bounded, or skip this fold.

src/winml/modelkit/optim/pipes/algebraic.py
```diff
@@ -490,6 +799,277 @@ def _channel_preserving_view_output(
...
+        or bias.dtype != scale.dtype
+        or not np.isfinite(bias).all()
+        or not np.isfinite(scale).all()
+        or not np.all(scale > 0)
+    ):
+        return None
+    try:
+        np.broadcast_to(bias, add_shape)
+        np.broadcast_to(scale, add_shape)
+        combined_bias = np.asarray(bias + np.log(scale), dtype=bias.dtype)
```
xieofxie: **Medium:** This can still create a huge initializer through orthogonal broadcasting. For example, bias `(128,1,128)` plus scale `(1,128,1)` materializes and serializes `(128,128,128)` even though both inputs are compact; larger dimensions can allocate multiple GiB or OOM. Please reject materially expanding combinations or preserve the existing bias and introduce the compact log-scale as a separate `Add`.

xieofxie (commented):
The stale-node fix and overflow-semantics disclosure are resolved. Two issues remain in scale-shape handling.

src/winml/modelkit/optim/pipes/algebraic.py
DingmaomaoBJTU: Fixed in `7a4434d3`. The no-existing-bias insert path no longer assumes that a post-view broadcast shape is equivalent just because it also broadcasts to the pre-`Exp` shape. `_compact_log_scale_for_input` now only preserves a compact scale directly when the `Exp` input/output shapes match, when the scale is scalar, or when the scale already has the full element count and can be reshaped back to the pre-`Exp` shape. Otherwise the fold is skipped. I added `test_post_view_scale_with_different_flattened_pattern_is_unchanged` for the `(2,2,3) -> (3,2,2)` / `(1,2,1)` case.

src/winml/modelkit/optim/pipes/algebraic.py
DingmaomaoBJTU: Fixed in `7a4434d3`. Existing-bias folding now computes the combined bias only when the broadcast result does not grow beyond the larger existing compact operand. If bias and log-scale would broadcast orthogonally into a larger initializer, the rewrite preserves the original compact bias and inserts a second compact pre-`Exp` `Add` for `log(scale)` instead of serializing the expanded shape. I added `test_orthogonal_bias_and_scale_broadcast_uses_compact_log_scale_add` to verify the generated log-scale keeps shape `(1,2,1)` and no `(2,2,3)` expanded initializer is introduced.

src/winml/modelkit/optim/pipes/algebraic.py
```diff
@@ -490,6 +767,375 @@ def _channel_preserving_view_output(
...
+        if insert_candidate is None:
+            continue
+        log_scale_name = _new_initializer(
+            model,
+            allocator,
+            insert_candidate.log_scale,
+            "algebraic_exp_log_scale",
+        )
+        adjusted_name = allocator.new("algebraic_exp_adjusted")
+        add = onnx.helper.make_node(
```
xieofxie, line 1227: **Medium:** This generated `Add` assumes multidirectional broadcasting, but the rewrite has no opset guard. In ONNX opset <= 6, `Add` defaults to `broadcast=0`; a valid legacy `Exp -> Mul` using `Mul(broadcast=1)` can therefore be rewritten into an invalid/non-broadcasting `Add`. Please skip Exp folding for the standard domain below opset 7 (and preferably when the opset is unknown), or emit the correct legacy `broadcast`/`axis` attributes.

src/winml/modelkit/optim/pipes/algebraic.py
```diff
@@ -211,6 +314,20 @@ def _static_shape(index: _GraphIndex, name: str) -> tuple[int, ...] | None:
...
+        return False
+    padded_shape = (1,) * (len(target_shape) - len(shape)) + shape
+    return all(
+        source_dimension in (1, target_dimension)
+        for source_dimension, target_dimension in zip(padded_shape, target_shape, strict=True)
+    )
+
+
+def _shape_element_count(shape: tuple[int, ...]) -> int:
+    return int(np.prod(shape, dtype=np.int64))
```
xieofxie: **Medium:** `np.prod(..., dtype=np.int64)` can wrap for legal int64 dimension values, causing the later size checks to accept incompatible shapes and potentially raise an uncaught `ValueError` during `reshape`. For example, `(7, 5270498306774157605)` wraps to `3`, so a three-element scale reaches `reshape(input_shape)`. Please use overflow-safe `math.prod` consistently for shape products, reject counts beyond NumPy indexability, and fail closed on reshape/allocation errors.

xieofxie (commented):
The reshape mapping and orthogonal-broadcast initializer issues are resolved. Two additional edge cases remain.

src/winml/modelkit/optim/pipes/algebraic.py
DingmaomaoBJTU: Fixed in `ca3a2d61`. `exp-positive-scale-folding` is now gated on the standard ONNX opset being present and >= 7 before it inserts any new `Add`, so legacy opset-6 or missing-opset models are left unchanged instead of relying on multidirectional broadcast semantics. I added `test_legacy_or_missing_standard_opset_is_unchanged` for both missing opset and opset 6.

src/winml/modelkit/optim/pipes/algebraic.py
DingmaomaoBJTU: Fixed in `ca3a2d61`. Shape element counts now use Python `math.prod` through `_shape_element_count` instead of `np.prod(..., dtype=np.int64)`, reject counts beyond NumPy indexability, and use fail-closed reshape handling. The new `test_huge_shape_product_overflow_is_unchanged` covers the wrapped-product case `(7, 5270498306774157605)` that previously reached `reshape(input_shape)` and raised.

src/winml/modelkit/optim/pipes/algebraic.py
```diff
@@ -211,6 +316,36 @@ def _static_shape(index: _GraphIndex, name: str) -> tuple[int, ...] | None:
...
+    left: tuple[int, ...],
+    right: tuple[int, ...],
+) -> bool:
+    left_count = _shape_element_count(left)
+    return left_count >= 0 and left_count == _shape_element_count(right)
+
+
+def _standard_opset_version(model: onnx.ModelProto) -> int | None:
+    for opset in model.opset_import:
+        if opset.domain in ("", "ai.onnx"):
```
xieofxie: **Medium:** Returning the first `""` or `"ai.onnx"` import is not fail-closed for ambiguous opset lists. For imports such as `[("ai.onnx", 17), ("", 6)]` or duplicate empty-domain imports `[("", 17), ("", 6)]`, this enables the rewrite at 17 even though default-domain nodes may resolve as opset 6, reintroducing the non-broadcasting `Add` problem. Please resolve only the serialized standard domain (`""`) and reject missing, duplicate, or conflicting entries. The sibling-Slice-to-Split lookup around line 598 should reuse the same strict resolver, since first-match ambiguity can also emit Split-13 input form into a legacy graph.

xieofxie (commented):
The shape-product overflow issue is resolved. One opset edge case remains.

src/winml/modelkit/optim/pipes/algebraic.py
DingmaomaoBJTU: Fixed in `6fa5343a`. Standard-domain opset resolution now uses `_strict_default_opset_version`, which only accepts exactly one serialized default-domain (`""`) import. It rejects missing default opset, duplicate default-domain imports, and `ai.onnx`-only/ambiguous combinations instead of treating the first `""` or `"ai.onnx"` as authoritative.

The sibling Slice-to-Split path now uses the same strict resolver before emitting Split-13 input form. I added regression coverage for ambiguous Exp folding cases (`ai.onnx` 17 + default 6, duplicate default 17/6) and ambiguous sibling Slice-to-Split cases (`ai.onnx` 17 + default 12, duplicate default 17/12).

xieofxie (commented):
The strict default-opset resolution issue is resolved. Four additional compatibility/performance issues remain.

src/winml/modelkit/optim/pipes/algebraic.py
```diff
@@ -140,30 +225,50 @@ def _attribute(node: onnx.NodeProto, name: str, default: Any = None) -> Any:
     return default
 
 
+def _is_standard_onnx_node(node: onnx.NodeProto) -> bool:
+    return node.domain in ("", "ai.onnx")
```
xieofxie: **Medium:** Serialized standard ONNX nodes use the empty domain; treating explicit `ai.onnx` nodes as interchangeable can apply built-in semantics to an extension domain. With a normal empty-domain opset plus an `ai.onnx` import, `ai.onnx::Exp`/`Mul` can be rewritten and the custom `Mul` removed. Please require `node.domain == ""` so custom-domain operators always fail closed.

src/winml/modelkit/optim/pipes/algebraic.py
```diff
@@ -490,6 +780,381 @@ def _channel_preserving_view_output(
...
+    model: onnx.ModelProto,
+    allocator: _NameAllocator,
+) -> bool:
+    index = _GraphIndex.build(model)
+    for add in model.graph.node:
+        bias_candidate = _exp_scale_candidate(index, add)
+        if bias_candidate is None:
+            continue
+        if bias_candidate.combined_bias is not None:
+            combined_name = _new_initializer(
```
xieofxie, line 1191: **Medium:** For IR versions before 4, every initializer must also be declared as a graph input. This and the other generated-initializer paths append only to `graph.initializer`, so a valid IR-v3 model that triggers a rewrite becomes invalid. Please either skip initializer-producing rewrites for `model.ir_version < 4` or add matching graph inputs according to the legacy IR rules.

src/winml/modelkit/optim/pipes/algebraic.py
```diff
@@ -712,93 +1450,54 @@ def _fold_channel_affine(
     index = _GraphIndex.build(model)
     for original_conv in list(model.graph.node):
         if (
-            original_conv.op_type != "Conv"
+            not _is_standard_onnx_node(original_conv)
```
xieofxie, line 1570: **Medium:** This folding path assumes opset-7 multidirectional broadcasting and ignores legacy `broadcast`/`axis` attributes on the affine nodes. In opset 6, a `[2,2,1,1]` Conv output multiplied by `[2,1,1]` with `broadcast=1, axis=0` scales batches, but the operand is right-aligned as `[1,2,1,1]` and folded as channel scales, changing results. Please gate Conv affine folding on a strict default opset >= 7 or implement legacy broadcast semantics.

src/winml/modelkit/optim/pipes/algebraic.py
```diff
@@ -490,6 +780,381 @@ def _channel_preserving_view_output(
...
+        return True
+    return False
+
+
+def _fold_exp_positive_scales(
+    model: onnx.ModelProto,
+    allocator: _NameAllocator,
+) -> None:
+    """Fold eligible positive post-Exp constants into the Exp input."""
+    while _fold_existing_exp_bias_scale(model, allocator):
```
xieofxie: **Medium:** Each successful iteration rebuilds the entire graph index but folds only one candidate, making this quadratic in the number of independent Exp/Mul chains. Measurements scale from about 0.25 s for 100 chains to 2.7 s for 300 and 9.4 s for 600. Please collect and apply non-overlapping live candidates in batches (then rebuild between batches) so large generated graphs do not incur quadratic optimization time.

src/winml/modelkit/optim/pipes/algebraic.py
DingmaomaoBJTU: Fixed in 786a0fd5. `_is_standard_onnx_node` now only treats serialized empty-domain nodes as standard ONNX; explicit `ai.onnx` and other non-empty domains fail closed. Added regressions for Split, Constant, and Exp/Mul route nodes to ensure these graphs remain unchanged.

src/winml/modelkit/optim/pipes/algebraic.py
DingmaomaoBJTU: Fixed in 786a0fd5. The pipe now returns IR versions below 4 unchanged before any algebraic rewrite can generate initializer-only tensors, avoiding invalid IR-v3 models. Added an IR-v3 generated-initializer regression.

src/winml/modelkit/optim/pipes/algebraic.py
DingmaomaoBJTU: Fixed in 786a0fd5. Conv channel affine folding now requires a strict default ONNX opset >= 7, so opset-6 `broadcast`/`axis` semantics are not interpreted as multidirectional broadcasting. Added an opset-6 `Mul(broadcast=1, axis=0)` regression that stays unchanged.

src/winml/modelkit/optim/pipes/algebraic.py
DingmaomaoBJTU: Fixed in 786a0fd5. Exp scale folding now collects non-overlapping live candidates from one graph index and applies each batch in a single graph rewrite for both existing-bias and inserted-Add cases. Added a multi-chain regression with a bounded graph-index build count to catch one-candidate-per-rebuild behavior.

src/winml/modelkit/optim/pipes/algebraic.py
```diff
@@ -490,6 +823,395 @@ def _channel_preserving_view_output(
...
+    allocator: _NameAllocator,
+) -> bool:
+    index = _GraphIndex.build(model)
+    insert_candidates = _select_exp_scale_insert_candidates(model, index)
+    if not insert_candidates:
+        return False
+
+    removed: set[int] = set()
+    insert_before: dict[int, onnx.NodeProto] = {}
+    for insert_candidate in insert_candidates:
```
xieofxie, line 1219: **Medium:** Batching removes repeated graph-index builds for independent candidates, but each candidate allocates the same three prefixes and `_NameAllocator.new` restarts suffix probing from zero every time. That still performs Θ(N²) set lookups (500,500 probes for 1,000 allocations of one prefix). Please track the next suffix per prefix so repeated allocations are amortized O(1).

src/winml/modelkit/optim/pipes/algebraic.py
```diff
@@ -490,6 +823,395 @@ def _channel_preserving_view_output(
...
+    model.graph.node.extend(rewritten)
+    return True
+
+
+def _fold_exp_positive_scales(
+    model: onnx.ModelProto,
+    allocator: _NameAllocator,
+) -> None:
+    """Fold eligible positive post-Exp constants into the Exp input."""
+    while _fold_existing_exp_bias_scales(model, allocator):
```
xieofxie: **Medium:** This still rebuilds the full graph once per scale in a serial chain such as `Add(bias) -> Exp -> Mul(s1) -> Mul(s2) -> ...`: only the first `Mul` is eligible in each index, and removing it exposes the next. Instrumentation shows N+2 index builds (130 for 128 scales), so total work remains Θ(N²). Please collect maximal consecutive positive-scale `Mul` chains in one pass or update the index incrementally.

xieofxie (commented):
The four compatibility fixes are resolved and the batched rewrites are graph-safe. The quadratic-runtime fix remains incomplete in two paths.

src/winml/modelkit/optim/pipes/algebraic.py
DingmaomaoBJTU: Fixed in 69599177. `_NameAllocator` now tracks the next suffix per prefix, so repeated allocations for `algebraic_exp_log_scale`, `algebraic_exp_adjusted`, and `algebraic_exp_log_add` do not restart probing from zero. Added `TestNameAllocator.test_repeated_prefix_allocation_does_not_restart_suffix_probe`, which failed at 528 membership probes for 32 allocations before the fix and now stays bounded.

src/winml/modelkit/optim/pipes/algebraic.py
DingmaomaoBJTU: Fixed in 69599177. `_post_exp_scale` now collects a compact-combinable serial `Mul` scale prefix from one graph index, returns the combined scale plus all Mul nodes, and the rewriter removes that whole prefix in one graph rewrite. It fails closed if combining scales would require orthogonal broadcast expansion or incompatible dtypes. Added `test_serial_exp_mul_scales_are_folded_without_per_scale_rebuilds`, which failed at 11 index builds for 6 scales before the fix and now stays bounded while preserving ORT outputs.

src/winml/modelkit/optim/pipes/algebraic.py
```diff
@@ -490,6 +825,448 @@ def _channel_preserving_view_output(
...
+    except ValueError:
+        return None
+    combined_count = _shape_element_count(combined_shape)
+    if (
+        combined_count < 0
+        or not _shape_broadcasts_to(combined_shape, target_shape)
+        or combined_count > max(int(left.size), int(right.size))
+    ):
+        return None
+    combined_scale = np.asarray(left * right, dtype=left.dtype)
```
xieofxie: **Medium:** Collapsing sequential `Mul` constants changes intermediate finite-precision underflow/rounding beyond the documented single-scale Exp relaxation. For float32 `Exp(x) -> Mul(1e-30) -> Mul(1e30)` with `x = log(1e-20)`, the original underflows to `0` at the first Mul, while combining the scales to `1` yields about `1e-20`. Please preserve sequential Muls unless this broader relaxation is explicitly accepted, or fold only the first Mul.

src/winml/modelkit/optim/pipes/algebraic.py
```diff
@@ -490,6 +825,448 @@ def _channel_preserving_view_output(
...
+        try:
+            np.broadcast_to(scale, output_shape)
+        except ValueError:
+            break
+        next_scale = (
+            scale
+            if combined_scale is None
+            else _compact_combined_scale(combined_scale, scale, output_shape)
+        )
+        if next_scale is None:
```
xieofxie: **Medium:** When a serial chain cannot be compacted because the combined scale becomes zero/nonfinite, collection stops after one Mul. The outer fold loop then removes one node and rebuilds the entire index for every remaining Mul; 256 scalar `1e30` scales cause roughly 261 index builds. Please collect the full chain independently of compactability and process safe groups once, or stop folding the remainder instead of repeatedly rebuilding.

src/winml/modelkit/optim/pipes/algebraic.py
```diff
@@ -592,89 +1412,119 @@ def _collect_routed_affine_candidates(
...
-    weight = next(
-        (
-            initializer
-            for initializer in model.graph.initializer
-            if initializer.name == conv.input[1]
-        ),
-        None,
-    )
-    if weight is None:
+    weights = _initializer_array(index, conv.input[1])
```
xieofxie, line 1507: **Medium:** This parameter-copy path does not reject nonfinite accumulated affine values or final casted parameters. A float32 Conv weight `1e-38` followed through a nested Split route by two `Mul(1e20)` nodes originally produces a finite result, but accumulated scale overflow can create infinite weights and a NaN bias, changing output to NaN. Please require accumulated scale/offset and final casted weights/biases to be finite before mutating the graph.

xieofxie (commented):
Per-prefix name allocation is resolved. Three issues remain in serial scale handling and Conv parameter synthesis.

src/winml/modelkit/optim/pipes/algebraic.py
DingmaomaoBJTU: Fixed in 3267e4bf. Exp scale folding now preserves serial post-Exp `Mul` boundaries: `_post_exp_scale` only returns the first positive scale Mul from the original graph, and generated exp-folding nodes/initializers act as a persistent marker so the second public `optimize_onnx` pass does not fold the later serial Mul. Added direct and public `optimize_onnx` regressions for `Exp -> Mul(1e-30) -> Mul(1e30)` at the float32 underflow boundary.

src/winml/modelkit/optim/pipes/algebraic.py
DingmaomaoBJTU: Fixed in 3267e4bf. The exp folding pass no longer loops over newly exposed serial Muls. Existing-bias and no-bias candidates are selected from one original `_GraphIndex` and applied in one graph rewrite, so a noncompact/nonfinite serial scale chain folds only the first safe Mul and leaves the remainder without N rebuilds. Added a regression that previously folded all six scales and now keeps the remaining five Muls with bounded index builds.

src/winml/modelkit/optim/pipes/algebraic.py
DingmaomaoBJTU: Fixed in 3267e4bf. Conv channel affine folding now rejects nonfinite accumulated scale/offset values, nonfinite source weights/biases, and nonfinite final casted folded weights/biases before mutating the graph. I also fixed the partial-mutation case where folded weights were inserted before folded bias validation. Added regressions for nonfinite synthesized parameters and for avoiding partial Conv mutation when only folded bias overflows.

src/winml/modelkit/optim/pipes/algebraic.py
```diff
@@ -490,6 +831,401 @@ def _channel_preserving_view_output(
...
+        (position, values)
+        for position, input_name in enumerate(node.input)
+        if (data_name is None or input_name != data_name)
+        and (values := _constant_array(index, input_name)) is not None
+    ]
+    return constants[0] if len(constants) == 1 else None
+
+
+def _is_generated_exp_folding_name(name: str) -> bool:
+    return any(name.startswith(prefix) for prefix in EXP_FOLDING_GENERATED_PREFIXES)
```
xieofxie: **Medium:** Inferring generated provenance with `startswith` makes user-authored tensor/node names affect rewrite eligibility, violating the architecture-agnostic naming rule. For example, an otherwise eligible `Add -> Exp -> Mul` whose bias initializer is named `algebraic_exp_adjusted_user` is silently left unchanged, while renaming it to `bias` enables folding. Please use explicit namespaced metadata or an architecture-neutral structural condition instead of name-prefix inference.

src/winml/modelkit/optim/pipes/algebraic.py
```diff
@@ -490,6 +831,401 @@ def _channel_preserving_view_output(
...
+
+
+def _has_generated_exp_folding_marker(node: onnx.NodeProto) -> bool:
```
xieofxie: **High:** These names are the only provenance preventing a later serial `Mul` from being folded. A normal ONNX renaming pass can erase them; rerunning optimization on `Exp -> Mul(1e-30) -> Mul(1e30)` then folds the second Mul and recreates the underflow regression (zero becomes about `1e-20`). Please preserve the serial boundary structurally rather than relying on mutable names; metadata can only be supplementary.

src/winml/modelkit/optim/pipes/algebraic.py
```diff
@@ -140,30 +237,50 @@ def _attribute(node: onnx.NodeProto, name: str, default: Any = None) -> Any:
...
+        if initializer.data_location == onnx.TensorProto.EXTERNAL and not initializer.raw_data:
+            return None
+        try:
+            return np.asarray(onnx.numpy_helper.to_array(initializer))
+        except (TypeError, ValueError, RuntimeError, onnx.checker.ValidationError):
+            return None
 
     producer = index.producers.get(name)
-    if producer is None or producer.op_type != "Constant":
+    if producer is None or not _is_standard_onnx_node(producer) or producer.op_type != "Constant":
```
xieofxie, line 254: **Medium:** Unloaded external-data protection is applied to initializers but not to a tensor stored in `Constant.value`. `onnx.numpy_helper.to_array(value)` can raise `onnx.checker.ValidationError`, aborting the optimization instead of skipping the candidate. Please apply the same external-data guard here and catch `RuntimeError` and `onnx.checker.ValidationError` as in the initializer path.

xieofxie (commented):
The three prior findings are resolved. Three new issues remain in generated provenance and external Constant handling.

src/winml/modelkit/optim/pipes/algebraic.py
DingmaomaoBJTU: Fixed in 954a8d5f by removing the generated-name/prefix provenance checks entirely. Exp positive scale folding eligibility is now structural, so user-authored node/tensor names like `algebraic_exp_*` cannot suppress or enable a rewrite. I added `test_user_prefix_names_do_not_block_single_exp_scale_folding` to cover the collision case.

src/winml/modelkit/optim/pipes/algebraic.py
DingmaomaoBJTU: Fixed in 954a8d5f by replacing mutable name provenance with structural serial-boundary detection. If the first post-`Exp` scale `Mul` feeds any downstream standard `Mul`, including through order-preserving views (`Reshape`, `Squeeze`, `Unsqueeze`, `Flatten`, `Identity`), graph-output-observed intermediates, or branch consumers, the pass skips folding that scale. This preserves the separate floating-point evaluation boundaries for serial scale chains, while still folding single-scale cases using `Exp(x + bias) * scale = Exp(x + bias + log(scale))` for finite positive `scale`. Added public `optimize_onnx` regressions for direct serial chains, renamed tensors between optimizer runs, view/Flatten/Identity routes, observed intermediates, and branches.

src/winml/modelkit/optim/pipes/algebraic.py
DingmaomaoBJTU: Fixed in 954a8d5f. `_constant_array` now treats `Constant.value` tensors stored as unloaded external data the same way as external initializers: it returns non-foldable instead of trying to load them. It also catches ONNX loading/validation failures from malformed `Constant.value` tensors. Added a regression that an external-data `Constant.value` is skipped without throwing.

src/winml/modelkit/optim/pipes/algebraic.py
```diff
@@ -6,48 +6,78 @@
...
 from .base import BasePipe, PipeConfig, caps_dict
 
 
+if TYPE_CHECKING:
+    from collections.abc import Iterable
+
+
 ALGEBRAIC_CAPABILITIES: dict[str, Any] = caps_dict(
     algebraic.STATIC_SPLIT_TO_SLICE,
+    misc.GATHER_SLICE_TO_SPLIT_FUSION,
```
xieofxie, line 27: **Medium:** `gather-slice-to-split-fusion` is already registered on `ORTGraphPipe`, so adding the same capability here makes analysis probe each owner independently. A model with both an ORT-eligible Gather route and algebraic sibling Slices can produce two findings with the same name/flag, each containing only one pipe implementation, even though enabling the flag in real optimization runs both. Please give the implementations distinct capabilities or aggregate shared-capability owners into one full-pipeline probe/finding; the new test should also assert that reporting is unique and represents the complete enabled result.

xieofxie (commented):
The three prior findings are resolved. One analysis-reporting issue remains.

src/winml/modelkit/optim/pipes/algebraic.py
DingmaomaoBJTU: Fixed in 00cfe33c by changing analysis for shared default-off capabilities to a single full-pipeline probe instead of per-owner findings. For `gather-slice-to-split-fusion`, `--check-optim` / `iter_optimization_outputs` now reports one finding with `pipe_name` set to `ort_graph+algebraic_rewrite`, and the produced model is the complete result of enabling that public flag across the pipeline.

The regression now asserts uniqueness for this flag and verifies the produced graph contains the combined Slice-to-Split result. This keeps the report aligned with real optimization behavior, where one `--enable-gather-slice-to-split-fusion` flag enables both owners together.

src/winml/modelkit/optim/analysis.py
```diff
@@ -408,6 +448,87 @@ def complete_probe(cap_name: str) -> None:
...
+                            kebab = auto_enable_dependencies(kebab, capabilities)
+                            probe_kwargs = {
+                                capabilities[name].python_name: value
+                                for name, value in kebab.items()
+                                if name in capabilities
+                            }
+                            probe_kwargs.update(optimizer_kwargs)
+
+                            try:
+                                base_out = full_pipeline_baseline()
```
xieofxie: **High:** Assigning the full-pipeline baseline back to `base_out` corrupts the normal per-pipe cursor. At the end of this owner pipe, line 625 sets `current = base_out`, so analysis jumps from the current pipe directly to the output of every downstream default pipe, then runs those downstream pipes again. This can produce incorrect baselines/findings and non-idempotent double processing. Please keep the shared baseline in a separate local (for example, `shared_base_out`) and leave the per-pipe `base_out` unchanged.

xieofxie (commented):
Quick incremental review found one analysis-state regression.

src/winml/modelkit/optim/analysis.py
DingmaomaoBJTU: Fixed in 8412d8dc. The shared full-pipeline baseline now stays in a separate `shared_base_out` local, so the per-pipe `base_out` remains the current pipe's default output and `current = base_out` advances the cursor exactly one pipe at a time.

I also added `test_shared_capability_probe_does_not_advance_pipeline_cursor`, which reproduces the previous cursor corruption by checking that a downstream observer sees the default downstream marker once, not twice, after the shared full-pipeline probe.

xieofxie (approved):
Reviewed the latest head. Previous findings are resolved and the focused final pass found no remaining blockers.

PR #1304, opened 2026-08-12
Fix analyze schema fallback for runtime-specific ops

fangyangci:
## Summary
- handle missing ONNX schema lookups in runtime checker by converting SchemaError into OpUnsupportedError
- keep analysis running and classify those nodes as unknown/unsupported instead of failing the whole command
- reduce noisy per-node unsupported logs by downgrading OpUnsupportedError logging from error to debug
- add regression coverage for missing-schema behavior in runtime checker query helper tests

## Validation
- uv run pytest tests/unit/analyze/core/test_runtime_checker_query_helpers.py
- uv run winml analyze -m "C:\Users\fangyangci\.foundry\cache\models\Microsoft\qwen3-0.6b-cuda-gpu-2\v2\model.onnx" --debug

## Notes
- uv.lock is intentionally excluded from this PR.

### DIFF
--- a/src/winml/modelkit/analyze/core/runtime_checker_query.py
+++ b/src/winml/modelkit/analyze/core/runtime_checker_query.py
@@ -2697,7 +2697,19 @@ class RuntimeCheckerQuery:
         ) as e:
             conditions_ms = _elapsed_ms(conditions_start)
             exception_type = type(e).__name__
-            log_fn = logger.debug if isinstance(e, OpUnsupportedError) else logger.error
+            reason = "optional_input_properties_not_found"
+            log_fn = logger.error
+
+            if isinstance(e, OpUnsupportedError):
+                error_message = str(e)
+                if "has no registered schema for domain" in error_message:
+                    reason = f"schema_not_registered:{node.op_type}"
+                else:
+                    reason = f"unsupported_op:{node.op_type}"
+                log_fn = logger.debug
+            elif isinstance(e, OpLackOfRequiredInformationError):
+                reason = "required_information_missing"
+
             log_fn(
                 "%s caught for op %s (node: %s): %s",
                 exception_type,
@@ -2723,7 +2735,7 @@ class RuntimeCheckerQuery:
                         compile=False,
                         run=False,
                         no_data=True,
-                        reason="optional_input_properties_not_found",
+                        reason=reason,
                         node_tags=node_tags,
                         debug_details=conditions_error_debug_details,
                     ),
--- a/src/winml/modelkit/analyze/core/runtime_checker_query.py
+++ b/src/winml/modelkit/analyze/core/runtime_checker_query.py
@@ -20,6 +20,7 @@ import numpy as np
 import onnx
 import pandas as pd
 from onnx import numpy_helper
+from onnx.defs import SchemaError
 
 from ...onnx import (
     ONNXDomain,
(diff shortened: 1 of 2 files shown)


src/winml/modelkit/analyze/core/runtime_checker_query.py
```diff
@@ -590,7 +591,18 @@ def get_query_conditions_for_node(
...
     conditions = {}
-    schema = domain.get_op_schema(node.op_type, opset_version)
+    try:
+        schema = domain.get_op_schema(node.op_type, opset_version)
+    except SchemaError as e:
+        # Some runtime-specific models emit extension ops in the default ONNX
+        # domain where no standard schema exists (for example
+        # SimplifiedLayerNormalization). Treat this as unsupported instead of
+        # aborting the whole analyze command.
+        raise OpUnsupportedError(
```
DingmaomaoBJTU, line 601: Converting the schema miss to OpUnsupportedError avoids the crash, but run_for_node currently maps every exception of this type to reason='optional_input_properties_not_found'. That makes the emitted UNKNOWN result claim an optional-input problem rather than a missing schema, and the helper-only test does not catch it. Could this path preserve a schema/unsupported-specific reason and add an end-to-end RuntimeCheckerQuery assertion for the returned result?

src/winml/modelkit/analyze/core/runtime_checker_query.py
```diff
@@ -2685,7 +2697,8 @@ def get_pattern_id(is_qdq: bool) -> str:
         ) as e:
             conditions_ms = _elapsed_ms(conditions_start)
             exception_type = type(e).__name__
-            logger.error(
+            log_fn = logger.debug if isinstance(e, OpUnsupportedError) else logger.error
```
DingmaomaoBJTU: This downgrades every OpUnsupportedError, including the existing path where get_runtime_checker_op has no implementation for an otherwise valid schema. With for_debug=False, debug_details is absent and the result uses the unrelated optional-input reason, so the accurate unsupported-op diagnostic disappears entirely. Please either downgrade only the new schema-miss case or preserve the specific unsupported cause in the normal result payload before lowering the log level.

DingmaomaoBJTU (changes requested):
The schema fallback keeps analyze running, but the current exception mapping loses the actual unsupported cause in normal results and suppresses unrelated unsupported-op diagnostics.

src/winml/modelkit/analyze/core/runtime_checker_query.py
fangyangci: Addressed in d2c35ab5. We now preserve specific reasons in run_for_node (schema_not_registered:<op>, unsupported_op:<op>, required_information_missing) instead of mapping everything to optional_input_properties_not_found. Also added end-to-end RuntimeCheckerQuery assertions in tests/unit/analyze/core/test_runtime_checker_query_helpers.py (TestRunForNodeUnsupportedReasons).

src/winml/modelkit/analyze/core/runtime_checker_query.py
fangyangci: Addressed in d2c35ab5. Unsupported reasons are now preserved in normal payloads (for_debug=False) as schema_not_registered:<op> / unsupported_op:<op>. This keeps diagnostics accurate while still reducing noisy unsupported-op log output.

DingmaomaoBJTU (approved):

PR #1305, opened 2026-08-12
Add weight sharing and support for AMD (Vitis AI)

spalne:
Weight sharing for the decoder — Enables the Qwen3 decoder's context (prefill) and iterator (decode) stages to be compiled through a single shared EP context. Since both stages derive from the same weights, they now emit one shared weight .bin referenced by both graphs instead of duplicating the weights per stage reducing on disk bundle size and load-time memory. The grouping logic is execution-provider-agnostic, so it applies automatically to QNN, VitisAI, and OpenVINO.

Adds VitisAI as a supported execution provider for the Qwen3 transformer (context/iterator) stages, alongside the existing QNN. Building with now emits the correct per-stage session_options into genai_config.json, routing inference to the AMD NPU via the waic_target_vaiml_cpp_me VAIML C++ backend.


src/winml/modelkit/session/genai_session.py
```diff
@@ -858,17 +901,23 @@ def _encode_prompt(self, prompt: str | list[int]) -> list[int]:
...
+        generation loops enforce ``max_new_tokens`` as the output-token limit.
 
         The prompt is **not** appended — callers decide whether to time
         ``append_tokens`` separately (see :meth:`generate_timed`).
         """
         og = self._import_og()
         assert self._context_length is not None, "_new_generator called before load()"
-        max_length = min(prompt_len + cfg.max_new_tokens, self._context_length)
+        max_length = (
+            self._context_length
```
DingmaomaoBJTU, line 998: This makes the existing decoder-pipeline test `test_max_length_is_prompt_plus_max_new_tokens` receive `max_length=131072` instead of `69`, so the unit test now fails deterministically. Please update the affected test/behavior contract together with this change.

tests/unit/models/qwen3/test_genai_config.py
```diff
@@ -542,6 +542,25 @@ def test_custom_soc_model(self) -> None:
...
+        stage_map = {s.name: s for s in stages}
+        assert stage_map["embeddings"].session_options is None
+        assert stage_map["lm_head"].session_options is None
+        ctx_opts = stage_map["context"].session_options
+        itr_opts = stage_map["iterator"].session_options
+        assert ctx_opts is not None
+        assert itr_opts is not None
+        vitisai_opts = ctx_opts["provider_options"][0]["vitisai"]
+        assert vitisai_opts["target"] == "waic_target_vaiml_cpp_me"
+        assert vitisai_opts["xmc_runner_config"] == 1
```
DingmaomaoBJTU: `vitisai_stage_session_options()` emits the provider option as the string `"1"`, so this assertion fails (`"1" != 1`); the `no_linear_slice` assertion below has the same mismatch. Please align both expectations with the emitted string values.

DingmaomaoBJTU (commented):
Found two deterministic unit-test regressions in the current PR head.

tests/unit/models/qwen3/test_genai_config.py
DingmaomaoBJTU: Aligned both VitisAI assertions with the emitted string values (`"1"`) for `xmc_runner_config` and `no_linear_slice`.

src/winml/modelkit/session/genai_session.py
DingmaomaoBJTU: Updated the behavior contract and tests: static decoder pipelines now explicitly assert the full configured `context_length`, while dynamic models retain coverage for `prompt_len + max_new_tokens` and context-length capping.

xieofxie (approved):
Reviewed the current changes; no blocking issues found.

PR #1308, opened 2026-08-13
feat: Add automatic perf runtime selection

xieofxie:
## Summary

- add `auto` as the default `perf --runtime` choice
- select `winml-genai` for model folders containing `genai_config.json`
- fall back to `winml` for all other model inputs while preserving explicit runtime choices
- cover automatic selection and CLI dispatch behavior with unit tests

### DIFF
--- a/src/winml/modelkit/commands/perf.py
+++ b/src/winml/modelkit/commands/perf.py
@@ -2288,7 +2288,9 @@ def _autobuild_genai_bundle(
     return bundle_dir, True
 
 
-def _run_genai_runtime(ctx: click.Context, *, console: Console, json_mode: bool) -> None:
+def _run_genai_runtime(
+    ctx: click.Context, *, model: str, console: Console, json_mode: bool
+) -> None:
     """Validate folder input and dispatch to the winml-genai benchmark path.
 
     The genai imports are function-local so ``winml perf --help`` does not pay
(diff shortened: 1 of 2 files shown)


src/winml/modelkit/commands/perf.py
```diff
@@ -2752,6 +2765,7 @@ def perf(
     except Exception as e:
         raise click.ClickException(f"Failed to resolve Hub-hosted ONNX path {model!r}: {e}") from e
     model = hf_model
+    runtime = _resolve_runtime(runtime, model)
```
DingmaomaoBJTU, line 2768: Auto detection uses the normalized model path here, but _run_genai_runtime later reads the original value from ctx.params['model']. For example, normalize_model_arg expands ~/bundle, so this selects winml-genai, then the runner sees the unexpanded ~/bundle, Path(...).is_dir() is false, and it attempts to auto-build it as a model ID. Please pass the normalized model into the GenAI runner or synchronize ctx.params['model'] after normalization, and add a CLI-level regression test.

DingmaomaoBJTU (changes requested):
Auto runtime selection works for ordinary paths, but normalized local paths are not propagated to the GenAI dispatch path, causing valid bundle inputs such as ~/bundle to be misrouted after detection.

src/winml/modelkit/commands/perf.py
xieofxie: Fixed in b8e16db1. `_run_genai_runtime` now receives the normalized model path directly instead of rereading the original `ctx.params["model"]` value. I also added a CLI regression test that normalizes `~/bundle` to a bundle directory and verifies the GenAI path receives that normalized directory.

DingmaomaoBJTU (approved):
Re-reviewed the latest changes. The normalized model path is now propagated directly to the GenAI runtime, resolving the prior dispatch mismatch for paths such as ~/bundle. Explicit runtime choices and model-ID auto-build behavior remain intact, and no new blocking issues were found.

PR #1312, opened 2026-08-14
Report memory metadata in winml sys JSON

KayMKM:
## Summary

- add `schema_version: 1` to the full `winml sys --format json` payload
- add installed physical memory as `memory.physical_total_mib`
- keep the memory probe best-effort: failures or non-positive capacities emit a warning, return `null`, and preserve the rest of the `winml sys` report
- document the versioned physical-memory field and cover successful conversion plus warning-backed fallback behavior with unit and hardware-backed E2E tests

GPU memory is intentionally unchanged in this PR. Existing execution-provider inventory already reports available memory through entries such as `DmlExecutionProvider.entries[].devices[].facts`; this change does not add a second GPU-memory field sourced from `Win32_VideoController.AdapterRAM`.

## Validation

- `uv run --no-sync mypy -p winml.modelkit` (435 source files, no issues)
- `uv run --no-sync ruff check src/winml/modelkit/commands/sys.py tests/e2e/test_sys_e2e.py tests/unit/commands/test_sys_device_info.py`
- `uv run --no-sync pytest tests/unit/commands/test_sys_device_info.py -q` (11 passed)
- `uv run --no-sync pytest tests/e2e/test_sys_e2e.py::TestSysJsonShape tests/e2e/test_sys_e2e.py::TestSysFlagVariations -m e2e -q` (12 passed)

### DIFF
--- a/src/winml/modelkit/commands/sys.py
+++ b/src/winml/modelkit/commands/sys.py
@@ -60,6 +60,8 @@ if TYPE_CHECKING:
 
 logger = logging.getLogger(__name__)
 
+_SYS_JSON_SCHEMA_VERSION = 1
+
 # Rich is imported lazily — Console/Table/Panel only matter at render time,
 # and pulling them at module scope adds ~50 ms to every `winml sys`
 # invocation (including --format json/compact, which never render anything
@@ -201,6 +203,18 @@ def _get_platform_info() -> dict[str, Any]:
     }
 
 
+def _get_memory_info() -> dict[str, int | None]:
+    """Gather installed physical memory without making ``sys`` fail."""
+    try:
+        import psutil
+
+        total_mib = round(psutil.virtual_memory().total / (1024 * 1024))
+    except Exception as exc:
+        logger.warning("Failed to get physical memory details: %s", exc)
+        total_mib = 0
+    return {"physical_total_mib": total_mib or None}
+
+
 def _get_library_versions() -> dict[str, str | None]:
     """Gather versions of key ML libraries."""
     libraries: dict[str, str | None] = {}
@@ -332,8 +346,10 @@ def _gather_system_info(verbose: bool = False) -> dict[str, Any]:
         Dictionary containing all system information
     """
     info = {
+        "schema_version": _SYS_JSON_SCHEMA_VERSION,
         "python": _get_python_info(),
         "platform": _get_platform_info(),
+        "memory": _get_memory_info(),
         "libraries": _get_library_versions(),
         "torch": _get_torch_info(verbose=verbose),
         "backends": {
@@ -586,6 +602,8 @@ def _gather_device_info(
                     "driver": item.driver_version,
                     "manufacturer": item.manufacturer,
                 }
+                if device_label == "GPU":
+                    entry["details"]["dedicated_memory_mib"] = item.vram_mib or None
             elif device_label == "CPU":
                 entry["details"] = {
                     "cores": item.core_count,
--- a/src/winml/modelkit/commands/sys.py
+++ b/src/winml/modelkit/commands/sys.py
@@ -215,6 +215,46 @@ def _get_memory_info() -> dict[str, int | None]:
     return {"physical_total_mib": total_mib or None}
 
 
+def _get_nvidia_memory_by_name() -> dict[str, list[int]]:
+    """Return dedicated memory reported by nvidia-smi, grouped by GPU name."""
+    import csv
+    import shutil
+    import subprocess
+
+    executable = shutil.which("nvidia-smi")
+    if executable is None:
+        return {}
+    try:
+        completed = subprocess.run(  # noqa: S603
+            [
+                executable,
+                "--query-gpu=name,memory.total",
+                "--format=csv,noheader,nounits",
+            ],
+            check=True,
+            capture_output=True,
+            text=True,
+            encoding="utf-8",
+            timeout=10,
+        )
+    except (OSError, subprocess.SubprocessError):
+        logger.debug("nvidia-smi memory query failed", exc_info=True)
+        return {}
+
+    memory_by_name: dict[str, list[int]] = {}
+    for row in csv.reader(completed.stdout.splitlines()):
+        if len(row) != 2:
+            continue
+        name = row[0].strip().casefold()
+        try:
+            memory_mib = int(float(row[1].strip()))
+        except ValueError:
+            continue
+        if name and memory_mib > 0:
+            memory_by_name.setdefault(name, []).append(memory_mib)
+    return memory_by_name
+
+
 def _get_library_versions() -> dict[str, str | None]:
     """Gather versions of key ML libraries."""
     libraries: dict[str, str | None] = {}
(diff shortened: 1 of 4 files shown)


src/winml/modelkit/commands/sys.py
```diff
@@ -332,8 +346,10 @@ def _gather_system_info(verbose: bool = False) -> dict[str, Any]:
         Dictionary containing all system information
     """
     info = {
+        "schema_version": _SYS_JSON_SCHEMA_VERSION,
```
DingmaomaoBJTU, line 351: **Required lint is failing here.** Adding scalar `schema_version` alongside nested dictionaries makes mypy infer `info` as `dict[str, object]`; the later `libs.get(...)` and `info["backends"][...]` accesses then fail type checking (5 errors in the current lint job). Please declare `info: dict[str, Any] = {...}` (or use a typed schema) so the existing nested accesses remain valid.

src/winml/modelkit/commands/sys.py
```diff
@@ -586,6 +602,8 @@ def _gather_device_info(
                     "driver": item.driver_version,
                     "manufacturer": item.manufacturer,
                 }
+                if device_label == "GPU":
+                    entry["details"]["dedicated_memory_mib"] = item.vram_mib or None
```
DingmaomaoBJTU: **This can report incorrect capacity for GPUs above 4 GiB.** `item.vram_mib` is derived from `Win32_VideoController.AdapterRAM`, whose WMI type is `uint32`; its maximum representable value is about 4095 MiB, and many modern adapters therefore appear capped/truncated. Exposing that value as `dedicated_memory_mib` makes an 8/16/24 GiB GPU look like ~4 GiB. Please source capacity from a 64-bit-capable API/property (or omit/null the field when the WMI value is not trustworthy) before publishing it as dedicated memory.

DingmaomaoBJTU (commented):
Reviewed the JSON schema additions and memory probes. The new fields are scoped correctly to the full system/device report, but there are two blocking issues: the current head fails required mypy, and the exposed GPU capacity is sourced from a 32-bit WMI field that truncates/caps modern GPU memory. Details inline.

src/winml/modelkit/commands/sys.py
KayMKM: Fixed in 1778c549 by declaring `info` as `dict[str, Any]`. The required `uv run mypy -p winml.modelkit` check now passes with no issues across 435 source files.

src/winml/modelkit/commands/sys.py
KayMKM: Fixed in 1778c549. `dedicated_memory_mib` no longer uses the 32-bit `Win32_VideoController.AdapterRAM` value. NVIDIA capacity now comes from `nvidia-smi --query-gpu=name,memory.total`; missing commands, failed queries, and unmatched or non-NVIDIA GPUs report `null`. Tests cover a 24 GiB GPU, duplicate names, missing `nvidia-smi`, mismatches, and the non-NVIDIA path.

KayMKM: Follow-up in 06be14ba: after validating on NVIDIA hardware, the additional `dedicated_memory_mib` field and `nvidia-smi` query were removed entirely. Existing EP inventory already reports VRAM through entries such as `DmlExecutionProvider.devices[].facts = ["Memory: 1.9 GB"]`, so this PR now focuses only on the versioned system RAM metadata and avoids duplicating GPU memory.

src/winml/modelkit/commands/sys.py
```diff
@@ -331,9 +345,11 @@ def _gather_system_info(verbose: bool = False) -> dict[str, Any]:
     Returns:
         Dictionary containing all system information
     """
-    info = {
+    info: dict[str, Any] = {
+        "schema_version": _SYS_JSON_SCHEMA_VERSION,
         "python": _get_python_info(),
         "platform": _get_platform_info(),
+        "memory": _get_memory_info(),
```
DingmaomaoBJTU, line 354: The PR summary says the JSON exposes dedicated GPU memory, but this payload still cannot contain it: `_gather_device_info()` only copies `driver` and `manufacturer` for GPUs even though `GPU.vram_mib` is already populated from `AdapterRAM`. Please add the reported value to the GPU `details` (when nonzero) and cover/document that field; otherwise this part of the advertised schema is absent.

DingmaomaoBJTU (commented):
Found one gap between the advertised JSON schema and the implemented payload.

src/winml/modelkit/commands/sys.py
KayMKM: Updated the PR description to match the final scope. It now advertises only `schema_version: 1` and `memory.physical_total_mib`, and explicitly notes that GPU memory is unchanged because the existing execution-provider inventory already reports it through device `facts`. No `dedicated_memory_mib` field is part of this PR.

DingmaomaoBJTU (commented):
Second-pass review found one additional contract mismatch in the E2E coverage.

tests/e2e/test_sys_e2e.py
```diff
@@ -169,6 +171,11 @@ def test_default_json_python_version_is_well_formed(self):
         data = _run_sys_json()
         assert re.match(r"^\d+\.\d+\.\d+", data["python"]["version"])
 
+    def test_default_json_has_versioned_physical_memory(self):
+        data = _run_sys_json()
+        assert data["schema_version"] == 1
+        assert data["memory"]["physical_total_mib"] > 0
```
DingmaomaoBJTU: This assertion rejects the supported fallback introduced by this PR. `_get_memory_info()` intentionally returns `null` when the probe fails, and the documentation explicitly promises that outcome. On a restricted runner, the CLI can behave correctly while this E2E test fails. Please accept either `null` or a positive integer here; the deterministic unit test already verifies the successful MiB conversion.

KayMKM: Fixed in 67de9b5c. The E2E contract now accepts either `null` or a positive integer for `physical_total_mib`, while the deterministic unit test continues to verify the successful bytes-to-MiB conversion.

KayMKM: Follow-up in fcbbc68d: the best-effort fallback is now observable as well as schema-valid. Unit tests assert that probe exceptions and non-positive totals both emit a warning and return `null`, while `winml sys` preserves the rest of the report. The PR description and validation counts have been updated accordingly.

src/winml/modelkit/commands/sys.py
DingmaomaoBJTU: Clarification after reviewing the commit history: please do **not** expose `GPU.vram_mib` directly here—`Win32_VideoController.AdapterRAM` is a 32-bit value and is unreliable for GPUs above 4 GiB. Since the later commit intentionally removed that field, the appropriate fix is to remove the dedicated-GPU-memory claim from the PR summary (or use a trustworthy 64-bit source in a separate implementation).

DingmaomaoBJTU (approved):
Reviewed the latest changes. The fallback contract is covered, all CI checks pass, and I found no remaining blocking issues.

PR #1313, opened 2026-08-17
Fix multi-GPU performance monitoring

KayMKM:
## Summary

- use the LUID from the concrete ORT EP device selected for the inference session
- pass the bound adapter LUID through both standard and module performance monitoring paths
- retain safe monitoring fallback behavior when no concrete adapter can be identified
- add unit coverage for LUID conversion and ambiguous multi-adapter routing

## Context

Daily E2E performance tests disabled accelerator monitoring when one execution provider exposed multiple GPU adapters. EP-wide discovery could not safely select one adapter even though the session had already selected a concrete device. This change uses that session binding as the authoritative adapter identity instead of guessing from the EP.


src/winml/modelkit/session/monitor/_pdh.py
```diff
@@ -674,6 +681,8 @@ def _resolve_adapter(
         """
         if requested == "cpu":
             return None, None
+        if adapter_luid is not None and requested in ACCELERATOR_DEVICE_TYPES:
```
DingmaomaoBJTU: `adapter_luid` is documented as taking precedence over discovery, but with the public default `device="auto"` this branch ignores it and falls through to NPU-first discovery below. That can monitor a different accelerator (or raise an ambiguity error) even though the caller supplied the exact adapter. Please honor the bound LUID for `auto` as well—e.g. resolve its device kind before constructing the poller—and add coverage for `HWMonitor(adapter_luid=..., device="auto")`.

DingmaomaoBJTU (changes requested):
Found one correctness issue in the new adapter-binding path.

src/winml/modelkit/session/monitor/_pdh.py
KayMKM: Fixed in `77500b93`. With `device="auto"`, the poller now classifies the exact bound LUID from PDH instead of falling through to EP-wide discovery, with regression coverage through `HWMonitor`.

xieofxie (approved):
Reviewed the current revision. No actionable issues found; the prior blocking thread is resolved and all checks pass.

src/winml/modelkit/session/monitor/_pdh.py
```diff
@@ -674,6 +681,17 @@ def _resolve_adapter(
...
+            if requested in ACCELERATOR_DEVICE_TYPES:
+                return adapter_luid, requested
+            try:
+                adapter_info = enumerate_adapters().get(adapter_luid)
+            except RuntimeError:
+                logger.debug("Failed to classify bound adapter LUID", exc_info=True)
+                return None, None
+            if adapter_info is None:
+                return None, None
+            return adapter_luid, ("npu" if adapter_info.is_npu else "gpu")
```
DingmaomaoBJTU: This classification can still route an authoritative bound LUID to the wrong query. `AdapterInfo.is_npu` treats any adapter exposing `3D` as a GPU, while `build_npu_query()` explicitly includes `3D` for OpenVINO NPUs; conversely, an OpenVINO GPU exposing only Compute engines can look like an NPU. The added test only stubs an unambiguous GPU, so it does not cover either supported case. Please propagate the concrete device kind alongside the bound LUID (the caller that selected the EP device already knows it) instead of re-inferring it from PDH engine fingerprints.

DingmaomaoBJTU (changes requested):
The bound LUID is now honored, but its device kind is inferred with a heuristic that is not reliable for the providers this monitor supports.

src/winml/modelkit/session/monitor/_pdh.py
KayMKM: Fixed in `2d0e53df`. The concrete ORT EP-device binding now propagates both its LUID and device kind, so `device="auto"` never reclassifies OpenVINO GPU/NPU devices from PDH engine fingerprints; the regression test covers both device kinds.

src/winml/modelkit/commands/perf.py
```diff
@@ -304,6 +304,22 @@ def _pre_bench_kwargs_from_ep_device(
...
+def _get_ep_device_binding(
+    ep_device: WinMLEPDevice | None,
+) -> tuple[str | None, str | None]:
+    """Return the LUID and device kind from a concrete EP-device binding."""
+    if ep_device is None:
+        return None, None
+
+    from ..sysinfo import get_ep_device_luid
+
+    luid = get_ep_device_luid(ep_device.device.ort_handle)
```
DingmaomaoBJTU: `ep_device` is not necessarily the adapter the session ultimately uses. ORT applies the selected device's default EP options first and then lets user options override them; DirectML's factory reads the resulting `device_id` and creates that DXGI adapter rather than honoring the supplied device handle. Therefore `--ep-options device_id=1` can run inference on GPU 1 while this records the original binding's LUID, causing utilization and VRAM metrics to come from the wrong GPU. Can we derive the monitored adapter from the effective provider options, or avoid this shortcut when an option overrides device selection?

KayMKM: Fixed in `c7720b4a`. The monitor now overlays provider options against the selected `OrtEpDevice.ep_options`, follows the uniquely matching EP device when options such as `device_id` redirect execution, and disables the bound-LUID shortcut when no unique match exists.

src/winml/modelkit/commands/perf.py
```diff
@@ -304,6 +304,58 @@ def _pre_bench_kwargs_from_ep_device(
...
+            key for _, options in candidate_options for key in options
+        }
+        selected_options = {
+            str(key): str(value)
+            for key, value in device.ort_handle.ep_options.items()
+        }
+        device_overrides = {
+            key: str(value)
+            for key, value in provider_options.items()
+            if key in advertised_keys and selected_options.get(key) != str(value)
```
DingmaomaoBJTU, line 354: Filtering overrides to `advertised_keys` misses provider-level device selectors that are not present in an `OrtEpDevice`'s option map. In particular, QNN gives `backend_type`/`backend_path` precedence over deriving the backend from the bound device; with an NPU `ep_device` plus `--ep-options backend_type=gpu`, this set is empty and the helper returns the NPU LUID even though QNN runs its GPU backend. Please treat backend-changing options conservatively (resolve the corresponding candidate when possible, otherwise return `(None, None)` so discovery is used) and add coverage for a QNN device whose advertised options are empty but whose user options select another backend.

DingmaomaoBJTU (changes requested):
The latest provider-option fix handles options advertised by EP devices, but QNN backend overrides can still redirect inference while telemetry remains bound to the original adapter.

src/winml/modelkit/commands/perf.py
KayMKM: Fixed in `e4ea49de`. Backend selectors are now resolved from the shared EP/device catalog even when `OrtEpDevice.ep_options` is empty; the monitor follows a unique matching device, falls back to discovery for the selected device kind when needed, and disables accelerator telemetry for unknown or conflicting selectors.

src/winml/modelkit/commands/perf.py
```diff
@@ -304,6 +304,117 @@ def _pre_bench_kwargs_from_ep_device(
...
+        provider_options,
+    )
+    has_provider_selector, provider_device = _get_provider_selected_device(
+        ep_device,
+        provider_options,
+    )
+    if has_provider_selector and provider_device is None:
+        return "cpu", None, None
+
+    monitor_device = adapter_device or requested_device
```
DingmaomaoBJTU, line 410: When a provider selector changes the device kind but no matching `OrtEpDevice` exists, `_get_ep_device_binding()` returns `(None, provider_device)`, so this continues with (for example) `monitor_device="gpu"` and no authoritative LUID. `HWMonitor` then calls `resolve_adapter_luid("gpu", ep_name="QNNExecutionProvider")`; with no QNN GPU device that falls back to `discover_gpu_luid()`, which explicitly chooses the first GPU on multi-GPU hosts. A QNN `backend_type=gpu` session can therefore run on the Qualcomm GPU while telemetry/VRAM comes from a different discrete or external GPU. Please only use the PDH fallback when the matching adapter is unambiguous; otherwise disable adapter-specific monitoring rather than reporting a potentially unrelated device.

src/winml/modelkit/session/ep_device.py
```diff
@@ -430,6 +439,39 @@ def lookup_device_spec(ep: str, device: str) -> EPDeviceSpec | None:
...
+        actual = str(options[key]).replace("\\", "/").rsplit("/", 1)[-1]
+        key_matches = {
+            spec.device
+            for spec in specs
+            if key in spec.provider_option_hints
+            and actual.casefold() == spec.provider_option_hints[key].casefold()
+        }
+        if not key_matches:
+            return True, None
+        matches = key_matches if matches is None else matches.intersection(key_matches)
```
DingmaomaoBJTU, line 470: This assignment currently fails the required mypy check: `key_matches` is inferred as `set[Literal["npu", "gpu", "cpu"]] | set[str]`, which is not assignable to invariant `set[str] | None`. Please give `key_matches` an explicit `set[str]` annotation (or otherwise normalize its inferred type). The PR lint job is failing on this line.

DingmaomaoBJTU (changes requested):
The original QNN backend override is handled, but one ambiguous multi-GPU fallback can still report the wrong adapter, and the current HEAD fails the required mypy check.

src/winml/modelkit/commands/perf.py
KayMKM: Fixed in `0412b59e`. When provider selectors redirect execution but no exact adapter LUID is available, both hardware monitoring and VRAM tracking now disable adapter-specific discovery instead of falling back to an arbitrary GPU.

src/winml/modelkit/session/ep_device.py
KayMKM: Fixed in `0412b59e` by explicitly typing `key_matches` as `set[str]`. The full required `uv run mypy -p winml.modelkit` check now passes.

DingmaomaoBJTU (approved):

PR #1314, opened 2026-08-17
Hide EP installation status from sys JSON output

xieofxie:
## Summary
- suppress EP installation notices and download progress while `winml sys` gathers JSON output
- preserve existing installation status for human-readable formats
- add regression coverage for CLI JSON mode and provider readiness


DingmaomaoBJTU (approved):
Reviewed the structured-output suppression path and its call sites. The scoped ContextVar is restored safely, JSON gathering suppresses both the notice and progress bar, and human-readable formats retain the existing behavior. No blocking issues found.

PR #1316, opened 2026-08-19
Use structured perf results in E2E evaluation

KayMKM:
## Summary
- run `winml perf` with `--output` and persist the structured report under `perf.result`
- preserve structured results per sub-model and fail runs that exit successfully without a valid JSON report
- document the result contract and add focused unit coverage

## Validation
- `uv run --frozen ruff check --fix scripts/e2e_eval/run_eval.py scripts/e2e_eval/utils/reporter.py tests/unit/eval/test_run_eval_script.py`
- `uv run --frozen pytest tests/unit/eval/test_run_eval_script.py -q` (148 passed)
- `run_eval.py --eval-type both` smoke run with `microsoft/resnet-18` on CPU and 5 accuracy samples (perf PASS, accuracy PASS)

### DIFF
--- a/scripts/e2e_eval/run_eval.py
+++ b/scripts/e2e_eval/run_eval.py
@@ -48,6 +48,7 @@ import functools
 import hashlib
 import json
 import logging
+import math
 import os
 import platform
 import re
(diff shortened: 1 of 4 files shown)


xieofxie (changes requested):
The op-trace artifact is lost when structured perf output is enabled. Current `winml perf` writes `result_op_trace.json` beside the requested `--output` file and prints `Op-trace JSON:`, but `_copy_op_trace` only recognizes the obsolete `Op-trace saved to:` marker. Because the source is inside `TemporaryDirectory`, the trace is then deleted at context exit and is never copied to the model result directory. Please copy the deterministic sibling trace path (or recognize the current output contract) before cleanup, and add coverage for op tracing with structured output.

scripts/e2e_eval/run_eval.py
```diff
@@ -1738,6 +1738,29 @@ def _record(
...
+    try:
+        result = json.loads(output_path.read_text(encoding="utf-8"))
+    except (OSError, json.JSONDecodeError) as e:
+        message = f"Failed to load structured winml perf output: {e}"
+        proc["stderr"] = "\n".join(filter(None, [proc.get("stderr", ""), message]))
+        proc["exit_code"] = 1
+        proc["error_summary"] = "invalid structured perf output"
+        return None
+
+    if not isinstance(result, dict):
```
DingmaomaoBJTU, line 1818: Checking only that the root is an object still accepts malformed reports such as `{}` or `{"latency_ms": []}` as successful perf runs. Those values are then persisted under `perf.result`, where downstream reporting cannot consume them despite the run being marked passed. Please validate the supported single-model/composite `winml perf` shape and required field types before preserving exit code 0.

DingmaomaoBJTU (commented):
Found one structured-output contract gap.

xieofxie (approved):
Re-reviewed the updated revision. The structured output is now validated against the supported perf schemas, and op-trace artifacts are copied from the deterministic sibling path before temporary cleanup. No blocking issues remain.

DingmaomaoBJTU (approved):
Re-reviewed the latest changes. The structured perf validation issue is addressed, CI passes, and I found no remaining blocking issues.

PR #1317, opened 2026-08-19
feat(optim): preserve fast QNN grouped-conv regions

DingmaomaoBJTU:
## Summary

Add an opt-in `trim-split-grouped-conv` optimizer surgery for a strictly guarded static grouped Conv followed by a tail Slice. The surgery trims kernel taps proven to multiply only zero padding, decomposes an even grouped Conv into two equal group-aligned branches, and records a versioned quantization-region hint. Static quantization runs through the existing generic QDQ path, then consumes only the generated branches' output Q/DQ pairs so Concat remains inside one region. The compiler is unchanged.

## Root cause

QNN QHAS attributed 99.94% of accelerator time to the positional grouped Conv, with a 20.58-second dominant path. Controlled variants isolated the mechanism: separate scale names measured 40.913 ms; independent branch weight encodings measured 40.025 ms; branch input QDQ alone measured 40.264 ms; branch output QDQ alone measured 39.558 ms; complete branch-local DQ-Conv-Q units regressed to 7,073.188 ms. Scale identity and values are not causal; QDQ node-unit boundaries are.

A broad pre-quantization policy using ORT `OpTypesToExcludeOutputQuantization` also worked, but changed unrelated Conv boundaries and measured 52.037 ms. The versioned hint keeps the policy local to optimizer-generated regions without introducing a QNN compiler option.

## Safety

The capability is disabled by default and appears as `optim.trim_split_grouped_conv` in build config JSON; it changes cache identity. The FP32 matcher requires static 1D shapes, explicit pads, unit stride and dilation, constant weights and optional bias, an even group count, an exact prefix-retaining tail Slice, exclusive topology, and a strict kernel reduction proven by reachability math. Unsupported graphs are no-ops. Source initializers are removed only when no top-level or nested-subgraph reference remains.

The hint consumer validates the complete region before cloning or mutation. It requires exact Conv -> Q -> DQ -> Concat data-input positions, exclusive consumers, no graph-output escape, unique names and roles, standard Conv/Concat domains, and standard or ORT `com.microsoft` QDQ domains. StaticPass stages quantization and all post-processing, path-validates the external-data result, and publishes the model and sidecar transactionally.

## Validation

- 443 focused and adjacent tests passed in separate repository-supported test processes: 116 optimizer/quantization/metadata, 115 optimizer/config, 175 build-config, and 37 cache tests.
- Ruff check and format check passed; touched-file editor diagnostics were clean.
- FP32 target rewrite, 12 samples: logits mean/min cosine 0.9999999999999966/0.9999999999999876, max absolute difference 1.19e-7.
- Final W8A16 CPU, 12 samples: logits mean/min cosine 0.999652/0.998877, max absolute difference 0.052961.
- Final W8A16 QNN, 12 samples: logits mean/min cosine 0.999238/0.998084, max absolute difference 0.045169.
- Compiled graph: one EPContext; all seven unrelated feature-extractor Conv output QDQ boundaries preserved.
- Final QNN performance, 100 iterations: p50 42.606 ms, p90 47.278 ms, versus the original 20,786.384 ms p50.
- Balanced A/B, B/A, A/B, B/A confirmation: every gain positive, mean paired gain 99.7947%, Student-t 95% CI [99.7851%, 99.8043%].
- Independent final review found no remaining Critical or Important issues.

## Rollback

Disable `trim-split-grouped-conv`. Unhinted models retain the existing optimization, quantization, and compile behavior.

### DIFF
--- a/src/winml/modelkit/optim/grouped_conv.py
+++ b/src/winml/modelkit/optim/grouped_conv.py
@@ -132,6 +132,7 @@ def _match_plan(
     *,
     consumers: dict[str, list[NodeProto]],
     graph_outputs: set[str],
+    graph_inputs: set[str],
     initializers: dict[str, TensorProto],
     shapes: dict[str, tuple[int | None, ...]],
     used_names: set[str],
@@ -150,6 +151,8 @@ def _match_plan(
     if tail_slice.op_type != "Slice" or tail_slice.domain not in ("", "ai.onnx"):
         return None
     if len(tail_slice.input) != 5 or len(tail_slice.output) != 1:
+        return None
+    if graph_inputs.intersection((*conv.input[1:], *tail_slice.input[1:])):
         return None
 
     conv_attributes = {attribute.name: attribute for attribute in conv.attribute}
@@ -363,6 +366,7 @@ def trim_split_grouped_convs(
     initializers = {initializer.name: initializer for initializer in graph.initializer}
     consumers = _consumer_nodes(graph)
     graph_outputs = {value.name for value in graph.output}
+    graph_inputs = {value.name for value in graph.input}
     shapes = _static_shapes(model)
     used_names = {
         name for node in graph.node for name in (*node.input, *node.output, node.name) if name
@@ -378,6 +382,7 @@ def trim_split_grouped_convs(
                 node,
                 consumers=consumers,
                 graph_outputs=graph_outputs,
+                graph_inputs=graph_inputs,
                 initializers=initializers,
                 shapes=shapes,
                 used_names=used_names,
--- a/src/winml/modelkit/optim/grouped_conv.py
+++ b/src/winml/modelkit/optim/grouped_conv.py
@@ -12,8 +12,8 @@ from dataclasses import dataclass
 
 import numpy as np
 import onnx
-from onnx import helper, numpy_helper
 
+from ..onnx import get_captured_tensor_names
 from ..quant.hints import (
     QUANTIZATION_REGION_HINT_KEY,
     STANDARD_ONNX_DOMAINS,
@@ -35,6 +35,7 @@ class _RewritePlan:
     kernel_end: int
     pad_begin: int
     pad_end: int
+    group: int
     branch_names: tuple[str, str]
     split_name: str
     concat_name: str
@@ -44,10 +45,6 @@ class _RewritePlan:
     bias_names: tuple[str, str] | None
 
 
-def _attributes(node: onnx.NodeProto) -> dict[str, object]:
-    return {attribute.name: helper.get_attribute_value(attribute) for attribute in node.attribute}
-
-
 def _constant_vector(
     name: str,
     initializers: dict[str, onnx.TensorProto],
@@ -55,7 +52,7 @@ def _constant_vector(
     initializer = initializers.get(name)
     if initializer is None:
         return None
-    values = np.asarray(numpy_helper.to_array(initializer))
+    values = np.asarray(onnx.numpy_helper.to_array(initializer))
     if values.ndim != 1:
         return None
     return [int(value) for value in values]
@@ -89,7 +86,7 @@ def _replace_attribute(node: onnx.NodeProto, name: str, value: object) -> None:
     attributes = [
         copy.deepcopy(attribute) for attribute in node.attribute if attribute.name != name
     ]
-    attributes.append(helper.make_attribute(name, value))
+    attributes.append(onnx.helper.make_attribute(name, value))
     del node.attribute[:]
     node.attribute.extend(attributes)
 
@@ -105,6 +102,21 @@ def _referenced_tensor_names(graph: onnx.GraphProto) -> set[str]:
                 for subgraph in attribute.graphs:
                     names.update(_referenced_tensor_names(subgraph))
     return names
+
+
+def _consumer_nodes(graph: onnx.GraphProto) -> dict[str, list[onnx.NodeProto]]:
+    consumers: dict[str, list[onnx.NodeProto]] = {}
+    for node in graph.node:
+        consumed_names = {input_name for input_name in node.input if input_name}
+        for attribute in node.attribute:
+            if attribute.type == onnx.AttributeProto.GRAPH:
+                consumed_names.update(get_captured_tensor_names(attribute.g))
+            elif attribute.type == onnx.AttributeProto.GRAPHS:
+                for nested_graph in attribute.graphs:
+                    consumed_names.update(get_captured_tensor_names(nested_graph))
+        for input_name in consumed_names:
+            consumers.setdefault(input_name, []).append(node)
+    return consumers
 
 
 def _match_plan(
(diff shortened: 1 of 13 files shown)


src/winml/modelkit/optim/grouped_conv.py
```diff
@@ -0,0 +1,489 @@
...
+"""Conservative static grouped-Conv graph surgery."""
+
+from __future__ import annotations
+
+import copy
+import logging
+from dataclasses import dataclass
+
+import numpy as np
+import onnx
```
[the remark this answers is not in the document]
DingmaomaoBJTU: Fixed in `e6691b3b` by replacing the third-party `import onnx` with explicit symbol imports while retaining the internal ONNX facade import. CodeQL and all PR checks now pass.

src/winml/modelkit/quant/hints.py
```diff
@@ -0,0 +1,310 @@
...
+# --------------------------------------------------------------------------
+"""Consume optimizer-authored quantization region hints."""
+
+from __future__ import annotations
+
+import copy
+import json
+from dataclasses import dataclass
+
+import onnx
```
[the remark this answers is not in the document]
DingmaomaoBJTU: Fixed in `e6691b3b` by replacing the third-party `import onnx` with explicit symbol imports while retaining the internal ONNX facade import. CodeQL and all PR checks now pass.

src/winml/modelkit/quant/passes/static.py
```diff
@@ -148,52 +181,55 @@ def run(
...
+            prefix=f".{output_path.stem}-",
+            dir=output_path.parent,
+        ) as staging_directory:
+            staged_output = Path(staging_directory) / output_path.name
+            original_cwd = Path.cwd()
+            try:
+                os.chdir(staged_output.parent)
+                quantize(
+                    model_input=input_model,
+                    model_output=str(staged_output.resolve()),
```
xieofxie, line 200: **High:** Resolve the staged path before changing the working directory. When `output_path` is relative, `TemporaryDirectory(dir=output_path.parent)` returns a relative staging path; after `chdir(staged_output.parent)`, this `resolve()` duplicates the staging component (for example `.out-abc\.out-abc\out.onnx`). The documented/default `quantize_onnx("model.onnx")` path therefore sends ORT an output whose parent does not exist. The new tests only exercise absolute `tmp_path` outputs.

src/winml/modelkit/optim/grouped_conv.py
```diff
@@ -0,0 +1,492 @@
...
+    if not plans:
+        return model
+
+    replacements: dict[int, list[NodeProto]] = {}
+    removed_node_ids: set[int] = set()
+    new_initializers: list[TensorProto] = []
+    regions: list[QuantizationRegion] = []
+    removable_initializers: set[str] = set()
+    for plan in plans:
+        weights = np.asarray(numpy_helper.to_array(initializers[plan.conv.input[1]]))
```
xieofxie, line 412: **High:** Skip the rewrite when any consumed weight, bias, or Slice-parameter initializer is also a graph input. ONNX allows callers to override such initializers, but these new branch initializers bake the default values and leave the caller-provided input unused, silently changing inference semantics. This also makes matching IR <= 3 models invalid because the newly added initializers are not added to `graph.input`.

src/winml/modelkit/quant/hints.py
```diff
@@ -0,0 +1,310 @@
...
+                    f"Hinted branch {branch_name!r} output is a graph output"
+                )
+
+            q_index = _one_consumer(
+                consumers,
+                branch_output,
+                context=f"Hinted branch {branch_name!r} output",
+            )
+            q_node = nodes[q_index]
+            if q_node.op_type != "QuantizeLinear":
```
xieofxie, line 211: **Medium:** A valid quantization configuration can leave this branch output feeding the hinted Concat directly (for example `op_types_to_quantize=["MatMul"]` or excluding the generated Conv). `StaticPass` still always consumes the hint, so successful ORT quantization then fails here and publishes no model. Please accept an exclusively direct branch-to-Concat edge as already canonical (and handle partially quantized branches), while still rejecting malformed topology.

xieofxie (changes requested):
Found three correctness regressions that should be addressed before merge.

src/winml/modelkit/quant/hints.py
DingmaomaoBJTU: Fixed in `81c578b3`. Each hinted branch is now validated independently as either a direct branch-to-Concat edge or an exact Q/DQ chain; partially quantized regions remove only the Q/DQ pair that exists. Fanout and malformed topology remain rejected. Added unit coverage for both partial directions and fully direct regions, plus a StaticPass integration regression. All PR checks pass.

src/winml/modelkit/quant/passes/static.py
DingmaomaoBJTU: Fixed in `81c578b3`. The staged output path is now resolved before changing the working directory, so relative output paths do not duplicate the staging component on Python 3.11/Windows. Added a regression using a relative `quantized.onnx` output. All PR checks pass.

src/winml/modelkit/optim/grouped_conv.py
DingmaomaoBJTU: Fixed in `81c578b3`. The matcher now skips the rewrite whenever any consumed weight, bias, or Slice-control value is also a graph input. Regressions cover all six initializer roles plus legacy IR initializer inputs. All PR checks pass.

src/winml/modelkit/optim/grouped_conv.py
```diff
@@ -0,0 +1,497 @@
...
+            bias = np.asarray(numpy_helper.to_array(initializers[plan.conv.input[2]]))
+            for index, values in enumerate(np.split(bias, [output_midpoint], axis=0)):
+                new_initializers.append(
+                    numpy_helper.from_array(
+                        np.ascontiguousarray(values),
+                        plan.bias_names[index],
+                    )
+                )
+
+        split = helper.make_node(
```
xieofxie, line 433: **High:** This emits an invalid `Split` for models using the default ONNX opset 18 or newer. Since Split-18, the node must provide either the optional `split` input or `num_outputs`; this generated two-output node provides neither. `onnx.checker.check_model(..., full_check=True)` rejects an otherwise valid opset-18 matched model, while the current tests only construct opset 17 models. Please emit `num_outputs=2` for the applicable opsets (while preserving compatibility with older schemas) and add an opset-18 checker/inference regression.

src/winml/modelkit/quant/passes/static.py
```diff
@@ -22,6 +25,36 @@
 logger = logging.getLogger(__name__)
 
 
+def _publish_staged_model(staged_path: Path, output_path: Path) -> None:
+    """Publish a validated staged ONNX model while preserving prior output on failure."""
+    staged_sidecar = staged_path.parent / f"{staged_path.name}.data"
+    output_sidecar = output_path.parent / f"{output_path.name}.data"
+    backup_model = staged_path.parent / "previous-output.onnx"
```
xieofxie: **High:** These fixed backup names can alias legal staged artifact names. If the requested output is `previous-output.onnx`, `backup_model == staged_path`, so backing up an existing output overwrites the newly validated model and the pass silently republishes the old file. With output `previous-output.onnx.data`, an existing sidecar makes `backup_sidecar == staged_path`, allowing sidecar bytes to overwrite the staged ONNX before publication. Please use collision-free backup paths (for example, a dedicated nested backup directory) and cover both filename collisions.

src/winml/modelkit/quant/passes/static.py
```diff
@@ -22,6 +25,36 @@
...
+    backup_sidecar = staged_path.parent / "previous-output.onnx.data"
+    backed_up_model = False
+    backed_up_sidecar = False
+    published_sidecar = False
+    try:
+        if output_path.exists():
```
xieofxie: **High:** `Path.exists()` is also true for directories. If either the output model path or its `.data` sidecar path is an existing directory, `replace()` moves that directory into the `TemporaryDirectory`; after successful publication, temporary-directory cleanup recursively deletes the moved directory and all of its contents. Reject non-file destinations before moving anything and add regressions proving such directories remain untouched.

xieofxie (changes requested):
The previous three findings are fixed, but the updated full review found three additional correctness and data-loss edge cases that should be addressed before merge.

src/winml/modelkit/optim/grouped_conv.py
DingmaomaoBJTU: Fixed in `3098f8a2`. The rewrite now reads the standard ONNX opset and adds `num_outputs=2` only for opset 18+, while retaining the opset 17 form. The regression runs full ONNX checking and CPU inference for both opsets and verifies identical outputs. All PR checks pass.

src/winml/modelkit/quant/passes/static.py
DingmaomaoBJTU: Fixed in `3098f8a2`. Both the model and sidecar destinations are validated as files before any move; existing directories and other non-files are rejected. Regressions verify model/sidecar directories and nested sentinel contents remain untouched. All PR checks pass.

src/winml/modelkit/quant/passes/static.py
DingmaomaoBJTU: Fixed in `3098f8a2`. Publication backups now live in a unique nested temporary directory, so legal staged basenames cannot alias them. Regressions cover both `previous-output.onnx` and `previous-output.onnx.data` and verify that the new model and sidecar are published. All PR checks pass.

xieofxie (approved):
Re-reviewed the complete latest change. The previously reported issues are fixed with focused regression coverage, and no blocking findings remain.

PR #1320, opened 2026-08-20
fix(serve): block cross-origin CLI requests

xieofxie:
## Summary
- replace wildcard CORS with same-origin request enforcement
- protect both CLI-only and inference serve applications
- add regression coverage for hostile preflight and POST requests plus valid same-origin dispatch


xieofxie:
in the wrong direction

PR #1321, opened 2026-08-20
fix(serve): harden CLI API security boundaries

timenick:
## Summary

- replace wildcard CORS with same-origin request protection for WinML serve apps
- reject HTTP attempts to enable remote code and enforce the policy at model/config loading boundaries
- centralize CLI command allowlist validation for both server modes
- add regression coverage for origin parsing, DNS rebinding, malformed headers, nested configuration, and direct-CLI compatibility

## Validation

- 23 security regression tests passed
- 44 serve tests passed
- 512 loader tests passed
- 191 build/config/direct-CLI tests passed
- Ruff and git diff checks passed
- the full suite was attempted; an unrelated OpenVINO NPU integration test timed out after its worker process pool crashed

### DIFF
--- a/src/winml/modelkit/serve/_security.py
+++ b/src/winml/modelkit/serve/_security.py
@@ -20,14 +20,6 @@ if TYPE_CHECKING:
     from starlette.types import ASGIApp, Receive, Scope, Send
 
 
-_CLI_API_LOOPBACK_ONLY_ROUTES: dict[str, set[str] | None] = {"/v1/cli": None}
-_MODEL_API_LOOPBACK_ONLY_ROUTES: dict[str, set[str] | None] = {
-    "/v1/cli": None,
-    "/v1/ep": {"POST"},
-    "/v1/models": {"POST", "DELETE"},
-}
-
-
 class SameOriginMiddleware:
     """Enforce browser-origin and local-only route boundaries."""
 
--- a/src/winml/modelkit/serve/_security.py
+++ b/src/winml/modelkit/serve/_security.py
@@ -1,0 +1,114 @@
+# -------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+# --------------------------------------------------------------------------
+"""Security middleware for local WinML HTTP servers."""
+
+from __future__ import annotations
+
+import ipaddress
+from typing import TYPE_CHECKING
+from urllib.parse import urlsplit
+
+from starlette.responses import PlainTextResponse
+
+
+if TYPE_CHECKING:
+    from starlette.types import ASGIApp, Receive, Scope, Send
+
+
+class SameOriginMiddleware:
+    """Reject browser requests whose origin differs from the target server."""
+
+    def __init__(self, app: ASGIApp) -> None:
+        self.app = app
+
+    async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
+        if scope["type"] == "http" and not _is_same_origin(scope):
+            response = PlainTextResponse("Cross-origin requests are not allowed.", status_code=403)
+            await response(scope, receive, send)
+            return
+        await self.app(scope, receive, send)
+
+
+def _is_same_origin(scope: Scope) -> bool:
+    headers = {key.lower(): value for key, value in scope.get("headers", [])}
+    origin_header = headers.get(b"origin")
+    if origin_header is None:
+        return True
+
+    try:
+        origin = urlsplit(origin_header.decode("ascii"))
+    except (UnicodeDecodeError, ValueError):
+        return False
+    if origin.scheme not in {"http", "https"} or origin.path or origin.query or origin.fragment:
+        return False
+    if origin.username is not None or origin.password is not None:
+        return False
+    if not _is_allowed_http_host(origin.hostname):
+        return False
+    try:
+        origin_port = _effective_port(origin.port, origin.scheme)
+    except ValueError:
+        return False
+
+    target_scheme = scope.get("scheme", "http").lower()
+    if target_scheme not in {"http", "https"}:
+        return False
+    host_header = headers.get(b"host")
+    if host_header is None:
+        server = scope.get("server")
+        if server is None:
+            return False
+        target_host, target_port = server
+        target_host = _normalize_host(target_host)
+        target_port = _effective_port(target_port, target_scheme)
+    else:
+        try:
+            target = urlsplit(f"{target_scheme}://{host_header.decode('ascii')}")
+        except (UnicodeDecodeError, ValueError):
+            return False
+        if (
+            target.path
+            or target.query
+            or target.fragment
+            or target.username is not None
+            or target.password is not None
+        ):
+            return False
+        target_host = _normalize_host(target.hostname)
+        try:
+            target_port = _effective_port(target.port, target_scheme)
+        except ValueError:
+            return False
+
+    return (
+        origin.scheme == target_scheme
+        and _normalize_host(origin.hostname) == target_host
+        and origin_port == target_port
+    )
+
+
+def _default_port(scheme: str) -> int:
+    return 443 if scheme == "https" else 80
+
+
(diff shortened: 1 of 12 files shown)


src/winml/modelkit/serve/_security.py
```diff
@@ -0,0 +1,114 @@
...
+
+
+class SameOriginMiddleware:
+    """Reject browser requests whose origin differs from the target server."""
+
+    def __init__(self, app: ASGIApp) -> None:
+        self.app = app
+
+    async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
+        if scope["type"] == "http" and not _is_same_origin(scope):
```
DingmaomaoBJTU: This makes every cross-origin server selection fail, but the bundled `/demo` UI still exposes an editable **Server URL** and assigns it to `serverUrl` before fetching `/v1/health` and all subsequent endpoints. For example, a demo served at `localhost:8000` can no longer connect to `127.0.0.1:9000` (or a frontend dev server to WinML), and the UI only reports “Cannot reach server.” Please make the UI consistent with the new security boundary—prefer removing/locking the selector to `window.location.origin`; alternatively, if cross-origin local development remains supported, add an explicit trusted-origin allowlist rather than restoring wildcard CORS. Add a regression test for the intended demo behavior.

DingmaomaoBJTU (changes requested):
The security hardening is generally sound, but strict same-origin enforcement currently conflicts with a shipped demo workflow.

src/winml/modelkit/serve/_security.py
```diff
@@ -0,0 +1,114 @@
...
+            await response(scope, receive, send)
+            return
+        await self.app(scope, receive, send)
+
+
+def _is_same_origin(scope: Scope) -> bool:
+    headers = {key.lower(): value for key, value in scope.get("headers", [])}
+    origin_header = headers.get(b"origin")
+    if origin_header is None:
+        return True
```
xieofxie, line 89: Origin checking only mitigates browser CSRF; accepting every request without an Origin header leaves the HTTP CLI unauthenticated. This is exploitable when the documented --host 0.0.0.0 mode is used: an origin-less client can call /v1/cli/compile with compiler=qairt and an attacker-controlled qnn_sdk_root, after which nsure_venv executes <root>\venv-winml\Scripts\python.exe. The remote-code ContextVar does not cover this path. Please either restrict CLI/management routes to loopback, require an unguessable per-launch token for all callers, or remove executable/toolchain-capable commands and path arguments from the HTTP surface.

xieofxie (changes requested):
The same-origin change does not secure origin-less HTTP clients, leaving a remote command-execution path when the server is exposed beyond loopback.

src/winml/modelkit/serve/_security.py
timenick: Fixed in de8de41c. The bundled demo now pins requests to window.location.origin, makes the Server URL field read-only, and applies a connect-src self CSP. The regression test exercises the served /demo response and verifies the CSP and locked selector.

src/winml/modelkit/serve/_security.py
timenick: Fixed in de8de41c. Privileged routes now enforce the actual ASGI client address: all /v1/cli/* requests, POST /v1/ep, and POST or DELETE /v1/models* require loopback, independent of Origin. Route matching is root-path-aware for mounted apps. Remote multi-model inference can use an existing slot but cannot lazy-load an attacker-selected model. Regression tests cover origin-less remote CLI access, management routes, mounted apps, and inference loading.

src/winml/modelkit/serve/_security.py
```diff
@@ -0,0 +1,173 @@
...
+from starlette.responses import PlainTextResponse
+
+
+if TYPE_CHECKING:
+    from collections.abc import Collection, Mapping
+
+    from starlette.types import ASGIApp, Receive, Scope, Send
+
+
+_CLI_API_LOOPBACK_ONLY_ROUTES: dict[str, set[str] | None] = {"/v1/cli": None}
```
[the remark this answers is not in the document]
timenick: Fixed in 637ed464. The CLI route policy now lives directly at the cli_api.py middleware registration, removing the cross-module private global while keeping the same loopback-only behavior.

src/winml/modelkit/serve/_security.py
```diff
@@ -0,0 +1,173 @@
...
+_MODEL_API_LOOPBACK_ONLY_ROUTES: dict[str, set[str] | None] = {
```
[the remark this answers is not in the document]
timenick: Fixed in 637ed464. The model-server route policy now lives directly at the app.py middleware registration, removing the cross-module private global while preserving the CLI and management route restrictions.

src/winml/modelkit/serve/_security.py
```diff
@@ -0,0 +1,165 @@
...
+        for prefix, methods in self._loopback_only_routes.items():
+            if path != prefix and not path.startswith(f"{prefix}/"):
+                continue
+            if methods is None or method in methods:
+                return True
+        return False
+
+
+def _is_loopback_client(scope: Mapping[str, Any]) -> bool:
+    client = scope.get("client")
```
xieofxie, line 70: This value is not necessarily the transport peer: both uvicorn.run calls retain Uvicorn 0.51.0's proxy_headers=True default, so ProxyHeadersMiddleware can replace scope["client"] from X-Forwarded-For. If FORWARDED_ALLOW_IPS=* (or otherwise trusts the connecting proxy/client), a remote caller can send X-Forwarded-For: 127.0.0.1 and regain access to /v1/cli/compile and the attacker-selected qnn_sdk_root execution path. Please launch both server modes with proxy_headers=False so this authorization uses the socket peer, or preserve and authorize against the raw peer separately if proxy support is required.

xieofxie (changes requested):
The direct remote-client bypass is fixed, but trusted proxy headers can still spoof the loopback address used for authorization.

src/winml/modelkit/serve/_security.py
timenick: Fixed in 9d5cfcb9. Both Uvicorn startup paths now set `proxy_headers=False`, so loopback authorization uses the socket peer rather than a value rewritten from trusted forwarding headers. Added a parameterized regression test covering the CLI-only and `--multi` startup modes.

DingmaomaoBJTU (approved):

src/winml/modelkit/serve/static/index.html
```diff
@@ -917,7 +918,7 @@ <h4>Commands</h4>
 // ══════════════════════════════════════════════════
 // State
 // ══════════════════════════════════════════════════
-let serverUrl = window.location.origin;
+const serverUrl = window.location.origin;
```
xieofxie: This drops an application path prefix. When the app is mounted (for example at /proxy, a case already exercised by the security tests) or exposed with root_path, the demo is served at /proxy/demo and its APIs at /proxy/v1/..., but window.location.origin makes every request target /v1/... at the origin root. Since the Server URL field is now read-only, users also cannot work around it. Please derive the API base from the demo document URL while preserving the prefix before /demo, or inject the configured root path when serving the page, and cover the mounted demo behavior.

xieofxie (changes requested):
The proxy-header authorization bypass is fixed, but pinning the demo to the bare origin breaks mounted and root-path deployments.

src/winml/modelkit/serve/static/index.html
timenick: Fixed in af94a6b6. The demo now derives its API base from the document URL, so `/demo` stays at the origin root while `/proxy/demo` targets `/proxy/v1/...` without relaxing the same-origin policy. Regression tests execute the actual JavaScript for both root and mounted paths. I also added `tests/unit/serve` to CI and pinned Node 22 for that shard so these browser-path tests are enforced.

xieofxie (approved):
The mounted/root-path demo regression is fixed, and the prior security findings remain addressed. No further issues found.

PR #1324, opened 2026-08-21
Eval: Add independent device and EP for ONNX references

zhenchaoni:
## Summary

Two-ONNX compare mode already supported `--reference`, but it created raw ORT sessions inside `TensorSimilarityEvaluator` and used the candidate's execution settings for both models.

This PR:

- loads the candidate and reference ONNX models through the standard `WinMLAutoModel` loading path before constructing the evaluator
- adds `--reference-device` (default: `cpu`) and `--reference-ep` so the reference can run independently from the candidate
- preserves the existing Hugging Face reference behavior when `--reference` is omitted
- reports and serializes the effective reference settings, with updated documentation and unit coverage

## Example

Run the candidate on CPU and the reference ONNX model on DML, then compare their output tensors on identical random inputs:

```bash
winml eval --mode compare \
  -m candidate.onnx \
  --device cpu \
  --reference baseline.onnx \
  --reference-device gpu \
  --reference-ep dml \
  --samples 10
```

### DIFF
--- a/src/winml/modelkit/eval/evaluate.py
+++ b/src/winml/modelkit/eval/evaluate.py
@@ -29,6 +29,7 @@ if TYPE_CHECKING:
     from ..models.winml.composite_model import WinMLCompositeModel
     from ..models.winml.genai_causal_lm import WinMLGenaiCausalLM
     from .base_evaluator import WinMLEvaluator
+    from .tensor_similarity_evaluator import TensorSimilarityEvaluator
 
 logger = logging.getLogger(__name__)
 
@@ -375,7 +376,8 @@ def _load_model(
         rebuild=config.rebuild,
     )
 
-    if config.model_id is None and loader is not _ModelLoaderKind.ONNX:
+    model_id = config.model_id
+    if model_id is None and loader is not _ModelLoaderKind.ONNX:
         raise ValueError("model_id is required.")
 
     if loader is _ModelLoaderKind.EVALUATOR_MANAGED:
@@ -402,10 +404,10 @@ def _load_model(
             hf_config = (
                 load_hf_config(
                     AutoConfig,
-                    config.model_id,
+                    model_id,
                     trust_remote_code=config.trust_remote_code,
                 )
-                if config.model_id is not None
+                if model_id is not None
                 else None
             )
             model = WinMLAutoModel.from_onnx(
@@ -423,6 +425,8 @@ def _load_model(
             model.config = hf_config
             return model
 
+        assert model_id is not None
+
         # HuggingFace build path — export overrides (--input-specs/
         # --export-config/--dynamic-axes) are merged under the build config's
         # ``export`` section as a sparse dict so from_pretrained routes them
@@ -438,7 +442,7 @@ def _load_model(
             build_override = override_dict
 
         return WinMLAutoModel.from_pretrained(
-            config.model_id,
+            model_id,
             ep_device,
             task=config.task,
             device=config.device,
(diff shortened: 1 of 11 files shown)


xieofxie:
should update title and description to match code

src/winml/modelkit/eval/evaluate.py
```diff
@@ -707,12 +742,21 @@ def evaluate(
                 "to see supported role=path model options.",
             ) from error
 
+    reference_model = (
+        _load_model(_make_reference_config(config))
```
xieofxie: **[P1] Preserve the FP32 Hugging Face reference**

This now routes the implicit Hugging Face reference through `_load_model` / `load_native_hf_model`, which loads with `torch_dtype="auto"`. Previously `TensorSimilarityEvaluator._load_reference_model` explicitly used `dtype=torch.float32`, so FP16/BF16 checkpoints now run as reduced-precision CPU references. That changes the reported similarity baseline (and can make CPU inference unsupported or unstable), despite the PR promising unchanged Hugging Face-reference behavior. Please add a dtype override/reference-loading path that keeps this reference on CPU FP32.

xieofxie (changes requested):
The implicit Hugging Face reference path has a precision regression that needs to be addressed before merge.

src/winml/modelkit/eval/evaluate.py
zhenchaoni: Fixed

src/winml/modelkit/eval/evaluate.py
```diff
@@ -654,8 +696,8 @@ def evaluate(
     mode = config.mode if config.mode is not None else "onnx"
     if mode not in EVAL_MODES:
         raise ValueError(f"Invalid mode {mode!r}; expected one of {EVAL_MODES} or None.")
-    # Two-ONNX compare: both candidate and reference run as raw ORT sessions, so
-    # HF task resolution / model_id are not required — keep task as-is.
+    # Two-ONNX compare does not require HF task resolution or model metadata.
+    # With no explicit task, WinMLAutoModel selects its generic raw-output model.
```
xieofxie, line 660: **[P1] Keep two-ONNX comparison on raw outputs even when --task is set**

The following assignment preserves an explicit task in two-ONNX mode, making both files load through task-specific wrappers rather than the generic raw-output model. Those wrappers can rename or discard graph outputs (for example, image and sequence classification return only logits, falling back to the first output), so a multi-output comparison can silently omit tensors or compare unrelated first outputs under one name. The previous raw-session implementation always compared all named ONNX outputs. Please force 	ask=None for both sides and cover an explicit-task, multi-output case.

src/winml/modelkit/eval/tensor_similarity_evaluator.py
```diff
@@ -36,68 +36,19 @@
...
-    # Either an HF-backed model (default path) or a raw-ORT wrapper (two-ONNX
-    # compare); annotated here so both __init__ branches type-check.
-    model: WinMLPreTrainedModel | _ONNXSessionModel
+    """Per-output tensor parity between a candidate and a reference model."""
 
     def __init__(
         self,
         config: WinMLEvaluationConfig,
         model: WinMLPreTrainedModel | WinMLCompositeModel,
+        reference_model: Any,
```
xieofxie: **[P2] Preserve the exported evaluator's two-argument constructor**

TensorSimilarityEvaluator is exported from winml.modelkit.eval, and existing callers can construct it as TensorSimilarityEvaluator(config, model) with the reference loaded internally. Making this argument required now raises TypeError for those callers. Please retain the two-argument API (for example, make injection optional and preserve the old loading behavior) or keep injected-reference construction behind an internal factory.

xieofxie (changes requested):
The FP32 reference regression is fixed, but two correctness/compatibility issues remain in the current diff.

src/winml/modelkit/eval/tensor_similarity_evaluator.py
zhenchaoni: Fix this by calling load_model inside the evaluator. So no constructor changes

src/winml/modelkit/eval/evaluate.py
zhenchaoni: Resolved this comment by your suggestion

xieofxie (approved):
Re-reviewed the latest head. The prior FP32, raw-output comparison, and constructor compatibility issues are resolved. No further blocking issues found.

PR #1330, opened 2026-08-21
fix: resolve VED export regressions for TrOCR and Manga-OCR on Transformers 5

fangyangci:
## Background
This PR addresses three deterministic EXPORT_FAIL regressions on OpenVINO NPU:
- microsoft/trocr-large-printed (fp16)
- kha-white/manga-ocr-base (w8a8)
- kha-white/manga-ocr-base (w8a16)

## Changes
- Make the TrOCR sinusoidal positional embedding patch compatible with Transformers 5 by handling missing _float_tensor and materializing meta-device weights safely.
- Normalize position_ids to [batch, seq] for BERT-like decoders to avoid gather rank mismatch.
- Pass a precomputed 4D additive decoder attention mask to avoid the dynamic cache_position mask path during export.

## Validation
- pytest tests/unit/export/test_vision_encoder_decoder_onnx_config.py -q (passed)
- Re-ran the three models with original build parameters; export no longer fails with EXPORT_FAIL.

## Update: Transformers 5.x only (2026-08-24)

- **Dependency Update**: Changed dependency to require `transformers>=5,<6`.
- **Logic Cleanup**: Removed all legacy `transformers 4.*` compatibility and conditional logic (including `use_auth_token` translation and Marian 4.x condition branches).
- **Validation**: All tests and static analysis passed completely (validated via `pytest` and `mypy` commands).

### DIFF
--- a/src/winml/modelkit/models/hf/vision_encoder_decoder.py
+++ b/src/winml/modelkit/models/hf/vision_encoder_decoder.py
@@ -209,8 +209,8 @@ def _patched_trocr_sinusoidal_positional_embedding_forward(
             self.weights = weights
         ref_tensor = getattr(self, "_float_tensor", None)
         if ref_tensor is not None and not ref_tensor.is_meta:
-            return cast("torch.Tensor", weights.to(ref_tensor))
-        return cast("torch.Tensor", weights.to(input_ids.device))
+            return weights.to(ref_tensor)
+        return weights.to(input_ids.device)
 
     abs_pos = getattr(self, "position_id", None)
     if abs_pos is not None:
--- a/src/winml/modelkit/models/hf/vision_encoder_decoder.py
+++ b/src/winml/modelkit/models/hf/vision_encoder_decoder.py
@@ -209,8 +209,10 @@ def _patched_trocr_sinusoidal_positional_embedding_forward(
             self.weights = weights
         ref_tensor = getattr(self, "_float_tensor", None)
         if ref_tensor is not None and not ref_tensor.is_meta:
-            return weights.to(ref_tensor)
-        return weights.to(input_ids.device)
+            runtime_weights: torch.Tensor = weights.to(ref_tensor)
+            return runtime_weights
+        runtime_weights = weights.to(input_ids.device)
+        return runtime_weights
 
     abs_pos = getattr(self, "position_id", None)
     if abs_pos is not None:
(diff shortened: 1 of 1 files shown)


src/winml/modelkit/models/hf/vision_encoder_decoder.py
```diff
@@ -470,11 +493,28 @@ def _invoke_hf(self, cache: Any, inputs: dict[str, torch.Tensor]) -> torch.Tenso
...
+            extra["position_ids"] = position_ids
+
+        # Pass a precomputed 4D additive mask so decoder internals do not
+        # rebuild a dynamic causal mask from cache_position.
+        decoder_attention_mask = inputs["decoder_attention_mask"]
+        if decoder_attention_mask.dim() == 2:
+            expanded_mask = decoder_attention_mask[:, None, None, :].to(
+                dtype=inputs["encoder_hidden_states"].dtype
+            )
+            decoder_attention_mask = (1.0 - expanded_mask) * torch.finfo(
```
DingmaomaoBJTU, line 511: This unconditional additive 4D mask is incompatible with the published `transformers>=4.57` dependency. In TrOCR 4.57, `_prepare_4d_causal_attention_mask()` treats any correctly shaped 4D input as a binary mask and computes `1.0 - attention_mask`, so the already-additive `0 / finfo.min` values are inverted and valid positions become masked. BertGeneration 4.57 passes the value to `get_extended_attention_mask()`, whose supported paths are only 2D/3D, causing a shape error. The development override only exercises Transformers 5.14, and the current VED tests do not invoke this wrapper. Please gate/precompute the 4D mask only for model/version paths that consume additive masks unchanged, retain the 2D mask for 4.57, and add wrapper-forward coverage for TrOCR and BertGeneration under both supported major versions.

DingmaomaoBJTU (changes requested):
The export fix works with the Transformers 5 development lock, but the unconditional 4D additive mask breaks the still-published Transformers 4.57 compatibility range.

src/winml/modelkit/models/hf/vision_encoder_decoder.py
fangyangci: This PR has switched to transformers>=5,<6. The 4.x compatibility branches/logic have been removed, and the relevant tests and type checks have been updated and are passing successfully. Consequently, the 4.57 compatibility concern is no longer applicable within the current scope of this PR.

DingmaomaoBJTU (approved):

PR #1351, opened 2026-08-24
sys: Use DXCore for accelerator discovery

xieofxie:
## Summary
- use Windows DXCore as the primary source of GPU/NPU adapter identity in `winml sys`
- keep WMI/PnP as descriptive enrichment for driver and manufacturer, and as a compatibility fallback when DXCore is unavailable
- preserve one native row per physical adapter, so identical display names no longer require WMI-to-ORT matching
- surface the DXCore adapter LUID in text, compact, and JSON output as a side product of native system discovery

## Discovery behavior

DXCore returns the adapter name, hardware IDs, and LUID together. `winml sys` now creates accelerator rows from that inventory, then joins WMI/PnP metadata by hardware ID or name. ONNX Runtime remains responsible only for execution-provider inventory and EP-specific facts; it no longer assigns the top-level device LUID.

CPU discovery remains WMI-based and reports `LUID: N/A` because CPU devices do not have a Windows graphics/compute adapter LUID.

## Hardware observations

| Feature | Selected configuration | Discovered device | Driver | LUID | Result |
|---|---|---|---|---|---|
| AMD GPU | `--ep migraphx --device gpu` | AMD Radeon(TM) 860M Graphics | `32.0.22032.6002` | `0x00000000_0x00012C8B` | Passed |
| AMD NPU | `--ep vitisai --device npu` | NPU Compute Accelerator Device | `32.0.20101.3760` | `0x00000000_0x00016055` | Passed |
| OpenVINO GPU | `--ep openvino --device gpu` | Intel Arc 140V GPU | `32.0.101.8132` | `0x00000000_0x0000FCD8` | Passed |
| OpenVINO NPU | `--ep openvino --device npu` | Intel AI Boost | `32.0.100.4621` | `0x00000000_0x00010120` | Passed |
| QNN GPU | `--ep qnn --device gpu` | Adreno X1-85 | `31.0.152.1` | `0x00000000_0x0488D1C4` | Passed |
| QNN NPU | `--ep qnn --device npu` | Snapdragon X Elite Hexagon NPU | `30.0.220.3000` | `0x00000000_0x00012972` | Passed |
| TensorRT RTX GPU | `--ep nvtensorrtrtx --device gpu` | NVIDIA GeForce RTX 5060 Laptop GPU | `32.0.16.1062` | `0x00000000_0x00014CD0` | Passed |

Each feature-specific inference E2E passed on its target hardware. OpenVINO NPU measured 0.40 ms mean latency and 2,528.66 samples/sec. The physical-device and EP views reported matching device identities where the provider published metadata.

## Validation

- all 7 feature-specific inference E2E tests passed
- all feature workers passed the 21-test `winml sys` E2E suite
- PR-focused unit scopes passed: 23 tests on AMD GPU and 25 tests on each remaining feature host
- OpenVINO NPU additionally passed 121 alias/session unit tests
- after merging latest `main`, 125 focused sysinfo/lazy-import tests and 22 combined sys/AMD GPU E2E tests passed
- Ruff is clean on the affected Python scope

## Notes

- VitisAI emitted two non-fatal `Invalid NodeIndex` diagnostics during initialization; inference still completed successfully with the requested provider and expected output.
- One QNN GPU mixed-process run hit a transient native WindowsML access violation; the affected tests passed in a clean process.
- An additional OpenVINO reshape compatibility run timed out after its generated-case worker pool crashed; this is unrelated to the system-discovery changes.


src/winml/modelkit/commands/sys.py
```diff
@@ -637,9 +639,14 @@ def _gather_device_info(
                     ep_info,
                     match_type,
                     sysinfo_name,
+                    allow_unnamed=sum(
```
zhenchaoni: Could we avoid attaching a LUID when multiple same-type devices share the same display name? `_find_matching_device()` independently returns the first named EP row for every sysinfo row, so two identical GPUs with LUIDs A and B are both reported as A. I reproduced this with two same-name GPU rows. Since there is no reliable WMI-to-ORT identity bridge here, it would be safer to leave the LUID null unless the match is one-to-one, or to consume matched candidates so one EP device cannot be assigned twice.

src/winml/modelkit/commands/sys.py
```diff
@@ -672,6 +686,14 @@ def _find_matching_device(
                 hw = dev.get("hardware_name", "") or ""
                 if hw == sysinfo_name or sysinfo_name in hw or hw in sysinfo_name:
                     return cast("dict[str, Any]", dev)
+                if not hw or hw == "<unknown>":
+                    unnamed_candidates.append(cast("dict[str, Any]", dev))
+
+    candidates_by_luid = {
```
zhenchaoni: The LUID aggregation only covers unnamed candidates. A named device still returns immediately on the first fuzzy name match, even when that row has `luid=None`; a later source for the same name with a valid LUID is ignored. I reproduced this with entries `[same name / None, same name / valid LUID]`, which leaves the top-level device LUID null. Could we collect named matches as well and prefer the sole non-null LUID?

zhenchaoni (changes requested):
I found two device-to-LUID matching issues that can produce missing or incorrect identifiers.

src/winml/modelkit/commands/sys.py
xieofxie: Addressed in 46e2e6c3. We now use the display name to narrow EP candidates and only attach a LUID when that sysinfo identity is unique. If multiple same-type sysinfo devices share the same display name, their LUIDs remain null because the WMI rows carry no LUID or other bridge to map them reliably to individual ORT adapters.

src/winml/modelkit/commands/sys.py
xieofxie: Addressed in 46e2e6c3. Named matches are now collected across EP sources, and the sole non-null LUID is preferred, so an earlier same-name source with no LUID no longer masks a later valid value. Multiple distinct LUIDs remain ambiguous and are not guessed.

xieofxie:
The review discussion exposed that top-level LUID assignment should not depend on ORT-to-WMI name matching. Commit 65636595 replaces that design: DXCore now owns each GPU/NPU row and returns its name, hardware IDs, and LUID together; WMI/PnP only enriches descriptive fields. This also handles multiple physical adapters with identical display names because each DXCore row already carries its own identity.

xieofxie:
Latest Qualcomm retest confirms the native path: DXCore returned Adreno X1-85 (`0x00000000_0x00012AC1`) and Snapdragon X Elite Hexagon NPU (`0x00000000_0x000135AA`). QNN GPU inference measured 0.743 ms mean latency / 1,345.89 samples/sec; QNN NPU selection succeeded. The prior lint failure was a mypy-only ctypes annotation issue, fixed in a2ab36ef; `mypy -p winml.modelkit` now passes all 440 source files.

xieofxie:
Tested `a2ab36ef` for **ov_npu** on an Intel Core Ultra 7 258V / Intel AI Boost machine (PR changes applied on the managed branch atop current `main`).

- `uv run pytest -m e2e tests/e2e/test_sys_e2e.py -q`: **21 passed**
- `uv run pytest -m e2e "tests/e2e/test_perf_e2e.py::TestPerfONNXDirect::test_benchmark_ep_device_npu[openvino]" -q -s`: **1 passed**; OpenVINO NPU inference works
- `winml sys --list-device --list-ep --format json`: OpenVINO correctly reports Intel AI Boost with LUID `0x00000000_0x0001030D`, but the top-level NPU device row reports `"luid": null`
- Direct `enumerate_compute_adapters()` result contains only the Intel Arc GPU (`0x00000000_0x0000FED2`) and does not enumerate the NPU

So the OpenVINO NPU execution path works, but the PR's DXCore-based top-level LUID discovery does **not** work for the Intel NPU on this host. The existing E2E suite does not catch this because it permits a null device LUID.

xieofxie:
Tested **OV GPU** on PR head `a2ab36e` merged with current `main` (`6b16162`) on Intel(R) Arc(TM) 140V GPU (16GB).

- `TestPerfONNXDirect::test_benchmark_ep_device_gpu[openvino]`: **passed** using `--ep openvino --device gpu`
- `test_default_json_devices_include_luid` and `test_devices_match_across_default_and_list_device_views`: **2 passed**
- `winml sys --format json` reported GPU LUID `0x00000000_0x0000FED2`, matching the OpenVINO GPU device entry
- Direct ResNet FP32 perf completed successfully: **1.32 ms** mean latency, **754.85 samples/sec** (10 iterations)

OV GPU is working on this branch.

xieofxie:
Tested the current PR head `7b608fdc22faa9e59a36fe5daa736f4b02e93186` on the AMD NPU worker: **PASS**.

- `winml sys --format json` discovered `NPU Compute Accelerator Device` (AMD) as priority 1 with LUID `0x00000000_0x0001628B`.
- `VitisAIExecutionProvider` reported the same NPU name and LUID, confirming physical-device/EP matching.
- `uv run pytest -m e2e tests/e2e/test_sys_e2e.py -q`: **21 passed**.
- Feature validation with `--ep vitisai --device npu` completed 5 measured inferences on the bundled W8A8 ResNet model: **0.974 ms mean latency**, **1,026.21 samples/sec**.

Hardware: AMD Ryzen AI 7 H 350; Vitis AI EP package 1.8.75.0.

xieofxie:
Reviewed and independently verified 7b608fdc. The switch to official DXCore GPU/NPU hardware attributes is SDK-correct and fixes Intel NPU enumeration. On RTX 4080, DXCore and NvTensorRTRTX both reported `0x00000000_0x00016C5B`; selected TensorRT RTX inference passed at 2.147 ms mean / 465.74 samples/sec over 3 iterations. Local validation: Ruff clean, mypy 440 files, 135 unit tests, and 2 hardware E2Es.

zhenchaoni:
Overall, the revised design looks correct and is consistent with the device-selection model: DXCore owns physical adapter identity and LUID, WMI/PnP provides descriptive enrichment, and ORT remains responsible for EP inventory.

I verified the focused unit scope (25 passed), the full `winml sys` E2E suite (21 passed), Ruff, mypy across 439 source files, and actual GPU JSON/text/compact output. On the test host, the automatically selected DML adapter also matched the LUID reported by `winml sys`.

One small robustness suggestion: in `_gather_device_info`, when a GPU/NPU WMI/PnP query raises, the exception branch continues before considering the already-successful DXCore result. This can hide an accelerator even though DXCore returned a valid name and LUID. Please keep the DXCore row and leave driver/manufacturer unavailable in that case, and only omit the device type when both discovery paths fail.

xieofxie:
Addressed in 4c171995. _gather_device_info now checks the successful DXCore inventory before handling a WMI/PnP exception. Native GPU/NPU rows remain visible with their DXCore name and LUID, while unavailable driver/manufacturer enrichment is left null; an accelerator type is omitted only when both paths fail. Added a regression test for the WMI-failure case. Validation: 17 focused unit tests and all 21 winml sys E2E tests passed; Ruff is clean.

zhenchaoni (approved):
Verified the DXCore/WMI failure fix. Native adapters now remain visible when descriptive enrichment fails. The focused unit tests, full winml sys E2E suite, Ruff, mypy, and host device output all pass. The overall design is consistent with the device-selection model.

PR #1357, opened 2026-08-25
Handle Hugging Face fetch failures and align Transformers with Optimum

fangyangci:
## Summary

This PR improves E2E recovery for Hugging Face fetch failures and keeps dependency resolution consistent between published wheel installations and local/CI `uv sync` environments.

The branch initially moved development to Transformers 5 through a uv-only override. Current `optimum-onnx` releases still require `transformers<4.58`, so that override has been removed and the supported runtime is reverted to Transformers 4. Both wheel/pip installations and `uv sync` now resolve from the published `transformers>=4.57` requirement together with Optimum's upper bound, selecting the compatible Transformers 4.57.x line.

### E2E evaluation

- Add `HF_FETCH_FAIL` as a dedicated failure type for Hugging Face connectivity, timeout, and local cache-miss errors.
- Allow `--retry-failed HF_FETCH_FAIL` so transient model download failures can be retried independently from generic environment failures.
- Extend `--clean-cache` with optional `winml`, `huggingface`, and `others` targets. Passing the flag without targets preserves the existing clean-all behavior.
- Document the new retry classification and targeted cache-cleaning options.

### Transformers and Optimum compatibility

- Revert the effective Transformers target from 5.x to 4.57.x because `optimum-onnx 0.1.x` currently declares `transformers>=4.36,<4.58`.
- Keep `transformers>=4.57` in project dependencies so the constraint is included in the published wheel metadata.
- Remove the uv-only `transformers>=5,<6` override. `uv sync` now honors the same dependency graph as wheel/pip installations instead of forcing an incompatible 5.x version.
- Retain the config-loading and model-export hardening covered by the updated AutoConfig, Marian, and vision encoder-decoder tests.

### Dependency and CI updates

- Stop tracking `uv.lock` and remove `--locked` from CI and contributor setup commands so environments resolve from the declared dependency ranges.
- Constrain `plotext` to `>=5.3.2,<6` because plotext 6.0 has breaking API changes that conflict with the current live chart code.
- Keep live chart clearing compatible with the supported plotext 5.x APIs.
- Apply Ruff-compatible union ordering and regex cleanup discovered by CI.

## Validation

- `uv sync --locked --all-extras --all-groups` completes successfully against the locally resolved dependency set.
- The latest PR checks validate lint, WinML CLI test groups, CodeQL, and CLA.


DingmaomaoBJTU (commented):
I found three issues that should be addressed before merging.

pyproject.toml
```diff
@@ -80,7 +82,7 @@ dependencies = [
   "torchmetrics[detection]>=1.0",
   "torchvision>=0.24",
   "tqdm>=4.67",
-  "transformers>=4.57",
+  "transformers>=5,<6",
```
DingmaomaoBJTU: This makes the published package impossible to install outside this repository: `optimum-onnx` 0.1.0 requires `transformers>=4.36,<4.58`, while this line requires `transformers>=5`. The `[tool.uv]` override is local workspace configuration and is not included in wheel metadata, so pip/uv consumers hit an unsatisfiable resolution. Please avoid publishing these mutually exclusive requirements.

scripts/e2e_eval/run_eval.py
```diff
@@ -1475,6 +1507,9 @@ def _run_build_only(entries: list[ModelEntry], args: argparse.Namespace) -> None
     results_path = output_dir / "build_only_results.json"
     results: dict = _load_results(results_path)
     run_stamp = _slugify_version(args.run_stamp or date.today().strftime("%Y%m%d"))
+    clean_cache_targets: set[str] = getattr(
+        args, "clean_cache_targets", _resolve_clean_cache_targets(args.clean_cache)
```
DingmaomaoBJTU: This fallback conflates the legacy boolean `False` with argparse's no-values `[]`: direct callers/tests still construct `Namespace(clean_cache=False)` without `clean_cache_targets`, so `_resolve_clean_cache_targets(False)` returns every target. `_run_build_only` then deletes the HF, WinML, and VAIP caches even though cleanup was disabled. Please preserve `False` as no cleanup rather than treating it as the clean-all form.

scripts/e2e_eval/utils/classifier.py
```diff
@@ -79,6 +80,16 @@ class FailureType(str, Enum):
             "cuda error",
         ],
     ),
+    (
+        FailureType.HF_FETCH_FAIL,
+        [
+            "we couldn't connect to 'https://huggingface.co' to load the files",
+            "couldn't find them in the cached files",
+            "thrown while requesting head https://huggingface.co",
```
DingmaomaoBJTU, line 125: This recognizes metadata HEAD failures but misses failures while streaming model weights. `huggingface_hub` 1.24 logs `Error while downloading from ...` and re-raises `httpx.ReadTimeout`/`ConnectError`; none of those strings match these rules, so the classifier returns `UNKNOWN` and `--retry-failed HF_FETCH_FAIL` skips the transient download failure. Please include the HF weight-download forms, scoped to Hugging Face URLs to avoid classifying unrelated HTTP failures.

scripts/e2e_eval/run_eval.py
fangyangci: Fixed in d4114c33. `_resolve_clean_cache_targets` now preserves the legacy boolean semantics: `None`/`False` disable cleanup, while `True` and argparse's no-values `[]` retain clean-all behavior. `_run_build_only` now uses the pre-resolved `clean_cache_targets` when present and only evaluates the legacy fallback when needed. Added parameterized regression coverage for all forms.

scripts/e2e_eval/utils/classifier.py
fangyangci: Fixed in d4114c33. Streaming failures are now classified as HF_FETCH_FAIL only when the output contains a Hugging Face host (huggingface.co/subdomains or hf.co/subdomains) together with a download or transient HTTPX signal (Error while downloading from, ReadTimeout, ConnectTimeout, ConnectError, or RemoteProtocolError). Added positive HF CDN/Xet cases and negative non-HF cases to prevent broad HTTP misclassification.

pyproject.toml
fangyangci: version revert back to 4.57

src/winml/modelkit/analyze/core/runtime_checker_query.py
```diff
@@ -591,18 +590,7 @@ def get_query_conditions_for_node(
...
-        # Some runtime-specific models emit extension ops in the default ONNX
-        # domain where no standard schema exists (for example
-        # SimplifiedLayerNormalization). Treat this as unsupported instead of
-        # aborting the whole analyze command.
-        raise OpUnsupportedError(
-            "Node "
-            f"{node.op_type} has no registered schema for domain "
-            f"'{domain.schema_domain}' at opset {opset_version}: {e}"
-        ) from e
+    schema = domain.get_op_schema(node.op_type, opset_version)
```
DingmaomaoBJTU: Removing this conversion lets `onnx.defs.SchemaError` escape for runtime-specific operators without a registered ONNX schema. `run_for_node()` only catches `OpUnsupportedError` on this path, so one such node aborts the whole analysis instead of producing a no-data result. Please restore the `SchemaError` to `OpUnsupportedError` handling and preserve the schema-specific reason.

scripts/e2e_eval/run_eval.py
```diff
@@ -2509,10 +2579,16 @@ def _should_skip_existing(existing: dict, retry_types: set[str] | None, eval_typ
...
+    # HF fetch retry matching is deliberately independent of the exclusive
+    # pipeline-stage classification and also covers accuracy-phase fetches.
+    if "HF_FETCH_FAIL" in retry_types and _matches_hf_fetch_retry(existing):
+        return False
+
     # Check perf failure (only when perf ran)
     if eval_type != "accuracy" and not perf.get("passed"):
         cls = classify_result(existing) or "UNKNOWN"
-        if not retry_types or cls in retry_types:
+        perf_retry_types = retry_types - {"HF_FETCH_FAIL"}
```
DingmaomaoBJTU: `classify_result()` can return `HF_FETCH_FAIL` for cache misses and Hugging Face streaming timeouts, but subtracting it here removes that classification from the normal retry match. `_matches_hf_fetch_retry()` recognizes only the explicit marker subset, so `--retry-failed HF_FETCH_FAIL` skips other failures already classified as `HF_FETCH_FAIL`. Please keep the classification eligible while retaining the cross-stage marker override.

src/winml/modelkit/loader/_autoconfig.py
```diff
@@ -241,29 +241,15 @@ def load_hf_config(
...
-                FutureWarning,
-                stacklevel=2,
-            )
-            if load_kwargs.get("token") is not None:
-                raise ValueError(
-                    "`token` and `use_auth_token` are both specified. Please set only the "
-                    "argument `token`."
-                )
-            load_kwargs["token"] = use_auth_token
+    if "use_auth_token" in load_kwargs:
```
DingmaomaoBJTU: `pyproject.toml` now resolves Transformers 4.57, but this unconditionally applies the Transformers 5 rule. It rejects the legacy `use_auth_token` path under 4.57, including `use_auth_token=None`, whereas the previous version-aware conversion handled it. Please restore version-aware behavior or align the dependency with this check.

DingmaomaoBJTU (commented):
I found three regressions that should be addressed before merging.

src/winml/modelkit/analyze/core/runtime_checker_query.py
fangyangci: Restored the accidentally removed `SchemaError` to `OpUnsupportedError` conversion, including the schema-specific no-data reason and its regression coverage. The focused runtime-checker helper suite passes (18 tests).

scripts/e2e_eval/run_eval.py
fangyangci: Fixed in 84e44e63. `HF_FETCH_FAIL` now remains eligible for normal primary-classification matching, while the independent failed perf/accuracy log scan continues to catch supplemental HF fetch evidence when another perf classification or accuracy `FAIL` also applies. Updated the CLI help, E2E README, and regression tests to document and verify that these retry criteria can overlap. The focused suites pass (196 tests).

src/winml/modelkit/loader/_autoconfig.py
fangyangci: Fixed in ed334565. The project is now explicitly constrained to Transformers 4.57 (`transformers>=4.57,<4.58`), and the remaining Transformers 5-specific paths in this PR were removed. `load_hf_config` again follows the Transformers 4 legacy `use_auth_token` behavior, including conversion to `token`, conflict detection, and accepting `use_auth_token=None`. Marian now uses the native Transformers 4.57 `cache_position` path, and processor registry lookup uses the Transformers 4 tuple shape. Related descriptions and tests were updated; the focused loader, Marian, and inspect suites pass (119 tests).

DingmaomaoBJTU (commented):
One blocking dependency regression remains.

pyproject.toml
```diff
@@ -119,20 +121,6 @@ environments = [
...
-#    `transformers<4.58.0` ceiling. The transformers_compat shim bridges the
-#    few 4.x internals optimum-onnx reaches for under transformers 5.
-# 2. Neutralize onnxruntime-qnn's transitive standard `onnxruntime` requirement.
-#    onnxruntime-windowsml ships the same `onnxruntime/` module; installing plain
-#    onnxruntime alongside it overwrites windowsml's files on disk and kills the
-#    DirectML EP. The `sys_platform == 'unobtainium'` marker is never true, so uv
-#    never installs plain onnxruntime while still keeping onnxruntime-qnn.
-override-dependencies = [
-  "transformers>=5,<6",
-  "onnxruntime ; sys_platform == 'unobtainium'",
```
DingmaomaoBJTU, line 134: Removing this override reintroduces standard `onnxruntime` through `onnxruntime-qnn`. The latest CI resolves `onnxruntime==1.29.0`, `onnxruntime-qnn==2.5.0`, and `onnxruntime-windowsml==1.24.6.202605042033` together. Reproducing that dependency set loads ORT 1.29 with only Azure/CPU providers, while WindowsML alone exposes `DmlExecutionProvider`; the distributions share the same `onnxruntime/` module, so standard ORT overwrites DirectML support. Please retain the false-marker `onnxruntime` override and remove only the Transformers override.

DingmaomaoBJTU (approved):
Re-reviewed the latest changes. The reported regressions are fixed, the QNN dependency no longer overwrites WindowsML ONNX Runtime, and the current checks pass.

PR #1361, opened 2026-08-26
fix(perf): support multiple QNN EP context partitions

xieofxie:
## Summary
- coalesce QNN profiling blocks emitted per EP-context partition into inference samples
- preserve partition-local operator IDs and timing conversion
- infer runtime partition counts when raw model metadata does not expose EPContext nodes
- add regression coverage for metadata-backed and runtime-inferred multi-partition traces

### DIFF
--- a/src/winml/modelkit/commands/perf.py
+++ b/src/winml/modelkit/commands/perf.py
@@ -3399,7 +3399,7 @@ def perf(
         # misleading JSON artifact on disk for CI consumers.
         # =================================================================
         if op_tracing:
-            from ..session.monitor.report import display_op_trace_report, write_op_trace_json
+            from ..session.monitor.report import display_op_trace_report
 
             # Both ONNX and HF inputs run through the same PerfBenchmark
             # instance, which exposes its perf context as ``_perf_ctx``.
@@ -3464,16 +3464,10 @@ def perf(
                 display_op_trace_report(trace_result, console, top_n=top_k)
             else:
                 display_op_trace_report(trace_result, console)
-            # Write the op-trace report next to the requested benchmark output
-            # file (same directory + stem, with an ``_op_trace`` suffix) so the
-            # two artifacts stay paired instead of landing under an unrelated
-            # fixed name.
-            trace_output = output.with_name(f"{output.stem}_op_trace{output.suffix}")
-            write_op_trace_json(trace_result, trace_output)
             profiling_csv = trace_result.artifacts.get("csv")
             _print_save_to_footer(
                 console,
-                trace_json=str(trace_output),
+                trace_json=None,
                 profiling_csv=profiling_csv,
             )
         else:
--- a/src/winml/modelkit/commands/perf.py
+++ b/src/winml/modelkit/commands/perf.py
@@ -105,6 +105,10 @@ def _detail_fallback_guidance(reason: TraceFallbackReason | None) -> str:
         ),
         TraceFallbackReason.QHAS_OUTPUT_MISSING: "the requested QHAS output was not found",
         TraceFallbackReason.QHAS_PARSE_FAILED: "the QHAS output could not be parsed",
+        TraceFallbackReason.MULTIPLE_PARTITIONS: (
+            "the model contains multiple EPContext partitions; QHAS currently "
+            "reports only one partition"
+        ),
     }
     if reason is None:
         return "QHAS post-processing was unavailable"
--- a/src/winml/modelkit/commands/perf.py
+++ b/src/winml/modelkit/commands/perf.py
@@ -2144,17 +2144,9 @@ def _io_specs_from_config(
 def _print_save_to_footer(
     console: Console,
     *,
-    trace_json: str | None,
     profiling_csv: str | None,
 ) -> None:
-    """Print save-to footer lines after the op-trace report.
-
-    Each line is rendered only when its path is supplied; if both are
-    ``None`` the helper emits nothing. The ``[dim]...[/dim]`` markup
-    softens the label so the path itself is the visual anchor.
-    """
-    if trace_json:
-        console.print(f"[dim]Op-trace JSON:[/dim] {trace_json}")
+    """Print the raw profiling artifact path after the op-trace report."""
     if profiling_csv:
         console.print(f"[dim]Profiling CSV:[/dim] {profiling_csv}")
 
@@ -3467,7 +3459,6 @@ def perf(
             profiling_csv = trace_result.artifacts.get("csv")
             _print_save_to_footer(
                 console,
-                trace_json=None,
                 profiling_csv=profiling_csv,
             )
         else:
--- a/src/winml/modelkit/session/monitor/qnn_monitor.py
+++ b/src/winml/modelkit/session/monitor/qnn_monitor.py
@@ -584,9 +584,9 @@ class QNNMonitor(WinMLEPMonitor):
 
         parsed = parse_qnn_profiling_csv(csv_path)
         samples = parsed.get("samples", [])
+        partition_count = self._epcontext_partition_count()
         if self._expected_measured_samples is not None:
             expected_total = self._warmup_samples + self._expected_measured_samples
-            partition_count = self._epcontext_partition_count()
             expected_profile_blocks = expected_total * partition_count
             if (
                 len(samples) != expected_profile_blocks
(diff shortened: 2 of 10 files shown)


xieofxie:
a follow up https://github.com/microsoft/winml-cli/issues/1362

src/winml/modelkit/commands/perf.py
xieofxie, line 3456: trace_result is part of the result here?

src/winml/modelkit/commands/perf.py
```diff
@@ -3464,16 +3464,10 @@ def perf(
...
-            # file (same directory + stem, with an ``_op_trace`` suffix) so the
-            # two artifacts stay paired instead of landing under an unrelated
-            # fixed name.
-            trace_output = output.with_name(f"{output.stem}_op_trace{output.suffix}")
-            write_op_trace_json(trace_result, trace_output)
             profiling_csv = trace_result.artifacts.get("csv")
             _print_save_to_footer(
                 console,
-                trace_json=str(trace_output),
+                trace_json=None,
```
xieofxie: if we don't need this, should we just update this function to remove this parameter?

src/winml/modelkit/commands/perf.py
xieofxie: Yes. The monitor result is serialized into the main benchmark JSON at `hw_monitor.ep_proof` by `_monitor_to_json_dict`, so the separate op-trace file duplicated that payload.

src/winml/modelkit/commands/perf.py
xieofxie: Updated. I removed the `trace_json` parameter and simplified the footer helper/tests to handle only the raw profiling CSV path.

src/winml/modelkit/session/monitor/qnn_monitor.py
```diff
@@ -58,6 +58,47 @@
...
+
+    combined_samples: list[dict[str, Any]] = []
+    for offset in range(0, len(samples), partition_count):
+        partition_samples = samples[offset : offset + partition_count]
+        combined: dict[str, Any] = {
+            "metadata": {
+                "hvx_threads": max(
+                    sample["metadata"]["hvx_threads"] for sample in partition_samples
+                ),
+                "accel_execute_cycles": sum(
```
DingmaomaoBJTU, line 92: Could we normalize each partition's metadata before aggregating it? `_parse_artifacts()` explicitly accepts QNN numeric values such as `"120000.7"` via `_to_int()`, but this new `sum()` runs first. With multiple partitions, summing those strings raises `TypeError`, the monitor reports `parse_failed`, and `winml perf --op-tracing` exits 4. A regression test combining multiple partitions with float-string metadata would catch this.

tests/unit/commands/test_perf_optracing.py
```diff
@@ -1378,15 +1374,11 @@ def test_cli_op_tracing_basic_on_qnn(tmp_path):
...
-    assert trace_files, (
-        f"Expected *_op_trace.json next to {output_path}; got: {list(tmp_path.iterdir())}"
-    )
-
     import json
 
-    trace_data = json.loads(trace_files[0].read_text(encoding="utf-8"))
+    report_data = json.loads(output_path.read_text(encoding="utf-8"))
+    trace_data = report_data["hw_monitor"]["ep_proof"]
+    assert not list(tmp_path.glob("*_op_trace.json"))
```
DingmaomaoBJTU, line 1381: This behavior change also leaves `tests/e2e/test_perf_e2e.py::TestPerfOnnx::test_op_tracing_basic_qnn_npu` expecting and reading `perf_op_tracing_qnn_npu_op_trace.json` (lines 805/823). Since `perf.py` no longer writes that artifact, the QNN hardware E2E now fails. Please update that E2E to read `output_file["hw_monitor"]["ep_proof"]` and assert the standalone trace file is absent, as this test does.

DingmaomaoBJTU (commented):
Found two actionable regressions in this change.

src/winml/modelkit/session/monitor/qnn_monitor.py
xieofxie: Fixed. QNN metadata normalization is now shared by partition coalescing and downstream conversion, including float-string values. Added a multi-partition float-string regression test.

tests/unit/commands/test_perf_optracing.py
xieofxie: Updated the existing QNN NPU E2E to read `hw_monitor.ep_proof` from the main perf JSON and assert the standalone `_op_trace.json` is absent. The enabled hardware E2E passes.

src/winml/modelkit/session/monitor/qnn_monitor.py
```diff
@@ -523,65 +586,76 @@ def _parse_artifacts(self, qhas_override: Path | None = None) -> OpTraceResult:
...
                 raise ValueError(
                     "profiling CSV sample count mismatch: "
-                    f"expected {expected_total} total samples "
+                    f"expected {expected_profile_blocks} profiling blocks "
                     f"({self._warmup_samples} warmup + "
-                    f"{self._expected_measured_samples} measured), got {len(samples)}"
+                    f"{self._expected_measured_samples} measured) across "
+                    f"{partition_count} EPContext partition(s), got {len(samples)}"
                 )
+            samples = _coalesce_partition_samples(samples, partition_count)
```
DingmaomaoBJTU, line 607: This coalesces all partitions for the CSV path, but `detail` mode later replaces those complete operators with QHAS output from one `<csv_stem>_qnn.log` and the main schematic. ORT currently overwrites that log for each segment ([onnxruntime#28166](https://github.com/microsoft/onnxruntime/issues/28166)), so a multi-partition run can return `status="ok"` while silently reporting only one partition. Please preserve the coalesced CSV result as an explicit `basic_fallback` when `partition_count > 1`, or process and merge a distinct QHAS artifact for every partition.

src/winml/modelkit/commands/perf.py
```diff
@@ -3399,7 +3391,7 @@ def perf(
         # misleading JSON artifact on disk for CI consumers.
         # =================================================================
         if op_tracing:
-            from ..session.monitor.report import display_op_trace_report, write_op_trace_json
+            from ..session.monitor.report import display_op_trace_report
```
DingmaomaoBJTU, line 3398: Removing the sibling trace file breaks the repository's eval runner. `scripts/e2e_eval/run_eval.py::_copy_op_trace()` still looks for `<result>_op_trace.json`, and `--op-tracing` still promises to copy it to each model's output folder; because the temp result directory is then deleted, eval runs now silently produce no `op_trace.json`. Please update the runner to serialize `proc["result"]["hw_monitor"]["ep_proof"]` before cleanup (and revise its test), or retain the sibling artifact until all consumers are migrated.

DingmaomaoBJTU (changes requested):
Two integration regressions remain before this is ready to approve.

src/winml/modelkit/commands/perf.py
xieofxie: yes, just copy result.json is enough

src/winml/modelkit/session/monitor/qnn_monitor.py
xieofxie: yes. let's use basic_fallback and QHAS as a new followup work item https://github.com/microsoft/winml-cli/issues/1367

xieofxie: Fixed in 05c216ac. Detail tracing now returns the complete coalesced CSV operators as `basic_fallback` with `multiple_partitions`, and does not invoke QHAS for multi-partition runs. Added regression coverage and verified the real QNN/NPU detail command.

src/winml/modelkit/commands/perf.py
xieofxie: Fixed in 05c216ac. The eval runner now copies the main perf result JSON, which contains `hw_monitor.ep_proof`, before the temporary directory is removed. Updated the runner test and help text.

xieofxie: Updated as requested: the eval runner now copies `result.json` directly.

src/winml/modelkit/session/monitor/qnn_monitor.py
xieofxie: Implemented `basic_fallback` for multiple partitions and left multi-partition QHAS handling to follow-up #1367.

DingmaomaoBJTU (approved):
Re-reviewed the complete updated diff. The prior multi-partition QHAS and eval-runner regressions are addressed, and no blocking issues remain.

PR #1368, opened 2026-08-27
fix(optrace): include CPU fallback operators in QNN results

### LINKED ISSUE #1362: optrace: include cpu op result (from onnxruntime.json) into the final result
To support cases where EP could not take all ops

xieofxie:
## Summary
- enable ONNX Runtime profiling for QNN op-trace sessions and merge CPU fallback node timings into the final operator result
- preserve optimized node identity, exclude warmup samples, recalculate mixed-provider percentages, and label CPU operators
- persist and support offline parsing of the ORT profile artifact

## Validation
- 260 affected unit tests passed
- DenseNet QNN NPU trace completed with 27 QNN operators and 15 CPU fallback operators

Fixes #1362

### DIFF
--- a/src/winml/modelkit/session/monitor/qnn_monitor.py
+++ b/src/winml/modelkit/session/monitor/qnn_monitor.py
@@ -214,6 +214,11 @@ class QNNMonitor(WinMLEPMonitor):
         self._csv_path: Path = (
             self._output_dir / f"profiling_output_{uuid.uuid4().hex}.csv"
         ).resolve()
+        self._ort_profile_prefix: Path = (
+            self._output_dir / f"onnxruntime_profile_{uuid.uuid4().hex}"
+        ).resolve()
+        self._ort_profile_path: Path | None = None
+        self._require_ort_profile: bool = False
         self._extra: dict[str, str] = dict(extra_provider_options or {})
         self._entered: bool = False
         self._result: OpTraceResult | None = None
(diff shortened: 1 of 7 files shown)


src/winml/modelkit/session/session.py
```diff
@@ -1563,7 +1566,7 @@ def _restore_baseline() -> Exception | None:
                 so = _build_session_options(
                     self._ep_device,
                     self._ep_config,
-                    None,
+                    effective_monitor,
```
DingmaomaoBJTU, line 1569: This hook is only reached when `_session_rebuilt` is true, but that predicate compares only provider options and session-config entries. A monitor whose sole contribution is `configure_session_options()` therefore reuses the old session and never applies its native settings. Please make native monitor configuration participate in the rebuild decision.

src/winml/modelkit/session/monitor/qnn_monitor.py
```diff
@@ -746,6 +775,92 @@ def _metadata_mean(field: str) -> float:
...
+                },
+            )
+            entry["samples_us"].append(float(duration))
+
+        operators: list[OperatorMetrics] = []
+        for op_path, entry in samples_by_node.items():
+            samples_us = entry["samples_us"]
+            if self._expected_measured_samples is not None:
+                expected_total = self._warmup_samples + self._expected_measured_samples
+                if len(samples_us) != expected_total:
```
DingmaomaoBJTU, line 842: A CPU fallback node inside `Loop` or `Scan` can emit multiple kernel events per top-level inference, so this count can legitimately exceed the perf-window iteration count. The current equality check rejects those valid mixed-provider traces as `parse_failed`; please account for repeated node executions within each run before validating/removing warmup samples.

PR #1375, opened 2026-08-31
Fix eager attention selection during model loading

fangyangci:
## Summary
- select the configured Transformers attention implementation before `from_pretrained` constructs attention modules
- thread export compatibility through both build and direct export paths
- cover loader and build propagation while keeping export test doubles compatible

## Motivation
DistilBERT exports could instantiate `DistilBertSdpaAttention` and only switch the config to eager immediately before ONNX export. This left the module implementation and config inconsistent, causing integer attention masks to reach SDPA and fail with:

```
Expected attn_mask dtype to be bool or float or to match query dtype
```

## Validation
- `uv run pytest tests/unit/loader/test_load_hf_model.py tests/unit/build/test_hf.py tests/unit/export/test_htp_exporter_attention_compat.py tests/unit/commands/test_config_value_priority.py tests/unit/commands/test_export.py -q --tb=short`
- 175 tests passed
- retried all 6 affected QNN GPU models: the attention-mask dtype error was eliminated in all 6; 5 completed perf and accuracy successfully, while one progressed to a separate duplicate `metadata_props` ONNX validation failure

### DIFF
--- a/src/winml/modelkit/build/hf.py
+++ b/src/winml/modelkit/build/hf.py
@@ -350,6 +350,9 @@ def _load_model(
             path (PR #719 dedup pattern).
     """
     task = config.loader.task
+    attn_implementation = (
+        config.export.compatibility.transformers_attention if config.export else None
+    )
 
     if random_init:
         from transformers import AutoConfig
@@ -402,15 +405,15 @@ def _load_model(
 
         model_label = model_id or config.loader.model_type
         logger.info("Creating random-weight model: %s (from %s)", model_class.__name__, model_label)
-        return model_class.from_config(hf_config)
+        model_kwargs: dict[str, Any] = {}
+        if attn_implementation is not None:
+            model_kwargs["attn_implementation"] = attn_implementation
+        return model_class.from_config(hf_config, **model_kwargs)
 
     if model_id is not None:
         from ..loader import load_hf_model
 
         effective_trust = trust_remote_code or config.loader.trust_remote_code
-        attn_implementation = (
-            config.export.compatibility.transformers_attention if config.export else None
-        )
         pytorch_model, _, _ = load_hf_model(
             model_name_or_path=model_id,
             task=task,
--- a/src/winml/modelkit/build/hf.py
+++ b/src/winml/modelkit/build/hf.py
@@ -408,6 +408,9 @@ def _load_model(
         from ..loader import load_hf_model
 
         effective_trust = trust_remote_code or config.loader.trust_remote_code
+        attn_implementation = (
+            config.export.compatibility.transformers_attention if config.export else None
+        )
         pytorch_model, _, _ = load_hf_model(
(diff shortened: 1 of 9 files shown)


src/winml/modelkit/build/hf.py
```diff
@@ -408,13 +408,17 @@ def _load_model(
         from ..loader import load_hf_model
 
         effective_trust = trust_remote_code or config.loader.trust_remote_code
+        attn_implementation = (
```
DingmaomaoBJTU: Could we move this compatibility lookup before the `random_init` branch and pass it to `model_class.from_config(...)` too? That branch returns at line 405, so it still constructs the default SDPA module. With DistilBERT, `from_config(config)` creates `DistilBertSdpaAttention`, whereas `from_config(config, attn_implementation="eager")` creates `MultiHeadSelfAttention`; the later export-time config mutation cannot replace the already-instantiated module. `build_hf_model(..., random_init=True)` therefore retains the same implementation/config mismatch this PR is fixing.

src/winml/modelkit/commands/export.py
```diff
@@ -464,7 +464,11 @@ def _run_component_export(component_task: str | None, out_path: Path) -> None:
             raise click.ClickException(f"Configuration error: {e}") from e
 
         # Load model with task detection (CLI is the orchestration layer)
-        pytorch_model, _, detected_task = load_hf_model(model, task=component_task)
+        pytorch_model, _, detected_task = load_hf_model(
+            model,
+            task=component_task,
+            attn_implementation=cfg.compatibility.transformers_attention,
```
DingmaomaoBJTU, line 470: This wires the CLI path, but the documented `HTPExporter.export(model=None, model_name_or_path=...)` auto-load path still calls `load_hf_model(model_name_or_path)` without the export config's attention setting (`export/htp/exporter.py:216`). Those users still construct SDPA modules and only mutate the config during export, so the original mismatch remains. Please forward `export_config.compatibility.transformers_attention` there as well and cover that path.

src/winml/modelkit/build/hf.py
fangyangci: Fixed in 844bf13b. The compatibility lookup now happens before the random-init branch, and a non-null attention implementation is forwarded to `model_class.from_config(...)`. I also added a lightweight real DistilBERT regression test that asserts random-init constructs `MultiHeadSelfAttention` instead of `DistilBertSdpaAttention`.

src/winml/modelkit/commands/export.py
fangyangci: Fixed in 844bf13b. The `HTPExporter.export(model=None, model_name_or_path=...)` auto-load path now forwards `export_config.compatibility.transformers_attention` to `load_hf_model`, with a focused regression test covering that call path.

DingmaomaoBJTU (approved):

PR #1377, opened 2026-09-01
Add OpenVINO operator profiling to perf

xieofxie:
## Summary
- add basic OpenVINO operator tracing for CPU and NPU perf runs
- enable ONNX Runtime profiling and parse detailed OpenVINO child kernel events
- exclude warmup invocations and report the raw profiling JSON path
- validate minimum ONNX Runtime and OpenVINO EP profiling support

## ORT compatibility
ORT 1.27 validates supplied and default local-function graph attributes even when the function body does not reference or execute them. FP16 conversion now checks those attributes before conversion and rejects captured tensors whose outer binding becomes FP16 while the skipped graph remains FP32. This prevents the `Tensor element type mismatch. 10 != 1` session-loading failure and remains compatible with ORT 1.24 by applying the stricter validation conservatively.


src/winml/modelkit/session/monitor/openvino_monitor.py
```diff
@@ -2,23 +2,137 @@
...
 
 if TYPE_CHECKING:
     from typing import Self
 
+    import onnxruntime as ort
+
+
+logger = logging.getLogger(__name__)
+
+_MINIMUM_ORT_VERSION = Version("1.26")
```
DingmaomaoBJTU, line 29: This version gate makes the new feature unusable in a normal winml-cli install: `pyproject.toml` still pins `onnxruntime-windowsml==1.24.5.202604171637`, so `_create_monitor()` always raises before OpenVINO profiling starts. Please update the packaged runtime in this PR or avoid requiring 1.26.

src/winml/modelkit/session/monitor/openvino_monitor.py
```diff
@@ -27,13 +141,198 @@ def __exit__(
...
+        return OpTraceResult(
+            model=str(self._onnx_model_path) if self._onnx_model_path is not None else None,
+            device=self._device,
+            tracing_level=self._level,
+            operators=operators,
+            ep="openvino",
+            tracing_backend="onnxruntime",
+            num_samples=self._measured_iterations
+            or max(operator.sample_count for operator in operators),
+            summary={"execute_us": total_us},
```
DingmaomaoBJTU: `OpenVinoMonitor` only supports `tracing_level="basic"`, but `_display_basic_report()` reads `accel_execute_us`; `execute_us` is only used by the detail renderer. As a result this total is silently omitted from the CLI summary. Please emit the basic-mode key or teach the basic renderer to handle the OpenVINO summary.

src/winml/modelkit/session/monitor/openvino_monitor.py
xieofxie: do we have some tests needed for that? I am considering we could relax the toml range

xieofxie: Also I see this pr pins it https://github.com/microsoft/winml-cli/pull/1357 any reason ?

src/winml/modelkit/session/monitor/openvino_monitor.py
xieofxie: fixed

src/winml/modelkit/session/monitor/openvino_monitor.py
xieofxie: updated

DingmaomaoBJTU (approved):

pyproject.toml
```diff
@@ -51,7 +51,7 @@ dependencies = [
   # form of what WinML's wasdk metadata pins as `1.24.5.post202604171637` (PEP 440
   # normalized) — an exact `==` can't reconcile the two. A bounded range keeps the
   # 1.24.x series while resolving to the published wheel.
-  "onnxruntime-windowsml>=1.24.5,<1.25 ; sys_platform == 'win32'",
+  "onnxruntime-windowsml>=1.24.5,<1.28 ; sys_platform == 'win32'",
```
DingmaomaoBJTU, line 54: This range now resolves to ORT 1.27.1, but the pinned `windowsml==2.0.300` declares `onnxruntime-windowsml==1.24.5.202604171637` as its matching runtime; `windowsml==2.3.42` is the package paired with ORT 1.27.1. Please update the WindowsML package in lockstep rather than installing a mismatched WinML C API/runtime pair.

src/winml/modelkit/session/monitor/openvino_monitor.py
```diff
@@ -2,23 +2,137 @@
...
+        if normalized_device not in ("cpu", "npu"):
+            raise ValueError(
+                f"OpenVINO profiling only supports device 'cpu' or 'npu', got {device!r}"
+            )
+
+        self._level = level
+        self._device = normalized_device
+        self._output_dir = Path(output_dir) if output_dir is not None else Path.cwd()
+        self._output_dir.mkdir(parents=True, exist_ok=True)
+        self._profile_prefix = (self._output_dir / "onnxruntime_profile").resolve()
```
DingmaomaoBJTU, line 59: Could we make this prefix unique per monitor? `output_dir` is normally the save directory or cwd, and ORT only appends a millisecond timestamp. Overlapping perf processes can therefore create the same filename or have `_find_fresh_profile()` select the other process's newer profile. A per-run UUID in the prefix would keep discovery scoped to this trace.

src/winml/modelkit/session/monitor/openvino_monitor.py
```diff
@@ -2,23 +2,137 @@
...
+            installed_version = Version("0")
+        if installed_version < _MINIMUM_ORT_VERSION:
+            raise RuntimeError(
+                "OpenVINO op-tracing requires ONNX Runtime 1.26 or newer. "
+                'Install it with: uv pip install "onnxruntime-windowsml>=1.26"'
+            )
+
+    def configure_session_options(self, session_options: ort.SessionOptions) -> None:
+        """Enable ORT profiling on the monitored inference session."""
+        session_options.enable_profiling = True
```
DingmaomaoBJTU, line 116: For an NPU session, ORT OpenVINO 1.27.1 explicitly falls back to OV CPU when compilation fails unless `session.disable_cpu_ep_fallback` is set. Because that fallback stays inside `OpenVINOExecutionProvider`, these events still pass the parser and are returned as `status="ok"`, `device="npu"` even though they are CPU timings. Please disable this fallback for NPU profiling or validate/report the actual OV device.

src/winml/modelkit/session/monitor/openvino_monitor.py
```diff
@@ -27,13 +141,198 @@ def __exit__(
...
+        return OpTraceResult(
+            model=str(self._onnx_model_path) if self._onnx_model_path is not None else None,
+            device=self._device,
+            tracing_level=self._level,
+            operators=operators,
+            ep="openvino",
+            tracing_backend="onnxruntime",
+            num_samples=self._measured_iterations
+            or max(operator.sample_count for operator in operators),
+            summary={"accel_execute_us": total_us},
```
DingmaomaoBJTU, line 200: `accel_execute_us` is a per-inference mean in the existing basic/QNN schema, but `total_us` sums every retained run. The added 4 us / 6 us test therefore displays 10 us, and this value grows with `--iterations` even when latency is unchanged. Please compute the mean of the per-inference accelerator totals here.

src/winml/modelkit/session/monitor/openvino_monitor.py
```diff
@@ -27,13 +141,198 @@ def __exit__(
...
+
+        retained: dict[str, list[tuple[float, float]]] = {}
+        for parent_name, occurrences in spans.items():
+            occurrences.sort()
+            stop = (
+                self._warmup_iterations + self._measured_iterations
+                if self._measured_iterations is not None
+                else None
+            )
+            retained[parent_name] = occurrences[self._warmup_iterations : stop]
```
DingmaomaoBJTU, line 293: These occurrences are partition executions, not necessarily top-level inference runs. A partition inside `Loop` or `Scan` can execute multiple times per inference, so this slice removes only the first loop execution from the warmup and can retain the rest of the warmup as measured samples. Please group spans by their enclosing inference run before applying the warmup/sample window.

PR #1380, opened 2026-09-01
Remove preload ORT DLL (undo PR 689)

zhenchaoni:
This PR is to remove the code of preloading onnxruntime DLL. This reverts PR #689

**About PR 689**
PR #689 originally aims to workaround an issue where TRT-RTX ep loads the onnxruntime from "C:\Windows\System32\onnxruntime.dll" instead of Python dependency packages. This issue causes an old onnxruntime loaded.

**We revert the above workaround because**
Firstly, the original issue is fixed since 7D TRT-RTX EP. Since 7D, TRT-RTX no longer loads the wrong onnxruntime

More importantly, the workaround code creates a duplicate onnxruntime instance. 
-  "ctypes.WinDLL(str(dll))" creates an ort instance from onnxruntime.dll
- "import onnxruntime" creates another ort instance from "onnxruntime_pybind11_state.pyd"

In Runtime API, when it detects 2 ort instances, it will raise an error. This is the main purpose of removing the workaround code.


chinazhangchao (approved):

xieofxie:
> Since 7D, TRT-RTX no longer loads the correct onnxruntime

XD

zhenchaoni:
> > Since 7D, TRT-RTX no longer loads the correct onnxruntime
> 
> XD
 My PR description typo. It should be "Since 7D, TRT-RTX no longer loads the **wrong** onnxruntime"

PR #1382, opened 2026-09-01
Lazy-load session backends and monitors

### LINKED ISSUE #1381: Lazy-load session backends and monitors to avoid unrelated optional dependencies
## Description

Importing a leaf session monitor currently executes the aggregate `winml.modelkit.session` package initializer, which eagerly imports every session backend and monitor. As a result, code that only needs hardware telemetry can fail because an unrelated optional dependency such as ONNX Runtime is not installed.

This is an import-boundary issue, not an ONNX model-loading failure. Backend/provider dependencies should be loaded only when their corresponding workflow is selected.

## Reproduction

Use an environment where `winml-cli` source dependencies for the requested leaf module are available, but `onnx` and `onnxruntime` are not installed:

```powershell
$env:PYTHONPATH = "C:\path\to\winml-cli\src"
python -X utf8 -c "import importlib.util; print('onnx_installed=', importlib.util.find_spec('onnx') is not None); from winml.modelkit.session.monitor.hw_monitor import HWMonitor; print(HWMonitor)"
```

Observed output:

```text
onnx_installed= False
ModuleNotFoundError: No module named 'onnxruntime'
```

The import stack reaches the failure through:

```text
winml.modelkit.session.monitor.hw_monitor
  -> winml.modelkit.session.__init__
  -> winml.modelkit.session.ep_device
  -> import onnxruntime as ort
```

`session/__init__.py` also eagerly imports GenAI, Runtime, QAIRT, QNN, OpenVINO, VitisAI, and monitor implementations even when none of them are requested.

## Expected behavior

Importing a leaf module should require only that module's dependencies. For example:

```python
from winml.modelkit.session.monitor.hw_monitor import HWMonitor
```

should not import or require ONNX Runtime, QAIRT/QNN, GenAI, OpenVINO, or VitisAI.

Likewise, selecting one normal ONNX execution path should not import unrelated provider-specific backends. Missing optional dependencies should be reported only when the user selects the feature that needs them.

## Suggested scope

- Replace eager public re-exports in `winml.modelkit.session.__init__` with lazy imports, for example module-level `__getattr__` or equivalent call-site imports.
- Ensure `session.monitor` does not eagerly import provider-specific monitors for a leaf hardware-monitor import.
- Preserve the existing public symbols and `__all__` contract.
- Add import-isolation tests that block or omit optional dependencies and verify unrelated leaf imports still succeed.
- Add provider-specific tests confirming that selecting QNN/QAIRT, GenAI, Runtime, OpenVINO, or VitisAI still loads the intended backend and produces an actionable missing-dependency error when necessary.

## Acceptance criteria

- `HWMonitor` can be imported without `onnx` or `onnxruntime` installed.
- CPU/DML ONNX workflows do not import QNN/QAIRT or other unrelated provider modules.
- Optional backend imports remain lazy until the matching runtime/provider is selected.
- Existing `winml.modelkit.session` public imports remain compatible.

timenick:
## Summary

- Lazy-load the session, monitor, and sysinfo public facades while preserving their existing exports.
- Defer QAIRT and provider-specific monitor modules until the matching backend is selected.
- Add subprocess import-isolation coverage for Runtime, GenAI, QNN, OpenVINO, VitisAI, and QAIRT, including actionable missing-dependency diagnostics.

Closes #1381

## Validation

- 15 import-isolation tests passed.
- CI-equivalent commands/session shard: 3667 passed.
- CI-equivalent remaining shard: 960 passed.
- Mypy checked 441 source files.
- Pre-commit passed.


xieofxie (approved):
Reviewed the changes and found no actionable issues.

PR #1387, opened 2026-09-02
Add Windows build details to winml sys

xieofxie:
## Summary

- read detailed Windows version metadata from the native `CurrentVersion` registry key
- include product, display version, build, UBR, branch, and build lab values in text and JSON output
- tolerate missing registry keys or individual values
- document the additional output and add unit coverage

image


src/winml/modelkit/commands/sys.py
```diff
@@ -114,6 +114,16 @@ def _get_python_info() -> dict[str, Any]:
     0x14C: "x86",
 }
 
+_WINDOWS_CURRENT_VERSION_KEY = r"SOFTWARE\Microsoft\Windows NT\CurrentVersion"
+_WINDOWS_VERSION_VALUES = {
+    "ProductName": "product_name",
```
DingmaomaoBJTU: Could we avoid exposing `ProductName` verbatim? Windows 11 commonly leaves this registry value as `Windows 10 ...` for compatibility—for example, build 26310 reports `Windows 10 Enterprise` here while WMI reports Windows 11. Because this function separately normalizes `release` to `11`, the text and JSON output become internally contradictory. Please derive or normalize the product name from a reliable OS API/build value instead.

xieofxie: yes, will remove

DingmaomaoBJTU (approved):

PR #1392, opened 2026-09-03
fix(export): support LayoutLM QA and BLIP decoder

### LINKED ISSUE #1391: LayoutLM QA and BLIP decoder fail during ONNX export
## Description

Two Hugging Face model/task combinations fail during ONNX build/export with the current Transformers stack:

- `impira/layoutlm-document-qa` / `question-answering`
- `Salesforce/blip-image-captioning-base` / `image-to-text`

## Current behavior

- LayoutLM QA is loaded through `AutoModelForQuestionAnswering`, which rejects `LayoutLMConfig`. Loading the concrete class then exposes invalid random `token_type_ids` for checkpoints with `type_vocab_size=1`.
- The BLIP decoder wrapper passes a 4D additive attention mask, while the current HF BLIP implementation expects a 2D/3D binary mask and fails shape validation during export.

Representative errors:

```text
Unrecognized configuration class LayoutLMConfig for AutoModelForQuestionAnswering
index out of range in self
Wrong shape for input_ids [1,1] or attention_mask [1,1,1,512]
```

## Expected behavior

Both model/task combinations should export successfully and run with `NvTensorRTRTXExecutionProvider` on GPU.

## Environment

- Python 3.11.14
- Transformers 4.57.6
- NvTensorRTRTXExecutionProvider 2.30.43.0

fangyangci:
## Summary

- resolve LayoutLM question answering with `LayoutLMForQuestionAnswering` instead of the incompatible AutoModel mapping
- generate LayoutLM `token_type_ids` directly in `[0, 1)` so checkpoints with `type_vocab_size=1` remain valid in generated build configs
- pass a 3D binary attention mask to the BLIP decoder to match the current Transformers contract
- add regression coverage for model resolution, LayoutLM value ranges, and BLIP mask shape/dtype

Fixes #1391

## Validation

- `38 passed` across the related LayoutLM/BLIP unit tests
- `impira/layoutlm-document-qa` / `question-answering`: real build PASS; TRT-RTX perf PASS at 6.014 ms mean
- `Salesforce/blip-image-captioning-base` / `image-to-text`: encoder and decoder real builds PASS; TRT-RTX perf PASS at 7.468 ms / 6.748 ms mean

Full `--eval-type both` additionally exposed accuracy-evaluator issues outside this export fix: the generic QA evaluator does not produce LayoutLM `bbox` inputs, and the image-to-text evaluator passes an unsupported `text` keyword and skips all BLIP samples.

### DIFF
--- a/tests/unit/export/test_blip_onnx_config.py
+++ b/tests/unit/export/test_blip_onnx_config.py
@@ -136,7 +136,7 @@ class TestBlipDecoderIO:
         assert tuple(inputs["past_0_key"].shape) == expected
         assert tuple(inputs["past_0_value"].shape) == expected
 
-    def test_decoder_passes_four_dimensional_additive_attention_mask(self, monkeypatch) -> None:
+    def test_decoder_passes_three_dimensional_attention_mask(self, monkeypatch) -> None:
         from types import SimpleNamespace
         from unittest.mock import MagicMock
 
@@ -165,10 +165,9 @@ class TestBlipDecoderIO:
         wrapper._invoke_hf(object(), inputs)
 
         mask = captured["attention_mask"]
-        assert mask.shape == (1, 1, 1, 2)
-        assert mask.is_floating_point()
-        assert mask[0, 0, 0, 0] == 0
-        assert mask[0, 0, 0, 1] == torch.finfo(mask.dtype).min
+        assert mask.shape == (1, 1, 2)
+        assert mask.dtype == torch.int64
+        assert torch.equal(mask, torch.tensor([[[1, 0]]], dtype=torch.int64))
 
     def test_decoder_cache_uses_position_input_when_model_omits_cache_kwargs(
         self, blip_config
--- a/tests/unit/export/test_blip_onnx_config.py
+++ b/tests/unit/export/test_blip_onnx_config.py
@@ -169,6 +169,43 @@ class TestBlipDecoderIO:
         assert mask.dtype == torch.int64
         assert torch.equal(mask, torch.tensor([[[1, 0]]], dtype=torch.int64))
 
+    def test_decoder_runs_with_real_model_and_binary_mask(self, blip_config) -> None:
+        import torch
+        from transformers import BlipForConditionalGeneration
+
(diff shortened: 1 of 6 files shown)


tests/unit/export/test_blip_onnx_config.py
```diff
@@ -165,10 +165,9 @@ def decode(**kwargs):
         wrapper._invoke_hf(object(), inputs)
 
         mask = captured["attention_mask"]
-        assert mask.shape == (1, 1, 1, 2)
-        assert mask.is_floating_point()
-        assert mask[0, 0, 0, 0] == 0
-        assert mask[0, 0, 0, 1] == torch.finfo(mask.dtype).min
+        assert mask.shape == (1, 1, 2)
+        assert mask.dtype == torch.int64
+        assert torch.equal(mask, torch.tensor([[[1, 0]]], dtype=torch.int64))
```
DingmaomaoBJTU, line 170: Non-blocking: Could we add a real decoder regression using the existing tiny `blip_config` fixture? `text_decoder` is a `MagicMock`, so the previous 4-D-mask assertion could pass even though Transformers rejects that mask. Updating the expected shape still leaves that contract untested. We can instantiate `BlipForConditionalGeneration(blip_config)` locally and compare the wrapper's logits/KV against the same decoder with its supported 2-D binary mask, using a nonzero cache position and masked cache slots. This can run on CPU without downloading weights.

DingmaomaoBJTU (approved):
LGTM. The inline suggestion is non-blocking.

tests/unit/export/test_blip_onnx_config.py
fangyangci: Thanks for the suggestion. Added a regression using the existing tiny blip_config and a real randomly initialized BlipForConditionalGeneration. It exercises the wrapper forward path with a non-zero cache position and masked cache slots, then validates the logits and per-layer KV outputs. The full BLIP export test file passes (14 tests), and Ruff passes as well. Addressed in 41b9889d.

DingmaomaoBJTU (approved):

PR #1398, opened 2026-09-07
feat(perf): select physical adapters with --device-luid

### LINKED ISSUE #1319: feat: select the device when multiple devices available for same ep + device combination
consider using device.metadata.DxgiAdapterNumber https://github.com/microsoft/onnxruntime/blob/e858164c9fbdb61d5d0be980d6a2a0c161db3584/onnxruntime/core/platform/windows/device_discovery.cc#L371

So also display it in winml sys

xieofxie:
## Summary
- Add `winml perf --device-luid` to select an ORT adapter using the LUID displayed by `winml sys`.
- Warn when multiple adapters match without a pin, preserving the first-device default.
- Keep inference and monitoring on the selected adapter; reject unavailable LUIDs and conflicting provider options.
- Replace PCI-ID deduplication with LUID-based identity so identical physical GPUs remain selectable.
- Add regression coverage, real DML adapter-selection coverage, and documentation.

## Scope
- Supports ONNX, HuggingFace, composite, and per-module runtime inference.
- Explicitly rejects `--runtime ort-genai`; does not pin the separate build compilation stage.
- Retains WMI/PDH fallbacks still needed after #1351.

Fixes #1319.


src/winml/modelkit/commands/perf.py
```diff
@@ -441,6 +442,60 @@ def _get_monitor_binding(
...
+            raise ValueError(
+                "--ep-options conflict with --device-luid or do not identify the pinned "
+                "adapter unambiguously. Remove the device-selecting provider options."
+            )
+    else:
+        candidates = [
+            device for device in selected.ep.devices if device.device_type.lower() == target.device
+        ]
+        # Multiple EP routes to one known adapter are not multiple adapters.
+        identities = {get_ep_device_luid(device.ort_handle) or id(device) for device in candidates}
```
zhenchaoni: `provider_options` can already redirect the binding to another adapter (for example, DML `device_id=1`), and `_get_ep_device_binding()` handles that case elsewhere. This branch still counts all devices and reports that the first ORT device will be used, even though the session may bind a different one. Please resolve the effective binding from `provider_options` before emitting the ambiguity warning; if it uniquely identifies an adapter, skip the warning (or report that adapter). Please also add a regression test for unpinned `--ep-options device_id=1` with two adapters.

src/winml/modelkit/commands/perf.py
```diff
@@ -1610,15 +1671,15 @@ def _perf_modules(
...
+    from ..session import EPDeviceTarget, resolve_device
     from .build import _instantiate_parent_model
 
     request_device = (device or "auto").lower()
     request_ep = ep
     resolved_target = resolve_device(
         EPDeviceTarget(ep=request_ep or "auto", device=request_device, source=ep_source)
     )
-    resolved_ep_device = WinMLEPRegistry.instance().auto_device(resolved_target)
+    resolved_ep_device = _resolve_perf_ep_device(resolved_target, device_luid, ep_options)
```
zhenchaoni: `_resolve_perf_ep_device()` can raise `DeviceNotFound` or `ValueError` for an unavailable LUID or conflicting provider options, but this call runs before `_perf_modules()`'s `try` block, and the module branch returns before the single-model catch-all. Consequently, `winml perf --module ... --device-luid <unavailable>` can leak a raw traceback instead of a Click-formatted error. Please move this resolution under CLI error handling (or convert these exceptions to `ClickException`) and cover the module CLI failure path.

src/winml/modelkit/session/ep_registry.py
```diff
@@ -554,7 +561,11 @@ def auto_device(self, target: EPDeviceTarget) -> WinMLEPDevice:
                 last_error = e
                 continue
             for device in winml_ep.devices:
-                if device.device_type == target_device_upper:
+                if device.device_type == target_device_upper and (
+                    device_luid is None
+                    or (get_ep_device_luid(device.ort_handle) or "").casefold()
```
zhenchaoni, line 622: The final exception currently depends only on the last candidate. If an earlier source registers successfully but does not expose the requested LUID, and a later source fails registration, this reports that all sources failed and raises `WinMLEPRegistrationFailed`; the correct result is `DeviceNotFound`, because at least one source was usable but lacked the requested adapter. Please track whether any candidate registered successfully separately from the last registration error, and add a regression test for “successful source without the LUID, followed by a registration failure.”

src/winml/modelkit/commands/perf.py
xieofxie: interesting.. from the ort device ep loop, it will report two in this case. What if select the first dml but set device_id to another? let me test

xieofxie: I don't think it has this option https://onnxruntime.ai/docs/execution-providers/DirectML-ExecutionProvider.html let me test..

src/winml/modelkit/session/ep_registry.py
```diff
@@ -554,7 +561,11 @@ def auto_device(self, target: EPDeviceTarget) -> WinMLEPDevice:
                 last_error = e
                 continue
             for device in winml_ep.devices:
-                if device.device_type == target_device_upper:
+                if device.device_type == target_device_upper and (
+                    device_luid is None
```
zhenchaoni, line 621: `device_luid is None` still makes enumeration order the default-selection policy, but that order is not defined. `register_ep()` preserves `ort.get_ep_devices()` order, while ORT's Windows hardware inventory is built through unordered containers; `winml sys` independently preserves the unsorted DXCore adapter-list order. On an Intel iGPU + NVIDIA dGPU system, this can therefore select either adapter, and it is not guaranteed to select the first GPU printed by `winml sys`.

Please define one deterministic GPU ranking and apply it in both paths before choosing/displaying the first GPU. For example, sort by the ORT hardware-device metadata field `DxgiHighPerformanceIndex` as an integer (ascending, so index 0 wins), with a stable LUID fallback when that metadata is unavailable. Please also add a regression test with the same iGPU/dGPU devices supplied in reversed enumeration order, asserting that `winml sys`'s first GPU and unfiltered `auto_device()` resolve to the same LUID.

src/winml/modelkit/commands/perf.py
xieofxie: oh, it working. worth check https://github.com/microsoft/onnxruntime/blob/9f913ae524b50217e9f7c09ce35b363832a44928/onnxruntime/core/providers/dml/dml_provider_factory.cc#L470

src/winml/modelkit/session/ep_registry.py
xieofxie: good idea! ref https://github.com/microsoft/onnxruntime/blob/9f913ae524b50217e9f7c09ce35b363832a44928/onnxruntime/core/platform/windows/device_discovery.cc#L421

src/winml/modelkit/commands/perf.py
xieofxie: Fixed in `6c1d25f6`. The ambiguity warning now evaluates the effective binding from provider options across candidates. If they uniquely identify a known adapter, the warning is suppressed; non-selector or unresolved options still warn without claiming the first device will be used.


Added regression coverage for `device_id` selecting either GPU with either default, plus real DML subprocess coverage. One intermittent monitor-LUID mismatch occurred during development and could not be reproduced; its cause remains uncertain.

src/winml/modelkit/commands/perf.py
xieofxie: Fixed in `6c1d25f6`. Module-mode device resolution now converts unavailable-device/LUID, conflicting-option, source-selection, and EP-registration errors into `ClickException` before model building. Added CLI regression coverage confirming a formatted error and nonzero exit instead of a raw traceback.

src/winml/modelkit/session/ep_registry.py
xieofxie: Fixed in `6c1d25f6`. `auto_device()` now tracks successful registrations independently of the last registration error. If any source registered successfully but none exposed the requested device/LUID, it raises `DeviceNotFound`. `WinMLEPRegistrationFailed` is reserved for cases where every candidate failed registration.


Added regressions for a successful source without a matching device/LUID followed by a registration failure.

src/winml/modelkit/session/ep_registry.py
xieofxie: Fixed in `6c1d25f6`. Runtime selection and `winml sys` now share GPU ordering: numeric `DxgiHighPerformanceIndex` ascending, then LUID as a stable tie-breaker. Missing or invalid indices sort after ranked GPUs, in LUID order.


`winml sys` retains DXCore as the physical inventory source and joins only ORT ranking metadata by LUID. Added reversed-enumeration regressions confirming that sys and unpinned runtime selection choose the same first GPU when their device inventories match. Explicit LUID selection still overrides the ranking; an EP exposing only a subset of installed GPUs may have a different default from the system-wide first GPU.

src/winml/modelkit/sysinfo/gpu_priority.py
```diff
@@ -0,0 +1,30 @@
...
+    """
+    index = None
+    if isinstance(high_performance_index, (str, int)) and not isinstance(
+        high_performance_index, bool
+    ):
+        try:
+            parsed = int(high_performance_index)
+            if parsed >= 0:
+                index = parsed
+        except ValueError:
```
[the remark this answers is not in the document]
xieofxie: Invalid optional ranking metadata leaves this GPU unranked; use LUID ordering.

src/winml/modelkit/commands/perf.py
```diff
@@ -441,6 +442,69 @@ def _get_monitor_binding(
...
+        if len(identities) > 1:
+            if provider_options:
+                effective_bindings = {
+                    _get_ep_device_binding(replace(selected, device=device), provider_options)
+                    for device in candidates
+                }
+                if len(effective_bindings) == 1:
+                    effective_luid, _ = next(iter(effective_bindings))
+                    if effective_luid is not None:
+                        return selected
```
zhenchaoni: `provider_options` currently only decides whether to suppress the ambiguity warning; when the unique effective binding is a different adapter, this still returns `selected`. For example, if `auto_device()` resolves Intel GPU 0 but `--ep-options device_id=1` binds NVIDIA GPU 1, the new test deliberately keeps `result is selected`. As a result, `_pre_bench_kwargs_from_ep_device()` and `WinMLSession`'s device identity/logging report the Intel adapter even though the provider options and monitor target NVIDIA.

Please resolve the unique binding back to its candidate and return a `WinMLEPDevice` for that adapter (or carry an explicit effective device through every downstream consumer). Add a regression asserting that the returned device/hardware name/LUID matches the provider-selected adapter, rather than only checking warning suppression and the separately inferred monitor LUID. Provider selectors that change device kind, such as `--device npu` with QNN `backend_type=gpu`, should likewise either update the resolved target before the build or be rejected so build and runtime cannot disagree.

xieofxie: Fixed in `f0981295`. Provider-option resolution now returns a `WinMLEPDevice` wrapping the actual selected adapter, so the pre-benchmark hardware name, session identity/logging, and monitoring all refer to the same device.


Selectors that change the resolved device kind—such as `--device npu --ep qnn --ep-options backend_type=gpu`—are rejected before the build, with guidance to use the matching `--device`.


Updated regression coverage to assert the returned adapter, hardware name, LUID, and pre-benchmark identity, rather than preserving `result is selected`. Also covered adapter selection without LUID metadata and cross-kind rejection in single-model and module paths.

src/winml/modelkit/commands/perf.py
```diff
@@ -393,11 +392,23 @@ def _get_ep_device_binding(
...
+    from ..sysinfo import get_ep_device_luid
+
+    device, device_kind = _get_provider_bound_device(ep_device, provider_options)
+    if device is None:
+        return None, device_kind
+    has_provider_selector, _ = _get_provider_selected_device(ep_device, provider_options)
     if device_kind not in ACCELERATOR_DEVICE_TYPES:
         return (None, device_kind) if has_provider_selector else (None, None)
-    return luid, device_kind
+    return get_ep_device_luid(device.ort_handle), device_kind
```
zhenchaoni: This return value loses the distinction between “provider options uniquely selected a concrete adapter that has no LUID” and “no concrete adapter was identified.” For example, `device_id=1` can make `_get_provider_bound_device()` return GPU 1, but DML's `device_id` is not a catalog device-kind hint; with no LUID this becomes `(None, "gpu")`, and `_get_monitor_binding()` returns `("gpu", None, None)`. `PdhPoller` then calls `resolve_adapter_luid("gpu", ep_name=...)`; because ORT supplied no LUIDs, that path falls back to `discover_gpu_luid()` and can monitor GPU 0 while inference runs on GPU 1. The same case also retains the multiple-device warning even though the provider option uniquely selected a device.

Please carry whether provider options uniquely identified a concrete candidate independently of whether that candidate has a LUID. If it has no monitorable LUID, disable accelerator monitoring (CPU/RAM only) instead of performing heuristic adapter discovery, and suppress/reword the default-device warning accordingly. Extend `test_provider_selection_preserves_identity_without_luid` to assert the monitor binding so this cannot silently report metrics for a different GPU.

xieofxie: Agreed—the code incorrectly treated missing LUID metadata as missing device selection.


Provider resolution now tracks a uniquely selected candidate independently of its LUID. An explicit selection suppresses the ambiguity warning even without LUID metadata. If the bound adapter has no LUID, adapter-specific monitoring and VRAM sampling are disabled rather than guessing another GPU.


Extended the no-LUID regression to cover the returned identity, warning suppression, CPU/RAM monitor binding, and absence of heuristic VRAM discovery.


Separately, `--device-luid` combined with `--ep-options device_id=...` is now rejected early with a Click usage error; other tuning options remain allowed.

zhenchaoni (approved):

PR #1400, opened 2026-09-07
Make E2E timeouts download-aware and prioritize model execution

### LINKED ISSUE #1399: Make E2E timeouts download-aware and prioritize model execution
## Problem

The E2E evaluation runner currently counts Hugging Face model and dataset download time against each subprocess timeout. Large downloads can therefore report TIMEOUT even while network progress is being made, while stalled downloads are not consistently retryable as HF_FETCH_FAIL.

Evaluation order also follows registry input order, so lower-priority models can run before P0 models.

## Expected behavior

- Monitor Hugging Face partial-download cache files for size or mtime progress.
- Exclude active Hugging Face download time from `--timeout` and restart the full execution budget after the download completes.
- Fail a download after 600 seconds without progress by default, expose the threshold as a CLI option, and classify the result as `HF_FETCH_FAIL` so retry filtering can select it.
- Sort all models by P0, P1, P2, P3, then by case-insensitive Hugging Face model ID.
- Preserve the existing task and precision order within the same model.

## Acceptance criteria

- Timeout and download-stall behavior is covered by subprocess tests.
- Scheduling order is covered by a shuffled model-list test.
- Runner unit tests and Ruff checks pass.

fangyangci:
## Summary

- monitor Hugging Face `.incomplete` cache files and exclude active downloads from the normal subprocess timeout
- restart the full execution timeout after a Hugging Face download completes
- fail downloads after 600 seconds without progress by default and classify them as `HF_FETCH_FAIL` for retry selection
- schedule E2E models by P0, P1, P2, P3, then by case-insensitive model ID
- document the new timeout option and add regression tests

## Testing

- `D:\Dev\MKs\ModelKit\.venv\Scripts\python.exe -m pytest tests/unit/eval/test_run_eval_script.py -q` (182 passed)
- `D:\Dev\MKs\ModelKit\.venv\Scripts\python.exe -m ruff check scripts/e2e_eval/run_eval.py scripts/e2e_eval/utils/classifier.py tests/unit/eval/test_run_eval_script.py`
- `git diff --check`

Fixes #1399

### DIFF
--- a/scripts/e2e_eval/run_eval.py
+++ b/scripts/e2e_eval/run_eval.py
@@ -623,19 +623,34 @@ def _kill_process_tree(pid: int) -> None:
             pass  # Process already exited; nothing to kill
 
 
+def _expand_cache_path(path: str | os.PathLike[str]) -> Path:
+    return Path(os.path.expandvars(os.fspath(path))).expanduser()
+
+
+def _hf_cache_roots(env: dict[str, str]) -> tuple[Path, Path, Path]:
+    """Resolve cache roots using Hugging Face's environment precedence."""
+    if "HF_HOME" in env:
+        hf_home = _expand_cache_path(env["HF_HOME"])
+    else:
+        xdg_cache = _expand_cache_path(env.get("XDG_CACHE_HOME", Path.home() / ".cache"))
+        hf_home = xdg_cache / "huggingface"
+
+    hub_cache_value = env.get("HF_HUB_CACHE")
+    if hub_cache_value is None:
+        hub_cache_value = env.get("HUGGINGFACE_HUB_CACHE")
+    hub_cache = _expand_cache_path(hub_cache_value or hf_home / "hub")
+    datasets_cache = _expand_cache_path(env.get("HF_DATASETS_CACHE") or hf_home / "datasets")
+    xet_cache = _expand_cache_path(env.get("HF_XET_CACHE") or hf_home / "xet")
+    return hub_cache, datasets_cache, xet_cache
+
+
 def _snapshot_hf_downloads(env: dict[str, str]) -> dict[Path, tuple[int, int]]:
     """Return observable Hugging Face partial downloads as size/mtime pairs."""
-    hf_home = Path(env.get("HF_HOME", Path.home() / ".cache" / "huggingface")).expanduser()
-    hub_cache = Path(
-        env.get("HF_HUB_CACHE")
-        or env.get("HUGGINGFACE_HUB_CACHE")
-        or hf_home / "hub"
-    ).expanduser()
-    datasets_cache = Path(env.get("HF_DATASETS_CACHE") or hf_home / "datasets").expanduser()
+    hub_cache, datasets_cache, xet_cache = _hf_cache_roots(env)
     searches = (
         (hub_cache, ("*/blobs/*.incomplete", "*.incomplete")),
         (datasets_cache, ("downloads/*.incomplete",)),
-        (hf_home / "xet", ("**/*.incomplete",)),
+        (xet_cache, ("**/*.incomplete",)),
     )
     snapshot: dict[Path, tuple[int, int]] = {}
     for root, patterns in searches:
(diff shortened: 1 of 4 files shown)


scripts/e2e_eval/run_eval.py
```diff
@@ -619,22 +623,89 @@ def _kill_process_tree(pid: int) -> None:
...
+        self.last_progress = now
+
+    def poll(self, now: float) -> bool:
+        current = _snapshot_hf_downloads(self._env)
+        progressed = {
+            path for path, state in current.items() if self._previous.get(path) != state
+        }
+        if progressed:
+            self._active_paths.update(progressed)
+            self.last_progress = now
```
DingmaomaoBJTU, line 726: **[P2] Only suspend the timeout for this subprocess's downloads**

The snapshot covers the shared Hugging Face cache, so files changed by an unrelated process also enter this tracker's active set. With a sibling downloader writing there, a child that only sleeps returns successfully after about 1.2 seconds despite `timeout=0.5`. A hung build or perf job can therefore avoid its execution timeout for as long as unrelated downloads keep progressing. Please associate download activity with the current job/process tree before suspending or resetting its budget, and cover a downloader running outside the monitored process tree.

scripts/e2e_eval/run_eval.py
```diff
@@ -619,22 +623,89 @@ def _kill_process_tree(pid: int) -> None:
             pass  # Process already exited; nothing to kill
 
 
+def _snapshot_hf_downloads(env: dict[str, str]) -> dict[Path, tuple[int, int]]:
+    """Return observable Hugging Face partial downloads as size/mtime pairs."""
+    hf_home = Path(env.get("HF_HOME", Path.home() / ".cache" / "huggingface")).expanduser()
```
DingmaomaoBJTU: **[P2] Honor Hugging Face's XDG cache fallback**

When `HF_HOME` is unset, `huggingface_hub` resolves its home under `XDG_CACHE_HOME` if that variable is set. This fallback instead always uses `~/.cache/huggingface`. With only `XDG_CACHE_HOME` configured, the watcher misses the actual Hub `.incomplete` files, so a progressing download still hits the normal execution timeout and stalled downloads bypass the new stall classification. Please match Hugging Face's cache-resolution precedence and cover the XDG-only configuration; the current subprocess cases all explicitly set `HF_HOME`.

scripts/e2e_eval/run_eval.py
fangyangci: Fixed in 8eae21a2. The download tracker now binds to the monitored subprocess PID and only treats changed partial files opened by that process or its descendants as active. Added a regression test where an unrelated sibling downloader writes the shared cache while the monitored child sleeps; the child now hits its 0.5s execution timeout.

scripts/e2e_eval/run_eval.py
fangyangci: Fixed in 8eae21a2. Cache root resolution now follows Hugging Face precedence: HF_HOME, otherwise XDG_CACHE_HOME/huggingface, otherwise ~/.cache/huggingface. It also honors HF_HUB_CACHE over the legacy HUGGINGFACE_HUB_CACHE and supports HF_XET_CACHE. Added XDG-only tests for both progressing and stalled downloads.

scripts/e2e_eval/run_eval.py
```diff
@@ -619,22 +623,143 @@ def _kill_process_tree(pid: int) -> None:
...
+    def bind(self, pid: int) -> None:
+        self._pid = pid
+
+    def poll(self, now: float) -> bool:
+        current = _snapshot_hf_downloads(self._env)
+        open_paths = _process_tree_open_paths(self._pid) if self._pid is not None else set()
+        progressed = {
+            path
+            for path, state in current.items()
+            if self._previous.get(path) != state and _normalized_path(path) in open_paths
```
DingmaomaoBJTU: **[P2] Keep ownership and progress observations in sync**

`current` is captured before the process-tree handle scan, which takes about 0.65–0.8 s on my Windows host. In the restart cases, the first poll snapshots no partial file, then sees it open; the next snapshots the growing file, but enumeration finishes after it closes. The intersection is empty both times, so the download never pauses/resets the execution budget and the child is reported as `TIMEOUT`. Both existing `test_execution_timeout_restarts_after_hf_download` and its XDG variant fail this way on `8eae21a2`. Please make the ownership/progress sampling tolerate this interleaving without accepting unrelated downloads, and cover a slow handle scan.

DingmaomaoBJTU (commented):
One remaining issue with download ownership/progress sampling.

scripts/e2e_eval/run_eval.py
fangyangci: Fixed in 03d5d0aa. The tracker now completes the potentially slow process-tree handle scan before taking the cache progress snapshot, so ownership and file state are sampled at adjacent points while the process-tree ownership filter remains unchanged. Added a regression test that delays the real handle scan by 0.7s; the full subprocess timeout test class passes (7 tests).

DingmaomaoBJTU (approved):
Re-reviewed 03d5d0aa. The previous download ownership, XDG cache resolution, and sampling-order findings are addressed. No remaining blocking findings.

PR #1403, opened 2026-09-09
Fix SAM wrapper resolution after task normalization

### LINKED ISSUE #1402: SAM wrappers fail explicit model-class resolution after task normalization
## Problem

Explicit SAM wrapper resolution fails when `task="mask-generation"` and `model_class` is `SAMMaskGeneration` or `SAM2MaskGeneration`. The loader normalizes the task to `feature-extraction` before querying the ModelKit custom model registry, so it misses the wrappers registered under the original task and incorrectly falls back to Transformers.

This produces errors such as:

```text
Model class SAMMaskGeneration not found for task feature-extraction
```

## Root cause

The explicit `model_class` path in `resolve_task` only queries `_get_custom_model_class` with the normalized Optimum task. ModelKit custom wrappers are registered against the original `mask-generation` task.

## Expected behavior

Custom model-class lookup should try the user-provided task before the normalized task, while preserving the normalized task for the existing Optimum fallback.

## Acceptance criteria

- Explicit `SAMMaskGeneration` resolves from `winml.modelkit.models.hf.sam`.
- Explicit `SAM2MaskGeneration` resolves from `winml.modelkit.models.hf.sam`.
- Existing resolver tests continue to pass.
- Intel OpenVINO NPU E2E reruns no longer report the model-class lookup error.

## Validation

- `tests/unit/loader/test_resolve_task.py`: 22 passed.
- Targeted E2E rerun: 0/20 results retain the class lookup error; 18/20 perf jobs pass. The remaining `wanglab/medsam-vit-base` failures are unrelated TensorFlow-only weight export failures.

fangyangci:
## Summary

- resolve explicit ModelKit custom wrappers using the original user task before the normalized Optimum task
- keep the normalized-task lookup as a fallback
- add regression coverage for `SAMMaskGeneration` and `SAM2MaskGeneration`

## Validation

- `D:\Dev\MKs\ModelKit\.venv\Scripts\python.exe -m pytest tests\unit\loader\test_resolve_task.py -q`
  - 22 passed
- Intel OpenVINO NPU targeted rerun for 10 SAM/SAM2 model keys across `w8a8` and `w8a16`
  - original model-class lookup error: 0/20
  - perf: 18/20 passed
  - remaining two failures are `wanglab/medsam-vit-base` TensorFlow-only weight export failures and are unrelated to wrapper resolution

Fixes #1402


DingmaomaoBJTU (commented):
Please address the two issues noted inline.

tests/unit/loader/test_resolve_task.py
```diff
@@ -106,6 +106,27 @@ def test_user_class_unknown_raises_friendly_error():
...
+def test_user_class_custom_wrapper_uses_original_task(
+    model_type, architecture, model_class
+):
+    r = resolve_task(
+        _cfg(model_type, [architecture]),
+        task="mask-generation",
+        model_class=model_class,
+    )
+
+    assert r.task == "image-feature-extraction"
```
DingmaomaoBJTU, line 168: [P1] Preserve the registered task in the returned resolution

This assertion locks in a task/class pair that the loader cannot use. With `task="mask-generation"` and `model_class="SAM2MaskGeneration"`, `resolve_loader_config()` now returns `image-feature-extraction` alongside the decoder wrapper. Feeding its returned task and class back into `resolve_task()` raises the original class-not-found error again. The returned task also makes `resolve_export_config()` select `Sam2ImageEncoderIOConfig` (only `pixel_values`) for a decoder that requires prompts, embeddings, and mask inputs. SAM and SAM2-video have the same round-trip failure. Once the custom wrapper matches the original registered task, please preserve that task in `TaskResolution`, including `optimum_task`, and cover loader-config round-tripping and decoder I/O selection rather than asserting image feature extraction.

src/winml/modelkit/loader/resolution.py
```diff
@@ -536,7 +536,9 @@ def resolve_task(
         # CLIP --model-class override).
         resolved = None
         if model_type_norm:
-            custom = _get_custom_model_class(model_type_norm, opt_task)
+            custom = (
+                _get_custom_model_class(model_type_norm, task) if task is not None else None
+            ) or _get_custom_model_class(model_type_norm, opt_task)
```
DingmaomaoBJTU: [P2] Check the requested class name before short-circuiting the fallback

The `or` only tries `opt_task` when the original lookup is empty, not when it returns a different class from the explicit override. For `sam2`/`sam2_video`, `task="mask-generation", model_class="Sam2VisionEncoder"` resolves on the base revision through the normalized `feature-extraction` registration. Here the original lookup finds `SAM2MaskGeneration`, short-circuits the fallback, then fails the name check and falls into Transformers, where `Sam2VisionEncoder` does not exist. Please try the normalized registration after an original-candidate name mismatch too, so the existing explicit-class override behavior is preserved instead of raising a new `ValueError`.

tests/unit/loader/test_resolve_task.py
fangyangci: Fixed in a96aee9c. A custom wrapper matched on the original registered task now preserves that task and its optimum_task. Added SAM, SAM2, and SAM2-video loader-config round-trip coverage plus decoder ONNX config/input selection checks.

src/winml/modelkit/loader/resolution.py
fangyangci: Fixed in a96aee9c. Custom task candidates are now checked individually against the requested class name, so a mismatch on the original task continues to the normalized task. Added sam2 and sam2_video encoder-override regression coverage.

src/winml/modelkit/loader/resolution.py
```diff
@@ -536,9 +536,16 @@ def resolve_task(
...
-                resolved = custom
+            candidate_tasks = [opt_task]
+            if task is not None:
+                candidate_tasks.insert(0, task)
+            for candidate_task in candidate_tasks:
+                custom = _get_custom_model_class(model_type_norm, candidate_task)
+                if custom is not None and custom.__name__ == model_class:
+                    resolved = custom
+                    if task is not None and candidate_task == task:
+                        surfaced = task
```
DingmaomaoBJTU: [P2] Keep the modality upgrade for canonical feature-extraction

This also overwrites `surfaced` when the original and normalized tasks are identical. With a `Sam2Model` config, `task="feature-extraction", model_class="Sam2VisionEncoder"` returns `image-feature-extraction` on the base revision but `feature-extraction` here; SAM2-video has the same regression. The loader propagates that task into quantization, where it selects `TextDataset` instead of `ImageDataset`, losing the intended image calibration and potentially falling back to random data. Please retain the modality upgrade for these existing encoder overrides while preserving the decoder's registered task, for example by applying `_resolve_task_modality(config, task)` rather than assigning the raw task. The new fallback test starts from `mask-generation`, so it does not cover this equal-task case.

fangyangci: Fixed in e5334afd. The original-task custom-wrapper match now uses `_resolve_task_modality(config, task)`, retaining `image-feature-extraction` for canonical encoder overrides while preserving the registered `mask-generation` decoder task. Added config-only regressions for `sam2` and `sam2_video` that compare canonical-task resolution with the existing normalized-task fallback; both failed before the fix and pass now. Validation: 71 focused loader/task-resolution tests passed, including the existing decoder loader-config round-trip and ONNX I/O selection coverage. Ruff lint and format checks also passed.

DingmaomaoBJTU (approved):

PR #1404, opened 2026-09-09
fix: honor device and EP in GenAI perf auto-builds

KayMKM:
## Summary

Fix `winml perf --runtime ort-genai --device cpu` incorrectly building the HuggingFace model with QNN/NPU before attempting CPU inference.

- Resolve the runtime EP and concrete device before auto-building and pass the same target to the transformer build. Retain the concrete device selected by `auto`, including multi-device EPs.
- Keep explicit EP/device bundle caches separate from each other and the legacy default QNN bundle. Preserve no-override defaults and prebuilt-bundle behavior.
- Declare CPU support in the Qwen3 bundle recipe and reject unsupported automatic build targets before export rather than silently producing a CPU-routed accelerator bundle.
- Cover target propagation, cache reuse/rebuild, aliases, multi-device EPs, unavailable/incompatible targets, and CPU optimized builds; update the command and sample documentation.

## Reproduction

```powershell
uv run winml perf -m Qwen/Qwen3-0.6B --runtime ort-genai --device cpu `
  --no-compile --warmup 2 --iterations 10 --max-new-tokens 20 `
  --prompt "What is the capital of France?" `
  --output temp\qwen3-genai-cpu\perf-genai-cpu.json --overwrite --no-color
```

Before: the auto-build logged `device=npu, precision=w8a16` and failed with `WinMLEPNotDiscovered: No EPEntry discovered for ep='QNNExecutionProvider'`.

After: ran from an isolated Python 3.11.9 environment with fresh WinML/HuggingFace caches. Both the initial CPU build/perf run and the subsequent cache-hit run exited 0. The JSON reports `ep=cpu`, `effective_device=cpu`, compilation disabled, 2 warmup requests and 10 timed requests, with 20 generated tokens per request. The final run measured mean model TTFT 446.68 ms and steady-state decode throughput 7.33 tokens/s on the local test machine.

## Validation

218 targeted pytest cases passed, covering GenAI perf/build, bundle registry/orchestration/configuration, related eval behavior, and CLI help import isolation. Targeted Ruff and mypy checks passed. No full-suite or hardware accelerator validation was performed.


DingmaomaoBJTU (approved):
Looks good to me. The auto-build now uses the selected runtime target, and EP/device-specific caches avoid reusing incompatible bundles. The no-override default and prebuilt-bundle routing are preserved. No blocking issues found.

DingmaomaoBJTU (approved):
Re-checked the latest commit. The new end-to-end coverage exercises a fresh CPU auto-build, canonical EP alias cache reuse without Hugging Face access, and automatic runtime detection for prebuilt bundles. No blocking issues found.

PR #1406, opened 2026-09-09
Add TensorRT RTX operator tracing to perf

xieofxie:
## Summary

- Add basic TensorRT RTX GPU op-tracing, including warmup exclusion, per-run statistics, and raw trace artifacts.
- Preserve native layer names and timings; label multi-node layers as `Fused` and include contributing ONNX node names/types in JSON.
- Report TensorRT RTX EP 2.30.49 as the minimum profiling version.
- Default all op-tracing to 10 measured iterations while preserving explicit overrides.
- Add unit and GPU integration coverage, plus documentation.

## Usage

```powershell
winml perf -m model.onnx --device gpu --ep nv_tensorrt_rtx --op-tracing basic
```

The default run uses 10 warmup iterations followed by 10 measured iterations.
Results are included under `hw_monitor.ep_proof` in the benchmark JSON.


src/winml/modelkit/session/monitor/trtrtx_monitor.py
```diff
@@ -0,0 +1,250 @@
...
+            if self._measured_iterations is not None
+            else None
+        )
+        for pid, runs in contexts.items():
+            if stop is not None and len(runs) < stop:
+                raise ValueError(
+                    f"TensorRT RTX context {pid} has {len(runs)} runs; "
+                    f"expected at least {stop} including warmup"
+                )
+            retained = list(runs.values())[self._warmup_iterations : stop]
```
DingmaomaoBJTU, line 226: **[P2] Align warmup filtering with full model runs**

The [native profiler](https://github.com/NVIDIA/TensorRT-RTX-EP-ABI/blob/7bc9addc928590802c4d95589ac3caa07a9cf517/src/tensorrt_rtx_profiler.h#L39-L75) increments `tid` for each subgraph enqueue, not necessarily once per model inference. The EP also supports [partially offloading control-flow bodies](https://github.com/NVIDIA/TensorRT-RTX-EP-ABI/blob/7bc9addc928590802c4d95589ac3caa07a9cf517/src/tensorrt_rtx_execution_provider.cc#L3444-L3446), so a loop body can produce multiple `tid`s per `session.run()`.

For a body executed twice per inference, `warmup=1, measured_iterations=2` produces six invocations, but this slice retains only `tid=1,2`: one is still warmup, and most measured work is discarded. The result nevertheless reports two samples and `status="ok"`.

Please correlate events with whole-inference boundaries before filtering and aggregating, or explicitly reject contexts whose invocation counts cannot be aligned with the perf window instead of silently accepting extra runs.

xieofxie: fixed

DingmaomaoBJTU (approved):
Reviewed the TensorRT RTX monitor, perf dispatch/default changes, metadata serialization, and session teardown/perf-window integration. No blocking findings. All currently reported PR checks pass. Locally, 47 parser/lifecycle tests passed in an isolated harness (24 dispatch/availability cases deselected; the CLI serializer adapter was substituted). Native TensorRT RTX GPU inference was not rerun locally. The documented constant-invocation assumption and its warning make the variable-control-flow attribution limitation explicit.

PR #1412, opened 2026-09-11
Integrate device selection in eval command

zhenchaoni:
## Summary

Add physical GPU selection by adapter LUID to `winml eval`, consistent with the existing `winml perf --device-luid` behavior.

Candidate and reference models can select GPUs independently, enabling tensor-similarity comparisons between two physical adapters using the same ONNX model.

## Changes

- Add `--device-luid` to select the candidate model's GPU.
- Add `--reference-device-luid` to select the reference model's GPU.
- Persist both LUIDs in eval configuration and JSON reports.
- Forward the selected LUID through the EP registry when creating the model session.
- Prevent an explicitly selected candidate GPU from silently falling back to CPU.
- Share lightweight LUID option validation between `winml perf` and `winml eval`.
- Document single-GPU and cross-GPU evaluation examples.

## Validation

- Ruff checks passed.
- Eval-related unit and CLI tests passed.
- Existing ResNet-50 CPU and streaming eval scenarios passed.
- Compared ResNet-50 inference across NVIDIA and Intel DML devices using 10 samples.
- NVIDIA candidate versus Intel reference produced a minimum cosine similarity of `0.999999999999987`.


src/winml/modelkit/eval/evaluate.py
```diff
@@ -391,7 +391,12 @@ def load_model(
     # when either is 'auto'.
     device = (config.device or "auto").lower()
     target = resolve_device(EPDeviceTarget(ep=config.ep or "auto", device=device))
-    ep_device = WinMLEPRegistry.instance().auto_device(target)
+    registry = WinMLEPRegistry.instance()
+    ep_device = (
+        registry.auto_device(target, device_luid=config.device_luid)
+        if config.device_luid is not None
+        else registry.auto_device(target)
```
DingmaomaoBJTU, line 400: [P2] Handle LUID selection on the loaders that return before this block

The `GENAI` and `EVALUATOR_MANAGED` branches return before this registry lookup. Consequently, `winml eval -m <bundle> --task text-generation --device gpu --device-luid <LUID>` accepts the pin but `_load_genai_causal_lm()` only forwards `ep`/`device`. Composite mask-generation evaluation similarly constructs its own ORT sessions without consulting `config.device_luid`. Both can evaluate on a different adapter while the JSON report records the requested LUID.

Please reject a non-null LUID for loaders that cannot honor it (as `perf` already does for `ort-genai`), or wire the selected adapter through their session creation. This needs to happen before the early returns; coverage of the normal `WinMLAutoModel` path won't catch it.

src/winml/modelkit/commands/eval.py
```diff
@@ -598,6 +609,7 @@ def _resolve_model_loader_task(cfg: WinMLEvaluationConfig) -> None:
     "input_data",
     "reference_path",
     "ep",
+    "device_luid",
```
DingmaomaoBJTU, line 612: [P2] Apply the PyTorch LUID restriction at the Python API boundary too

Adding `device_luid` to this CLI-only list rejects the command/config-file cases, but `evaluate(WinMLEvaluationConfig(runtime="pytorch", device="gpu", device_luid=..., ...))` bypasses it. The shared `_validate_pytorch_runtime_config()` in `eval/evaluate.py` checks `ep`, precision, and the other ONNX-only settings, but does not check this new field. The native loader then chooses the normal CUDA device without using the LUID, and the result still serializes the requested pin.

Please add `device_luid` to that shared validator as well and cover direct `evaluate()` calls, so the public API rejects unsupported adapter selection rather than silently evaluating on another device.

src/winml/modelkit/eval/evaluate.py
zhenchaoni: These 2 evaluator loader didn't integrate with the entire ep registry feature since day 1 of ep registry. Integrating them will make this PR even larger. I will track this in #1413

src/winml/modelkit/commands/eval.py
zhenchaoni: Fixed in c19b99df. Added `device_luid` to `_validate_pytorch_runtime_config()` so direct `evaluate()` calls now reject LUID selection with the PyTorch runtime before model loading. Added direct API regression coverage in `TestEvaluate::test_pytorch_runtime_rejects_device_luid`. The focused PyTorch eval suite passes (43 tests).

DingmaomaoBJTU (commented):
Re-reviewed c19b99df. The direct Python API / PyTorch LUID validation finding is addressed by the shared validator change and direct evaluate() regression coverage.

I am withholding approval because the other finding remains: GenAI and evaluator-managed loaders still accept the newly introduced --device-luid without honoring it, and the result can record a pin that was not used. Tracking full EP-registry integration in #1413 is reasonable, but the silent acceptance is introduced by this PR, not by the pre-existing lack of integration.

This does not require integrating those loaders here: please reject a non-null device_luid before the unsupported loader branches return, with coverage for both branches. Full adapter support can remain in #1413.

DingmaomaoBJTU (approved):
Approved with the remaining GenAI/evaluator-managed LUID handling limitation tracked in #1413.

PR #1414, opened 2026-09-11
fix(perf): preserve memory baselines and distinguish unavailable samples

KayMKM:
## Problem and behavior

Failed or absent process memory counters currently look like measured zero, and the existing post-factory baseline excludes allocations made while loading the model. This change preserves the legacy baseline, deltas and three-checkpoint peak while adding a separate pre-model snapshot and deltas. If no GPU process-memory instance exists before loading, the added GPU delta is null; valid legacy-baseline deltas remain available.

## Changes

- Add `*_before_model_load_mb`, `*_model_factory_delta_mb` and `*_total_from_before_model_load_delta_mb` for RSS and local/shared GPU memory. Keep the original baseline after the model factory and before input generation/explicit compilation. Preloaded components do not inherit an invented pre-model baseline.
- Emit `memory_measurement` v2 with process identity, selected LUID, phase timestamps, raw bytes, source and validity/reason metadata. Keep private commit distinct from RSS.
- Preserve null for unavailable readings and dependent deltas/peaks; preserve valid zeros. Accept both valid PDH data statuses, retain counter diagnostics, enumerate matching PID/LUID physical memory instances and close failed queries.
- Keep main's device-selection behavior. Make monitor means/peaks, console/live output and the GenAI consumer safe for unavailable memory.

## Validation

- 420 related tests passed: memory tracker, EP monitor, perf CLI/module/GenAI and device-LUID suites.
- Whole-package mypy (-p winml.modelkit), Ruff, changed-file license headers and git diff --check passed.
- Real DirectML runs pinned separately to AMD Radeon 890M and NVIDIA RTX 4070 Laptop: deterministic 4/64 MiB FP32-weight ONNX MatMul models, three fresh-process large-model repetitions per GPU, 5 warmups and 3-second monitored windows.
- Main's original helper plus original PDH implementation, the new sampler and an independent PowerShell counter reader agreed in each live worker. Large-model local/shared values were 138.5390625/133.2890625 MiB on AMD and 79.296875/91.15625 MiB on NVIDIA. Querying the other adapter returned unavailable rather than another GPU's usage. Output checks against NumPy passed.

## Compatibility and limits

Legacy field names retain their sampling boundary and formulas; unavailable values intentionally become null. Added measurements are checkpoint differences, not continuous peaks or weights-only footprint. GPU local/shared accounting differs on UMA and must not be added blindly to RSS. Hardware validation used the existing local ORT Windows ML 1.28.0 environment, outside main's <1.28 dependency constraint; other EPs and the full classic matrix were not certified by these runs.

### DIFF
--- a/src/winml/modelkit/session/monitor/_pdh.py
+++ b/src/winml/modelkit/session/monitor/_pdh.py
@@ -21,6 +21,7 @@ import ctypes
 import ctypes.wintypes as wintypes
 import logging
 import os
+import re
 import statistics
 import sys
 import threading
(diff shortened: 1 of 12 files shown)


src/winml/modelkit/session/monitor/_pdh.py
```diff
@@ -309,16 +353,18 @@ def build_adapter_query(
...
-    query.add_counter(
-        "memory_shared_bytes",
-        rf"\GPU Process Memory(pid_{pid}_luid_{luid}_phys_0)\Shared Usage",
-        fmt="large",
-    )
+    try:
+        instances = memory_instances(pid, luid)
+    except RuntimeError:
+        instances = []
+    for index, instance in enumerate(instances):
```
DingmaomaoBJTU: [P2] Keep monitoring memory instances that appear after model load. GenaiPerfBenchmark enters HWMonitor before _run_benchmark_body calls session.load(), so the current PID may have no GPU Process Memory instance when this enumeration runs. If instances is empty, no memory counters are registered, and subsequent polling never discovers them. Consequently, device-memory means and peaks remain unavailable for the entire run even after loading and generation allocate GPU memory. Please support late-created instances, for example by retrying discovery or registering a counter path that PDH can resolve once the instance appears, while preserving unavailable readings until data becomes valid. Add coverage for an initially absent process-memory instance becoming available after model load.

DingmaomaoBJTU (approved):
Re-reviewed the changes since a953429, including periodic PID/LUID instance discovery, registration retries, disappearance handling, sample coverage reporting, checkpoint retries, and the deterministic timeout-test update. The prior late-created GPU memory instance finding is addressed: discovery now refreshes during collection, and recovered measurements retain the earlier missing-coverage evidence. No remaining blocking findings. Seven late-memory unit tests passed locally in an isolated harness (the GenAI integration case was excluded), and git diff --check passed. All reported PR CI checks are green. Native GPU/GenAI hardware execution was not rerun locally.

PR #1415, opened 2026-09-13
feat(skills): add model support contribution workflow

ssss141414:
## Summary
- migrate the operational `adding-model-support` skill from `gim-home/ModelKitArtifacts` PR #352; the imported corpus originated at `9b51a3ea025e4789cf00ba7a8cf0c7844e7e1de5`, and shared workflow improvements are synchronized back to source commit `c9903e407207cd7f43f95c71b31c1df02f04fe7d`
- preserve 45 workflow, agent, evaluation, model-knowledge, methodology, cleanup, and contract-test files while removing the unreferenced stale story snapshot
- adapt active instructions to the repo-local `skills/adding-model-support/` layout and preserve historical provenance
- use a concise dispatcher plus explicit fresh-subagent, optional dependency, and fail-closed promotion contracts
- align response evals with the active reviewer verdict contract

## Validation
- trigger metadata classified all 18 trigger fixtures correctly in an independent frontmatter-only smoke evaluation
- response smoke evaluation correctly returned a normal-comment `REQUEST_CHANGES` verdict and ownership routing for a fixable compatibility break
- passed all 5 promotion, dispatch, and reviewer contract assertions
- passed Ruff check and format verification for the Python contract test
- parsed all skill JSON and YAML frontmatter successfully
- verified relative Markdown links, EOF newlines, trailing whitespace, no exact duplicate files, and `git diff --check`
- compared both repositories: 45/45 paths match; the remaining nine content differences are intentional repository/dependency adaptations
- full `uv run pytest` and pre-commit bootstrap were attempted but dependency downloads were blocked by a local PyPI TLS handshake failure; equivalent available checks were run offline


DingmaomaoBJTU (commented):
Reviewed the workflow entry/role contracts and the executable cleanup helper. All reported CI checks are green and the five promotion-contract tests passed locally with repository plugin requirements disabled (missing torch warning-filter warnings). A dry-run of the unmodified cleanup script reproduces the issue below, so I am withholding approval. No cleanup was executed and no PR files were changed.

skills/adding-model-support/scripts/cleanup-run-cache.ps1
```diff
@@ -0,0 +1,424 @@
...
+    if (-not (Test-Path -LiteralPath $Path)) {
+        return [int64]0
+    }
+    $item = Get-Item -LiteralPath $Path -Force
+    if (-not $item.PSIsContainer) {
+        return [int64]$item.Length
+    }
+    $measurement = Get-ChildItem -LiteralPath $Path -Force -Recurse -File |
+        Measure-Object -Property Length -Sum
+    if ($null -eq $measurement.Sum) {
```
DingmaomaoBJTU, line 109: [P2] Handle an empty directory before reading the measurement property. Under Set-StrictMode -Version Latest, an empty Get-ChildItem pipeline can leave $measurement without a Sum property, so this null check itself throws PropertyNotFoundStrict. I reproduced this by passing a valid dry-run manifest whose candidate is an empty __pycache__ directory: Get-PathBytes aborts at this line instead of reporting zero bytes. Empty bytecode directories are explicitly eligible, so a normal terminal cleanup inventory can fail before writing its record and block the next model. Handle the no-measurement case before accessing .Sum, and add executable cleanup tests for empty directories as well as non-empty candidates.

ssss141414: Fixed in 55cc9628. `Get-PathBytes` now checks that the measurement and its `Sum` property exist before reading the value, so an empty directory records zero bytes under StrictMode. Added an executable regression covering both empty and non-empty `__pycache__` candidates, and added the skill test path to the existing Windows CI `remaining` partition. I also synchronized the reference-conditioned prediction-diversity fix from PR #352. Validation: 6 focused tests passed; Ruff, JSON validation, and diff checks passed.

DingmaomaoBJTU (approved):

PR #1416, opened 2026-09-14
fix(ci): stabilize GPU E2E tests on shared RDP agents

KayMKM:
RDP transitions on the shared NPU-OV2 agent leave duplicate GPU routes in ORT, including a route that fails with DXGI_ERROR_DEVICE_REMOVED. This causes DML evaluation failures and makes adapter-selection tests fail even when the physical GPU remains usable.

This change opts NPU-OV2 into physical-GPU selection in the E2E harness. DML Eval resolves the sole DXCore GPU and passes its current LUID to winml perf. The perf pytest step uses that GPU while preserving the requested EP, including OpenVINO GPU. Both adapter-selection cases still execute inference independently through --device-luid and --ep-options device_id, and verify the monitor binding. Missing hardware/LUIDs and unrelated extra devices still fail; normal agents retain strict inventory and default-selection checks. No tests are skipped or marked xfail, and shipped product code is unchanged.

The keepalive installer also includes the enhanced worker already deployed on the agent: network-source filtering, Session 1 support, occupied-console protection, verified tscon retries, and cleanup limited to non-present RDP display nodes. The task now passes the source address directly, and the documentation explains that console recovery does not prove DML readiness.

Validation:
- Eight targeted hardware E2E cases passed on NPU-OV2 with RDP connected: both selectors, DML binding/inference, DML and OpenVINO monitoring, and CLI argument validation; no skips (September 11).
- 201 evaluation-harness unit tests passed (September 11).
- 42 targeted checks passed before publication, including PowerShell parsing, mocked keepalive transitions, worker equivalence, and pinning regressions (September 14).
- Ruff and git diff --check passed; pipeline YAML parsed successfully.

The pin is opt-in through dmlPinnedAgents (initially NPU-OV2); an empty list disables it. It does not guarantee that in-flight inference survives a driver reset during an RDP switch.

### DIFF
--- a/.pipelines/Modelkit E2E Test.yml
+++ b/.pipelines/Modelkit E2E Test.yml
@@ -178,6 +178,11 @@ parameters:
     displayName: 'OV: self-hosted agent name (Agent.Name)'
     type: string
     default: NPU-OV2
+  - name: dmlPinnedAgents
+    displayName: 'Agents whose DML perf eval pins the sole physical GPU (empty disables)'
+    type: object
+    default:
+      - NPU-OV2
   - name: ovPairs
     displayName: 'OV: EP/device pairs (any of: dml_gpu, mlas_cpu, ov_cpu, ov_gpu, ov_npu)'
     type: object
@@ -223,6 +228,7 @@ stages:
         parameters:
           agentName: ${{ parameters.ovAgentName }}
           agentSuffix: ov
+          dmlPinnedAgents: ${{ parameters.dmlPinnedAgents }}
           pairNames: ${{ parameters.ovPairs }}
           models: ${{ parameters.models }}
           modelTimeout: ${{ parameters.modelTimeout }}
--- a/.pipelines/templates/e2e-test-jobs.yml
+++ b/.pipelines/templates/e2e-test-jobs.yml
@@ -21,6 +21,10 @@ parameters:
   #   mlas_cpu, ov_cpu, ov_gpu, ov_npu, trtrtx_gpu
   - name: pairNames
     type: object
+  # Opt-in for shared RDP development agents; affects DML perf eval and perf pytest.
+  - name: dmlPinnedAgents
+    type: object
+    default: []
   # Inline list of models to test. Each entry: { hf_id: <str>, task: <str> }.
   # `task` is optional — omit (or leave empty) to let winml infer.
   - name: models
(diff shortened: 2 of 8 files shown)


DingmaomaoBJTU (commented):
Reviewed the physical-GPU pinning and keepalive changes. One actionable issue in the new retry path is noted below. Validation included a Windows PowerShell 5.1 native-stderr reproduction; the shared-agent hardware E2E suite was not rerun.

scripts/agent_setup/setup_rdp_gpu_keepalive.ps1
```diff
@@ -73,36 +171,74 @@ try {
...
+      exit 0
     }
-    else {
-        # Non-fatal: the session state can change between the check and tscon.
-        Write-Log "WARN: tscon $target /dest:console exited $code (session likely changed state); ignoring."
+
+    Write-Log "Remote disconnect from $sourceAddress left session $target Disc; redirecting immediately."
+    $redirected = $false
+    for ($attempt = 1; $attempt -le 3; $attempt++) {
+      $out = & tscon $target /dest:console 2>&1
```
DingmaomaoBJTU: [P2] Keep native stderr from bypassing the tscon retry loop. The scheduled task runs powershell.exe (Windows PowerShell 5.1), and this worker sets $ErrorActionPreference to Stop. In that runtime, a native command writing stderr through 2>&1 can raise a terminating NativeCommandError before $LASTEXITCODE is examined. A failed tscon therefore goes straight to the outer catch and exits 0 instead of reaching the verification/retry logic. This was reproduced with a harmless native command writing stderr under the same PowerShell settings. Handle native stderr without terminating the loop (while retaining exit-code and session-state checks), and test a real native-stderr failure rather than only a mocked return code.

DingmaomaoBJTU (approved):
Re-reviewed the changes since 090f9cb. The earlier native-stderr finding is addressed for both tscon and pnputil: Continue is scoped to native invocation, the exit code is captured, and the original error preference is restored in finally. A harmless native-stderr reproduction in Windows PowerShell confirmed that three attempts complete, exit code 7 is retained, and Stop is restored. All reported PR checks pass; no remaining blocking findings. Shared-agent GPU/RDP end-to-end behavior was not rerun locally.

PR #1417, opened 2026-09-14
Fix fp16 quantization name collision

zhenchaoni:
## Problem

ORT’s FP16 converter generates deterministic I/O Cast names when preserving FP32 public I/O. Existing models, including the Affinity encoder and decoder, may already contain these names. Our collision guard consequently rejects conversion.

## Solution

Add a preprocessing step that renames conflicting internal tensors and nodes before calling ORT:

- Use inferred I/O types to identify the exact names ORT will generate.
- Allocate deterministic `__existing_N` names, avoiding existing and newly allocated names.
- Update tensor references and metadata, including nested captures, while respecting local scope.
- Preserve public I/O names and types, and leave conversion-neutral nodes unchanged.
- Retain the existing collision guard as a safety check.

This change only resolves naming conflicts; it does not change precision-conversion policy or prevent ORT from adding Casts during repeated conversion.


src/winml/modelkit/quant/fp16.py
```diff
@@ -2891,6 +2995,11 @@ def convert_to_fp16(
 
     _reject_sparse_initializer_tensor_metadata(model, op_block_list)
     _reject_duplicate_float_initializer_names(model, op_block_list)
+    _rename_generated_io_cast_name_collisions(
```
DingmaomaoBJTU, line 3001: [P2] Stage collision renames until conversion validation succeeds. This call mutates the caller-owned model before the remaining preflight guards, and before the existing safe-conversion deepcopy. A model containing an internal graph_input_cast_0 / graph_input_cast0 collision plus an unloaded external FLOAT weight now gets renamed and then raises the unloaded-external-data RuntimeError. I reproduced that SerializeToString() differs after rejection; without this preprocessing, rejection leaves the model unchanged. This regresses the existing fail-before-mutation behavior and leaves callers that catch conversion failures with changed tensor/node identifiers despite receiving no converted model. Perform the renaming and subsequent validation on a working copy, publishing it back only after successful conversion, and add a combined collision + rejected-conversion regression case.

DingmaomaoBJTU (commented):
Reviewed the collision preprocessing and its integration. The standalone FP16 test module passed all 148 tests with ONNX 1.18 and ORT 1.30 (outside the repository-wide conftest/native EP setup). A separate reproduction identified the failure-path mutation below; no product files were changed.

src/winml/modelkit/quant/fp16.py
zhenchaoni: Fixed in 5e4add07. With keep_io_types enabled, collision renaming and subsequent conversion/validation now run on a working copy. The caller-owned model is updated only after all validation succeeds; the existing safe-conversion path reuses this copy instead of making another one.

Added a regression combining internal I/O Cast name collisions with an unloaded external FLOAT weight. It reproduced the mutation before the fix and now verifies byte-for-byte preservation of the original model after rejection.

Validation: all 149 FP16 unit tests pass, and ruff check src/ tests/ passes. Also reran six real-model checks using both Affinity encoder and decoder: rename-only output equivalence, CLI FP16 conversion with preserved FP32 public I/O, and a second CLI conversion. Rename-only and first-versus-second conversion outputs were bitwise identical across three generated input batches per model on CPU; artifacts passed ONNX checker and strict type inference, and repeated conversion kept names unique.

DingmaomaoBJTU (approved):
Re-reviewed the update since 35d04da. The previous failure-path mutation finding is addressed: keep-I/O preprocessing and conversion operate on a copy, and the caller model is updated only after successful validation. The original unloaded-external-weight reproduction now leaves the caller model byte-for-byte unchanged. All 149 standalone FP16 tests passed locally with ONNX 1.18 / ORT 1.30, outside the repository-wide conftest, and diff checking passed. No remaining blocking findings. CI optim/models/CLA checks have passed; other CI jobs are still pending at review time, so this approval does not assert a fully green CI run.

PR #1418, opened 2026-09-15
test(e2e): isolate EP CLI invocations in subprocesses

KayMKM:
Loading VitisAI before MIGraphX in a shared pytest process causes MIGraphX registration to fail, even though standalone CLI invocations succeed. This blocks the daily E2E pipeline with three perf failures and masks the expected unsupported-compilation error in one compile test.

Run the parametrized perf EP cases and the unsupported-compilation rejection cases through the real CLI in fresh Python processes. Keep the existing result checks, require the expected process exit, capture stdout/stderr on failure, and add brief comments documenting the EP loading interaction. No skips, xfails, or product-code changes are introduced.

Validation on the AMD agent with Python 3.11.14 / ORT 1.27.1:
- Five targeted cases passed on the PR branch: VitisAI followed by the three MIGraphX perf cases and the MIGraphX compile rejection case.
- Before porting the patch to latest main, the full compile and perf suites passed together: 60 passed, 72 skipped, 0 failed. The four originally failing cases all passed.
- Ruff check, Ruff format check, and git diff --check passed.

This isolates CLI invocations to match terminal usage; it does not fix native EP coexistence within a single application process.


DingmaomaoBJTU (approved):
Reviewed the subprocess migration, existing CLI subprocess helpers, EP availability gates, and retained assertions. No blocking findings: perf still requires a clean exit and validates output JSON/monitor results; unsupported compilation still checks the expected error and source-model integrity. A local direct check of the new helper verified argument/timeout forwarding, nonzero-exit rejection, and stdout/stderr diagnostics; diff checking passed. All reported PR CI checks are green. The AMD VitisAI/MIGraphX hardware sequence was not rerun locally; this review does not claim native in-process EP coexistence is fixed.

PR #1420, opened 2026-09-16
fix(quant): clamp DistilBERT mask constants before calibration

KayMKM:
DistilBERT fill-mask evaluation can return NaN metrics when its default w8a16 build runs through OpenVINO. Its attention-mask sentinel (approximately -3.4e38) produces extreme calibration scales because DistilBERT is missing the constant-clamping policy already used by BERT.

Reuse BERT's build configuration for DistilBERT so the existing optimizer bounds the sentinel before calibration and recomputes quantization parameters normally. Keep w8a16 quantization enabled and preserve explicit user overrides. Add regression coverage for direct Where and additive attention masks, including finite scales, probability accuracy, and suppression of masked positions.

### Validation

- 20 related unit checks passed, including the four new regression cases.
- The original fill-mask E2E passed after a cold rebuild on OpenVINO NPU.
- Three additional artifact/accuracy checks passed: only the mask sentinel changed, the quantized graph produced finite CPU outputs, and the same 10-sample NPU evaluation stayed close to the unquantized reference.
- Pseudo-perplexity: 10.2133 quantized vs. 10.0471 unquantized (1.65% difference); NLL: 2.3237 vs. 2.3073. This accuracy comparison covers the E2E's 10 samples.
- Ruff lint, Ruff format, and git diff checks passed.


DingmaomaoBJTU (approved):
Reviewed the DistilBERT registry addition, the reused BERT_CONFIG, configuration assembly/override behavior, the existing constant-clamping implementation, and the four new regression cases. No blocking findings. BERT_CONFIG contributes the existing clamping policy; configuration assembly deep-copies it, and the tests preserve explicit opt-out plus w8a16 weight/activation types. The numerical regressions cover both Where and additive masks, finite positive scales, probability parity and masked-position suppression. git diff --check passed. Reported CI checks are passing except test (remaining), which is still pending at review time. The new tests and OpenVINO hardware evaluation were not rerun locally; hardware results cited in the PR remain author-provided evidence.

PR #1428, opened 2026-09-17
feat(skills): add auto-optimize workflow

DingmaomaoBJTU:
## Summary

- check in the canonical `auto-optimize` Agent Skill from `gim-home/ModelKitArtifacts@2bcf29d9150ab3af2909f3cc2a0b278d7ca2387d`
- adapt the repository-layout contract and register `skills/auto-optimize/tests` in the existing CI matrix, following #1415
- normalize imported Python to this repository's Ruff policy while preserving script interfaces and generated output
- close inherited fail-open validation paths for non-finite hotspot evidence and quoted/escaped absolute reproduction paths

## Validation

- `uv run pytest skills/auto-optimize/tests --tb=short --no-cov` (`309 passed`)
- `uv run ruff check skills/auto-optimize`
- `uv run ruff format --check skills/auto-optimize`
- pre-commit hooks on all 13 target-adapted files
- all 7 bundled JSON documents parsed successfully
- all 29 pinned-source files are present; final diff is limited to the skill and CI registration


DingmaomaoBJTU (commented):
Detailed readiness review: all 309 bundled tests passed locally, and all reported CI checks are green. However, the executable validation and workflow ordering gaps below prevent a ready-for-check-in verdict. Additional local reproductions confirmed acceptance of NaN final-report evidence and a generated replay wrapper returning success after an earlier native failure. No PR files were modified. The current tests provide useful script/contract coverage, but do not cover these end-to-end failure paths.

skills/auto-optimize/scripts/render_report.py
```diff
@@ -0,0 +1,911 @@
...
+        [
+            f"leader.{field}"
+            for field in ("id", "status", "model_path", "correctness", "quality")
+            if _missing(leader.get(field))
+        ]
+    )
+    errors.extend(
+        [
+            f"leader.{field}"
+            for field in ("p50_ms", "gain_pct")
```
DingmaomaoBJTU, line 344: [P1] Reject non-finite final performance evidence. These isinstance checks accept float NaN/Infinity (and bool), and the final validator does not reject non-finite confidence bounds either. Using the bundled valid _report() fixture, replacing baseline.p50_ms, leader.p50_ms and both leader CI bounds with NaN still passes validate_report(final=True). finalize_output and promotion validation rely on this gate, so an invalid measurement can be published as a confirmed champion. Require finite numeric values with appropriate latency/range constraints and validate confidence bounds; add negative final-report and bundle/promotion tests, not only hotspot-planner NaN tests.

skills/auto-optimize/scripts/finalize_output.py
```diff
@@ -0,0 +1,854 @@
...
+    }
+}
+
+Get-Command winml -ErrorAction Stop | Out-Null
+
+if ($ValidateOnly) {
+    exit 0
+}
+
+& (Join-Path $PSScriptRoot 'repro-run.ps1')
```
DingmaomaoBJTU, line 127: [P2] Ensure an earlier native replay failure cannot be overwritten by a later success. The generated wrapper sets ErrorActionPreference=Stop but does not make nonzero native exit codes terminating errors. It returns only the last LASTEXITCODE after repro-run.ps1. A body containing $Root=$PSScriptRoot, a native exit 7, then a native exit 0 passes the current script validator and the generated wrapper exits 0; this was reproduced locally with harmless cmd.exe calls. A failed winml build followed by a successful command can therefore appear to reproduce successfully. Define and enforce checked native invocation for every replay step (or require a supported PowerShell native-error mode), and execute a multi-step failure regression through the real wrapper.

skills/auto-optimize/SKILL.md
```diff
@@ -0,0 +1,54 @@
...
+
+After structural, correctness, paired-evidence, and trace gates, the [Feature Gap Engineer](./roles/feature-gap-engineer.md) may modify `WINML_CLI_REPO` in an isolated current-main worktree. Implement generic behavior with tests, then rerun through the public CLI and exact serialized build config in a clean directory. Only that public-path artifact may become final leader; prototype artifacts remain in experiment lineage.
+
+After validation, run `promotion.py create` once for `promotion_handoff.json`; follow [PR Routing](./references/pr-routing.md). Auto-optimize owns the optimizer PR. Run `gh label list`; the target repo must contain `model-opt-by-skill`, and the skill must not create the label automatically. Create the Draft PR with `gh pr create --draft --label model-opt-by-skill`, then verify with `gh pr view <url> --json labels`. Missing or unavailable label blocks handoff, and missing post-create label verification blocks handoff. Use [Ponytail](./references/ponytail.md) or fallback, then invoke [Check-in Reviewer](./roles/checkin-reviewer.md), record its ready for check-in verdict; never merge or convert the Draft.
```
DingmaomaoBJTU: [P2] Finalize the bundle before creating its promotion handoff. The documented sequence invokes promotion.py create here, but that command immediately calls validate_output_bundle and hashes manifest.json. The first instruction to publish that bundle is later at line 50, after PR creation/review, so a fresh run following the sequence has no validated bundle to promote. Move replay/finalization/validation before handoff creation and define how later PR/review metadata is incorporated without silently invalidating the handoff manifest hash. Add a fresh-run workflow scenario that exercises the order rather than merely checking that both helper names occur in the skill.

ssss141414 (approved):

